PR #28, opened 2015-01-04
permit single quotes if they quote literal double quotes

jackwasey:
A legitimate use of a single quote pair in R is to quote a literal double quote. This tiny patch allows this. It's not immediately clear to me how to test a negative with your check_lint tests. Thanks.

### DIFF
--- a/R/single_quotes_linter.R
+++ b/R/single_quotes_linter.R
@@ -5,7 +5,7 @@ single_quotes_linter <- function(source_file) {
   lapply(ids_with_token(source_file, "STR_CONST"),
     function(id) {
       parsed <- with_id(source_file, id)
-      if (re_matches(parsed$text, rex(start, "'", anything, "'", end))) {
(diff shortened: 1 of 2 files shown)


jimhester:
Thank you very much for the patch.

Could you make the following changes.

1. Please use [rex](https://github.com/kevinushey/rex) to construct the regular expression.
  - The equivalent `rex` expression to the regex you wrote is 
  ```r
rex(start, single_quote, any_non_double_quotes, single_quote, end)
```
2. Please add at least one additional test to [tests/testthat/test-single_quotes_linter.R](https://github.com/jimhester/lintr/blob/1f850191680a8925d14b288b36ea74e1069d170e/tests/testthat/test-single_quotes_linter.R) showing the new functionality.
  - You can test that a given linter does not return any lints by passing `NULL` as the second argument.
  - So an acceptable test would be 
  ```r
expect_lint('quote "something"', NULL, single_quotes_linter)
```

Thanks!

jackwasey:
Think that should do it. Thanks.

jimhester:
Thanks!

PR #69, opened 2015-03-24
Respect "# nolint" comments with caching

### LINKED ISSUE #68: Caching and # nolint
The `# nolint` comments don't seem to work properly with caching. They are respected only on first execution (presumably, when the cache is dirty) -- from the second execution onwards, errors are reported.

Try

```
lintr::lint_package(".", cache = TRUE)
```

for [my package](https://github.com/krlmlr/kimisc).

krlmlr:
- includes test

Fixes #68.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 * `trailing_whitespace_linter` was reporting the incorrect line number
 * Use RStudio source marker API to display lints (#37, @jjallaire)
 * Permit single quotes if they quote literal double quotes (#28, @jackwasey)
+* # nolint comment are respected with caching (#68, @krlmlr)
 
 # lintr 0.2.0 #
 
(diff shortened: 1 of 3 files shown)


jimhester:
Thank a lot for adding a test for this Kirill, very helpful!

krlmlr:
Well... the test was supposed to fail, it was pushed before the fix. Also, I'm not sure -- could the `clear_cache()` call introduce a race condition or something?

krlmlr:
I think Travis CI got the wrong revision for this particular test. I'll create a bogus pull to test the test.

jimhester:
The main issue I could see with using `clear_cache()` with no arguments is it will clear the entire cache for all files.

It would probably be better to use `clear_cache("exclusions-test")` to only clear the cache for that file.

PR #78, opened 2015-04-09
Simplify getting started

krlmlr:
Please feel free to change wording. Adding this as a supported function to the package is probably too much.

Ref: https://github.com/jimhester/lintr/issues/73#issuecomment-90927745

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,13 +127,13 @@ You can create an configuration file for `lintr` that ignores all linters that s
 
 ```r
 library(magrittr)
+library(dplyr)
 lintr::lint_package() %>%
   as.data.frame %>%
-  plyr::count("linter") %>%
-  plyr::arrange(-freq) %>%
-  with(sprintf(
-    "linters: with_defaults(\n    %s\n  )",
-    paste(sprintf("%s = NULL # %s", linter, freq), collapse = "\n    , "))) %>%
+  group_by(linter) %>%
+  tally(sort = TRUE) %$%
+  sprintf("with_defaults(\n  %s\n  NULL\n)",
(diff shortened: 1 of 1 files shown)


jimhester:
Could also use `%$%` instead of the curly brackets, makes things a little simpler, although perhaps more obscure.

```r
library(dplyr)
lintr::lint_package() %>%
  as.data.frame %>%
  group_by(linter) %>%
  tally(sort = TRUE) %$%
  sprintf("with_defaults(\n  %s\n  NULL\n)",
          paste0(linter, " = NULL, # ", n, collapse="\n  ")) %>%
  cat(file = ".lintr")
```

krlmlr:
Thanks. Your code works for me, too. Updated PR.

(I'm not yet convinced about `%$%`. A big advantage of `magrittr` is that code is not only easy to write but also easy to maintain. Imagine I wanted to add a step after the `tally` call -- I have to remember that `%$%` is logically bound to the next step. That's why I personally prefer the `with` syntax. Something like the following would be better:

```
lintr::lint_package() %>%
  as.data.frame %>%
  group_by(linter) %>%
  tally(sort = TRUE) %>%
  expose %>%
  sprintf("with_defaults(\n  %s\n  NULL\n)",
          paste0(linter, " = NULL, # ", n, collapse="\n  ")) %>%
  cat(file = ".lintr")
```
)

jimhester:
That is fair, I am not sure I am sold on `%$%` myself.  I think using `with` as you did might be the most desirable as it keeps the internal variable exposure contained to the single statement.

I will merge this as is however, for a simple one off script I think the succinctness of `%$%` somewhat beneficial.

jimhester:
Thanks!

PR #83, opened 2015-04-13
Add linter for commented R code

jackwasey:
This linter looks for code-like comments, allowing roxygen blocks which may contain examples.

### DIFF
--- a/R/commented_code_linter.R
+++ b/R/commented_code_linter.R
@@ -1,0 +1,29 @@
+#' @describeIn linters checks that there is no commented code outside roxygen
+#'   blocks
+#' @export
+commented_code_linter <- function(source_file) {
+  res <- re_matches(source_file$file_lines,
+                    rex("#", except("'"), anything,
+                        or(some_of("{}[]"), # code-like parentheses
+                           "<-", "->", "==", # an assignment
+                           group(graphs, "(", anything, ")"), # a function call
+                           group("!", alphas) # a negation
+                        )
+                    ),
+                    global = TRUE, locations = TRUE)
+  line_number <- length(source_file$file_lines)
+  lints <- list()
+  while (line_number > 0L && !is.na(res[[line_number]][["start"]])) {
+    lints[[length(lints) + 1L]] <- Lint(
+      filename = source_file$filename,
+      line_number = line_number,
+      column_number = res[[line_number]][["start"]],
+      type = "style",
+      message = "Commented code should be removed.",
+      line = source_file$file_lines[[line_number]],
+      linter = "commented_code_linter"
+    )
+    line_number <- line_number - 1L
+  }
+  invisible(lints)
+}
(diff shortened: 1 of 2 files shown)


R/commented_code_linter.R
```diff
@@ -0,0 +1,29 @@
+#' @describeIn linters checks that there is no commented code outside roxygen
+#'   blocks
+#' @export
+commented_code_linter <- function(source_file) {
+  res <- re_matches(source_file$file_lines,
+                    rex("#", except("'"), anything,
+                        or(some_of("{}[]"), # code-like parentheses
+                           "<-", "->", "==", # an assignment
```
jimhester: Should this check for `=` assignment rather than `==` equality?

I assume you restricted the search to assignments only because it could be valid to have a comment like

```r
# checks that var > 2
if (var > 2) {
  x
}
```

jackwasey:
Yes, I did mean it to be equality, although my comment says otherwise. I thought too many people would legitimately use an equals symbol in a text comment. Overall, this linter is going to have trouble with false positives or false negatives, and I thought it better to err on the side of not giving spurious messages. I think it would be nice to have this, even if it only catches 75%, but prompts people to think about commented code, which, AFAIK is considered poor form.

jimhester:
I agree with you and think it is wise to restrict this to only a few operators rather than all of them.  I guess one way to go would be to include all of the operators in your regex and parse (but not execute) the code in the comment, then only generate a lint if the code parses without error.

jimhester:
The looping can be made a little simpler by the fact that you don't need to search globally on the line, (since once you start a comment the rest of the line is commented) and just using an lapply loop.  i.e.

```r
commented_code_linter <- function(source_file) {
  res <- re_matches(source_file$file_lines,
                    rex("#", except("'"), anything,
                        or(some_of("{}[]"), # code-like parentheses
                           "<-", "->", "==", # an assignment
                           group(graphs, "(", anything, ")"), # a function call
                           group("!", alphas) # a negation
                        )
                    ),
                    global = FALSE, locations = TRUE)
  line_numbers <- rownames(na.omit(res))
  lapply(line_numbers, function(line_number) {
    Lint(
      filename = source_file$filename,
      line_number = line_number,
      column_number = res[line_number, "start"],
      type = "style",
      message = "Commented code should be removed.",
      line = source_file$file_lines[line_number],
      linter = "commented_code_linter"
    )
  })
}
```

jackwasey:
I like it. Executing a parse on the comments would be more sensitive without losing much or any specificity, but I don't think I have time to learn how to do that. I assume you mean this would be in addition to the regex above.

jimhester:
Yes exactly, it didn't really take too much code to do in the end.

see https://github.com/jimhester/lintr/commit/0fb087149e45aeeb5f1f23d4a11cff980deb4feb if you are interested in the details, I changed some things around and added support for all the operators, but the bulk of it is your original pull.  The only thing I think would be nice is to detect multi-line code comments (as they won't be parse-able split into individual lines).  But it is more a nice to have rather than essential, so I think it is great as is.

Thanks again for the pull request!

jackwasey:
That's fantastic. Now I'll see if my lint is still clean with your version!

PR #93, opened 2015-06-02
expect_lint fix

rundel:
Changes include
* A fix for a bug that was preventing expect_lint from returning the results of any checks after the first. 
* Changed `expect_lint_free` to be a replacement for `expect_lint(content, NULL, checks)` which I think will make the tests cleaner. Existing `expect_lint_free` is renamed `expect_package_lint_free` which I think is more explicit.
* I've also included a few small formatting tweaks and cleanups


jimhester:
Thank you for the pull request!  I generally prefer pull requests to consist of as few changes as possible, each commit here would ideally be a separate pull request, as they are all independent.  However my comments on each are below.

1. I believe testthat can only handle 1 returned expectation per test, so there was no choice but to take the first one (which is why that was done).  Probably logic should be added to see if any of the expectations fail and if so to use that expectation as the return value.
2. I would expect the full package check to be used much more frequently in external packages than the per file check, so I would prefer it retain the same name of `expect_lint_free` and the less used function get a more descriptive name.
3. 4722335 seem like good independent changes, if you put them in a different pull request I can merge them now.

rundel:
Sure thing, I'll refactor and resubmit.

PR #95, opened 2015-06-02
Fixes bug where only the results of the first check condition being used

rundel:
Existing implementation only evaluates the first check condition, the following code should fail:

```
# should fail
lintr:::expect_lint(
    "a=1",
     list(c(line_number=1, column_number=3)),
     assignment_linter
)
```

### DIFF
--- a/R/expect_lint.R
+++ b/R/expect_lint.R
@@ -17,16 +17,16 @@ expect_lint <- function(content, checks, ..., file = NULL) {
     content <- readChar(file, file.info(file)$size)
   }
 
-    results <- expectation_lint(content, checks, ...)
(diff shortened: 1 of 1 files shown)


jimhester:
The issue with this is that while it fixes the bug you noted testthat blows up when you supply more than one expectation, i.e.

```r
# test.R
test_that("test", {
  expect_lint(
    "a=1\nb=2",
    list(c(line_number=1, column_number=3),
      c(line_nubmer=2)),
    assignment_linter
    )
})

# run with testthat::test_file("test.R")
```
The above works correctly on master but error's with this pull request.

PR #99, opened 2015-07-13
fix CI environment detection

paulstaab:
Currently, the `CI` environment variable is double dereferenced, first in `in_ci` and the result again in `logical_env`. For me, this makes `in_ci` always return `FALSE` and disables commenting. 

This fixes the problem and adds a short unittest.


jimhester:
Thank you!  Not sure how I missed this pull request, I have been meaning to fix the commenting, just haven't had time to track down the cause.

Should also close #100

PR #109, opened 2015-07-29
Exclude checking `!` from spaces_left_parentheses_linter

yu-iskw:
HI @jimhester, 

I tried to exclude checking `!` before a left parenthesis from `spaces_left_parentheses` in order to pass like `!(a == b)`. Could you review it?

Thanks,
Yu


jimhester:
LGTM, thank you for pull request and for adding a test for this behavior!

yu-iskw:
Thank you for merging it!

Issue #77, opened 2015-04-08
Space before parenthesis: c((x + y) * z)


krlmlr:
```
## R/id.r:51:14: style: Place a space before left parenthesis, except in a function call.
##     res <- c((mat - 1L) %*% combs + 1L)
##              ^
```

After adding the space:

```
## R/id.r:34:13: style: Do not place spaces around code in parentheses or square brackets.
##     res <- c( (mat - 1L) %*% combs + 1L)
##             ^
```

One of both has to give in here. Perhaps similar to #75.

jimhester:
Agreed, I guess of the two options I prefer the former, so we will have to special case multiple `(` I guess.

richfitz:
Might be worth a special case for things like `!` too? I feel that

```
if (!(foo && bar || baz)) {
```

is more common than

```
if (! (foo && bar || baz)) {
```

jimhester:
Yeah we need to rework the logic of this linter, `if (!(foo && bar || baz)) ` is definitely preferred.

yu-iskw:
How about `:` before a left parentheses, such as `1:(len(array) - 1)`? 
I think it would be more common than `1: (len(array) - 1)`.

jimhester:
All of these cases do not produce lints in the latest master, but I added tests for them to prevent regressions.

PR #110, opened 2015-08-12
#77 Add ":" and "[" into the rule for a space before a left parentheses

yu-iskw:
HI @jimhester,

I think it would be more natural to pass the rule for a space before a left parentheses with `:` and `[` as follows.

```
(x - 1):(y - 1)
x[(x - 1):(y - 1)]
```

Thanks,
Yu


jimhester:
Looks resonable, thanks!

yu-iskw:
Thank you for merging it!

Issue #118, opened 2015-11-04
A commented_code_linter bug: a comment after `#' `


yu-iskw:
It seems that `commented_code_linter` has a bug, when `#` is after '#' '.

```
#' take(foo, 2L) # list(1, 2)
```

I tried to add the below test into `tests/testthat/test-commented_code_linter.R`. So it failed.

```
+  expect_lint("#' take(foo, 2L) # list(1, 2)",
+    NULL,
+    commented_code_linter)
```

jimhester:
This seems to be doing the proper thing in the examples you gave. Roxygen comments are supposed to be ignored.

I would not expect any lints from `"#' blah <- 1 # inside comment"`, but I _would_ expect them from `#' take(foo, 2L) # list(1, 2)` as `# list(1, 2)` is commented code. This is exactly the behavior I see.

yu-iskw:
Sorry, it's my fault. I made a mistake to paste the code.

Anyway, I'll give it a try to fix.

jimhester:
There is a bug related to this, `## whatever` is treated as commented code because `# whatever` is parseable. The regular expression should strip leading spaces and `#` before trying to parse.

jimhester:
In the current devel roxygen comments `#'` are all ignored and `## whatever` does not create a lint, so we can close this.

PR #119, opened 2015-11-04
[#118] fix a bug of `commented_code_linter`: commented code inside roxygen docs

yu-iskw:
https://github.com/jimhester/lintr/issues/118

### DIFF
--- a/R/commented_code_linter.R
+++ b/R/commented_code_linter.R
@@ -47,7 +47,9 @@ commented_code_linter <- function(source_file) {
     is_parsable <- parsable(substr(line,
                                    res[line_number, "code.start"],
                                    res[line_number, "code.end"]))
-    if (is_parsable) {
+    is_roxygen_comment <- re_matches(line, rex(start, "#'", any_spaces))
+
+    if (is_parsable && !is_roxygen_comment) {
       Lint(
         filename = source_file$filename,
         line_number = line_number,
--- a/tests/testthat/test-commented_code_linter.R
+++ b/tests/testthat/test-commented_code_linter.R
@@ -22,7 +22,7 @@ test_that("returns the correct linting", {
     NULL,
     commented_code_linter)
 
-  expect_lint("#' foo(bar, 2L) # foo(1, 2)",
+  expect_lint("#' foo(bar, 2L) # 1 + 2",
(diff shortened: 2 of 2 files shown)


yu-iskw:
@jimhester could you review it when you have time?

jimhester:
I think this is already working as intended,
```r
#' foo(bar, 2L) # foo(1, 2)
```
Should have a lint for the `# foo(1, 2)` as that is commented code, why do you think it should not be linted?

yu-iskw:
For example, when I run the test on master, I got an error.

```
  expect_lint("#' blah <- 1 # 1 + 1",
    NULL,
    commented_code_linter)
```

```
commented_code_linter : ...../var/folders/j0/7y3gvsjn0bgbyfgj7dhqf0580000gn/T//RtmpuRSO1v/file47fde56a20b:1:16: style: Commented code should be removed.
#' blah <- 1 # 1 + 1
               ^~~~~
....
1. Failure: returns the correct linting ----------------------------------------
commented_code_linter returned list(filename = "/var/folders/j0/7y3gvsjn0bgbyfgj7dhqf0580000gn/T//RtmpuRSO1v/file47fde56a20b", line_number = 1, column_number = 16, type = "style", message = "Commented code should be removed.", line = "#' blah <- 1 # 1 + 1", ranges = list(c(16, 20)), linter = "commented_code_linter") lints when it was expected to return none!
```

jimhester:
`# 1 + 1` is also valid commented code and should produce a lint. Again why are you expecting these not to do so?

yu-iskw:
I think we shouldn't produce a lint if a commented code is in roxygen docs.

jimhester:
Why not?

fangly:
Commented code outside a roxygen section is presumably be stale code that is not current anymore, remains from debugging or throaway code generated to test an idea, etc... One may justifiably want to lint such commented code.

On the other hand, the goal of a roxygen section is to document something and any commented code within a roxygen section serves this purpose of documentation as well. Thus, I agree with @yu-iskw on this issue and my inclination would be to _not lint commented code within roxygen sections_.

naught101:
Code in an `@example` section in roxygen is particularly legit..

PR #122, opened 2015-12-03
Add an example image for Atom

Arcanemagus:
Adds an example image for Atom in a theme that somewhat matches the existing images.
Corrects the link to linter to the actual repository.

If you would prefer this against the `atom` branch just let me know.


jimhester:
Great, thanks!

Issue #75, opened 2015-04-08
Treat )( and ]( as function call


krlmlr:
```
+## R/each.r:71:45: style: Place a space before left parenthesis, except in a function call.
+##         for (i in 1:n) proto[[i]] <- fs[[i]](x, ...)
+##                                             ^
+## R/here.r:24:39: style: Place a space before left parenthesis, except in a function call.
+##   call <- substitute(function(...) (f)(...), list(f = f))
+##                                       ^
```

jimhester:
Both of these cases no longer produce lints in the devel version.

PR #128, opened 2015-12-29
Make spaces_left_parentheses_linter more robust when determining `(` type 

saurfang:
Currently `spaces_left_parentheses_linter` contains false negatives when it comes to
```r
for(i in seq_along(1:10)) { }
```
where `if`, `for`, `while` blocks could contain function calls.

It seems to me that we can improve this linter by stricter checking on the type of `(`'s owner. Previously if the line has any function calls, we completely let it go.

I have also incorporated the improved linter's suggestions in this PR. Hope that is okay.

Finally I'm not very familiar with this package yet. So definitely let me know if you see any concerns or potential improvements of my PR.

### DIFF
--- a/R/spaces_left_parentheses_linter.R
+++ b/R/spaces_left_parentheses_linter.R
@@ -7,15 +7,15 @@ spaces_left_parentheses_linter <- function(source_file) {
 
       parsed <- source_file$parsed_content[id, ]
 
-      # content before parsed parentheses
-      before_parsed <- source_file$parsed_content[1:(id - 1), ]
-      # only look terminal tokens
-      terminal_types <- before_parsed[before_parsed$terminal, "token"]
-      # find the type of last such token
-      type <- terminal_types[length(terminal_types)]
+      terminal_tokens_before <-
+        source_file$parsed_content$token[
+                                         source_file$parsed_content$line1 == parsed$line1 &
+                                         source_file$parsed_content$col1 < parsed$col1 &
+                                         source_file$parsed_content$terminal]
+      last_type <- tail(terminal_tokens_before, n = 1)
 
-      is_function <- length(type) %!=% 0L &&
-        (type %in% c("SYMBOL_FUNCTION_CALL", "FUNCTION", "'}'", "')'"))
+      is_function <- length(last_type) %!=% 0L &&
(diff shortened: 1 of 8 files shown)


R/spaces_left_parentheses_linter.R
```diff
@@ -7,14 +7,15 @@ spaces_left_parentheses_linter <- function(source_file) {
...
-          "token"]
-
-      is_function <- length(family_ids) %!=% 0L &&
-        any(types %in% c("SYMBOL_FUNCTION_CALL", "FUNCTION"))
+      # content before parsed parentheses
+      before_parsed <- source_file$parsed_content[1:(id - 1), ]
+      # only look terminal tokens
+      terminal_types <- before_parsed[before_parsed$terminal, "token"]
+      # find the type of last such token
+      type <- terminal_types[length(terminal_types)]
```
jimhester: Maybe change this to 
```r
last_type <- tail(terminal_types, n = 1)
```
Which might make it more explicit so a comment is unnecessary.

R/spaces_left_parentheses_linter.R
```diff
@@ -7,14 +7,15 @@ spaces_left_parentheses_linter <- function(source_file) {
...
-      family_ids <- family(source_file$parsed_content, parsed$id)
-
-      types <- source_file$parsed_content[
-          source_file$parsed_content$id %in% family_ids,
-          "token"]
-
-      is_function <- length(family_ids) %!=% 0L &&
-        any(types %in% c("SYMBOL_FUNCTION_CALL", "FUNCTION"))
+      # content before parsed parentheses
+      before_parsed <- source_file$parsed_content[1:(id - 1), ]
```
jimhester: Could these lines be replaced by the `tokens_before` assignment in the closed_curly_linter?

https://github.com/jimhester/lintr/blob/56c35a3a387b136f4087b92c4992d70d39edb9e8/R/closed_curly_linter.R#L13-L15

saurfang:
Thanks for the suggestion @jimhester. I have made corresponding changes. I have also added `']'` to the whitelist per #75, which I think makes sense and is relevant to this PR improving robustness of this particular linter. Let me know if you have any other thoughts or would like me to rebase this PR.

jimhester:
Looks great, thanks!

Issue #111, opened 2015-08-23
commas_linter does not recognize lints when preceded by a blank line


emallickhossain:
I've found that the commas_linter does not recognize lints when the line to be linted is preceded by a blank line or a commented line. Below is a minimal working example:

file: <b>Test.r</b>
    
    #random comment

    a <- c(4,5)
    b = 7

Then, I run the following:

    lintr::lint("./Test.r", linters = list(spacesAfterCommas = commas_linter, 
                                           useArrowForAssignments = assignment_linter))

The only error I get is the following:

    ./Test.r:4:3: style: Use <-, not =, for assignment.
    b = 7
      ^

Why is the commas_linter not catching my error? If I remove the blank line, commas_linter catches the lint.

zachmayer:
This happens to me too.  If I take the blank line out, it catches the error.

jimhester:
Closed by #129

PR #129, opened 2015-12-29
Get correct line_number in commas_linter

saurfang:
This is related to #111. It looks like the root cause is that when there is empty line(s), the parsed results' row number (id) is different from line number, which is provided in the `names`.

I also find it more intuitive for the missing space after lint to have position to be after comma so it is as easy as to press space to take the advice in RStudio for example. Hope you are okay with this change and me jamming it in this PR.


jimhester:
Looks great to me, thanks!

PR #130, opened 2015-12-29
infix_spaces_linter should check `=` inside function calls

saurfang:
If I'm not mistaken, according to the style guide and your current code implementation, `=` is meant to be linted. However it seems that `=` inside function calls is a differently named token versus plain `=` assignment. I have added `EQ_SUB` to the list to capture that.

In addition, `=` inside function calls treats the entire function call as its parent. Therefore I changed the siblings check to be "greater than one" rather than "equal to two". If I understand your intention correctly, this check is meant to prevent false positive like `-1, <2` etc. So far I haven't been able to come up or find any false positives my change would introduce.

Let me know if I missed anything or incorrectly interpreted the style guide.

### DIFF
--- a/R/absolute_paths_linter.R
+++ b/R/absolute_paths_linter.R
@@ -37,7 +37,7 @@ absolute_paths_linter <- function(source_file) {
       rex(quote, x)
     },
     character(1),
-    USE.NAMES=FALSE)
+    USE.NAMES = FALSE)
 
   lapply(ids_with_token(source_file, "STR_CONST"),
     function(id) {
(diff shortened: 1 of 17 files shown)


jimhester:
Can you rebase this on the new master, it must have had a conflict with the previous pull requests I merged.

saurfang:
Ah yes. I got my terminology mixed up. Thanks for the clarification! I have rebased the branch.

jimhester:
Great looks good! Thanks for rebasing!

Issue #147, opened 2016-05-12
Semicolons at the end of the line?


gaborcsardi:
Would you like to see a PR for it?

jimhester:
Sure,  that is a good one actually!

gaborcsardi:
How about non-trailing semicolons? Personally I think that these are quite often fine, but sometimes not.
Should we report these? Have a separate linter for them? Ignore them completely?

jimhester:
I think non trailing are fine as well

gaborcsardi:
Closed by #148.

PR #148, opened 2016-05-12
trailing_semicolon_linter #147

gaborcsardi:
A minor glitch is that if top level expressions are separated by semicolons, those are not picked up, because they are not in the parse data frame.

EDIT: if you want, I can pick them up from the source code, from `lines`.


jimhester:
Thanks Gabor, looks great.

Issue #151, opened 2016-05-17
Use sg lightweight instead of igraph


gaborcsardi:
If you are interested, I will provide a PR soonish. Either for switching to https://github.com/mangothecat/simplegraph which is R-only and has no dependencies, or custom code. It is fairly trivial.

jimhester:
Sounds good to me, I had initially used igraph as a performance improvement, but it is a heavy dependency for sure.

gaborcsardi:
Do you have an idea how big the graphs are?

jimhester:
Depends on the file you are linting I guess, I was testing on https://github.com/wch/r-source/blob/trunk/src/library/tools/R/QC.R which is ~8300 lines long. Lintr splits the parse data data frames by expression, looks like the largest dataframe is 6652 rows, which I guess is fairly small...

gaborcsardi:
Indeed, my quick and dirty R implementation will not do. For QC.R it increases running time from 70 sec to 223 sec. https://github.com/gaborcsardi/lintr/commit/7121cf8d86b547539f7ebc773910bd05667eccac

I'll try sg else in a minute.

Btw. is this graph component stuff only needed for splitting the file into expressions? Because then maybe I can start sg else.

Btw. 2. it is easy to write this in C as well, is that preferable to igraph? I am not sure myself. Maybe the best would be some micro-package dependency that only does this, but fast (= using C code).

gaborcsardi:
OK, the bottleneck was a stupid type conversion (did you know that `modifyList` is super slow? It makes sense if you look at its code...), and now it is actually faster than the current `lintr` code, it is ~61 sec for QC.R. Hopefully it is also correct. :) Let me run some more tests, and then I'll submit a PR.

PR #152, opened 2016-05-17
Replace igraph calls with our own R code, get rid of igraph

gaborcsardi:
It is also faster, at least the complicated QC.R case. I also profiled it for this file, it is as fast as it gets, `self.time: 0.00`, `self.pct: 0.00`. :)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -16,7 +16,6 @@ Imports:
     stringdist,
     testthat,
     digest,
-    igraph,
     rstudioapi (>= 0.2),
     httr,
     jsonlite,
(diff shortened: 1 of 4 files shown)


gaborcsardi:
This closes #151.

jimhester:
Looks great, could you add a note to NEWS.md then I will happily merge.

gaborcsardi:
I am not sure what you want in the NEWS, nothing has changed in the API, right? Just that it is not depending on igraph now?

jimhester:
Yeah just mentioning the dependency is removed is fine.

I don't typically restrict NEWS mentions to just API changes, but maybe that is a good practice...

gaborcsardi:
OK, added the bullet point to the NEWS. Thanks!

jimhester:
Thanks!

PR #155, opened 2016-05-21
Seq linter: 1:length(...) and similar expressions

gaborcsardi:
It finds 1:length(...), 1:nrow(...), 1:ncol(...), 1:NROW(...) and 1:NCOL(...) expressions, everywhere.

Maybe we should restrict it into `for` loop conditions, I am not sure.

It depends on the XML data, thus the #154 PR.

I'll add a bullet to the NEWS.


jimhester:
Thanks Gabor, this is awesome, now I need to go back and rewrite some of the old linters using XPath. I imagine it will make the code cleaner and probably faster as well...

gaborcsardi:
Thanks for merging! I'll add more linters that use this.

PR #156, opened 2016-05-23
Allow lint results to be exported as checkstyle xml

joshkgold:
Based on the popular [Checkstyle](http://checkstyle.sourceforge.net/) format, allow lints to be exported to an xml file.  Useful for CI servers like Jenkins that can display a Checkstyle report and visualize errors.

Note this is the only bit of R code I've written, out of necessity.  Comments and critiques on style and semantics are appreciated.


jimhester:
Thanks for the PR!

I ended up rewriting the majority of this, but your code gave me a good idea of the requirements. The major things I changed were

- Uses xml2 to create the XML rather than doing it by hand.
- Does not try to create a results directory (this does not belong in this function IMHO)
- More lengthy test including multiple source files in results.

Thanks again!

joshkgold:
Awesome!  I don't even want to know how many minutes you spent overhauling it compared to the hours it took me to write that small piece of code :-)

Thanks for reviewing and improving!

Get Outlook for iOS<https://aka.ms/o0ukef>

_____________________________

PR #157, opened 2016-05-24
seq_linter fixes: 1L:length(...) and length(...):1

gaborcsardi:
Some fixes for the new seq_linter:

* Find `1L:length(...)`, etc. expressions as well.
* Proper printout for the `length(...):1` expressions. Somewhat surprisingly, these are found, too, there is no order in the XPath search.


jimhester:
Thanks!

PR #165, opened 2016-08-26
Allow trailing_whitespace_linter to skip empty lines

flying-sheep:
Also included backwards compatible code for people putting `trailing_whitespace_linter` in their config

### DIFF
--- a/R/trailing_whitespace_linter.R
+++ b/R/trailing_whitespace_linter.R
@@ -1,34 +1,44 @@
 #' @describeIn linters check there are no trailing whitespace characters.
 #' @export
-trailing_whitespace_linter <- function(source_file) {
+trailing_whitespace_linter <- function(check_empty_lines = TRUE) {
+  # for backwards compatibility, check if this was called on the
+  # check_empty_lines option or directly on a source file
+  if (is.logical(check_empty_lines)) {
+    function(source_file) trailws_linter(source_file, check_empty_lines)
+  } else {  # call with old default
+    source_file <- check_empty_lines
+    trailws_linter(source_file, TRUE)
+  }
+}
+
+trailws_linter <- function(source_file, check_empty_lines) {
   res <- re_matches(source_file$lines,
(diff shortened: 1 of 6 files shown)


jimhester:
Can you explain the reason you want this? Extra whitespace in "empty" lines is just as useless as extra whitespace in any other line and should be linted and removed.

flying-sheep:
i like to be able to put my cursor there and type instead of having to adjust indentation first. this requires empty lines to be indented just as much as surrounding lines.

i plan to add a linter that checks for uneven indentation. it will get an option to also check empty lines, which would collide with a `trailing_whitespace_linter` that doesn’t have that option.

naught101:
I think this is a bad idea. A good editor should automatically indent a new line to the correct depth without needing to leave white space in empty lines. I know that vim does it (e.g. indents an extra 4 spaces after braces, and indents up to the parenthesis when splitting function calls over multiple lines). Is there any case other than these where pre-indentation is helpful?

flying-sheep:
It's simply a matter of preference. You don't need to think it's a good idea as long as it's optional.

AshesITR:
Note that RStudio is one of those editors that will happily keep "empty" lines padded with spaces when they are indented. Since that's a very common IDE for R development, I'd also like this feature.

Just type something like this in a new RStudio window, save & lint:

```r
for (i in 1:10) {
  x <- 2 * i
  # Hit Enter 2x:

  print(i)
}
```

MichaelChirico:
Anyone still interested in keeping this in lintr?

AshesITR:
Is this still a default behaviour of RStudio?
If so, I think it's worth adding the option to ignore pure-whitespace lines.

AshesITR:
NB currently, completely blank lines aren't caught because the linter operates on expressions and not globally, so only lines with expressions are considered for linting.

AshesITR:
Superseded by #1044

PR #170, opened 2016-09-09
I have divided commas_linter() to allow separation  and easier modification

rmsharp:
I would like to have the syntax my_df[ , 1:3] allowed. I thought a reasonable first step would be to divide the two primary functions found in commas_linter.R


jimhester:
The commas linter is one of the most resource intensive linters, because R code tends to have a large number of commas. Part of the reason the current code has two linters combined is so each comma only needs to be looked at once.

This PR duplicates a large amount of code for each linter, this will make future maintenance more challenging, ideally the common code should be put in a helper function that all of the linters can use.

Even if we address the second point I am not sure how to handle the performance implications of this change, lintr is already slower than I would like 😦 .

rmsharp:
Jim,

Thank you for the clear explanation. Sounds like for the short term, I can just put in my own comma lintr that enforces the style I have adopted.

Various styles are always going to be a challenge.

Mark
R. Mark Sharp
rmsharp@me.com

tyner:
I think in the case of consecutive commas (for example when indexing arrays), it should be more forgiving. Currently,

<code>x[,, ]</code>  gives "style: Commas should always have a space after."
<code>x[, , ]</code>  gives "style: Commas should never have a space before."

AshesITR:
Closing this for now because it diverged a lot from the main codebase. NB allowing this style requires adjustments to `spaces_inside_linter` and `commas_linter`:

```r
lintr::lint(text = "a[ , , ]")
#> <text>:1:3: style: Do not place spaces around code in parentheses or square brackets.
#> a[ , , ]
#>   ^
#> <text>:1:4: style: Commas should never have a space before.
#> a[ , , ]
#>   ~^
```

Both could be controlled via a flag `allow_empty_arg = FALSE` that would specifically allow `" ,"`.

Issue #171, opened 2016-09-09
Wrong arguments to httr::http_condition in github_comment


cdiener:
In `comments.R` line 79. `httr::http_condition` is passed a `message` argument, however the signature of `http_signature` in the latest httr (1.2.1) is

```
Usage:

     http_condition(x, type, task = NULL, call = sys.call(-1))
     
Arguments:

       x: a response, or numeric http code (or other object with
          ‘status_code’ method)

    type: type of condition to generate. Must be one of error, warning
          or message.

    task: The text of the message: either ‘NULL’ or a character vector.
          If non-‘NULL’, the error message will finish with "Failed to
          ‘task’".

    call: The call stored in the condition object.

```

So getting to that code point will raise an error due to the unused message argument.

PR #173, opened 2016-09-14
Correct call signature for http_condition.

cdiener:
This is a fix for #171. `httr::http_condition` was called with wrong arguments leading to an error.


jimhester:
Thanks!

cdiener:
Thanks to you as well :)

PR #177, opened 2016-10-17
Fix for syntastic error in VIM

FvD:
Syntastic expects a list of strings here. The current proposed settings leads to the error:

    syntastic: error: variable g:syntastic_r_checkers has to be a list of strings


jimhester:
Thanks!

PR #204, opened 2017-02-12
Add linter for no spaces before function calls

jrnold:
Apologies if I missed a linter that was already defined; I looked but I couldn't find it (althoug

This adds a new linter, `function_left_parentheses_linter`, which checks that for function calls there is no space before the parentheses. This is almost the inverse of the `spaces_left_parentheses_linter`, which checks for spaces before parentheses *except* when it is a function.

This only considers function calls and not after , brackets instead of parentheses, or after curly brackets, square brackets, or parentheses. I didn't know if they belonged in this or another linter.

This linter handles cases like `print (blah)` which should be `print(blah)`.


jimhester:
I think this looks good, could you add a note to NEWS.md?

jrnold:
I put the note under `1.0.0.9000`, but should I have created a new section `1.0.0.9001`?

jimhester:
Thanks!

Issue #205, opened 2017-02-14
expect_lint() escaping issue


fangly:
Consider a file "doublebackslash.R" that contains a single double-backslash string:
```
'\\\\'
```

And the following script:
```
verbose_assignment_linter <- function(source_file) {
  cat("file content is:")
  print(source_file$content)
  lintr::assignment_linter(source_file)
}

file <- "doublebackslash.R"
file_content <- "'\\\\'"

cat("\nMethod 1\n")
lintr::lint(file, linters=verbose_assignment_linter)

cat("\nMethod 2\n")
lintr:::expect_lint(content=NULL, checks=NULL, verbose_assignment_linter, file=file)

cat("\nMethod 3\n")
lintr:::expect_lint(content=file_content, checks=NULL, verbose_assignment_linter)
```

The result of running the script is:
```
Method 1
file content is:[1] "'\\\\\\\\'"
file content is:[1] "'\\\\\\\\'"

Method 2
file content is:[1] "'\\\\\\\\'"
file content is:[1] "'\\\\\\\\'" ""          

Method 3
file content is:[1] "'\\\\'"
file content is:[1] "'\\\\'"
```

I expected the same results for the three methods, but method 3 (`expect_lint` with string input) differed from the two others. I believe that `expect_lint` should be modified by escaping its string input.

Cheers,

Florent

fangly:
I do not have a proper fix for this issue, but a temporary workaround for the affected tests (those that contain some backslashes) is to call `encodeString`, e.g.
```
expect_lint(encodeString("'E:\\blah\\file.txt'"), msg, linter)
```
instead of 
```
expect_lint("'E:\\blah\\file.txt'", msg, linter)
```

PR #206, opened 2017-02-14
expect_lint escape fix

fangly:
When given a string, expect_lint needs to escape it. See #205


fangly:
This does not seem to work as intended. Closing

PR #202, opened 2017-02-02
Don't lint identifiers within library imports

mplough:
Fixes #201 for both camel case and snake case linters.

PR #207, opened 2017-02-20
Fix test error

jrnold:
Update the checkstyle.xml file because xml2 now explicitly sets the encoding of generated xml files. See the comment in PR #202.


jimhester:
This was included in https://github.com/jimhester/lintr/pull/208, so I am just going to close this. Thanks!

PR #208, opened 2017-02-20
Add Rstudio project

jrnold:
Sorry if this is an inappropriate/annoying request, since it is more programmer style. But can you add an Rproject file to this? It makes is easier for anyone working in Rstudio to contribute.


jimhester:
Yes this is fine, thanks!

PR #214, opened 2017-03-08
New linters and bug fixes

fangly:


R/undesirable_function_linter.R
```diff
@@ -0,0 +1,58 @@
...
+  "par"="use withr::with_par()",
+  "options"="use withr::with_options()",
+  "Sys.setenv"="use withr::with_envvar()",
+  "Sys.setlocale"="use withr::with_locale()",
+  ".libPaths"="use withr::with_libpaths()",
+  "attach"="use roxygen2's @importFrom statement in packages, or `::` in scripts",
+  "detach"="use roxygen2's @importFrom statement in packages, or `::` in scripts",
+  "library"= "use roxygen2's @importFrom statement in packages, or `::` in scripts",
+  "require"= "use roxygen2's @importFrom statement in packages, or `::` in scripts",
+  "loadNamespace"="use `::` or requireNamespace()",
```
jimhester: I do not think `loadNamespace` should be linted, I actually prefer using it in most cases, most of the time you use it for suggested packages and do want to throw an error if a package is not installed. See http://r-pkgs.had.co.nz/namespace.html#search-path for a discussion of the differences.

R/undesirable_function_linter.R
```diff
@@ -0,0 +1,58 @@
...
+  "source"=NA,
+  "sapply"="use vapply() or lapply()",
+  "mapply"="use Map()",
+  "ifelse"="use an if () {} else {} block",
```
jimhester: While it is true some (most?) of the time people use ifelse when it would be better to use `if()` there are cases where you do want to branch on a vectorized conditional, I think it is too aggressive to lint this.

R/undesirable_function_linter.R
```diff
@@ -0,0 +1,58 @@
...
+  "return"="let the last value of a function automatically be returned",
```
jimhester: `return()` is still very useful for early returns, you would need to only lint this at the end of a function body.

R/undesirable_function_linter.R
```diff
@@ -0,0 +1,58 @@
...
+  "substring"="use substr()"
```
jimhester: Hmm I think this is also too aggressive, substring recycles its first argument, which makes it useful for some things.

``` r
substring("abcdefghijklmnop", seq(1, 15, 5))
#> [1] "abcdefghijklmnop" "fghijklmnop"      "klmnop"
```

R/path_linters.R
```diff
@@ -0,0 +1,208 @@
...
+          TRUE  # either Unix path or strict (lax=FALSE)
+        }
+      },
+      split_path(str),
+      re_matches(str, rex(or(list(root_regex[["win32"]], root_regex[["unc"]], "\\"))))
+    )
+  )
+}
+
+is_long_path <- function(str) {
```
jimhester: I think `path` is a better name for this argument in all of these path functions.

R/object_name_linters.R
```diff
@@ -0,0 +1,193 @@
...
+    linter = type
+    )
+}
+
+
+#' @describeIn linters  Check that object names conform to a single naming style.
+#' @param style UpperCamelCase, lowerCamelCase, alllowercase, ALLUPPERCASE, snake_case or
+#'              dotted.case
+#' @export
+object_name_linter <- function(style = "lowerCamelCase") {
```
jimhester: The default needs to be `snake_case`.

R/comment_linters.R
```diff
@@ -0,0 +1,169 @@
...
+    regex <- rex(capture(name="match", one_or_more("#"), any_spaces, or(todo)))
+    code_locs <- integer()
+    function(source_file) {
+      file_lines <- source_file[["file_lines"]]
+      if (is.null(file_lines)) {
+        # code section: register where expression ends on the line
+        code_locs <<- register_code_locations(source_file, code_locs)
+      } else {
+        # non-code section: detect and report TODOs
+        if (length(file_lines) == 0L) {file_lines <- ""}  # workaround rex bug goo.gl/2rbZDy
```
jimhester: I would rather have a full link rather than short url, as the full link contains information that can be used to retrieve the issue even if the link breaks in the future, such as repo username and issue number.

R/comment_linters.R
```diff
@@ -0,0 +1,169 @@
...
+
+  !inherits(res, "try-error")
+}
+
+
+
+#' @describeIn linters  Check that the source contains no TODO comments (case-insensitive).
+#' @param todo  Vector of strings that identify TODO comments.
+#' @export
+todo_comment_linter <- function(todo=c("todo", "fixme")) {
```
jimhester: I feel like this function should just use the `getParseData()` and check for a `COMMENT` token that has text `^#[[:spaces:]]TODO`, why is all this complication needed?

fangly: The function is more complicated than anticipated because not all comments are reported within `source_file$parsed_content`.

```R
# TODO 1
x <- 3 # TODO 2
foo <- function(x) { # TODO 3
  # TODO 4
  x * 2 # TODO 5
}
y <- foo(x) # TODO 6
# TODO 7
```
In the example above, only TODO 3-5 are returned as tokens. In `semicolon_terminator_linter()`, there is a similar issue that is addressed by the `locate_other_sc` function.

R/undesirable_function_linter.R
fangly: If you are generally aiming to follow Hadley's style and guidelines, `loadNamespace` should be left in by default.

R/undesirable_function_linter.R
fangly: Ok, I'll remove it from the default. I will likely make it available a set of extended defaults for those that prefer to lint aggressively.

R/undesirable_function_linter.R
fangly: Sure, that just reflected my preference. I will remove if from the defaults. In the case of `return`, a separate linter would be preferable anyway,

R/undesirable_function_linter.R
fangly: No worries. Again, this only reflected my own personal biases.

R/path_linters.R
fangly: The three functions `is_valid_path()`, `is_long_path()` and `is_valid_long_path()` do indeed expect some sort of path as input. It makes sense to change the argument name as you suggested.

I am not so convinced for changing the argument name for the four functions `is_absolute_path()`, `is_root_path()`, `is_relative_path()`, `is_path()` because the input is a string and the functions determine if this is a path.

R/comment_linters.R
fangly: Yes, that is understandable.

jimhester:
> Note: there is no alias for camel_case_linter(), snake_case_linter() and multiple_dots_linter() because they identify names to remove, while object_name_linter() identifies names to keep.

Yes this is true, we will probably need to keep the old definitions around until these functions are completely deprecated and gone then.

R/path_linters.R
jimhester: Fine, `split_path()` arguments need to be `path` and `sep` though.

R/path_linters.R
```diff
@@ -0,0 +1,208 @@
...
+  )
+}
+
+
+unquote <- function(str, type=c("'", "\"", "`")) {
+  # Remove the outtermost quotes and unescape backslashes
+  for (q in type) {
+    str <- re_substitutes(str, rex(start, q, capture(zero_or_more(any)), q, end), "\\1")
+  }
+  stringi::stri_unescape_unicode(str)
```
jimhester: We cannot depend on stringi, you can look at base alternatives if you need this http://stackoverflow.com/a/25310392/2055486

R/path_linters.R
fangly: I thought we were splitting hairs, so I changed it to "path" anyway.
I will do the same for `split_path` as well.

fangly:
Hi Jim,

I think I have pretty much addressed all your comments now. Good input, thanks!

The only issue left is the stringi dependency. The problem is not to escape, but to **un**escape. I am not aware of anything that does this in base. Maybe you do?

Cheers,
Florent

fangly:
I have implemented an unescape() function that relies on base R's `gregexpr` and `regmatches()` instead of stringi. I think there are no more open issues, Jim.

jimhester:
Ok I think all that is left is updating the NEWS file and README explaining the changes / listing the new linters added.

fangly:
This is done. It is a minor commit, so, I merging this PR now.

jimhester:
Please let me merge major things like this.

fangly:
Ok, sorry for my impatience. Thank you, Jim.

PR #220, opened 2017-03-13
Fix case for line_length_linter message

mschilli87:
While other linters' messages start with upper case characters (as they are full sentences), `line_length_linter` prints a message starting with a lower case character.

This patch adjusts `line_length_linter` accordingly.


fangly:
Thanks. But I think you forgot to update the corresponding tests, which causes a Travis fail.

mschilli87:
Sorry, I did not realize.

Should be fixed now.

jimhester:
Thanks!

Issue #215, opened 2017-03-09
get_source_expressions() removes some tokens


fangly:
Linters like `todo_comment_linter` and `semicolon_terminator_linter()` (#214) are more complicated than could be because some tokens are omitted by `get_source_expressions()` (not available within `source_file$parsed_content`).

```R
# TODO 1
x <- 3 # TODO 2
foo <- function(x) { # TODO 3
  # TODO 4
  x * 2 # TODO 5
}
y <- foo(x) # TODO 6
# TODO 7
```
In this first example (above), only TODO 3-5 are returned as COMMENT tokens in `source_file$parsed_content`. 

```R
a <- 1; b <- 2;
foo <- function(a, b) {
  c <- a + b; cat(c);
}
```
In this second example, only the two semicolons within the function are returned as ';' tokens. The two first are missing.

Ideally, one would want to have all needed tokens exposed, and the overcomplicated linters simplified.

The variable `parsed_content` in `get_source_expression()` seems to contain all tokens, but later in the same function, variable `expressions` seems to have dropped some of them. This illustrates that the issue is likely in lintr.

jimhester:
I think the easiest solution to this is to include the full parsed content (and xml parsed content) in the global expressions https://github.com/thomaskisler/scripts/blob/68f286726141a34ce6ef4817a2f5c1227a3e9f24/ranger.plot.data.tree.R.

Alternatively I have thought about removing the caching / breaking into top level expressions, it complicates the implementation quite a bit and has corner cases for some linters that are I believe nearly impossible to handle.

jimhester:
Sorry forgot to copy the link properly, now fixed.

fangly:
Pull request was merged, so closing this issue.

PR #221, opened 2017-03-13
Add comment and semicolon tokens to source_file$parsed_content

fangly:
Hi Jim,
This is my attempt to fix the issue described at #215. The key change was a small modification of the `top_level_expressions()` function so that it returns also the ID of "COMMENT" and "';'" tokens. The rest is just adapting the linters to this change. All tests succeed.
What do you think?
Best,
Florent


R/get_source_expressions.R
```diff
@@ -309,12 +308,13 @@ next_with_parent <- function(pc, loc) {
...
   loc <- which(with_parent$id == id)
 
-  return(which(pc$id == with_parent$id[loc + 1L]))
+  which(pc$id == with_parent$id[loc + 1L])
 }
 
 top_level_expressions <- function(pc) {
   if (is.null(pc)) {
-    return(NULL)
+    character()
```
jimhester: This function is returning the output of `which()` from the other branch, which is a integer vector, so this should be `integer()`. You are correct the previous return value of `NULL` was incorrect.

R/get_source_expressions.R
```diff
@@ -309,12 +308,13 @@ next_with_parent <- function(pc, loc) {
...
+  } else {
```
jimhester: I would prefer not to have an else here

R/utils.R
```diff
@@ -76,14 +77,12 @@ quoted_blanks <- function(matches, shift_start = 0, shift_end = 0) {
...
-  loc <- which(fun(source_file$parsed_content$token, value))
-  if (loc %==% integer(0)) {
-    NULL
-  } else {
+  pc <- source_file$parsed_content
+  if (!pc$col1 %==% integer(0) &&
+      !(loc <- which(fun(pc$token, value))) %==% integer(0) ) {
     loc
+  } else {
+    character()
```
jimhester: Again this should be `integer()`

R/utils.R
```diff
@@ -76,14 +77,12 @@ quoted_blanks <- function(matches, shift_start = 0, shift_end = 0) {
...
-  if (source_file$parsed_content$col1 %==% integer(0)) {
-    return(NULL)
-  }
-  loc <- which(fun(source_file$parsed_content$token, value))
-  if (loc %==% integer(0)) {
-    NULL
-  } else {
+  pc <- source_file$parsed_content
+  if (!pc$col1 %==% integer(0) &&
+      !(loc <- which(fun(pc$token, value))) %==% integer(0) ) {
```
jimhester: Please don't use assignments within conditionals

fangly:
Jim, good point about the ids being integers. I always think of them as strings. I've made the appropriate changes.

jimhester:
Thanks!

PR #222, opened 2017-03-13
Pipe continuation linter

### LINKED ISSUE #216: Add pipe linting
i.e. ensure space before `%>%` and newline afterwards (unless the whole pipe fits on a single line)

jimhester:
Fixes #216 

@fangly This PR illustrates using the XML Xpath logic to write a linter, ideally we would move towards this approach for more linters in the future.

I am also not sure this is the best name for the linter, or if the message is clear, if you have any suggestions there I would appreciate them!


R/pipe_continuation_linter.R
```diff
@@ -0,0 +1,41 @@
+#' @describeIn linters check that all commas are followed by spaces, but do not
+#' have spaces before them.
+#' @importFrom xml2 xml_find_all as_list
+#' @export
+pipe_continuation_linter <- function(source_file) {
```
fangly: I think `%>%` is one operator amongst many. It would make sense to lint it in `infix_spaces_linter()`. In this instance, space is to be interpreted loosely as "\s" or "\n".

R/pipe_continuation_linter.R
```diff
@@ -0,0 +1,41 @@
...
+    return()
+  }
+
+  # Find all expressions that span more than one line
+  # and contain specials named '%>%' that are on the same line
+
+  expressions_spanning_multiple_lines <- "expr[@line1 != @line2]"
+  pipes_on_same_line <- "SPECIAL[text() = '%>%' and @line1 = following::SPECIAL[text() = '%>%']/@line1]"
+
+  exprs <- xml_find_all(x,
```
fangly: Not being familiar with Xpath, these statements look obscure, but powerful! Maybe in some future, all linters can reduce to a list of Xpath statements with an accompanying message.

R/pipe_continuation_linter.R
```diff
@@ -0,0 +1,41 @@
...
+  lapply(exprs,
+    function(expr) {
+      x <- as_list(expr)
+      line <- get_file_line(source_file, x@line1)
+      Lint(
+        filename = source_file$filename,
+        line_number = x@line1,
+        column_number = x@col2,
+        type = "style",
+        message = "Each `%>%` statement should be on it's own line.",
```
fangly: Perhaps two different messages make sense here?
"`%>%` should be preceded by a space.",
"`%>%` should be followed by a space (if the pipe fits on a single line) or a newline (otherwise)."

R/pipe_continuation_linter.R
jimhester: This text is actually wrong, it is residual from the commas linter. This linter does nothing with spaces, just determines if a pipe is either all on one line or spread across multiple lines with one action per line.

R/pipe_continuation_linter.R
jimhester: The [tidyverse style guide](http://style.tidyverse.org/pipes.html#spacing-and-indenting) uses

> %>% should always have a space before it and a new line after it

Maybe we can add `, unless the full pipeline fits on one line.`

R/pipe_continuation_linter.R
jimhester: We do include specials in the `infix_spaces_linter`, and this seems to be working fine with pipes, so I think the space aspect is covered.

Issue #212, opened 2017-02-24
Ambiguous files in cache cause lint crash


fangly:
Hi Jim,

I obtained a "Marker line is unspecified or invalidNA" error when using lintr with caching. I managed to find out what is causing this problem. It is reproducible using these steps:
1. Create two packages, "A" and "B", that have the identically-named files "tests/testthat.R". Let us assume that "A/tests/testthat.R" and "B/tests/testthat.R" are identical, except that "A/tests/testthat.R" has some trailing newlines.
2. In package A, run `expect_lint_free(cache=TRUE)`. This runs fine, yields a few lints because of the trailing newlines, and generates a "~/.R/lintr_cache/testthat.R" file.
3. In package B, run `expect_lint_free(cache=TRUE)`. This operation crashes with "Marker line is unspecified or invalidNA". I imagine this is because lintr sees the "~/.R/lintr_cache/testthat.R" file and wrongly assumes that it belongs to package B. Since there are some lints at line numbers after last line of file "B/tests/testthat.R", mayhem ensues...

My conclusion is that the cache should differentiate files belonging to different packages. This is important because many packages use the same names, e.g. "testthat.R"  or "zzz.R".

But the problem is probably not limited to identically-named files between packages. When I create a file "A/R/more/testthat.R", issue a `clear_cache()` and re-run `expect_lint_free(cache=TRUE)`, there is a single "~/.R/lintr_cache/testthat.R" file, meaning that the two "testthat.R" files in package A were likely cached in the same location, one of them overwriting the other.

A straightforward solution would be to create a subfolder for each package, and further subfolders mimicking the folder structure of the respective packages, i.e. 

Source file | Cache file
------------ | -------------
A/tests/testthat.R | ~/.R/lintr_cache/A/tests/testthat.R
A/R/more/testthat.R | ~/.R/lintr_cache/A/R/more/testthat.R
B/tests/testthat.R | ~/.R/lintr_cache/B/tests/testthat.R

Cheers,
Florent

russHyde:
Fixed in https://github.com/jimhester/lintr/pull/232 . Closing

PR #232, opened 2017-04-27
Fix crash due to ambiguous cache file paths

fangly:
Here is a fix for #212. It takes the approach described in the same ticket, i.e. give different packages a different cache directory, give each file its path relative to the package. This ensures that no two R source files end up having the same cache location.

One thing to note is that `find_package()` and `pkg_name()` are called frequently in this fix. I am not sure of the performance penalty (e.g. `pkg_name()` needs to read the DESCRIPTION file each time). It may be worth somehow saving the package name and location once they have been determined.


jimhester:
This solution is too complicated, you have the package root and the file path relative to the package root, you just need to append them and use that as the cache path.

Also I see no reason to rename the argument, we need to maximize backwards compatibility unless there is substantive improvement.

fangly:
Ok, I renamed the argument in the two functions, though I do think that its name is not precise enough and its use particularly confusing in `clear_cache(file, path)`.

I partially disagree with your statement about my solution being to complicated:

1. As far as I can tell, there is no guarantee that the filenames passed to `lint()` are relative paths. With `lint_package(".", relative_path = TRUE)`, I get absolute paths (on the master branch) `lint_package()` calls `normalizePath(files)`, which is the correct thing to do in my opinion. Calling `lint()` directly, one may get relative paths, but they are relative to the current directory and not necessarily the package directory.

2. An alternative solution (that could be considered simpler) is to create a hash on the absolute file path.

fangly:
Hi Jim,
I did implement the alternative solution mentionned above. Please do have a look at the code again. 
Going one little step farther, one could also merge `get_cache_file_path()` into `digest_content()`. Note that the last commit is not really related to caching.
Cheers,
Florent

R/lint.R
```diff
@@ -132,16 +133,19 @@ reorder_lints <- function(lints) {
...
 #' @export
 lint_package <- function(path = ".", relative_path = TRUE, ...) {
   path <- find_package(path)
 
   read_settings(path)
   on.exit(clear_settings, add = TRUE)
 
-  names(settings$exclusions) <- normalizePath(file.path(path, names(settings$exclusions)))
-  exclusions = force(settings$exclusions)
+  args <- list(...)
```
jimhester: We should add `exclusions` to the argument list of `lint_package()` after the `...` and `#' @includeParams lint`, then modify the exclusions as needed and just use a standard call to `lint()` as before.

fangly:
Very well, this is now done.

fangly:
Please, remember to merge this branch, Jim.

jimhester:
Thanks again for working on this!

PR #239, opened 2017-05-11
Unused import linter

jimhester:
This still needs some more testing, and tests written, but works with limited inputs.


AshesITR:
This looks quite useful. Should we try to get it up to speed and add tests?

If so, for packages we should check for functions not used but imported in `NAMESPACE` instead of loaded via `require` / `library`.

jimhester:
Yeah, I think that is right.

AshesITR:
Rebased onto current HEAD as a first step.

AshesITR:
@jimhester I think we have a problem for packages:
Without scanning all source files, we can never know if an import is truly _never_ used.

WDYT about keeping it restricted to `library()` and `require()` calls for the moment?

jimhester:
yeah that is true, we would have to scan for all source in the package. `R CMD check` already does this for packages, so I think it is fine to restrict this only to scripts.

AshesITR:
NB: `bit64` and `data.table` should also be exempt by default -- both cause other code to behave differently if loaded.

AshesITR:
@MichaelChirico PTAL I've updated this PR and rebased onto current master.

R/unused_import_linter.R
```diff
@@ -0,0 +1,60 @@
...
+      logical(1L)
+    )
+
+    lapply(
+      import_exprs[is_unused],
+      xml_nodes_to_lint,
+      source_file = source_file,
+      lint_message = function(import_expr) {
+        pkg <- strip_names(xml2::xml_text(xml2::xml_find_first(import_expr, "expr[STR_CONST|SYMBOL]")))
+        paste0("package '", pkg, "' is attached but never used.")
```
AshesITR, line 54: We could add `"If you want to state a package as necessary for running the code, use requireNamespace() instead."` to indicate that the `pkg::fun()` usages being ignored is not a bug. WDYT?

R/unused_import_linter.R
```diff
@@ -0,0 +1,60 @@
...
+        "expr[SYMBOL_FUNCTION_CALL[text() = 'library' or text() = 'require']]",
+        "and",
+        "not(SYMBOL_SUB[text() = 'character.only'])",
+        "]"
+      )
+    )
+    if (length(import_exprs) == 0) {
+      return(list())
+    }
+    # strip_names() needed to turn "'pkg'" -> "pkg" for STR_CONST calls.
```
MichaelChirico, line 26: I don't think we want `strip_names()` here, right? e.g. for `R"()"` strings

AshesITR: Yes, it's a quick-and-dirty solution. Should be revisited when we unify the string literal treatment.
Do we already have an issue for that? Then I'll mark it as a TODO for that issue.

MichaelChirico: true, but I had in mind consistency with missing_package_linter, which is in turn consistent with what library itself does.

MichaelChirico: #1034 is the issue btw

AshesITR: ok, will change to that. This approach also copes with raw strings, so no follow-up todo 🥳

inst/lintr/linters.csv
```diff
@@ -71,5 +71,6 @@ undesirable_function_linter,style efficiency configurable robustness best_practi
 undesirable_operator_linter,style efficiency configurable robustness best_practices
 unneeded_concatenation_linter,style readability efficiency
 unreachable_code_linter,best_practices readability
+unused_import_linter,best_practices correctness
```
MichaelChirico, line 74: i'm not sure `correctness` is a great tag, since there's nothing "incorrect" about having `library()` for a package you don't (or no longer) use, right?

MichaelChirico: maybe `robustness` (because your script will run in more environments with streamlined dependencies) and/or `common_mistakes`? should we have a script-oriented counterpart to `package_development` like `scripting`?

R/unused_import_linter.R
```diff
@@ -0,0 +1,64 @@
...
+  Linter(function(source_file) {
+    if (is.null(source_file$full_xml_parsed_content)) return(list())
+
+    import_exprs <- xml2::xml_find_all(
+      source_file$full_xml_parsed_content,
+      paste(
+        "//expr[",
+        "expr[SYMBOL_FUNCTION_CALL[text() = 'library' or text() = 'require']]",
+        "and",
+        "not(SYMBOL_SUB[text() = 'character.only'])",
```
MichaelChirico, line 19: this will also skip `library('dplyr', character.only = TRUE)`, right?

R/unused_import_linter.R
```diff
@@ -0,0 +1,64 @@
...
+        "]"
+      )
+    )
+    if (length(import_exprs) == 0) {
+      return(list())
+    }
+    imported_pkgs <- xml2::xml_text(xml2::xml_find_first(import_exprs, "expr[STR_CONST|SYMBOL]"))
+    imported_pkgs <- vapply(
+      imported_pkgs,
+      function(expr) as.character(parse(text = expr, keep.source = FALSE)),
```
MichaelChirico, line 29: can we vectorize this so we only have to run `parse()` once?

tests/testthat/test-unused_import_linter.R
```diff
@@ -0,0 +1,23 @@
...
+  expect_lint("library(not.a.package)\ntibble(a = 1)", NULL, linter)
+  # SYMBOL calls with character.only = TRUE are ignored, even if the argument is a package name
+  expect_lint("library(dplyr, character.only = TRUE)\n1 + 1", NULL, linter)
+
+  msg <- rex::rex("package 'dplyr' is attached but never used")
+
+  expect_lint("library(dplyr)\n1 + 1", msg, linter)
+  expect_lint("require(dplyr)\n1 + 1", msg, linter)
+  expect_lint("library('dplyr')\n1 + 1", msg, linter)
+  # ignore namespaced usages
```
MichaelChirico, line 21: Don't think I agree with this behavior. Some places (including at google, for our own peculiar reasons) encourage ns-qualified usage everywhere, for example.

At a minimum, let's have it set by an option?

R/unused_import_linter.R
MichaelChirico: Just saw this again... redirecting to the other comment thread: https://github.com/r-lib/lintr/pull/239#discussion_r842300096

MichaelChirico:
What about packages that are used to load methods? I can't think of any example off-hand, but I'm imagining `library(pkg)` is used to make methods like `print()` or `[` available. I guess it should be rare enough that users can just supply this to the `exception` argument?

tests/testthat/test-unused_import_linter.R
MichaelChirico: See also #453

AshesITR:
That's exactly why the `bit64` and `data.table` packages are excluded by default. LMK if there are more packages that need such an exclusion.

inst/lintr/linters.csv
AshesITR: My reason for correctness is the dev probably intended to use the package so maybe forgot a piece of code that acutally uses the package. But open to changing this to `common_mistakes`. Robustness is intended for "Linters highlighting code robustness issues, such as possibly wrong edge case behaviour.", not really fitting here IMO.

Regarding scripting: Might be useful. Let's do this in a separate PR.

R/unused_import_linter.R
AshesITR: Yes, it would. Do you suggest adding an `or *[1][self::STR_CONST]` condition to this exception?

R/unused_import_linter.R
AshesITR: Interesting enough, `as.character(parse(text = imported_pkgs, keep.source = FALSE))` seems to work in this context.
I interactively tested curious edge cases (extremely long strings, and the empty string).

tests/testthat/test-unused_import_linter.R
AshesITR: Are the lines improperly referenced here?
IIUC you want a new argument `allow_ns_usage = FALSE`?

IMO if you require a package and only use it by namespace, you shouldn't attach it, i.e. use `requireNamespace("pkg")` instead of `library(pkg)`.

AshesITR:
#1090 should fix the build issue.

inst/lintr/linters.csv
MichaelChirico: for `robustness`, I had in mind "this script will run in more places if it excludes unneeded `library()` calls", but agree it's a stretch.

I think `common_mistakes` is better -- the other situation is when the package _was_ used in the past, but that code's now deleted / migrated elsewhere, so this script can drop the requirement. I think `common_mistakes` covers the two cases better than `correctness`.

R/unused_import_linter.R
MichaelChirico: yes... similar to `missing_package_linter()`

tests/testthat/test-unused_import_linter.R
MichaelChirico: > IMO if you require a package and only use it by namespace, you shouldn't attach it, i.e. use requireNamespace("pkg") instead of library(pkg).

I see your point but a lot of people apparently disagree. The philosophy is "always ns-qualify non-base usage"; `library()` calls are added at the top to make the dependencies of the script more apparent (if `ns::` is only used once and on line 458, it can be hard to tell without a thorough scouring of the code that `ns` is a requirement -- especially if `ns::` is in a branch of the code that isn't necessarily executed except under certain parameterizations/conditions). This is very similar to C++.

Moreover, `library()` might be run at the top to induce intended side effects of `.onAttach()`. We can't tell statically if removing `library(foo)` will break the script because `.onAttach()` isn't run.

So overall, if we want to throw a lint in scripts with `library(ns)` where `ns` is only used in `ns::` form, that should only be an option that's off by default, IMO.

AshesITR: I think `library()` is more like `using namespace ...`, which should be used sparingly and substituted by a non-invasive `requireNamespace()`.
Otherwise, the R search path is unnecessarily polluted, which can cause name clash issues.

https://towardsdatascience.com/attach-packages-mindfully-in-r-b9c8151b3fb4

Regarding side-effects, that's why the package exceptions are provisioned. Maybe we need more by default?

I'd be happier with a default of `FALSE` for this reason, but willing to concede the default of `allow_ns_usage` if you really think it's better for most users.

MichaelChirico: > the R search path is unnecessarily polluted, which can cause name clash issues

hence always using `::`

> really think it's better for most users

That we've gotten #453 and #459 already tells me `TRUE` is the more common case... I just put it to a poll here:

https://twitter.com/michael_chirico/status/1519187094341517312

Hopefully we can get a decent # of responses. Let me know if the poll wording sounds fair.

AshesITR: I think many people don't know about `requireNamespace()`.
Maybe which is better when referring by ns::, library or requireNamespace?

MichaelChirico: `requireNamespace()` in particular got the most pushback:

https://twitter.com/michael_chirico/status/1519197746774769666

Overall using `library()` got the most support, but not overwhelmingly so (roughly 53-47 as of now): https://twitter.com/michael_chirico/status/1519187094341517312

AshesITR: Having read the comments on twitter I also see (the same) arguments against library(). The critique against `requireNamespace()` was that it doesn't stop, which I somehow though it would 😅
Seems `loadNamespace()`, which does stop for missing packages (?), would have been the correct replacement.

So I'm still convinced `FALSE` should be default.

MichaelChirico: That sounds like a good compromise! I would still leave it as an option but can accept a `FALSE` default.

renkun-ken: `requireNamespace()` does stop. `requireNamespace(quitely = TRUE)` does not stop.

AshesITR:
Ready for another round of review.

MichaelChirico (dismissed):
Remind me why `data.table` is excluded? maybe a simple example of a script that _needs_ library(data.table) but doesn't include any data.table exports. Can't think of one right now.

MichaelChirico:
Looks like my added test isn't caught, please check the test is right & adjust the `character.only` xpath if so

AshesITR:
> Remind me why `data.table` is excluded? maybe a simple example of a script that _needs_ library(data.table) but doesn't include any data.table exports. Can't think of one right now.

```
dt <- readRDS("a/data.table.RDS")
dt[col == 1]
```

It's about the `[.data.table` generic.

MichaelChirico:
that doesn't trigger a lint though:

```
lintr::lint("dt <- readRDR('a/data.table.RDS')\ndt[col == 1]", lintr::unused_import_linter())
```

AshesITR:
> that doesn't trigger a lint though:
> 
> 
> 
> ```
> 
> lintr::lint("dt <- readRDR('a/data.table.RDS')\ndt[col == 1]", lintr::unused_import_linter())
> 
> ```

Because it's a default?
Does `unused_import_linter(except_packages = character())`?

MichaelChirico:
OK, nvm. understood. Technically, that example is not the best because users _should_ be running `setDT()` on the `readRDS` object:

```
Rscript -e "saveRDS(data.table::as.data.table(iris), 'tmp')"
Rscript -e "data.table::truelength(readRDS('tmp'))"
# [1] 0
rm tmp
```

Maybe `data.table` should be dropped as an exclusion? It's kind of a code smell, but then again that's not the purpose of this particular linter at all.

AshesITR:
Well we actually have this in internal ETL code where we know for sure the RDS contains a data.table because it's written by an adjacent file.

MichaelChirico:
> Well we actually have this in internal ETL code where we know for sure the RDS contains a data.table because it's written by an adjacent file.

IINM without `setDT()`, we can load the table but it won't be a "properly" functioning data.table, e.g. we can't add columns:

https://stackoverflow.com/q/28078640/3576984

Anyway, besides the point of this PR, and I'm OK with the default exceptions list. Last thing to fix is the added test for `character.only` & we're good to go

AshesITR:
> > Well we actually have this in internal ETL code where we know for sure the RDS contains a data.table because it's written by an adjacent file.
> 
> IINM without `setDT()`, we can load the table but it won't be a "properly" functioning data.table, e.g. we can't add columns:
> 
> https://stackoverflow.com/q/28078640/3576984

good to know that readRDS can mess things up - we mostly use feather locally, getting the data tables using `fst::read_fst(..., as.data.table = TRUE)` so that would be a more valid example.

MichaelChirico (approved):

PR #245, opened 2017-05-18
User experience improvements

fangly:
Hi Jim,

Please have a look at the code in this branch. My main objective was to ensure a consistence interface to the different linters because the situation is currently confusing; some linters have to be called with an argument while other must just be identified by name. My proposed solution is to have all linters written so that they need to be called.

Before:
  * `semicolon_terminator_linter(semicolon = "compound")`
  * `T_and_F_symbol_linter`

After:
  * `semicolon_terminator_linter(semicolon = "compound")`
  * `T_and_F_symbol_linter()  # takes no argument`

An advantage of this change is that we can later decide to add optional arguments to any linter without breaking compatibility with prior versions of lintr. So, we introduce an incompatible change now for better user experience and will be able to introduce fewer incompatible changes in the future.

I also gave all the linters the "linter" class and check in `lint()` that the 'linters' argument only contains objects of class 'linter'. This is to warn users early when they use the 'linters' argument wrongly and prevent them from encountering obscure error messages.

To support these changes, I have added two deprecation messages and work-arounds in `lint()`. The users are thus informed of the changes but the code should continue to work as before.

Thoughts on the idea or implementation? Cheers,

Florent


fangly:
Jim, I'd be interested to read what you think of this branch.
Cheers,
Flroent

jimhester:
I think the ideas here are good, unfortunately it is now pretty far diverged from the current master :(

Would be great if @fangly or someone else would be able to update the code in this PR to merge cleanly with the current lintr codebase.

AshesITR:
By now, this has diverged even more.
IMO we should add a compatibility layer for old `.lintr` files to automagically call the linter functions if they have no arguments and issue a warning.

Let me know if you're still interested in this change and I'll try to get the PR going again.

AshesITR:
This changeset is absolutely huge.

I merged all R/ files by hand, ported newer linters added since 2017 to the `Linter()` interface and then fixed all tests that started failing.

I also updated the deprecation note @fangly added to 2.0.1.9001.

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# lintr 1.0.0.9001 #
+* Consistent access to linters through a function call, even for linters without
+  parameters (@fangly)
+* Fix line number sometimes wrongly reported by no_tab_linter() (#134, @fangly)
+* Fix line and column number sometimes wrongly reported by spaces_inside_linter()
+  (#203, @fangly)
+* Add `pipe_continuation_linter()` (#216).
+* Deprecated camel_case_linter(), snake_case_linter() and multiple_dots_linter()
+  in favor of object_name_linter() which enforce the given style: snake_case,
```
AshesITR, line 9: This obviously needs cleaning up.

MichaelChirico:
> This changeset is absolutely huge.

FWIW highly recommend viewing the diff with whitespace diffs off in the U. A lot more manageable:

https://github.com/jimhester/lintr/pull/245/files?diff=unified&w=1

R/assignment_linter.R
```diff
@@ -1,17 +1,19 @@
...
+    lapply(ids_with_token(source_file, "EQ_ASSIGN"),
+      function(id) {
+        parsed <- with_id(source_file, id)
+        Lint(
+          filename = source_file$filename,
+          line_number = parsed$line1,
+          column_number = parsed$col1,
+          type = "style",
+          message = "Use <-, not =, for assignment.",
+          line = source_file$lines[[as.character(parsed$line1)]],
```
MichaelChirico, line 14: most linters are using `[` not `[[` right?

R/lint.R
```diff
@@ -63,6 +63,32 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
     names(linters) <- auto_names(linters)
   }
 
+  linters <- Map(
+    function(obj, name) {
+      if (inherits(obj, "linter")) {
+      } else if (inherits(obj, "function")) {
```
MichaelChirico, line 69: I think `is.function` is more appropriate:

```
x = sum
class(x) = 'foo'
inherits(x, 'function')
# [1] FALSE
is.function(x)
# [1] TRUE
```

R/assignment_linter.R
AshesITR: I think so, yes. We should change it to `[` here as well then.

R/utils.R
```diff
@@ -216,3 +216,11 @@ set_lang <- function(new_lang) {
 reset_lang <- function(old_lang) {
   if (is.na(old_lang)) Sys.unsetenv("LANGUAGE") else Sys.setenv(LANGUAGE = old_lang)
 }
+
+#' Create a \code{linter} closure
+#' @param fun A function that takes a source file and returns \code{lint} objects.
+#' @return The same function with its class set to 'linter'.
+#' @export
+Linter <- function(fun) { # nolint: object_name_linter.
```
MichaelChirico, line 224: Since we're exporting, I'd add `stopifnot(is.function(fun))` here too

vignettes/creating_linters.Rmd
```diff
@@ -29,51 +30,54 @@ assignment_linter <- function(source_file) {
         linter = "assignment_linter"
         )
     })
+})
 }
 ```
 
-Lets walk through the parts of the linter individually.
+Let us walk through the parts of the linter individually.
```
MichaelChirico, line 37: That feels too formal, `Let's` will be fine IMO

MichaelChirico (changes requested):
I agree it's good to be more consistent about forcing `()` on the calls.

Mostly minor changes, I'll have another look at the NEWS file once that's cleaned up -- we'll definitely want to surface this high as a breaking change.

And with that, changing to 3.0.0 is definitely warranted (wasn't 100% sure before).

AshesITR:
Yep. NB the compatibility layer issues a warning for the old style, so at least lintr configurations and core linting functions shouldn't break.

R/lint.R
```diff
@@ -63,6 +63,32 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
...
+        if (is.null(formals(obj))) {
+          old <- "Passing linters as variables"
+          new <- "a call to the linters (see ?linters)"
+          lintr_deprecated(old = old, new = new, version = "2.0.1.9001",
+                           type = "")
+          obj <- obj()
+        } else {
```
AshesITR, line 76: Should we be stricter here and check that `formals(obj)` has exactly one argument, otherwise stop?

R/lint.R
```diff
@@ -63,6 +63,32 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
     names(linters) <- auto_names(linters)
   }
 
+  linters <- Map(
+    function(obj, name) {
+      if (inherits(obj, "linter")) {
+      } else if (inherits(obj, "function")) {
+        if (is.null(formals(obj))) {
```
AshesITR, line 70: A word of caution with builtins (not really important here IMO), but
```
> formals(sin)
NULL
```
Safer way:
```
> formals(args(sin))
$x

```

R/utils.R
AshesITR: Adding that and that `fun` must have exactly one argument.
We can re-use that in `lint.R` for the function case (`obj <- Linter(obj)` will then perform the check over there)

vignettes/creating_linters.Rmd
AshesITR: Feel free to roll back.

R/lint.R
AshesITR: Testing this now via `obj <- Linter(obj)`

R/utils.R
```diff
@@ -216,3 +216,14 @@ set_lang <- function(new_lang) {
...
+  if (!is.function(fun) || length(formals(args(fun))) != 1L) {
+    stop("`fun` must be a function of one argument.", call. = FALSE)
```
AshesITR, line 226: WDYT about the error message? I first thought of "`fun` must be a monadic function." but I'm not sure this is universally understood.

MichaelChirico: yea i wouldn't use such a CS term :smile: 

I tweaked it a bit, think it's fine now

MichaelChirico (dismissed):

AshesITR (approved):

PR #246, opened 2017-05-24
Recognize .R scripts that contain Roxygen inline expressions

WilDoane:
knitr::detect_pattern() incorrectly identifies .R scripts that contain Roxygen inline expressions as Markdown documents, causing a fatal error.

The following should lint 2 results (long var name and use of =), but errors due to either \`r\` expression.

```
#' ---
#' date: "`r Sys.Date()`"
#' ---

an_extremely_long_variable_name = Sys.Date()

#' Today is `r Sys.Date()` according to R.

```


WilDoane:
I want to find a better solution that will check both the top-level R script and the embedded Roxygen inline code.

PR #250, opened 2017-06-21
Nonlocal variables

fangly:
Hi Jim,
Here is a new linter (that uses eval and xpath), inspired by https://stackoverflow.com/questions/6216968/r-force-local-scope
Comments?
Best,
Florent


jimhester:
I think the idea is fine, but we are already doing this check with `codetools::checkUsage()` in the object usage linter https://github.com/jimhester/lintr/blob/master/R/object_usage_linter.R#L105.

If we want to do some of these checks without the need to evaluate R code, that would be a nice improvement.

fangly:
Didn't you have an idea on how to avoid eval?

Yes, there is code similar between object_usage_linter and nonlocal_variable_linter. Thinking of it, it would make sense to integrate `nonlocal_variable_linter()` into `object_usage_linter()`. I may try to do it when I have time, giving `object_usage_linter()` an argument to select which uses to report: unused variables, global variables, non local variables, etc.

AshesITR:
Closing this for now. Feel free to re-open a PR if you manage to remove the need for code evaluation from `object_usage_linter()` for this use-case.

PR #262, opened 2017-09-01
Fix #260

### LINKED ISSUE #260: Attenuate the verbosity in lint_package()
In `lint_package()`, it would be nice to have an option to condense the output. Rather than print out every lint, sometimes I may want to print the number of lints per file and omit all the clean files. I am picturing something like `print(covr::package_coverage())`.

wlandau:
- Add the `summary.lints()` function. Taken from @jimhester in #260 and modified. I have a mild preference to 
    1. sort by file name and
    1. make sure a predictable 4-column data frame is returned whether lints are found or not.
- Test `summary.lints()` on tidy and untidy code.
- Fix `tests/testthat/checkstyle.xml` (version given must agree with `lintr` version).
- Remove trailing whitespace from `test-methods.R`.


wlandau:
Sorry about the missing coverage report. I manually made sure all of `summary.lints()` was covered, so it should have only increased. Is there any action I can take on my end to fix this?

jimhester:
You can ignore the coverage report issues.

R/methods.R
```diff
@@ -141,3 +141,28 @@ as.data.frame.lints <- function(x, row.names = NULL, optional = FALSE, ...) {
...
   x
 }
+
+#' @export
+summary.lints <- function(object, ...) {
+  filenames <- vapply(object, `[[`, character(1), "filename")
+  types <- factor(vapply(object, `[[`, character(1), "type"),
+    levels = c("style", "warning", "error"))
+  tbl <- table(filenames, types)
+  filenames <- rownames(tbl)
```
jimhester: change this line to
```r
filenames <- rownames(tbl) %||% character()
```

Then you can remove the entire conditional at https://github.com/jimhester/lintr/pull/262/files#diff-a0a9d3c78e9f113319e5e069fb2697ecR155

tests/testthat/checkstyle.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<checkstyle version="lintr-1.0.0.9001">
+<checkstyle version="lintr-1.0.1">
```
jimhester: Try to keep unrelated changes out of PRs. If you do include them they should be in separate commits.

R/methods.R
wlandau: Now fixed.

tests/testthat/checkstyle.xml
wlandau: Sorry about that. And given that you just bumped the version, I think I also created a merge conflict. Should I go ahead and resolve it?

tests/testthat/test-methods.R
```diff
@@ -58,9 +58,35 @@ test_that("as.data.frame.lints", {
...
+    }"
+  bad <- "if(1>0){
+      print  ('bad')
+    }
+  "
+  file <- tempfile()
+  on.exit(unlink(file))
+
+  writeLines(good, con = file, sep = "\n")
+  expect_silent(out <- summary(lint(file)))
```
jimhester: I think this test would be more clear if you did something like

```r
no_lints <- lint(
"if (1 > 0){
   print(\"good\")
}")
no_lint_summary <- summary(no_lints)
# ... tests

has_lints <- lint(
"if(1>0){
  print  ('bad')
}")

has_lint_summary <- summary(has_lints)
# ... tests
```

Also I would explicitly pass linters to `lint()` rather than relying on the default ones. The defaults can change over time so they could potentially break this test if new linters are added.

wlandau: Good call. Just cleaned this up.

jimhester:
Ok, all this needs is a note in the NEWS.md file explaining the new feature and thanking yourself and this PR.

wlandau:
Done, thanks!

jimhester:
Thanks!

PR #264, opened 2017-09-05
Adds RStudio addins

JhossePaul:
This commit adds two addins to RStudio to bind shortcuts for easier usage. Now you can, for example, bind
Alt+Shift+L and lint the current source code.

### DIFF
--- a/R/addins.R
+++ b/R/addins.R
@@ -3,7 +3,18 @@ addin_lint <- function() {
   if (filename$path == "") {
     return("Current source has no path. Please save before continue")
   }
-  lintr::lint(filename$path)
+
+  config_file <- lintr:::find_config(filename$path)
+  config <- read.dcf(config_file, all = T)
+  config_linters <- gsub("\n", "", config[["linters"]])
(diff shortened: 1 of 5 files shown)


R/addins.R
```diff
@@ -0,0 +1,8 @@
+addin_lint <- function () {
+  filename <- rstudioapi::getActiveDocumentContext()
+  lintr::lint(filename$path)
+}
+
+addin_lint_package <- function () {
+  lintr::lint_package()
```
jimhester: Probably would be safer to retrieve the project directory (`rstudioapi::getActiveProject()`) and pass that to `lint_package()`?

jimhester:
Thanks this is a good idea! I made a minor comment, could you also add a note to `NEWS.md` with this change and mentioning your GitHub handle and this PR number?

Thanks!

R/addins.R
JhossePaul: Yep. Indeed I was thinking I completely forgot defensive programming. I'll add this and some more checks and then update the pull request

jimhester:
Great thanks!

PR #274, opened 2017-10-21
fix debian control fields link

rafazaya:
First PR so let me know if i'm doing it wrong - simply fixing a link in the README to the Debian control fields docs.


jimhester:
Perfect! Thanks!

PR #291, opened 2017-12-21
For fully ignored files, remove from `lint_package` prior to any further processing.

mwaldstein:
For my project, I have a lot of httptest cache files that should not be linted. While I can add them to the exclusions, the current implementation reads in the files before excluding, causing linting to take 30+ min.

This change removes fully excluded files as early as possible to avoid reading them in.


jimhester:
LGTM, can you please add a note to NEWS.md with the change and mentioning yourself, then I can merge this.

mwaldstein:
Will do.

mwaldstein:
Updated

jimhester:
Thanks!

PR #298, opened 2018-03-07
fix: Allow for (( when linting. Fixes #259

### LINKED ISSUE #259: "((" is a left paren space violation?
Currently this is a violation of the `spaces_left_parentheses_linter`:

```R
if ((var == "") | (other_var == "")) {
```

I think `((` should be legal. Is there a reason it isn't?

If this is a mistake, the solution would be to amend https://github.com/jimhester/lintr/blob/master/R/spaces_left_parentheses_linter.R#L27 to include `(`, right?

nathaneastwood:
There were failing tests beforehand. These are unrelated.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -85,6 +85,7 @@
 * Permit single quotes if they quote literal double quotes (#28, @jackwasey)
 * # nolint comments are respected with caching (#68, @krlmlr)
 * Properly handle all knitr document formats
+* Allow for (( when linting (#259, @nathaneastwood)
 
 # lintr 0.2.0 #
 
--- a/R/spaces_left_parentheses_linter.R
+++ b/R/spaces_left_parentheses_linter.R
@@ -24,7 +24,7 @@ spaces_left_parentheses_linter <- function(source_file) {
         before_operator <- substr(line, parsed$col1 - 1L, parsed$col1 - 1L)
 
         non_space_before <- re_matches(before_operator, rex(non_space))
-        not_exception <- !(before_operator %in% c("!", ":", "["))
+        not_exception <- !(before_operator %in% c("!", ":", "[", "("))
 
         if (non_space_before && not_exception) {
           Lint(


jimhester:
Great thanks, can you add a small note to NEWS.md with this change mentioning the issue number and your GitHub handle. Thanks!

nathaneastwood:
Done :) I even managed a nice `squash` for my sins

jimhester:
Great thanks!

daroczig:
Thanks a lot for this!

I see this is still in effect as per https://github.com/jimhester/lintr/blob/master/R/spaces_left_parentheses_linter.R, but the sources uploaded to CRAN doesn't include this fix? :thinking: 

image

VS

image

downloaded from https://cran.r-project.org/src/contrib/lintr_1.0.3.tar.gz

Sorry, probably I'm totally off with this one ... but any ideas what am I missing here?

daroczig:
Hm, it seems that it's rather due to the fact that v1.0.3 does not include this fix as per https://github.com/jimhester/lintr/blob/v1.0.3/R/spaces_left_parentheses_linter.R#L27 -- despite the commit was merged to `mater` in March 2018 and this version was tagged and released on CRAN in Nov 2018. I really don't get how this happened, but seems like I have to update to the development version to be able to use this ~1 yr old fix :(

jimhester:
The version of CRAN is based off of v1.0.0 code and has been updated only with minor bug fixes as necessitated to remain on CRAN. While the devel version contains additional bug fixes it also is [much slower](https://github.com/jimhester/lintr/issues/226) when linting and needs additional work to go to CRAN.

If you are interested in these changes going to CRAN there [are a number of things](https://github.com/jimhester/lintr/issues/318#issuecomment-389972589) that still need to be addressed to get the current devel codebase ready for a CRAN release.

daroczig:
Oh, got it, thanks a lot for the pointers, @jimhester :+1:

Issue #299, opened 2018-03-07
Allow `##` in commented_code_linter?


prosoitos:
In emacs ESS, `##` starts a commented line, while `#` starts a comment at the end of a code line. This is convenient for auto-indentation (`## comments` do not indent, while `# comments` automatically indent).

Lintr considers `##` to start a comment as a mistake. Would it create other issues down the line to add `##` to the commented_code_linter?

jimhester:
I am fine with changing this, would be happy to accept a PR for it!

prosoitos:
Great!

I just gave it a try. This is my very first PR, so please bear with me if I messed something up...

PR #301, opened 2018-03-10
Allow ## in comments (#299)

prosoitos:
"##" is emacs ESS way to start a commented line


R/comment_linters.R
```diff
@@ -28,7 +28,7 @@ ops <- list(
 #' @export
 commented_code_linter <- function(source_file) {
   res <- re_matches(source_file$file_lines,
-                    rex("#", any_spaces,
```
jimhester: Maybe a better solution would be `some_of("#")`, which corresponds to `#+` and would allow any number of `#`

jimhester:
Looks good, thanks! Could you also add a note to NEWS.md with the change and mentioning the issue number and your GitHub username; then I can merge this.

R/comment_linters.R
prosoitos: I thought of it, but I was worried to disrupt something else for people running lintr in RStudio or other interfaces that I don't use. So I opted for the minimal possible code change that would fix my problem. But either of these options would totally work for me :)

prosoitos:
Will do. Thank you for all the work on lintr and for your quick help!!

prosoitos:
Following your advice, I replaced `one_of("#", "##")` by `some_of("#")` and I added a line to `NEWS.md`. Because my branch is now 9 commits behind, there are conflicts on that last file. Please let me know if there is anything I should be doing on my end (again, this is my first PR, so I am not used to the proper protocol).

Thank you!

jimhester:
Everything looks good, thanks again and congratulations on your first successful pull request!

prosoitos:
Thank you!

PR #303, opened 2018-03-12
Don't add spaces to ^ infix operator

### LINKED ISSUE #302: infix operators don't conform to tidyverse standards
I have found a difference between `lintr` and [`styler`](https://github.com/r-lib/styler). `styler` will correct the following

```r
(10 ^ (x - 1)) + 1
```

to

```r
(10^(x - 1)) + 1
```

This gives the following `lintr` errors:

```
Put spaces around all infix operators.
Place a space before left parenthesis, except in a function call.
```

Looking at the current tidyverse style guide, [`styler` is correct here](http://style.tidyverse.org/syntax.html#spacing). To get around this you can use

```r
styler::style_pkg(".", math_token_spacing = list(one = c("'^'", "'+'", "'-'", "'*'", "'/'")))
```

A similar issue exists with

```r
break ()
```

which gets corrected to 

```r
break()
```

by `styler`. I am not sure which is correct with regards to the tidyverse here.

nathaneastwood:
Should close #302.

I also corrected a test from a previous commit that I overlooked.


jimhester:
Looks good, thanks! Could you add a note to NEWS.md detailing this change and referencing the issue number and your GitHub username; then I can merge this.

nathaneastwood:
Already done buddy 🙂

nathaneastwood:
A little nudge on this one bud :)

tvatter:
As mentioned in #302, I noticed that this fix is not present on the CRAN version, which is around 6 months more recent than the fix itself. Any idea what happened?

PR #317, opened 2018-05-11
Update Sublime Text configuration section of README

jmwerner:
The current documentation led to errors in Sublime Text 3. Also, it was challenging to find how to disable certain linters, so I included that in the example too.


jimhester:
Thanks!

PR #326, opened 2018-06-07
Add equals_na_linter

### LINKED ISSUE #143: x == NA instead of is.na(x)
It would be useful to have a linter that checks for `x == NA` where I meant to write `is.na(x)`. Is that possible?

jabranham:
Adds a linter that checks for e.g. `x == NA` rather than `is.na(x)`. Added to default linters.

Let me know if something isn't clear or if I need to change something.

Closes #143


jimhester:
Thanks! Could you add a note to NEWS.md mentioning this addition with your GitHub handle and the issue number?

jabranham:
Sure, done. I also regenerated the documentation. 

Just FYI, the NEWS.md file starts with `1.0.1.9001`, but that's smaller than what's in DESCRIPTION, which says `1.0.2.9000`.

jimhester:
It is just outdated in the NEWS file and needs to be updated.

jimhester:
Thanks!

Issue #335, opened 2018-07-04
Allow the user to pass the path of .lintr to lint()


mattiaciollaro:
Hey @jimhester, thanks again for this package!

I am trying to wrap `lintr`, together with linters for other languages, in [another project](https://github.com/spreemohealth/style).

Is it possible to manually pass the `.lintr` config file to `lint()` instead of having `lint()` look for it automatically in `current_dir -> project_dir -> user_dir` as implemented in the [`find_config`](https://github.com/jimhester/lintr/blob/89ec4178e6e76d94894cd97b31aa5b6a1a32c39e/R/settings.R#L52) function of the `settings.R` file?

I am currently trying to hack a temporary solution, because I think the answer to my question is *not at the moment*.
If it is in fact not possible to do so at the moment, do you think this could be made a feature?

I am thinking about the possibility of adding a `config_file` argument to `lint()` which, if not `NULL`, overrides the automatic search for the `.lintr` file (assuming that `parse_settings = TRUE`).

jimhester:
The config file location is looked up by the R option `lintr.linter_file`, so you could use that to change the location. But I agree it should probably just be an argument to `lint()`.

jimhester:
I think this was addressed with baddc5b

Issue #123, opened 2015-12-11
Make sure lintr finds `.lintr` files when called from plugins.


jimhester:
The plugins usually call `lint()` with a tempfile rather than the acutal filename, we need to change the `lint()` API to take a optional filepath as well, which can be used to find the project configuration file.

AshesITR:
#503 implements this via the `file` argument, whose contents will be ignored when `text` is also present.

PR #338, opened 2018-07-24
Allow users to set lintr.linter_file

jabranham:
This lets users specify the location of the linter file with
options(lintr.linter_file) before calling lintr::lint(). This should
be useful to allow the various plugins to specify the path to the
.lintr file

Related to #123 and #335

### DIFF
--- a/R/settings.R
+++ b/R/settings.R
@@ -56,6 +56,9 @@ find_config <- function(filename) {
   }
   linter_file <- getOption("lintr.linter_file")
 
+  ## if users changed lintr.linter_file, return immediately.
+  if (is_absolute_path(linter_file) && file.exists(linter_file)) return(linter_file)
+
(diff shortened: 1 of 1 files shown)


jabranham:
Checking if the file exists makes sense, but doesn't it make sense for users to be able to use a relative path? It could be just as simple as saying `.lintr-config` (or whatever) instead of `.lintr`.

jimhester:
Yes, you can already do that, but the code assumes the path is relative to the file being linted or the project rather than the working directory; which causes issues when the 3rd party plugins call lintr on a copy of the file in a temporary directory rather than the file itself.

jabranham:
Ah, I forgot that. Probably should have more coffee :-)

Changed, thanks

jimhester:
Thanks!

PR #342, opened 2018-08-13
Allow passing multiple styles to object_name_linter

### LINKED ISSUE #341: Accept multiple styles in object_name_linter
```
> library(lintr)
> txt <- c("var_one <- 1", "varTwo <- 2", "var.three <- 3")
> linter <- object_name_linter(c("snake_case", "lowerCamelCase"))
```
Current behavior:
```
> expect_lint(txt, list("should be snake_case", line_number=3), linter)
Error in vapply(x, `[[`, character(1), "message") : 
  values must be length 1,
 but FUN(X[[1]]) result is length 2
In addition: Warning messages:
1: In if (!matches_styles(name, style)) { :
  the condition has length > 1 and only the first element will be used
2: In if (!matches_styles(name, style)) { :
  the condition has length > 1 and only the first element will be used
3: In if (!matches_styles(name, style)) { :
  the condition has length > 1 and only the first element will be used
```

Proposed: 
```
> expect_lint(txt, list("should be snake_case or lowerCamelCase", line_number=3), linter)
> 
```

I realize it's usually far better to pick one naming style and stick to it, but today I'm writing a package whose interface is dictated by two incompatible upstream APIs, so I'd like to allow "snake_case or dotted.case, but no others". Since users would need to explicitly opt into it by specifying multiple styles, I submit that this change doesn't pose too much of a moral hazard :)

Since `matches_styles` already returns a vector, I think this should be easy to implement and I can take a stab at it if y'all are OK with the concept.

infotroph:
Fixes #341. Example usage: ` lint("inst/example/bad.R", object_name_linter(c("dotted.case", "lowerCamelCase"))`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,5 @@
 # lintr 1.0.2.9000 # 
+* Fixed error when object_name_linter is passed multiple styles (#341, @infotroph)
 * Config files are now also searched for in the users' home directory (#266, @randy3k)
 * Fixed crash caused by ambiguous cache file paths (#212, @fangly).
 * RStudio addins to lint current source and project (fixes #264, @JhossePaul)
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -132,11 +132,11 @@ object_name_linter <- function(style = "snake_case") {
   make_object_linter(
     function(source_file, token) {
       name <- unquote(token[["text"]])
-      if (!matches_styles(name, style)) {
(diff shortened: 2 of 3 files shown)


jimhester:
Thanks!

jimhester:
I added you to the list of contributors to lintr, see https://github.com/jimhester/lintr/issues/318#issuecomment-389972589 for details on how this works.

PR #349, opened 2018-09-17
fixes #252: Allow plain-code-blocks in RMarkdown

### LINKED ISSUE #252: Plain Code Blocks in RMarkdown
I have an issue when running lintr (v1.0.0 in vim or within R) on Rmarkdown files that contain plain code blocks (as defined [here](http://rmarkdown.rstudio.com/authoring_basics.html)). Running lintr on the following (temp.Rmd) throws an "Error: Malformed file!", despite plain code blocks being legal in Rmarkdown and lintr working fine on any other Rmarkdown files that contain R code. If the block containing "This isn't evaluated" is removed, lintr runs fine; when it's present, the Malformed file error comes up.

\---
title: "Plain Code Bug?"
output: pdf_document
\---
\```{r}
42
\```

\```
This isn't evaluated
\```

russHyde:
Added a test to the end of test.Rmd that included a plain-code-block like this:

```
This shouldn't cause a problem
```

That is, a code-block where the interpreter is not specified (these are allowable in Rmarkdown according to section 2.5.2 of Yihui's Rmarkdown book).

The code-block reproduced the `Malformed File` error described in issue #252

Plain-blocks match the knitr `chunk.end` pattern at both the start and the end
of the chunk. So when plain-blocks are present there was an extra 2k chunk ends
than starts.

A small function `filter_chunk_end_positions` was added to filter the positions
of the chunk-ends based on the chunk-starts and called from `extract_r_source`

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -1,3 +1,29 @@
+filter_chunk_end_positions <- function(starts, ends) {
+  # In a valid file, possibly with plain-code-blocks,
+  # - there should be at least as many ends as starts,
+  #   and there should be an even-number of extra ends (possibly zero)
+  #   since each plain-code-block should open & close, and the open/close
+  #   tags of a plain-code-block both match the chunk.end pattern
+
+  # starts (1, 3, 5, 7,        11)  --> (1, 3, 5, 7, 11)
+  # ends   (2, 4, 6, 8, 9, 10, 12)  --> (2, 4, 6, 8, 12) # return this
+  length_difference <- length(ends) - length(starts)
+
+  if(length_difference < 0 | length_difference %% 2 != 0) {
+    stop("Malformed file!", call. = FALSE)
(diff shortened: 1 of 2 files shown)


randy3k:
not the most beautiful way to find the blocks, but I beleive it would work.

jimhester:
Thanks! Could you add a 
    
    ```{python}
    ```

example like in https://github.com/jimhester/lintr/issues/322 as well to the Rmd.

Also please add a note to NEWS.md explaining the change and referencing the issue number and your GitHub username.

russHyde:
For consistency with the other helper functions in `extract.R`, would you prefer that the function I added was placed after the main `extract_r_source` function in the source file?

jimhester:
Thanks @russHyde for the PR and @randy3k for reviewing!

PR #350, opened 2018-09-18
fixes #322: allows Rmd chunks using {lang} engine format

### LINKED ISSUE #322: Lint fails with Rmd chunks using {lang} engine format
This issue is related to #32, where lintr should ignore non-R chunks as it lints. `lintr` currently only ignores chunks that have the following chunk header: `{r engine = "LANG"}`, vs. the newer and now more common `{LANG}` Refer [here](https://rmarkdown.rstudio.com/authoring_knitr_engines.html), and the discussion on knitr issue #963 [here](https://github.com/yihui/knitr/issues/963).

Currently, if you use the `{LANG}` language engine syntax, `lintr` will do 2 things:

1. attempt to lint the chunk and throw an error for the user
2. not lint any R code after that chunk.

Below is a minimal reprex, using the .Rmd test case in `lintr` as a base and adding a chunk with `{python}`.

````
text <- c('# Test #

          Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod
          tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At
          vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren,
          no sea takimata sanctus est Lorem ipsum dolor sit amet.
          
          ```{r}
          a = 1
```

          Test
          ====
          
          Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod
          tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At
          vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren,
          no sea takimata sanctus est Lorem ipsum dolor sit amet.
          
          ```{r}
          b <- function(x) {
          d = 1
          }
          
          ```
          
          ```{r engine="python"}
          a=[]
          
          a[0]=1
          ```
          
          ```{python}
          a=[]
          
          a[0]=1
          ```


​          
          ## Same code chunk as above, but lintr will not ouput this
          ```{r}
          
          b <- function(x) {
          d = 1
          }
          
          ```
          ')
  tmp <- tempfile()

  writeLines(text, tmp)

  lintr::lint(tmp)
````
The results of the lint are below:

image

The part of the code that would require an edit is [here](https://github.com/jimhester/lintr/blob/15fd8ee6685248b7f477c949c1e78dc13350cd15/R/extract.R#L31).

I _think_ the fix to this would involve adding another `&&` condition that checked that the chunk began with `r` and does NOT have `engine` following it. Or perhaps more simply, found any chunk that began with anything other than `r` and ignored it. Happy to try a fix on this and submit a PR if that would be useful.

### PR DESCRIPTION
**Pull request by russHyde**
Rmarkdown format allows the use of code blocks in the following format

~~~~
```{some_engine}
# not necessarily R syntax
foo = bar
```
~~~~

These blocks should be filtered out before linting the remaining R-code blocks.

A couple of python blocks were added to test.Rmd: one with R-like syntax and one with syntax that can't be parsed as R.

The function `defines_knitr_engine` was extracted from `extract_r_source` and extended. This tests for the presence of {r engine = some_engine} and {some_engine} in the start line of a code chunk. The names of valid engines are obtained from `knitr::knit_engines`

An unrelated, cosmetic change to `filter_chunk_end_positions` was made (which(grepl(...)) --> grep(...)).

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -31,8 +31,8 @@ extract_r_source <- function(filename, lines) {
         start + 1 < end &&
 
         # block does not set an engine (so is r code)
-        !rex::re_matches(lines[start],
-          rex::rex(boundary, "engine", any_spaces, "="))) {
+        !defines_knitr_engine(lines[start])
+      ) {
         output[seq(start + 1, end - 1)] <<- lines[seq(start + 1, end - 1)]
       }
     },
@@ -68,11 +68,24 @@ filter_chunk_end_positions <- function(starts, ends) {
   }
 
   positions <- sort(c(starts = starts, ends = ends))
-  code_start_indexes <- which(grepl("starts", names(positions)))
+  code_start_indexes <- grep("starts", names(positions))
   code_ends <- positions[1 + code_start_indexes]
 
   stopifnot(all(grepl("ends", names(code_ends))))
   code_ends
+}
+
+defines_knitr_engine <- function(line) {
+  # "scala|python|bash|..."
+  engine_choices <- paste(names(knitr::knit_engines$get()), collapse = "|")
+
+  # Looks for {some_engine}, {some_engine label, ...} or {some_engine, ...}
+  bare_engine_pattern <- paste0("\\{(", engine_choices, ")[, }]")
+
+  rex::re_matches(
+    line, rex::rex(boundary, "engine", any_spaces, "=")
+  ) ||
+    grepl(pattern = bare_engine_pattern, x = line)
 }
 
 replace_prefix <- function(lines, prefix_pattern) {
--- a/R/extract.R
+++ b/R/extract.R
@@ -26,7 +26,6 @@ extract_r_source <- function(filename, lines) {
   Map(
     function(start, end) {
       if (
-
         # block contains at least one line of code
         start + 1 < end &&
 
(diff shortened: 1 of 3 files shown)


R/extract.R
```diff
@@ -68,13 +68,26 @@ filter_chunk_end_positions <- function(starts, ends) {
...
   stopifnot(all(grepl("ends", names(code_ends))))
   code_ends
 }
 
+defines_knitr_engine <- function(line) {
+  # "scala|python|bash|..."
+  engine_choices <- paste(names(knitr::knit_engines$get()), collapse = "|")
+
+  # Looks for {some_engine}, {some_engine label, ...} or {some_engine, ...}
+  bare_engine_pattern <- paste0("\\{(", engine_choices, ")[, }]")
```
jimhester: please construct this pattern using rex, it would be something like

```r
rex::rex("{", or(names(knitr::knit_engines$get())), one_of("}", " ", ","))
```

russHyde: No problem. I've modified that and pushed. Hadn't used `rex` before.

russHyde:
Thanks again for a great package

Issue #318, opened 2018-05-17
Help collaborate on lintr!


jimhester:
As announced at https://twitter.com/jimhester_/status/997109466674819074 I am interested in having some help working on lintr. This issue can serve as a place for those interested to 'sign up' and also to discuss priorities. So reply below if you are interested and I will invite you to be a collaborator on the project.

brankokovac:
Branko Kovac

Jamesguy0121:
I am interested in helping out.

jimhester:
So do you all think it would be helpful to have a [gitter](https://gitter.im) chat room so you could have more real time access for questions etc, or are GitHub comments enough?

erleholgersen:
I'm also interested in helping out.

navdeep-G:
I am interested if there is still room. :)

adanvr:
I would like to help as well

dougmet:
Hi, a few of us from Mango might be able to muck in as it slots together quite well with mangothecat/goodpractice and friends. cc @hfrick @sellorm

jimhester:
@dougmet, that would be great, If the other Mango folks are interested please add a separate reply and I can add you to the collaborators.

jimhester:
Also most of the linters in goodpractice that would be great to have in lintr, so a good first contribution would be doing that!

wligtenberg:
I am interested in helping out as well.

randy3k:
I am the author of https://github.com/REditorSupport/languageserver
I’d also like to be a collaborator.

hfrick:
Hi @jimhester, some lintr/goodpractice crossover would be nice. 👋

chucheria:
I want to help too! 🙌

owenjonesuob:
Hiya, great to see interest in lintr-goodpractice collaboration! I'm definitely keen to help out.

jimhester:
Never too late!

DragosMG:
Same here. I'd be keen to help.

cryptomanic:
I would love to contribute.

gu-stat:
I'd like to help, if there's still room.

jimhester:
A great first issue PR for someone to review would be https://github.com/jimhester/lintr/pull/321, it mainly needs a note added to NEWS.md

gdequeiroz:
Hi @jimhester, what do you mean by "assess performance" in  `Assess performance of the devel version vs that of the current CRAN version (devel is quite a bit slower)`

[...]

Issue #226, opened 2017-03-21
Major speed deprecation with new object linters


jimhester:
``` r
# Run from project root directory

library(git2r)
r <- repository()

run_bench <- function(commit, code, ...) {
  checkout(Filter(function(x) startsWith(x@sha, commit), commits(r))[[1]])
  devtools::load_all()
  x <- microbenchmark::microbenchmark(eval(code), ...)
  x$expr <- commit
  x
}

code <- quote({lint("R/lint.R")})
x <- rbind(
  run_bench("fd1ec51d", code, times = 10),
  run_bench("7fe7e909", code, times = 10))
#> Loading lintr
#> Loading lintr

summary(x)
#>       expr      min       lq     mean   median       uq      max neval
#> 1 7fe7e909 1.595453 1.599098 1.620174 1.610035 1.621247 1.714548    10
#> 2 fd1ec51d 4.230639 4.302109 4.387139 4.361579 4.428514 4.681118    10
summary(x, "relative")
#>       expr      min       lq     mean   median       uq      max neval
#> 1 7fe7e909 1.000000 1.000000 1.000000 1.000000 1.000000 1.000000    10
#> 2 fd1ec51d 2.651685 2.690335 2.707819 2.708996 2.731549 2.730233    10
```

So the current master is ~2.7 times slower than the code as of Nov. 10th. Lintr was already too slow and we can't afford to be even slower than before.

jimhester:
New implementation of object_name_linter is an order of magnitude more performant than current master, https://github.com/jimhester/lintr/compare/name_performance?expand=1. Should remedy most of the slowdown.
```r
microbenchmark(
  {lint("R/lint.R", object_name_linter("snake_case"))}, 
  {lint("R/lint.R", object_name_linter_old("snake_case"))})
#> Unit: milliseconds
#>                                                         expr       min        lq      mean    median        uq      max neval
#>  {     lint("R/lint.R", object_name_linter2("snake_case")) }  225.9701  252.2269  265.9434  263.2129  273.7824  483.367   100
#>   {     lint("R/lint.R", object_name_linter("snake_case")) } 1275.5349 1360.3626 1430.6543 1425.3408 1486.9745 1735.298   100
```

fangly:
Hi Jim,
A few comments on this promising branch:
1. `declared_s3_generics()` is not found anywhere. I uncommented it locally so I could test your changes.
2. `lengths()` is new to R 3.2.0 but we are using 3.1.1 in lintr.
[...]

PR #236, opened 2017-05-04
Object name linter performance improvements

jimhester:
Fixes #226

PR #356, opened 2018-11-13
speed up make_object_linter - ref issue #226

russHyde:
As described in #226 , the speed of the `object_name_...` linters has reduced since introduction of the `object_name_linter("snake_case")` style linters

The function `make_object_linter` was refactored. Some conditional logic, that checked whether a given object was an operator, a generic or a base-function and bypassed that object if it was, was moved out of an lapply loop into a prefiltering step. This step is vectorised over all object-names in a given file, so speeds up linters that are built using `make_object_linter`.

No additional unit-tests were added

For analysis of the `check.R` file referenced in #318  using a) the CRAN version (1.0.2); b) the current master github version; c) the updated version

code:
```
system.time(lint("helpers/check.R", linters = object_length_linter(20), cache = FALSE))
```

a) cran:
   user  system elapsed
 40.312   0.156  40.820

b) master (ooft!; my laptop isn't great ..):
   user  system elapsed
577.812   0.984 586.789

c) update:
   user  system elapsed
324.437   0.078 326.421
326.063   0.109 327.099

I'll admit the improvement is less than two-fold and comes nowhere near the cran version. I get similar speed--up when comparing `lint_package` on the `lintr` source code (the only default_linters that were speeded up were `object_length_*` and `object_name_*` and they had equivalent degree of speedup).

Will have a play with other possible changes & try and work out why there's such a big difference between the sys-time speedup (~ 10 fold) and the actual speed up (< 2 fold)

### DIFF
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -1,14 +1,27 @@
 make_object_linter <- function(fun) {
   function(source_file) {
+
+    token_nums <- ids_with_token(
+      source_file, rex(start, "SYMBOL" %if_next_isnt% "_SUB"), fun=re_matches
+    )
+    if(length(token_nums) == 0){
+      return(list())
+    }
+    tokens <- with_id(source_file, token_nums)
+    names <- unquote(tokens[["text"]]) # remove surrounding backticks
+
+    keep_indices <- which(
+      !is_operator(names) &
+        !is_known_generic(names) &
+        !is_base_function(names)
+    )
+
     lapply(
-      ids_with_token(source_file, rex(start, "SYMBOL" %if_next_isnt% "_SUB"), fun=re_matches),
-      function(token_num) {
-        token <- with_id(source_file, token_num)
-        name <- unquote(token[["text"]])  # remove surrounding backticks
+      keep_indices,
+      function(i) {
+        token <- tokens[i, ]
+        name <- names[i]
         if (is_declared_here(token, source_file) &&
-            !is_operator(name) &&
-            !is_known_generic(name) &&
-            !is_base_function(name) &&
             !is_external_reference(source_file, token[["id"]])) {
           fun(source_file, token)
         }


russHyde:
I've updated my code now - replacing `object_name_linter` and `object_length_linter` with versions that are vectorised over all of the the object-names in a file (the above just performs a vectorised prefiltering step). 

This reduces the run times further

- `object_length` is similar to the bulk of the other linters;
- `object_name` is ~ 2-3 times slower than most other linters but ~ 5x faster than in the master branch 

`pipe_continuation_linter` remains a compute hog.

```
# times of default_linters on `check.R` using my work desktop
                             linter  master  vector
1                 assignment_linter  12.752  12.581
2               closed_curly_linter  13.924  13.814
3                     commas_linter  15.093  15.067
4             commented_code_linter  13.273  13.543
5                  equals_na_linter  14.608  14.759
6  function_left_parentheses_linter  18.200  18.600
7               infix_spaces_linter  15.189  15.237
8                line_length_linter  12.719  12.983
9                     no_tab_linter  14.654  14.917
10             object_length_linter 222.020  13.638
11               object_name_linter 224.244  45.003
12              object_usage_linter  22.166  22.355
13                open_curly_linter  13.776  14.089
14         pipe_continuation_linter 178.237 180.176
15             single_quotes_linter  13.970  14.186
16             spaces_inside_linter  14.345  14.601
17   spaces_left_parentheses_linter  18.413  17.653
18      trailing_blank_lines_linter  12.433  13.206
19       trailing_whitespace_linter  14.229  14.614
```


Should I add the new code to this pull-request?

jimhester:
I think as far as the object linter goes we should try to use the approach in https://github.com/jimhester/lintr/pull/236, IIRC it is pretty close to a drop in replacement, but there are a few corner cases that need to be fixed.

russHyde:
Cool. Thanks for looking into it; the lintr speed was a bit of a problem for me (mostly in vim/syntastic).

jimhester:
Thanks @russHyde, I will merge this for now, i the future we can consider moving to #236, but this is a good improvement for now.

Thanks!! You are da bomb!

russHyde:
@jimhester thanks for merging. I will try to have a look at your xpath-based code; I didn't quite follow it the first time round (due to my xpath inexperience). If there's any particular aspect of #236 that you want someone to look at, ping me. Very happy to help on lintr, since one of my packages (and my workflow more generally) depends really heavily on it.

I should state that I didn't end up pushing the whole of my work on this PR (the stuff on `object_[name|length]_linter` that is mentioned above); will have a look when I'm back in the office.

PR #360, opened 2018-12-18
Adding function lintr_dir to check R files for given path

arekbee:


peterskipper:
My team could definitely use this if it was added!

mattyb:
I'm seeing similar test failures on a different branch

frankschmitt:
This would be tremendously useful. Is there anything I can do to help to get this pull request merged  into master?

jimhester:
The main reason I haven't merged it is it duplicates logic that would be better off refactored into a function that could be used here and in `lint_package()`.

arekbee:
@jimhester  lint_package() has been refactored. 
There are no integration tests for lint_dir nor lint_package. Is it required?

R/lint.R
```diff
@@ -140,49 +140,37 @@ reorder_lints <- function(lints) {
...
   )
 
   # Remove fully ignored files to avoid reading & parsing
   to_exclude <- vapply(seq_len(length(files)),
-    function(i) {
-      file <- files[i]
-      file %in% names(exclusions) && exclusions[[file]] == Inf
-     },
-    logical(1))
+                       function(i) {
```
byapparov: @arekbee I think this file will benefit from styling with styler package - https://github.com/r-lib/styler

this will make review easier.

R/lint.R
```diff
@@ -125,64 +125,53 @@ reorder_lints <- function(lints) {
     ]
 }
 
-#' Lint a package
+#' Lint a directory
 #'
-#' Apply one or more linters to all of the R files in a package.
-#' @param path the path to the base directory of the package, if \code{NULL},
+#' Apply one or more linters to all of the R files in a directory
+#' @param path the path to the base directory, if \code{NULL},
```
russHyde: Could you check what actually happens when NULL is passed as the base-path, from the code it looks like the function will die (because `dir` will)

R/lint.R
```diff
@@ -125,64 +125,53 @@ reorder_lints <- function(lints) {
...
   # Remove fully ignored files to avoid reading & parsing
   to_exclude <- vapply(seq_len(length(files)),
-    function(i) {
-      file <- files[i]
-      file %in% names(exclusions) && exclusions[[file]] == Inf
-     },
-    logical(1))
+                       function(i) {
+                         file <- files[i]
+                         file %in% names(exclusions) && exclusions[[file]] == Inf
```
russHyde: `exclusions` is no longer defined within the scope of this function. So this should throw an error. Did you add any tests for `lint_dir`, or is `lint_dir` only tested through a call from `lint_package` (where `exclusions` is defined)?

R/lint.R
```diff
@@ -192,10 +181,10 @@ lint_package <- function(path = ".", relative_path = TRUE, ..., exclusions = NUL
...
-      function(x) {
-        x$filename <- re_substitutes(x$filename, rex(path, one_of("/", "\\")), "")
-        x
-      })
+                      function(x) {
+                        x$filename <- re_substitutes(x$filename, rex(path, one_of("/", "\\")), "")
+                        x
+                      })
     attr(lints, "path") <- path
   }
```
russHyde: None of the logic in lines 164-189 have changed

R/lint.R
```diff
@@ -205,6 +194,43 @@ lint_package <- function(path = ".", relative_path = TRUE, ..., exclusions = NUL
...
+#' Apply one or more linters to all of the R files in a package.
+#' @param path the path to the base directory of the package, if \code{NULL},
+#' it will be searched in the parent directories of the current directory.
+#' @param relative_path if \code{TRUE}, file paths are printed using their path
+#' relative to the package base directory.  If \code{FALSE}, use the full
+#' absolute path.
+#' @param ... additional arguments passed to \code{\link{lint}}, e.g.
+#' \code{cache} or \code{linters}.
+#' @param exclusions exclusions for \code{\link{exclude}}, relative to the
+#' package path.
```
russHyde: The docs for `path`, `relative_path`, `...` and `exclusions` are almost identical to their definitions for `lint_dir`. You could use `@inheritParams   lint_package` to deduplicate the docs and rewrite the docs for `lint_dir` to apply generally for both `lint_dir` and `lint_package`

tests/testthat/test-exclusions.R
```diff
@@ -198,4 +208,30 @@ test_that("it excludes properly", {
   }
 })
 
+test_that("it excludes properly by dir", {
+  read_settings(NULL)
```
russHyde: I'm afraid I don't understand what's happening in these tests. Should the user need to use `read_settings()` before using `lint_dir`? Could you add some comments to explain what is being setup and what is being tested when these tests run, please.

tests/testthat/test-knitr_formats.R
```diff
@@ -1,4 +1,14 @@
 context("knitr_formats")
+
+test_that("it handles dir", {
+  lints <- lint_dir(path = "knitr_formats", pattern = rex::rex(".R", one_of("html", "md", "nw", "rst", "tex", "txt")))
```
russHyde: perhaps `rex::one_of`

R/lint.R
arekbee: hi @russHyde , thank you for the tip. I haven't used @inheritParams yet.

arekbee:
How can I increase codecov/patch  from 66.66% to  81,74% ??
Is it really needed (codecov/patch)?

jimhester:

Thanks!! You are da bomb!

PR #362, opened 2019-01-15
Set github token from env

### LINKED ISSUE #64: GitHub comment fails on private repos
Travis.com allows tests to be run on private GitHub repos. However, the comments fail because the comment_token doesn't have access to the private repo. Here's the error message:

<pre><code>Error in stop(httr::http_condition(response, "error", message = httr::content(response,  : 
  {
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3"
}
Calls: print -> print.lints -> github_comment
</code></pre>

For our purposes, it'd be nice to simply turn off GitHub commenting entirely. However, it might also be a good idea to have this fail a bit more gracefully and not halt the program. When this happens, we don't see any of the lint results in our Travis CI logs.

mattyb:
Fixes #64

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,6 @@
 # lintr 1.0.3.9000 # 
+* Add support for overriding GitHub API Token via `GITHUB_TOKEN` environment
+  variable (#63, @mattyb)
 * Changed the default value of the `length` argument to `object_length_linter` to 30 for consistency (#325 @DragosMG) 
 * Fixed error when object_name_linter is passed multiple styles (#341, @infotroph)
 * Config files are now also searched for in the users' home directory (#266, @randy3k)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,10 @@
 # R for travis: see documentation at https://docs.travis-ci.com/user/languages/r
 
 language: R
+r:
+  - 3.4
```
jimhester: I would prefer to change the travis yaml independent of the other changes.

jimhester:
Thanks! Overall this looks good, can you add a note to NEWS.md describing the change and mentioning the issue number and your GitHub username, then I can merge this.

.travis.yml
mattyb: makes sense! I was trying to debug #363. Will revert.

mattyb:
updated!

jimhester:
Thanks!

Issue #242, opened 2017-05-12
Space between function right parenthesis and function open curly


jennybc:
It feels like lintr should complain about the lack of space between the right paren and opening curly brace here (occurs twice):

``` r
if (length(foo) == 0L){
    if (verbose){
      message("message")
    }
    invisible(NULL)
  } else
    req_tbl[keep_names, ]
```

jennybc:
Another issue brought up by that snippet: seems like the else statement should have brackets. Either because it always should or because the if used brackets. If that's correct, I can open a second issue or someone else can.

jimhester:
I agree about the spacing before the `{`, but I don't think any of the linters currently check for that.

My personal style is to always use braces for all conditionals, but some people feel the opposite and the tidyverse style is mostly silent on this matter (http://style.tidyverse.org/syntax.html#indenting).

I guess tidyverse style guide does say to use braces on all multiline conditionals, so it probably should be linted.

jennybc:
I say: he who controls the default linters has the power 👊.

xanderdunn:
I'd like to see a linter that checks for this as well

ashiklom:
Seconding (thirding? fourthing?) this. It could it its own linter, and maybe even one that isn't enabled by default. E.g. `paren_brace_space_linter`.

dschlaep:
I would like to see this implemented as well, either as stand-alone linter as previously suggested or folded into an existing one, e.g., `open_curly_linter`.

Interestingly, Rstudio (Version 1.1.463) shows above example as a lint case even though `lintr::lint()` does not, see attached screenshot.
![Screen Shot 2019-03-11 at 15 42 58](https://user-images.githubusercontent.com/4205019/54152863-9d439f00-4414-11e9-832a-edc3f5aefdd2.jpg)


Thanks!

jimhester:
RStudio does not use lintr, it uses a completely different codebase.

PR #380, opened 2019-04-19
add paren_brace_linter

bfgray3:
re: #242


jimhester:
Thanks, looks good! You also need to add the new linter to the default linters. Also
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jimhester:

Thanks!! You are da bomb!

PR #381, opened 2019-04-23
Add `commented_code_linter` to documentation of possible linters

andrewychoi:
Looks like this was added in https://github.com/jimhester/lintr/pull/83 but not added to documentation


jimhester:
Thanks!

Issue #384, opened 2019-05-06
New test failures with R 3.6


jimhester:
It looks like there are a number of tests failures exposed by R 3.6, we should get these fixed.

dschlaep:
#377 is likely one of them (it has R-devel in the title, but that version is by now R v3.6)

russHyde:
Had a look at this under R 3.6.1 (with current lintr master, xml2=1.2.2, rcpp=1.0.2, libxml=2.9.9)

```
Rscript -e "devtools::load_all(); testthat::test_dir('tests/testthat', reporter = 'summary')" > test_failures.txt
```

The first 10 failing tests were in these files:
```
absolute_path_linter: ...................................................
assignment_linter: ....1
clear_cache: ...
load_cache: ...
save_cache: ...........
cache_file: ..
retrieve_file: ..
cache_lint: ..
retrieve_lint: ........
has_lint: ...
find_new_line: .........
lint with cache: ............
checkstyle_output: .
closed_curly_linter: ............
commas_linter: ..............
commented_code_linter: ............2
comments: ...
defaults: ...........................
deprecated: ..........
equals_na_linter: ...3
error: ....
parse_exclusions: .......
normalize_exclusions: ................
exclude: 4
expect_lint: .5..67
extraction_operator_linter: .......................
function_left_parentheses_linter: .......................
get_source_expression: ...............
implicit_integer_linter: .......................
infix_spaces_linter: ...........8
knitr_formats: 9abcde...
```

Here's an example failure:
```
── 1. Error: returns the correct linting (@test-assignment_linter.R#12)  ───────
Opening and ending tag mismatch: expr line 8 and equal_assign [76]
1: expect_lint("blah=1", rex("Use <-, not =, for assignment."), assignment_linter) at tests/testthat/test-assignment_linter.R:12
2: lint(file, ...) at /home/ah327h/progs/pull_requests/lintr/R/expect_lint.R:45
3: get_source_expressions(filename) at /home/ah327h/progs/pull_requests/lintr/R/lint.R:36
4: lapply(top_level_expressions(parsed_content), function(loc) {
       line_nums <- parsed_content$line1[loc]:parsed_content$line2[loc]
       expr_lines <- source_file$lines[line_nums]
       names(expr_lines) <- line_nums
       content <- get_content(expr_lines, parsed_content[loc, ])
       id <- as.character(parsed_content$id[loc])
       edges <- component_edges(tree, id)
[...]

Issue #377, opened 2019-03-11
Function `%||%` fails on R-devel due to `_R_CHECK_LENGTH_1_LOGIC2_`


dschlaep:
R-devel introduces as new features:
"Experimentally, setting environment variable `_R_CHECK_LENGTH_1_LOGIC2_` will lead to warnings (or errors if the variable is set to a ‘true’ value) when && or || encounter and use arguments of length more than one. " (https://cran.r-project.org/doc/manuals/r-devel/NEWS.html)

The `lintr` function `%||%` will fail this upcoming check, e.g., 
```
x <- c("", "", "object_name_linter", "", "", "", "", "", "", "", "")
y <- c("", "", "", "", "", "", "", "", "", "", "")
lintr:::`%||%`(x, y)
```
with message 
```
Error in is.null(x) || is.na(x) : 
  'length(x) = 11 > 1' in coercion to 'logical(1)'
```

Expected result (as obtained with R v3.5.2):
```
> lintr:::`%||%`(x, y)
 [1] ""                   ""                  
 [3] "object_name_linter" ""                  
 [5] ""                   ""                  
 [7] ""                   ""                  
 [9] ""                   ""                  
[11] ""                  
```

Suggested fix: replace 
`if (is.null(x) || is.na(x) || length(x) <= 0) {...` 
with
`if (is.null(x) || anyNA(x) || length(x) <= 0) {...` 
(`anyNA` was introduced in R v3.1.0 (https://cran.r-project.org/doc/manuals/r-release/NEWS.html) which is earlier than the `lintr` required R version 3.1.1).

russHyde:
Ref https://github.com/jimhester/lintr/pull/395

russHyde:
Fixed in #395

PR #395, opened 2019-08-15
ensure lintr runs/installs/tests on R-3.6

russHyde:
The dev version of `xmlparsedata` has been updated to fix a parsing bug on R-3.6; before fixing that bug several lintr tests were failing (https://github.com/jimhester/lintr/issues/384). This change points `lintr` to the dev version of xmlparsedata

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,6 +30,8 @@ Imports:
 Suggests:
     rmarkdown,
     mockery
+Remotes:
+    r-lib/xmlparsedata
 License: MIT + file LICENSE
 LazyData: true
 VignetteBuilder: knitr
--- a/NEWS.md
+++ b/NEWS.md
@@ -61,6 +61,9 @@
 * New equals_na_linter() (#143, #326, @jabranham)
 * Fixed plain-code-block bug in Rmarkdown (#252, @russHyde)
 * Fixed bug where non-R chunks using {lang} `engine format` were parsed from R-markdown (#322, @russHyde)
+* Ensured `lintr` runs / installs / tests on R-3.6: pinned to github
+  `xmlparsedata`; ensure vectors are length-1 when compared using `&&` and `||`
+  (#363 #377 #384 #391, @russHyde).
 
 # lintr 1.0.3 #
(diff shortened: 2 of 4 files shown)


russHyde:
I'm not quite sure what the problem is here. I suspect the function `%||%` needs rewriting:

```
`%||%` <- function(x, y) {                                                        
  if (is.null(x) || is.na(x) || length(x) <= 0) {                                 
    y                                                                             
  } else {                                                                        
    x                                                                             
  }                                                                               
}
```

```
===> ?
`%||%` <- function(x, y) {                                                        
  if (is.null(x) || any(is.na(x)) || length(x) <= 0) {                                 
    y                                                                             
  } else {                                                                        
    x                                                                             
  }                                                                               
}
```

jimhester:
Maybe this, though we might want to rethink having the `is.na()` in there at all...

```r
`%||%` <- function(x, y) {                                                        
  if (is.null(x) || length(x) <= 0 || is.na(x[[1L]])) {                                 
    y                                                                             
  } else {                                                                        
    x                                                                             
  }                                                                               
}
```

russHyde:
Modified the definition of `%||%`. This should fix #377

russHyde:
A similar problem has occurred with the function `exclude()`. Within the following statement:

```
file %in% names(exclusions) &&
        exclusions[[file]] == Inf ||
        df$line_number[i] %in% exclusions[[file]]
```

Presumably `exclusions[[file]]` can have length 0 or > 1. 

Suggest:

```
file %in% names(exclusions) &&
    length(exclustions[[file]]) == 1L &&
    exclusions[[file]] == Inf ||
    df$line_number[i] %in% exclusions[[file]]
```


I'm not sure why these errors aren't being detected in my R 3.6.1 environment when I run devtools::check()

russHyde:
This should fix #384 and https://github.com/jimhester/lintr/issues/391 and #377 

Also, https://github.com/jimhester/lintr/issues/363 related to failing tests in master, so should be fixed now too

jimhester:
Thanks you can merge this whenever.

russHyde:
Cheers jim

PR #398, opened 2019-08-22
lint formals and ensure names match >= 1 style

russHyde:
Previous xml-based version of `object_name_linter`:

- required all names to match every one of the requested
  styles

- disregarded names of formals in newly-defined functions

Hence:

- added extra xpath clause to catch `SYMBOL_FORMALS`

- used or-reduction to ensure each name matches at least one
style.

Also:

- since multiple styles are allowed, the lint message is now
  like "Variable and function name style should be style_1,
  style_2 or style_3".

- tests for `object_name_linter` and lint message for
  `object_name_linter_old` were changed to match this
  format.

- added `...` argument to a unit test to ensure that
  `object_name_linter` does not flag dots.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,7 +36,8 @@ License: MIT + file LICENSE
 LazyData: true
 VignetteBuilder: knitr
 RoxygenNote: 6.1.1
-Collate: 
+Encoding: UTF-8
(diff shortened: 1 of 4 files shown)


russHyde:
sorry - forgot the docs

PR #400, opened 2019-08-28
361 cyclocomp linter

fabian-s:
Sorry, I just noticed that NA "ranges" trigger an error if rendered e.g. as a `reprex`... 
Fixed now.


fabian-s:
Sorry for being an idiot. continued at #401

PR #408, opened 2019-09-13
fix #366 : extend pipe-continuation linter to handle nesting

### LINKED ISSUE #366: pipes (%>%) in single lines get marked by lintr
I just installed this package and faced an issue with single line pipes. I have a function that basically looks like this.

```
myFun <- function() {
  a %>% b() %>% c()
}
```

`lintr::lint_package()` adds a style-marker here because of missing newlines. Is this behavior expected? Here is an actual repex: GregorDeCillia/suitr@4f6f97e

russHyde:
extends pipe-continuation linter to handle nesting

    fixes https://github.com/jimhester/lintr/issues/366

----

    pipe-continuation linter was throwing lints in nested expressions.
    For example, the nested pipeline in this function:
    
    ```
    f <- function() {
      this() %>% nested() %>% pipeline()
    }
    ```
    
    This happened because lintr was looking for 'an-expression spanning
    several lines, that contained multiple pipe characters on one of
    the lines'.
    
    In the above, the definition of `f` is a multi-line expression and
    the pipeline is flagged because it is part of that multi-line
    expression, even though the pipeline is contained within a
    single-line subexpression.

----
    
    Previous version of pipe-continuation linter threw lints on
    single-line pipelines that were nested inside an outer
    expression (eg, pipelines in the body of a function, or in a
    test_that clause)
    
    The previous version searched for any multi-line expression
    that contained several pipe-symbols on the same line.
    
    We now search for any pipe-expression, L %>% R, where;
    - no smaller expression surrounds `L %>% R` (to remove
      nesting problems)
    - L %>% R spans multiple lines; and
    - the expression L contains a pipe-symbol
    
    Then we filter these:
    - If R starts on the same line as the pipe-symbol, a lint
      is thrown if L ends on that line
    - If R starts on a subsequent line, then a lint is thrown
      if a pipe-symbol in the expression L sits on the same
      line as the pipe-symbol under scrutiny
    
    This has necessitated changing one of the existing tests
    
    `foo %>% bar %>% baz %>%\n qux()`
    
    only throws one pipe-continuation lint now (on the
    pipe-symbol before `qux`). Because of how the xml tree is
    constructed, the expressions in this are
    
    (foo)
    (foo %>% bar)
    ((foo %>% bar) %>% baz)
    (((foo %>% bar) %>% baz) %>% qux)
    
    Of these, the first three expressions are all single-line
    pipelines. I can't work out how to extract the whole of
    a pipeline as a single xml tree, in a way that doesn't
    cause problems with nested expressions. Hopefully this
    change to the tests is ok.
    
    Have modified the expected value for this test and have
    added a number of new tests for pipelines nested in other
    expressions.


jimhester:
Thanks! Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

russHyde:
@jimhester I've bumped the news. All the best

russHyde:
Apologies if the comments are a bit verbose, I was learning xpath while solving this. There may be more efficient ways to express some of the queries I wrote

russHyde:
Thanks @jimhester

PR #413, opened 2019-09-30
By default relative_path should be FALSE for lint_dir

arekbee:


jimhester:
Why do you think it should be `FALSE`? I think it should be `TRUE`.

arekbee:
> Why do you think it should be `FALSE`? I think it should be `TRUE`.

I wanted to open a discussion if default parameter should be TRUE or FALSE. More often I use lint_dir with absolute path then with relative path. If I am inside R project/package I would use lint_package, but if I am want to run it on multiple projects, different folders or outside package then I would use lint_dir. Of course these are only my preferences and don't want to have misleading default parameter for lint_dir and lint_package. Other solution could be that lint_dir will guess if you are using absolute or relative path and relative_path argument in function will be only for forcing the interpretation of path.

arekbee:
Old issue

PR #424, opened 2019-10-17
Set testthat dependency to 2.2.1

jeffwong-nflx:
`expect_lint_free` uses testthat::skip_on_covr, which was not introduced into testthat until 2.2.1

https://github.com/jimhester/lintr/blob/38279839efe0f685b281e80a77bcea1b04665f33/R/expect_lint.R#L110

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Imports:
     crayon,
     codetools,
     cyclocomp,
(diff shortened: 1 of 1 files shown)


jimhester:

Thanks!! You are da bomb!

jimhester:

Thanks!! You are da bomb!

PR #442, opened 2020-01-06
fix #433 : Double quotes in path linters

### LINKED ISSUE #433: absolute_path_linter only works on paths with single quotes
When I run the absolute_path_linter on my file, it only catches paths that are wrapped in single quotes, not double quotes. 

Example file:
```{r}
test_path1 <- 'C:/Documents'
test_path2 <- "C:/Documents"
```

Output:
```{r}
lint("lintr_testfile.R", linters = with_defaults(absolute_path_linter()))
```
* `Line 1 Only use double-quotes.`
* `Line 1 Do not use absolute paths.`
With no flags for the absolute path on line 2. 

Looking at the source code for absolute_path_linter, I think the issue comes from this line which does the unquoting: `path <- unquote(token[["text"]], "'")`

russHyde:
In issue #433 , filepaths that were surrounded with double-quotes were immune to the absolute-path-linter. This issue also affected the nonportable path linter.

Here, 

- some tests were added to ensure that double-quoted filepaths are linted by the path-linters.
- `absolute_path_linter` and `nonportable_path_linter` were refactored to place any duplicated logic in a function-builder (`make_path_linter`)
- strings that are surrounded by double (single, respectively) quotes were stripped of the surrounding double quotes and then the remaining double quotes were unescaped
- (outdated; unquote has not been modified) The latter change was achieved by modifying `unquote` to accept a vector of quote-characters;
- The change was achieved by getting the path-linter to select the appropriate quote-character before calling `unquote` (see below)

`unquote` is also called by some functions in `object_name_linter.R` (but defined in path_linter.R)
- hence `unquote` and some related functions were moved to utils.R

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,10 @@
 # lintr (development version)
 
+* `absolute_path_linter()` and `nonportable_path_linter()` now handle
+  file-paths that are wrapped with double-quotes (#433, #437, @russHyde).
 * `object_usage_linter` has been changed to ensure lint-position is indicated
   relative to the start of the file, rather than the start of a defining
-  function (#432, @russHyde)
+  function (#432, @russHyde).
 
 # lintr 2.0.0
 
--- a/R/path_linters.R
+++ b/R/path_linters.R
@@ -138,40 +138,37 @@ split_path <- function(path, sep="/|\\\\") {
   )
 }
 
-unquote <- function(str, q="`") {
-  # Remove surrounding quotes (select either single, double or backtick) from given character vector
-  # and unescape special characters.
-  str <- re_substitutes(str, rex(start, q, capture(anything), q, end), "\\1")
-  unescape(str, q)
-}
 
-escape_chars <- c(
-  "\\\\" = "\\",  # backslash
-  "\\n"  = "\n",  # newline
-  "\\r"  = "\r",  # carriage return
-  "\\t"  = "\t",  # tab
-  "\\b"  = "\b",  # backspace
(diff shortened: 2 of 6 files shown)


russHyde:
[the tests that fail in CI were already known about]

russHyde:
Since `unquote` (and implicitly unescape, excape_chars) are used by:

- functions in `path_linter.R` (`[absolute|nonportable]_path_linter`); and also
- functions in `object_name_linter.R` (`make_object_linter` (hence `object_name_linter_old` and `object_length_linter`) and `object_name_linter`)

but are defined in `path_linter.R`, `unquote` (et al) should be moved to `utils.R`

russHyde:
Jim appears to use `make_***_linter()` as the name for linter-builders (see `make_object_linter`), so rename `build_path_linter_function` as `make_path_linter`.

russHyde:
Rather than making `unquote` more complex (by allowing mutliple quote characters in the `q` argument), since the path linters only call `unquote` on IDs with the token "STR_CONST" (and on one ID at a time), we can assume that there will either be a single-quote or a double-quote at the start of the ID and so can choose the appropriate quote character before calling `unquote`.

This makes the definition of the `make_path_linter` more complicated, but should prevent the `unquote`-unit test failures

russHyde:
All tests now pass

russHyde:
Will do.

russHyde:
@jimhester I've rebased this onto master, added to NEWS, and squashed all the commits down to one. None of the code has changed since your comment, above. (But I need another OK before I can merge)

Issue #403, opened 2019-09-02
Error in doc_parse_raw: Opening and ending tag mismatch: expr line 28 and equal_assign [76]


step-:
In RStudio 1.2.1335 Linux, R 3.6.1:
```
> lintr::lint("R/lintr-bug.R")
Error in doc_parse_raw(x, encoding = encoding, base_url = base_url, as_html = as_html,  : 
  Opening and ending tag mismatch: expr line 28 and equal_assign [76]
```
lintr-bug.R attached. 
I had just built lintr from github with packrat. I'm attaching packrat.lock so you can see all package versions involved.
[attachments.zip](https://github.com/jimhester/lintr/files/3567094/attachments.zip)

russHyde:
Could you install `xmlparsedata` from github, eg, using `devtools::install_github("r-lib/xmlparsedata")`, please. On R.3.6 we were getting this `doc_parse_raw` error when parsing R code that contained equals-assignments; see here: https://github.com/r-lib/xmlparsedata/issues/5

step-:
Indeed `xmlparsedata` from github fixed this issue. Now lintr prints three markers "Use <- not = for assignment" and terminates correctly. Thank you.

csgillespie:
I'm now getting this error with R-devel on travis, but __not R-current (3.6). Any suggestions?

russHyde:
Do you have a travis log we could look at or is it a private repo?

csgillespie:
https://travis-ci.org/csgillespie/poweRlaw

russHyde:
@csgillespie had a look at your inteRgrate travis log. OTOH I'm not sure what the problem is.

Can I ask why you use `devtools::install(..., upgrade = "never")`.  Can you be sure you're using lintr>=2 and xmlparsedata >=1.0.3 in your travis run?.

FWIW: our logs looked like this when we had issues with `doc_parse_raw` https://travis-ci.org/jimhester/lintr/builds/561186819?utm_source=github_status&utm_medium=notification

russHyde:
Sorry, idsregard that. I can see the lintr 2 install in your poweRlaw log

russHyde:
As a minor point, you're installing two differrent versions of lintr when running CI on poweRlaw. The 2.0.0 version when travis works on the `r_packages` stub of your .travis.yml; then the dev version when you install `integRate`. So, the dev version should be responsible for the failing call.

russHyde:
[...]

PR #455, opened 2020-02-11
Fix the parsing of source-expressions on R-devel

russHyde:
The parsed_content generated in R-devel (to be R4.0) uses a token `expr_or_assign_or_help` in the same way that R3.6 used `equal_assign`. Parsing code with equals-assignments caused problems when running lintr on R-devel for this repo (see #403 ): https://github.com/csgillespie/poweRlaw/commit/ede33811f79376cb0a4ca29f8b0972a8b8b35910 

The existing `lintr` code had a function `fix_eq_assigns` that handled `equal_assign`; this PR just extends that code to handle `expr_or_assign_or_help` in the same way.

To work out what was going on, I extracted the contents of an lapply loop(in `get_source_expressions`) into a stand-alone function `get_single_source_expression` to aid debugging.

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -10,7 +10,6 @@ get_source_expressions <- function(filename) {
   source_file$content <- get_content(source_file$lines)
 
   lint_error <- function(e) {
-
     message_info <- re_matches(e$message,
       rex(except_some_of(":"),
         ":",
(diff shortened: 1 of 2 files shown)


russHyde:
@csgillespie this should fix the problem you raised in #403 . If you could do a quick review of the code that'd be really helpful.

russHyde:
TODO:
- [ ] collapse commits
- [ ] append to NEWS
- [ ] find out when `expr_or_assign_or_help` was added to parseData and why

jimhester:
As to the why, Tomas Kalibera sent this note a few weeks ago

> the R parser changed in 77722 (PR#16710), fixing the precedence of "?"
with respect to "=" to be as documented (lower) and as it seems it was
until 2001. The change affects these packages, which will start failing
their tests now:
> 
> lintr, parsetools, rlang, styler

csgillespie:
@russHyde Don't think that worked 

 - [travis](https://travis-ci.org/csgillespie/poweRlaw/jobs/648916668) output
 - [travis](https://travis-ci.org/csgillespie/poweRlaw/jobs/648916668/config) config

russHyde:
Haha, so the precedence has been incorrect for two decades?

Colin Gillespie flagged a problem in their CI runs yesterday

russHyde:
Hi @csgillespie thanks for the feedback. Could you change your config to 
```
cd /tmp/ && git clone https://github.com/russHyde/lintr.git && git checkout expr-or-assign-or-help
...
```

I ran `lint_package` on poweRlaw locally using R-devel and it ran (without the errors that current lintr master throws).

csgillespie:
@russHyde That works on travis - thanks

jimhester:
FWIW the easiest way to install this PR is putting `jimhester/lintr#455` in your `r_github_packages`

R/get_source_expressions.R
```diff
@@ -117,6 +100,30 @@ get_source_expressions <- function(filename) {
   list(expressions = expressions, error = e, lines = source_file$lines)
 }
 
+get_single_source_expression <- function(
```
jimhester: Generally I think we should follow the [tidyverse style guide](https://style.tidyverse.org/functions.html#long-lines-1) for this, which would be either

```r
get_single_source_expression <- function(loc, parsed_content, source_file, filename, tree) {
```

Or

```r
get_single_source_expression <- function(loc,
                                         parsed_content,
                                         source_file,
                                         filename,
                                         tree) {
```

R/get_source_expressions.R
```diff
@@ -280,7 +287,10 @@ fix_eq_assigns <- function(pc) {
   }
 
   eq_assign_locs <- which(pc$token == "EQ_ASSIGN")
-  if (length(eq_assign_locs) == 0L || "equal_assign" %in% pc$token) {
+  if (
```
jimhester: Maybe style this like this?

```r
if (length(eq_assign_locs) == 0L ||
   any(c("equal_assign", "expr_or_assign_or_help") %in% pc$token)) {
```

russHyde:
Thanks @jimhester

PR #457, opened 2020-02-13
lintr:::print.lint issue

jamieRowen:
fixed an issue that caused a
that caused an issue when printing results from the single_quotes_lintr
It was assumed in highlight_string that the range was such that
range[1] < range[2] which is not necessarily the case when
the string is split over a new line. Added a test case for this
too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # lintr (development version)
-
+* `single_quote_linter()` no longer causes a print issue when open quote 
+  appears at a column > than close quote (@jamieRowen)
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks! Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`. Then I can merge this.

jamieRowen:
Done, no issue number attached as don't think it was raised prior, just came across it earlier today

jimhester:
Thanks! 

Thanks a million!

jamieRowen:
No worries, it's a pleasure

PR #460, opened 2020-02-22
Use XDG config directory

klmr:
It would be great if the default lintr config would be preferentially looked up in the appropriate [XDG base configuration directory](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) instead of the user home directory. That is, if lintr defaulted to loading `${XDG_CONFIG_HOME-$HOME/.config}/lintr` (note lack of leading dot in the filename!) and only if that doesn’t exist, checked for `$HOME/.lintr`. This avoids cluttering the home directory with config files.

Of course this can already be achieved by appropriately setting the `lintr.linter_file` option. But defaults matter.

This PR implements the necessary change, and stays 100% backwards compatible. Note that this causes the cyclomatic complexity of the `find_config` function to exceed the linting threshold. I’m open for suggestions of how to change this — but I’m tempted to call it a false positive (and to add an appropriate `nolint` comment).

### DIFF
--- a/R/settings_utils.R
+++ b/R/settings_utils.R
@@ -69,7 +69,9 @@ find_config <- function(filename) {
     find_config2(path),
     # User directory
     # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
-    file.path(Sys.getenv("HOME", unset = "~"), linter_file)
+    file.path(Sys.getenv("HOME", unset = "~"), linter_file),
+    # Next check for a global config file
+    file.path(R_user_dir("lintr", which = "config"), "config")
   )
 
   # Search through locations, return first valid result
--- a/NEWS.md
+++ b/NEWS.md
@@ -122,6 +122,9 @@
 
 * `available_linters()` gives priority to `tags` over `exclude_tags` in the case of overlap. In particular, this means that `available_linters(tags = "deprecated")` will work to return deprecated linters without needing to specify `exclude_tags` (#1959, @MichaelChirico).
 
+* The {lintr} configuration file is now searched in the system's user configuration path; the lintr config filename can
+  also be configured explicitly by setting the environment variable `R_LINTR_LINTER_FILE` (#460, @klmr)
+
 ### New linters
 
 * `matrix_apply_linter()` recommends use of dedicated `rowSums()`, `colSums()`, `colMeans()`, `rowMeans()` over `apply(., MARGIN, sum)` or `apply(., MARGIN, mean)`. The recommended alternative is much more efficient and more readable (#1869, @Bisaloo).
--- a/R/settings.R
+++ b/R/settings.R
@@ -8,7 +8,8 @@
 #'  4. `linter_file` in the user home directory
 #'  5. [default_settings()]
 #'
-#' The default linter_file name is `.lintr` but it can be changed with option `lintr.linter_file`.
+#' The default linter_file name is `.lintr` but it can be changed with option `lintr.linter_file`
+#' or the environment variable `R_LINTR_LINTER_FILE`
 #' This file is a dcf file, see [base::read.dcf()] for details.
 #' @param filename source file to be linted
 read_settings <- function(filename) {
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -279,7 +279,7 @@ settings <- NULL
 .onLoad <- function(libname, pkgname) {
   op <- options()
   op_lintr <- list(
-    lintr.linter_file = ".lintr"
+    lintr.linter_file = Sys.getenv("R_LINTR_LINTER_FILE", ".lintr")
   )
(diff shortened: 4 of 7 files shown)


gaborcsardi:
How about the rappdirs package?

klmr:
Good idea! However, if I understand correctly rappdirs is particularly geared towards GUI applications, rather than cross-platform tools and/or code library packages. In particular, on macOS, `rappdirs::user_config_dir('lintr')` returns `/Users/‹user›/Library/Application Support/lintr`, which is generally *not* where a command line tool/library configuration would be expected to reside. Apple would probably disagree, but the overwhelming convention for such tools on macOS is to either follow XDG or to dump everything in the home dir.

jimhester:
I think using rappdirs is the best solution if we want to go down this route.

AshesITR:
Meanwhile, base R provides `tools::R_user_dir("lintr", "config")`, with `backports` providing this function to older R versions.

AshesITR:
@krlmlr Are you still interested in this? #1757 is about to change the config search behavior so now would be a good time to also include XDG_CONFIG_HOME if it still seems useful.

klmr:
@AshesITR (Wrong user tagged) Absolutely, and I had thought I had already submitted an updated PR. My bad! I’ll check the other PR and submit my changes ASAP!

AshesITR:
Thanks a ton and sorry for the wrong tag.
Maybe best to base against the new PR because it simplifies the code complexity by creating a candidate list.
This PR would then be as small as +1 line of code and some documentation around it.

klmr:
OK, I’ve added the necessary changes. This includes adding an environment variable to override the lintr config file location; I added this change because I think it belongs here: it is a widespread convention that the location of a configuration file can be overridden by an environment variable.

In the case of ‘lintr’ it made sense (I think) to make this environment variable act as the default value for `getOption('lintr.linter_file')`, rather than being yet another way of specifying the file location.

@AshesITR At your suggestion I rebased this PR onto #1757; however, this means that the outstanding changes to that PR will probably block this one.

R/settings_utils.R
```diff
@@ -68,24 +68,26 @@ find_config <- function(filename) {
...
+    file.path(path, linter_file),
+    # Next check for a file in the .github/linters directory
+    file.path(path, ".github/linters", linter_file),
+    # Next check for a file in higher directories
+    find_config2(path),
+    # Next check for a file in the user directory
+    # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
+    file.path(Sys.getenv("HOME", unset = "~"), linter_file),
+    # Next check for a global config file
+    file.path(R_user_dir("lintr", which = "config"), "config")
```
AshesITR, line 83: Maybe use `lintr_file` as the filename here as well?

Also, from the earlier discussion I though that XDG_CONFIG_HOME might also be a useful search location, right?

klmr: > Maybe use lintr_file as the filename here as well?

To clarify: you think the file path should be e.g. `~/.config/R/lintr/lintr_file` instead of `~/.config/R/lintr/config`? I’m happy with either but the latter is more common (e.g. Git has `~/.config/git/config`, etc).

> Also, from the earlier discussion I though that XDG_CONFIG_HOME might also be a useful search location, right?

`tool::R_user_dir` uses `XDG_CONFIG_HOME`, if set.

AshesITR: I was thinking it should be `getOption("lintr.linter_file")` like with all other candidate paths.

klmr: Hm, I disagree with this. Firstly, I don’t think there’s a benefit to making this configurable (the path *is* already configurable on multiple levels, including via `XDG_CONFIG_HOME` and `R_USER_CONFIG_DIR`).

And secondly, the filename given by `getOption("lintr.linter_file")` makes the file invisible on most platforms (leading dot by default). This is appropriate in the home directory or a project directory (…ish, to be honest). It is *not* appropriate or necessary in a dedicated config directory. Users (me included) might want to use an invisible filename for their projects (e.g. `.lintr`) but a regular file for the global configuration.

What do you think?

AshesITR: Okay, works for me.

AshesITR:
Could you write some tests to see if the XDG config is pulled as expected?

klmr:
@AshesITR Apologies, I had completely missed that. Test added.

AshesITR:
LGTM, thanks! Waiting for #1757 to hit the finish line and then rebasing #460 again.

klmr:
@AshesITR I’ve rebased the PR against the merged `main` branch now that #1757 has been merged. Please have a look.

AshesITR:
Sorry for taking so long to review, I'll try to get to this soon.

AshesITR (approved):

PR #462, opened 2020-02-24
Fix fix_eq_assigns with last token being EQ_ASSIGN

### LINKED ISSUE #461: Error values must be length 1 but FUN(X[[1]]) result is length 0
From https://github.com/REditorSupport/languageserver/issues/218:

Create a file `diagnostics.R` with the following content:

```r
sum(1:10, na.rm += 1)
```

The file is obviously not parse-able but calling `lintr::lint("diagnostics.R")` will result in the following error:

```
Error in vapply(eq_assign_locs, next_with_parent, pc = pc, integer(1)) : 
  values must be length 1,
 but FUN(X[[1]]) result is length 0
Backtrace:
1: vapply(eq_assign_locs, next_with_parent, pc = pc, integer(1))
2: fix_eq_assigns(fix_column_numbers(fix_tab_indentations(source_file)))
3: get_source_file(source_file, error = lint_error)
4: get_source_expressions(filename)
5: lintr::lint("diagnostics.R")
```

renkun-ken:
Closes #461 

This PR fixes the handling of source code such as 

```r
x=
```

```r
sum(1:10, na.rm += 1)
```

for which the parse data stops at the `=` in these cases where the last token is always `EQ_ASSIGN`. This PR adds a condition that if the last token is `EQ_ASSIGN` then `fix_eq_assigns` should not do anything because `next_with_parent` no longer works.

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -290,6 +290,7 @@ fix_eq_assigns <- function(pc) {
 
   eq_assign_locs <- which(pc$token == "EQ_ASSIGN")
   if (length(eq_assign_locs) == 0L ||
+    length(pc$token) %in% eq_assign_locs ||
     any(c("equal_assign", "expr_or_assign_or_help") %in% pc$token)) {
     return(pc)
   }
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -290,7 +290,7 @@ fix_eq_assigns <- function(pc) {
 
   eq_assign_locs <- which(pc$token == "EQ_ASSIGN")
   if (length(eq_assign_locs) == 0L ||
-    length(pc$token) %in% eq_assign_locs ||
+    nrow(pc) %in% eq_assign_locs || # check whether the equal-assignment is the final entry
     any(c("equal_assign", "expr_or_assign_or_help") %in% pc$token)) {
     return(pc)
   }
--- a/tests/testthat/test-commas_linter.R
+++ b/tests/testthat/test-commas_linter.R
@@ -43,8 +43,4 @@ test_that("returns the correct linting", {
   expect_lint("a[1, , 2, , 3]",
     NULL,
     commas_linter)
-
-  expect_lint("purrr::partial(list, 1, ... = , 2)",
-    rex("Commas should never have a space before."),
-    commas_linter)
 })
--- a/tests/testthat/test-commas_linter.R
+++ b/tests/testthat/test-commas_linter.R
@@ -43,4 +43,8 @@ test_that("returns the correct linting", {
   expect_lint("a[1, , 2, , 3]",
     NULL,
     commas_linter)
+  
+  expect_lint("purrr::partial(list, 1, ... = , 2)\n",
+    rex("Commas should never have a space before."),
+    commas_linter)
 })
--- a/tests/testthat/test-commas_linter.R
+++ b/tests/testthat/test-commas_linter.R
@@ -43,8 +43,8 @@ test_that("returns the correct linting", {
   expect_lint("a[1, , 2, , 3]",
     NULL,
     commas_linter)
-  
-  expect_lint("purrr::partial(list, 1, ... = , 2)\n",
+
+  expect_lint("purrr::partial(list, 1, ... = , 2)",
     rex("Commas should never have a space before."),
     commas_linter)
 })
--- a/tests/testthat/test-error.R
+++ b/tests/testthat/test-error.R
@@ -16,10 +16,16 @@ test_that("returns the correct linting", {
     b",
     rex("unexpected end of input"), (function(...) NULL))
 
-  expect_lint("x=\n",
-    rex("unexpected end of input")
+  expect_lint("x=",
+    list(
+      rex("Use <-, not =, for assignment"),
+      rex("unexpected end of input")
     )
-  expect_lint("sum(1:10, na.rm += 1)\n",
-    rex("unexpected '='")
+  )
+  expect_lint("x += 1",
+    list(
+      rex("Use <-, not =, for assignment"),
+      rex("unexpected '='")
     )
+  )
 })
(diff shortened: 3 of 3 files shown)


renkun-ken:
I'd like to also add some test cases for this but I'm not sure where's the best place to put them?

russHyde:
Hi, would you be able to check how this behaves for the uncompleted arguments in purrr::partial
```
my_list <- partial(list, 1, ... = , 2)
my_list("foo")
```

See the bottom of https://purrr.tidyverse.org/reference/partial.html

I'm not certain where would be best for the unit tests (initially thought that test-assignment_linter might have been a good place); does a spcific linter fail for your use case?

renkun-ken:
@russHyde  Do you mean the following?

```
> lintr::lint('my_list <- partial(list, 1, ... = , 2)\nmy_list("foo")\n')                                                                                                                                                        
<text>:1:35: style: Commas should never have a space before.
my_list <- partial(list, 1, ... = , 2)
                                 ~^
```

I thought linting code such as `x=` and `sum(1:10, na.rm += 1)` should be a straight error because they don't even parse correctly but they would cause lintr to produce error internally.

Are there any test cases for simple parse errors?

russHyde:
Sorry for any misunderstanding.
The dangling equals in `x=` makes for invalid R code (provided there's nothing on a subsequent line to be evaluated).
But `purrr::partial(list, 1, ... = , 2)` has a dangling equals in the arguments and is valid R code.

So, if you're aim is to ensure that
- `x=` 
- and `f(a += 1)` throw errors at the parsing stage (rather than during linting);

... could you add a test to ensure that `purrr::partial(list, 1, ... = , 2)` does not throw an error during parsing (even if it gets flagged by the default linters)*

----

* - I really don't like this `... =` syntax, but it's perfectly valid R code

renkun-ken:
It is interesting that linting and fixing code like `purrr::partial(list, 1, ... = , 2)` may end up in a loop like:

```
> lintr::lint('purrr::partial(list, 1, ... = , 2)\n')                                                                                                                        
<text>:1:31: style: Commas should never have a space before.
purrr::partial(list, 1, ... = , 2)
                             ~^

> lintr::lint('purrr::partial(list, 1, ... =, 2)\n')                                                                                                                         
<text>:1:29: style: Put spaces around all infix operators.
purrr::partial(list, 1, ... =, 2)
                            ^~
```

jimhester:
@russHyde when you think this is PR is ready feel free to merge it

R/get_source_expressions.R
```diff
@@ -290,6 +290,7 @@ fix_eq_assigns <- function(pc) {
 
   eq_assign_locs <- which(pc$token == "EQ_ASSIGN")
   if (length(eq_assign_locs) == 0L ||
+    length(pc$token) %in% eq_assign_locs ||
```
russHyde: Could you use `nrow(pc) %in% eq_assign_locs`, or add a comment to state that this is checking whether the equal-assignment is the final entry in the parsed content

tests/testthat/test-error.R
```diff
@@ -15,4 +15,17 @@ test_that("returns the correct linting", {
     function() {
     b",
     rex("unexpected end of input"), (function(...) NULL))
+
+  expect_lint("x=",
```
russHyde: If we modify the default linters, these tests might need rewriting. Rather than using all of the default linters in these tests, could you use a single specific linter, that *shouldn't* throw on the code, for example `equals_na_linter`. That way, only the parsing errors should be caught.

For example,
`expect_lint("x=", rex("unexpected end of input"), equals_na_linter)`

Also, could you add a couple of additional tests:
- `{x = }`
- and `x = ;`

should both throw parsing errors, but should throw no other linting errors when using `equals_na_linter`

tests/testthat/test-commas_linter.R
```diff
@@ -43,4 +43,8 @@ test_that("returns the correct linting", {
   expect_lint("a[1, , 2, , 3]",
     NULL,
     commas_linter)
+
+  expect_lint("purrr::partial(list, 1, ... = , 2)",
```
russHyde: Could you move this test into `test-error.R` as an example of some valid R code that has a (seemingly) incomplete equal-assignment statement. And rewrite it so that it tests that a parsing error does not throw:

```
# no parsing error is expected for the equals-assignment in this code:
expect_lint("purrr::partial(list, 1, ... = , 2)", NULL, equals_na_linter)
```

russHyde:
Thanks so much for this (especially since I was the last one to touch that bit of code)

I'm trying to summarise the logic behind the current implementation of `fix_eq_assigns`, and think it might be worth expanding the comment that precedes its definition.

```
`fix_eq_assigns` wraps equal assignments (x = 3) in a parent expression, so they have the same structure as the equivalent left / right assignment.

In R3.6 and above, equal-assignments are already wrapped in `equal_assign` or `expr_or_assign_or_help` expressions and need not be further modified.

Malformed code with incomplete assignments "x = " are not wrapped in an expression since no valid expression can be formed from these statements. The parsed code is
returned early, and handled by the caller.
```

The only discrepancy, is that other incomplete assignments "{x = }" and "x = ;" are not handled in the early-return, they are handled in the subsequent code for `fix_eq_assigns`, but I don't think that's a huge issue.

Could you add a couple more tests into test-error.R

R/get_source_expressions.R
renkun-ken: Sure

tests/testthat/test-error.R
renkun-ken: I was thinking about this too. Thanks for your nice suggestions!

tests/testthat/test-commas_linter.R
renkun-ken: Sure

renkun-ken:
Thanks for your nice suggestions! All suggested changes are made.

russHyde:
Thanks.

PR #467, opened 2020-03-16
Fix #441 : paren-brace, equals-NA and no-tab lints in strings or comments

### LINKED ISSUE #441: `paren_brace_linter` false positive within regex strings: "(match_group){quantifier}"
`paren_brace_linter` flags the "){" within regex definition below:

```
#' Checks if a file has a .tar or .tar.bz2 extension
#'
#' @noRd
.is_tar_file <- function(x) {
  grepl(pattern = ".*\\.tar(\\.bz2){0,1}", x)
}
```

Should `paren_brace_linter` be made more strict - so that it doesn't consider strings?

russHyde:
Some regex-based linters were throwing false-positives in irrelevant bits of source code (in strings, or in comments; see #441 ).
For example,
- the paren-brace linter was flagging the regex in `grepl("(iss){2}", "Mississippi")`,
- the equals-na linter was flagging `is.na(x) # test whether x == NA"`
- the no-tabs linter was flagging `multi_line_tabbed_string <- "lorem ipsum\n\tblah de blah"`

Tests based on these false-positives were added.

This PR abstracts some of the code that these regex-based linters use into a function "make_linter_from_regex"; this function can be tuned to ignore regex matches that occur in comments and/or string definitions. Then redefines `paren_brace_linter`, `equals_na_linter` and `no_tabs_linter` to use `make_linter_from_regex`.

I have had some trouble understanding comment-parsing and would like to continue working on this PR for a bit (I'm shifting it off my work computer since I might not be able to access my office for a while)


jimhester:
The main question I guess is if these linters should remain using regexes. The no tabs linter I guess has to, as you don't have access to whitespace in the AST. But definitely the `equals_na` linter could be implemented by looking at the AST and probably the paren brace linter as well, which would likely be more robust than relying on regexes for their behavior.

gaborcsardi:
I think even for the whitespace linter, it is a better solution to create the AST, and then get all the positions that are not accounted for by terminal tokens, plus the trailing whitespace. 

Otherwise it is pretty difficult to handle this with regular expressions, especially with the new raw character constants.

russHyde:
Totally agree. I originally tried (and stuggled) to implement the paren-brace linter using xpath; I'll have a look at it again

gaborcsardi:
> Totally agree. I originally tried (and stuggled) to implement the paren-brace linter using xpath; I'll have a look at it again

Let me know if you need help with the xpath way, it is really difficult to use it AFAIR. Unfortunately, apart from xpath, there isn't much software out there for attributed sub-graph isomorphism.

russHyde:
Thanks

russHyde:
The the equals-NA linter has been reimplemented using xpath. I will try to do the same for paren-braces linter (but will leave the tab-linter as is). 

xpath seems like a multi-dimensional regex.

Should we be concerned that making xpath-searches the norm for linter implementations may raise the bar for new contributors? As such, I'd still like to sure up the code for building linters based on regexes in a way that can optionally exclude comments, string literals etc. Will continue to work on this.

----

I'm not sure why the Travis run isn't being reported in the PR checks (that is, whether it's a github or a travis issue): it can be found here https://travis-ci.org/github/jimhester/lintr/builds/668856106

Observed failures in Travis on R3.3 and R3.2 because Rcpp-compilation fails (something I've recently seen for my own (lintr-dependent) package; I just dropped the tests/support for R < 3.4 in dupree when that happened).

jimhester:
I agree with you about using XPath making it more difficult for contributors. That is also the reason I didn't use C or C++ in lintr, I wanted to lower the bar to contributors. Unfortunately R is not terribly well suited to this particular task and the performance is not great, which drives us towards more esoteric solutions like XPath...

Unfortunately I don't have any great solutions :/

gaborcsardi:
Well, I would say that, yes, xpath is more difficult to write. But imo it is still easier to write a _correct_ linter with xpath than with regexes. Especially with the new raw string that is coming in R 4.0.0.

(Being the author of the xpath based search I am probably a bit (?) biased. :)

jimhester:
I agree that xpath is definitely better than regexes

russHyde:
@jimhester 
I've updated the code.
- No-Tab linter is still defined using a regex (and tabs within strings are ignored by this regex)
- Both paren-brace and equals-NA linter are defined using xpath searches

I have added tests that should capture all the original bugs (paren-braces in comments or strings, equals-NA in comments or strings, tab-prefixed lines in a multi-line string) and a couple of additional tests.

The travis run is here: https://travis-ci.org/github/jimhester/lintr/builds/671546861
It still fails on R3.2 and 3.3; due to an Rcpp installation issue

MichaelChirico:
superseded by #620 (starts from the same history, just from a branch originating on the main repo)

PR #476, opened 2020-04-02
Make `lint_package()` also lint vignettes and data-raw and make `lint_dir()` lint all lintable files by default (#447)

### LINKED ISSUE #447: lintr::lint_package should lint vignettes by default
Since linting is possible for vignettes, it should be the default for `lintr::lint_package()` to also lint package vignettes.

This line just needs `"vignettes"` as an additional directory and `"\\.[Rr](md|)$"` as a pattern for that to work:

https://github.com/jimhester/lintr/blob/dbc76246c244e89f04be3031ff6db1dbe26960ac/R/lint.R#L238

AshesITR:
I tried updating NEWS.md as well - not sure if that's how you like it.


R/lint.R
```diff
@@ -151,7 +152,8 @@ reorder_lints <- function(lints) {
 #'   )
 #' }
 #' @export
-lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = NULL, pattern = rex::rex(".", one_of("Rr"), end), parse_settings = TRUE) {
+lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = NULL,
+                     pattern = rex::rex(".", one_of("Rr"), "" %or% "md" %or% "nw", end), parse_settings = TRUE) {
```
jimhester, line 156: This gives you a little nicer regex.

```suggestion
                     pattern =  rex::rex(".", one_of("Rr"), or("", "md", "nw"), end), parse_settings = TRUE) {
```

AshesITR: Thanks, I'm not used to `rex::rex` at all :) The new one is just two characters longer than what I would produce by hand. Didn't notice how using `%or%` created two unnamed capture groups.

AshesITR:
Is there a way to force-restart travis? It seems to have hung on something.

AshesITR:
Ah, forgot to re-document the suggested change. Now all looks good.
NB codecov/patch complains about the line break in the function definition of `lint_dir`.

jimhester:
The codecov patch check is failing because there isn't a test which is exercising the changed code. If you would like you can try writing one...

PR #494, opened 2020-05-15
Add SNAKE_CASE style option

AshesITR:
I like to write variables in snake_case except when they are CONSTANTS.
To prevent the `object_name_linter` from complaining, I would like to allow (uppercase) "SNAKE_CASE" as an option.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * `extract_r_source` handles Rmd containing unevaluated code blocks with named
   format specifiers (#472, @russHyde)
+* New style SNAKE_CASE for `object_name_linter()` (#494, @AshesITR)
 
 # lintr 2.0.1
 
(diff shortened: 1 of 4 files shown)


russHyde:
Could you add several tests to ensure this works as anticipated and remake the docs to ensure SNAKE_CASE is added to the help page for `object_name_linter`.

It looks like it will match "MY_var" but not "DONT_MUTATE_MY_BIG_TABLE" at present.

AshesITR:
Thanks, will do once I have access to a RStudio instance.

AshesITR:
@russHyde It appears that man/ is .gitignored, but locally the docs got updated to include

```
A subset of ‘CamelCase’, ‘camelCase’, ‘snake_case’, ‘SNAKE_CASE’, ‘dotted.case’, ‘lowercase’, ‘UPPERCASE’. A name should match at least one of these styles.
```

I had to manually delete man/linters.Rd, though. Roxygen didn't automatically detect the change.

russHyde:
That's funny. I can't see man/ in the .gitignore for lintr.

AshesITR:
Ah, the .Rd files are actually the same:

```tex
\item{styles}{A subset of
\Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
match at least one of these styles.}
```

Anything else needs updating?

russHyde:
lgtm. recommend merging if that's okay with you @jimhester .

jimhester:
Merge away! Also feel free to merge other PRs in the future without needing to ping me unless you need an extra set of eyes

russHyde:
Thanks @AshesITR .

PR #503, opened 2020-06-25
Add text argument to lint

### LINKED ISSUE #475: Add text argument to lintr::lint
In languageserver, we have to pass the file content to `lintr::lint()` (since the file is often being edited but not saved, and we do provide live diagnostics on user input) but also want it to behave as if the content is from a filename so that the following cases could work:

* `linter_file` is respected (https://github.com/REditorSupport/languageserver/issues/235)
* `object_usage_linter` detects its package (https://github.com/REditorSupport/languageserver/issues/238)
* Exclusions in `.lintr` not working (https://github.com/REditorSupport/languageserver/issues/282)

One bad solution is to write a temp file inside the user workspace (e.g. original file `test.R` and we write a temp `test.Rtmp`) but Language Server Protocol seems to discourage any change in user workspace directly rather than via LSP communication.

Therefore, I'm wondering if we could add a `text` argument so that we could use 

```r
lintr::lint(filename, linters = NULL, cache = FALSE, ..., parse_settings = TRUE, text = NULL)
```

by supplying both `filename` and `text`:

```r
lintr::lint(filename, text = text)
```

so that both are respected and `lint` wouldn't have to read the file again if `text` is supplied.

renkun-ken:
Close #475 

This PR adds a `text=` argument to `lintr::lint` so that user could lint a file with supplied `text` as if the file is being linted. This is particularly useful for code editors to lively lint a file being edited but not saved while keeping the `.lintr` config (if found) effective.

* [x] Add `text=` argument to `lintr::lint`
* [x] Test cases


AshesITR:
If `filename` contains a newline, it's treated as source code.
I like an explicit `text` argument more, but I think we should remove the implicit and undocumented `inline_data` feature if there is a proper `text` argument.

https://github.com/jimhester/lintr/blob/42b49c1ba6d74a4910419d7ef78df804914d2b4e/R/lint.R#L30-L36

MichaelChirico:
@renkun-ken is this ready for review? if so please help to merge to HEAD :smile:

renkun-ken:
I update this PR to latest master. 

> If `filename` contains a newline, it's treated as source code.
> I like an explicit `text` argument more, but I think we should remove the implicit and undocumented `inline_data` feature if there is a proper `text` argument.
> 
> https://github.com/jimhester/lintr/blob/42b49c1ba6d74a4910419d7ef78df804914d2b4e/R/lint.R#L30-L36

I guess we might need to keep it that way for a while and might give a warning in the future if `filename` is detected to be used as inline code in the future. In languageserver, we use this `filename` as code for quite a while.

Now the behavior is:

1. If `filename` has a `\n` then it is regarded as code and `filename` becomes `tempfile()`.
2. If `text` is supplied, change `filename` to `tempfile()` if `filename` is missing, then lint `text` as if it is from `filename`.

If you think this behavior makes sense, then I'll update some test cases to fix the tests.

renkun-ken:
I test the latest commit with languageserver using the `text=` argument, the linters that rely on `filename` (e.g. `object_usage_linter`) work now.

renkun-ken:
> @renkun-ken is this ready for review? if so please help to merge to HEAD 😄

Yes, please. This PR is ready for review and merge now.

MichaelChirico:
please remove .DS_Store

R/get_source_expressions.R
```diff
@@ -39,24 +41,20 @@
...
-      }
-    }
-  )
+  source_file$lines <- if (is.null(lines)) {
+    read_lines(filename)
+  } else {
+    lines
+  }
+
+  terminal_newline <- !identical(attr(source_file$lines, "terminal_newline", exact = TRUE), FALSE)
```
MichaelChirico, line 57: !identical(., FALSE) seems overly complicated... isn't that just `isTRUE`?

R/lint.R
```diff
@@ -24,6 +24,9 @@ NULL
 #' character string, store the cache in this directory.
 #' @param ... additional arguments passed to \code{\link{exclude}}.
 #' @param parse_settings whether to try and parse the settings.
+#' @param text the string or lines to be used as if linting \code{filename}.
```
MichaelChirico, line 27: This reads a bit strange to me... how about

> Optional argument for supplying a string or lines or lines directly, e.g. if the file is already in memory or linting is being done _ad hoc_. If `text` is supplied...

R/get_source_expressions.R
renkun-ken: It is possible that the input does not have `terminal_newline` at all if the `lines` does not come with the attribute. Therefore I regard all cases other than `FALSE` as `TRUE`, rather than all cases other than `TRUE` as `FALSE`.

R/lint.R
```diff
@@ -34,18 +37,37 @@ NULL
...
+  if (is.null(text)) {
+    inline_data <- rex::re_matches(filename, rex::rex(newline))
+    if (inline_data) {
+      text <- gsub("\n$", "", filename)
+      filename <- NULL
+    }
+  } else {
+    inline_data <- TRUE
+    if (length(text) > 1) {
+      text <- paste0(text, collapse = "\n")
```
MichaelChirico, line 51: nit: i usually prefer `paste` to `paste0` when we're just aggregating. I don't think there's any technical difference, but I feel like I have to add "`sep=""`" mentally and then remove it when I see there's only one argument, if that makes sense?

R/lint.R
```diff
@@ -34,18 +37,37 @@ NULL
...
+    }
+  }
+
+  no_filename <- missing(filename) || is.null(filename)
+
+  if (inline_data && no_filename) {
     filename <- tempfile()
-    on.exit(unlink(filename), add = TRUE)
```
MichaelChirico, line 43: why remove `add=TRUE`? is there no other `on.exit`?

R/get_source_expressions.R
MichaelChirico: How about `!isFALSE` then?

R/lint.R
renkun-ken: Thanks, it is more clear in this way.

R/lint.R
renkun-ken: Yes it is more readable and clear about the intended behavior.

R/get_source_expressions.R
renkun-ken: I didn't use `isFALSE` because I'm not sure if we use backports or something to ensure compatibility with older R versions before `isFALSE` was introduced in 3.5. Do we have a policy on that?

R/lint.R
renkun-ken: Just my negligence, sorry, reverted to using `on.exit(add=TRUE)` as before.

R/lint.R
```diff
@@ -68,9 +91,13 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
 
   if (length(cache_path)) {
     lint_cache <- load_cache(filename, cache_path)
-    lints <- retrieve_file(lint_cache, filename, linters)
+    lint_obj <- if (is.null(text)) filename else list(
```
MichaelChirico, line 94: nit: i would either put this all on one line (it looks like it fits in 120) or else but the `if` branch on its own line for readability (easier to read when both branches have similar indentation)

R/lint.R
```diff
@@ -24,6 +24,10 @@ NULL
 #' character string, store the cache in this directory.
 #' @param ... additional arguments passed to \code{\link{exclude}}.
 #' @param parse_settings whether to try and parse the settings.
+#' @param text Optional argument for supplying a string or lines or lines directly,
+#'   e.g. if the file is already in memory or linting is being done ad hoc.
+#'   If \code{text} is supplied but \code{filename} is missing, then \code{tempfile()}
+#'   will be used as the \code{filename}.
```
MichaelChirico, line 30: it looks like `<text>` will be used as the file name, this reads like the output would have the tempfile path instead... could you clarify what we mean here?

MichaelChirico:
almost there, thanks!

R/lint.R
renkun-ken: Sure

R/get_source_expressions.R
MichaelChirico: Hmm let's just stick with adding a comment

R/lint.R
renkun-ken: It is indeed a bit ambiguous here. Here I mean that if `text` is provided but `filename` is not, then we'll write the text to a temp file (same behavior as before when `filename` is detected to be code), but the lint result will show as `<text>` which I believe should be the desired output rather than showing the tempfile in the lint result.

Looks like I should update this doc here, maybe just removing the `text` without `filename` case in the doc looks enough?

renkun-ken: `a string or lines or lines directly` looks a bit redundant?

MichaelChirico: I think so. Removed the last sentence

MichaelChirico (dismissed):

R/get_source_expressions.R
renkun-ken: Sure

tests/testthat/test-lint_file.R
```diff
@@ -119,6 +119,7 @@ test_that("lint() results from file or text should be consistent", {
     "x<-1",
     "x+1"
   )
+  on.exit(unlink(file))
```
renkun-ken, line 122: I didn't use `on.exit()` here because I want to ensure that lintr does not assume that the file exists when linting with both `filename` and `text` supplied.

MichaelChirico: ah ok, please revert

MichaelChirico (approved):

renkun-ken:
Thanks for the review!

renkun-ken:
Looks like we forgot to add a news item. I'll create a quick PR for this soon.

PR #514, opened 2020-08-20
Add dim() to the set of bad functions for seq_linter

### LINKED ISSUE #513: seq_linter should also catch 1:dim(x)[y]
This is conceptually the same problem as nrow() and ncol(), since

```
dim(data.frame())
#> [1] 0 0
```

michaelquinn32:
Fixes #513


jimhester:
Does this also lint `1:dim(x)[[1]]`?

michaelquinn32:
That's a really good point! I added a test for that case as well. We're still green. Thanks!

jimhester:
This looks fine to me, @AshesITR feel free to merge once you are satisfied it looks good.

michaelquinn32:
@AshesITR Thanks for the suggestion!

The data type of the index value should be independent from the functionality of the linter. The change only extends the XML query to look for dim(), which when parsed happens slightly deeper in the code tree.

How about we switch one of the two new tests to use an explicit integer value instead of a generic numeric value. Would that work for you?

AshesITR:
Sounds good to me, thanks :)

michaelquinn32:
Added some more integer checks. Jim, the Travis issues are unrelated. There's a "too many requests" error during installation. Otherwise, mind taking one last look? Thanks!

AshesITR:
LGTM @jimhester can you restart the failed travis builds? Thanks!

AshesITR (approved):

AshesITR (approved):

PR #524, opened 2020-09-26
clarify mismatched start/stop exclusion

MichaelChirico:
A user seeing this may not have called `lintr` directly (e.g. it's part of a suite of checks), so specifically calling out that it's related to linting will be helpful.

Also wanted to get the pluralization right, and specifically referring to "ranges" is a further clarification.

I think `covr` has a similar sequence, I will file there shortly.


jimhester:
Thanks! You are da bomb!

jimhester (approved):

PR #526, opened 2020-10-01
WIP: seq_linter detects 1:(length(x) - 1) as well

### LINKED ISSUE #525: 1:length(x) linter could also detect 1:(length(x) - 1L)-alike expressions
I think `(length(x) - literal_constant)` usages are frequent enough and also easy to convert to a `seq_len` call.

I don't think it makes sense to go to crazy with this -- just looking for the second argument to `:` to be `+` or `-` and `ncol/length/nrow` and a literal constant (no variables even right?).

MichaelChirico:
Closes #525 

Currently migrating to use parsed_content instead of the XML approach (may or may not be advisable).

Next will be to extend the logic for the `1:(length(x)-1)` case

### DIFF
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -8,51 +8,56 @@ seq_linter <- function(source_file) {
 
   if (!length(source_file$parsed_content)) return(list())
 
-  xml <- source_file$xml_parsed_content
+  content <- source_file$parsed_content
+  seq_idx <- which(content$token == "':'")
+  n_seq <- length(seq_idx)
+  if (n_seq == 0L) return(list())
 
+  # (probably) overallocate for now
+  out <- vector('list', n_seq)
   bad_funcs <- c("length", "nrow", "ncol", "NROW", "NCOL", "dim")
-  text_clause <- paste0("text() = '", bad_funcs, "'", collapse = " or ")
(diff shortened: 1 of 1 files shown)


jimhester:
`1:(length(x) - 1` is not good though, you should use `seq_along()`. `1:(length(x) - 1` does the wrong thing when the length of x is 0.

jimhester:
Or is that what you are saying, it currently doesn't detect `1:length(x) - 1`?

MichaelChirico:
I see, I totally misunderstood the point of `seq_linter` (assuming it was for efficiency :p )

Do think it's worthwhile to try and detect this anyway?

AshesITR:
AFAIK `seq_along` is preferrable in the following cases:

 * `1:length(x)` for any `x`
 * `1:ncol(x)` for data.frames `x`, but not for matrix `x` (in that case, `seq_len(ncol(x)) != seq_along(x)`)

Since `1:length(x)` is probably a very common beginner mistake, I'd support a different error message if and only if the LHS is 1 (or 1L or 1.0) and the RHS is `length(...)`.

MichaelChirico:
Now that I've been using `lintr` more, I don't think this fix is necessary; closing.

PR #529, opened 2020-10-03
Clarify seq_linter message

### LINKED ISSUE #522: Error message for seq_linter.R
Error message for seq_linter is slightly unclear - it implies that nrow() is not allowed.  

(I will provide suggestion for revision later.)

MichaelChirico:
Closes #522

I think basically we need to convey why seq_len is being suggested. I opted for a succinct version that works for all of the `bad_funcs` covered here. I think we could do better with a longer message, but I don't know how much you prefer the messages as terse as possible.


AshesITR:
NB `rex()` auto-escapes text, so the test fixes are broken, because the actual regex contains `\.\*` instead of `.*`.

jimhester:
you can use the `anything` shortcut to insert `.*` in the rex expression.

MichaelChirico:
Thanks for the `rex` pointers, it's passing now

AshesITR (approved):

jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

MichaelChirico:
NEWS added

jimhester:
Thanks a watt!

PR #533, opened 2020-10-12
improve spaces_left_parentheses_linter for some edge cases

### LINKED ISSUE #508: Incorrect "Place a space before left parenthesis, except in a function call."
Hello, I believe this is still an issue:

```
x^(y + z)
```

`lintr_2.0.1` still says 

> Place a space before left parenthesis, except in a function call.

_Originally posted by @bersbersbers in https://github.com/jimhester/lintr/issues/302#issuecomment-660983194_

MichaelChirico:
Closes #508 

Actually sort of two related issues with `spaces_left_parentheses_linter`

 1. caret (`^`) is not excluded from operators allowing a preceding `^`
 2. unary plus/minus are not excluded

The former is easy to fix (h/t @AshesITR), the latter was a bit more involved (there may be an easier way to check whether the `+`/`-` is unary)


AshesITR:
I suspect checking

```r
with(
  source_file$parsed_content,
  sum(parent == parent[source_file$parsed_content[id, "parent"] == id]) == 2L
)
```

To find unary operators might also work. Can you try that (instead of `nrow(source_file$parsed_content[is_sibling_expr, ]) == 1L)`)?
That should result in more compact code

MichaelChirico:
Is there a goal to migrate logic to be using the XPath tree where possible (or at least, where the `parsed_content` approach is not extremely simple)? If so I could also rewrite this that way now that I'm more comfortable w that approach.

jimhester:
Yeah, using XPath for this would be great!

MichaelChirico:
OK, have filed #628 as a follow-up to this; let's just merge this for now not to confuse the issues.

AshesITR (approved):

MichaelChirico:
Strange that `travis-ci` is still showing up. Maybe closing & reopening will help.

MichaelChirico:
Seems no... @jimhester is there anything to tweak in the repo settings? or are we hitting some cache issue

jimhester:
I just deleted the travis webhook, so they should no longer show up.

PR #537, opened 2020-10-18
vectorized approach for single_quotes_linter

MichaelChirico:
I tried re-writing `single_quotes_linter` using an XML XPath approach.

To compare performance, kept the HEAD version installed and then `source()`d the new version as `single_quotes_linter2` and ran the linter on all the R files in my `github` directory and all the R files in base R's source; it's 5-20% faster:

```
r_scripts = list.files('~/github', pattern = '\\.R$', full.names = TRUE, recursive = TRUE)
cat(length(r_scripts), '\n')
# 437
system.time(lapply(r_scripts, lint, single_quotes_linter))
#   user  system elapsed 
# 75.018   0.362  75.441 
system.time(lapply(r_scripts, lint, single_quotes_linter2))
#   user  system elapsed 
# 60.555   0.219  60.763 

base_r_scripts = list.files('~/svn/R-devel', pattern = '\\.R$', full.names = TRUE, recursive = TRUE)
cat(length(base_r_scripts), '\n')
# 997
system.time(lapply(r_scripts, lint, single_quotes_linter))
#   user  system elapsed 
# 76.986   0.335  77.319 
system.time(lapply(r_scripts, lint, single_quotes_linter2))
#   user  system elapsed 
# 71.869   0.295  72.156 
```

The difference is not blazingly fast so understand if it's not merged but since I went through writing it I figured it can't hurt to file.


jimhester:
I am surprised it is faster than the regex based linter actually. I would expect the XML approach to be faster than linters that are relying on the parse tokens, but generally the regex linters are pretty fast already.

MichaelChirico:
I think that's why it's only slightly faster. At a glance, I think the `HEAD` version is not vectorized enough:

```
  lapply(ids_with_token(source_file, "STR_CONST"),
    function(id) {
      parsed <- with_id(source_file, id)
      if (re_matches(parsed$text, rex(start, single_quote, any_non_double_quotes, single_quote, end))) {
        Lint(
          filename = source_file$filename,
          line_number = parsed$line1,
          column_number = parsed$col1,
          type = "style",
          message = "Only use double-quotes.",
          line = source_file$lines[as.character(parsed$line1)],
          ranges = list(sort(c(parsed$col1, parsed$col2))),
          linter = "single_quotes_linter"
          )
      }
    })
```

IIUC `rex` will be called once per string constant, as opposed to being called once on all string constants

jimhester:
yeah that sounds correct to me, I would expect the vectorized version to outperform the XPATH implementation.

MichaelChirico:
I'll test that out and update here

R/single_quotes_linter.R
```diff
@@ -2,20 +2,33 @@
...
-          line_number = parsed$line1,
-          column_number = parsed$col1,
+          line_number = line1,
+          column_number = col1,
           type = "style",
           message = "Only use double-quotes.",
-          line = source_file$lines[as.character(parsed$line1)],
-          ranges = list(sort(c(parsed$col1, parsed$col2))),
+          line = line,
+          ranges = list(c(col1, col2)),
```
MichaelChirico, line 28: note the difference in logic here. instead of landing on the final column, land on either the final column, or the end of the first line, whichever comes first.

lmk if (1) this is as intended and/or (2) we should write a test for the difference.

MichaelChirico:
Indeed a fully vectorized approach is fastest. `2` here is the vectorized approach, and `3` is that using XML:

```
system.time(lapply(r_scripts, lint, single_quotes_linter))
#    user  system elapsed 
#  59.590   0.060  59.638 
system.time(lapply(r_scripts, lint, single_quotes_linter2))
#    user  system elapsed 
#  52.805   0.052  52.847 
system.time(lapply(r_scripts, lint, single_quotes_linter3))
#    user  system elapsed 
#  54.083   0.044  54.118 

system.time(lapply(r_scripts, lint, single_quotes_linter))
#    user  system elapsed 
#  59.846   0.080  59.915 
system.time(lapply(r_scripts, lint, single_quotes_linter2))
#    user  system elapsed 
#  53.480   0.060  53.529 
system.time(lapply(r_scripts, lint, single_quotes_linter3))
#    user  system elapsed 
#  55.962   0.064  56.016 
```

Overall, 8-10% better than `HEAD` now (that 20% I got yesterday appears to have been a fluke -- perils of doing an end-to-end benchmark instead of focusing on the core change)

Meta-question -- out of curiosity, if we're going to used the `parsed_content` `data.frame`, is there a situation where we would prefer to use the expression-level linters instead of the file-level lint? I'm not quite following the logic of having both an expression-level pass over the file and a full-tree pass

R/single_quotes_linter.R
jimhester: I think how you have it seems ok, but probably worth having a test case

MichaelChirico: OK done. I adjusted one of the tests to include a check on the `ranges` output.

jimhester:
You should really use the expression level ones, that is how the cached lints work, only the expressions that have changed since the last linting are re-linted.

The whole file expressions are only used by a few linters which cannot be done with single top level expressions, or are fast enough that it isn't really worth caching them.

MichaelChirico:
oh that makes perfect sense. anything in mind for what counts as "fast
enough" btw?

PS I know this would be a bigger project, but would it make sense to build
towards either having _cache versions of each linter, or to pass cache as
an argument to the linter which basically switches between file- and
expression-level linting? sadly neither sound trivially backwards-compatible

MichaelChirico:
Ran this vs `HEAD` on ~1000 random CRAN packages; improvement is ~11.5%

```
library(lintr)
library(rex)
library(microbenchmark)
system_time = function(expr) {
  t0 = get_nanotime()
  expr
  get_nanotime() - t0
}
cran_packages = list.files(
  '/media/michael/69913553-793b-4435-ac82-0e7df8e34b9f/cran-mirror/src/contrib',
  full.names = TRUE, pattern = '\\.tar\\.gz$'
)
cran_sample = sample(cran_packages, 1000L)
old_time = 0
new_time = 0
dir.create(tdir <- file.path(tempdir(), 'src'), showWarnings = FALSE)
for (pkg in cran_sample) {
  pkg_nm = gsub('_.*', '', basename(pkg))
  if (match(pkg, cran_sample) %% 50L == 0L) {
    cat('Running on', pkg_nm, '\n')
  }
  untar(pkg, files = paste0(pkg_nm, '/R/*'), exdir = tdir, extras = '--wildcards')
  src_f = list.files(tdir, pattern = '\\.R$', recursive = TRUE, full.names = TRUE)
  tryCatch({
    # run this quickly to error out before running timing on either expr, see lintr#541
    lapply(lapply(src_f, readLines), nchar)
    old_time = old_time + system_time(lapply(src_f, lint, single_quotes_linter))
    new_time = new_time + system_time(lapply(src_f, lint, single_quotes_linter2))
  }, error=identity)
  unlink(file.path(tdir, pkg_nm), recursive=TRUE)
}
cat('Relative timing (new/old):', round(new_time/old_time, 3), '\n')
# Relative timing (new/old): 0.885 
```

jimhester (approved):

AshesITR (approved):

PR #538, opened 2020-10-21
added assignment_spaces lintr

f-ritter:
We have a former SPS programmer in our team, who likes to define variables like this:
image
He doesn't notice all the time so I created this check. I didn't add it to the default_linters.
Let me know if I can improve anything.


AshesITR:
Thanks for taking the time to make a Pull Request :)
Could you add a negative test case (i.e. "x <- 1" doesn't throw a lint)?

f-ritter:
I added all negative test cases and simplified the regex. Couldn't think of a case where it would be a false positive, is there?

AshesITR:
The simplified regex looks like there will be a false positive for indented code.
Can you try adding

```r
if (0 < 1) {
  x <- 42L
}
```

as a test case (no lints expected, if I understand your intentions)

Also, bad variable names might confuse the linter, such as

`"my  =  bad  =  variable  =  name" <- 2.0`

By the way: Your teammates code is okay with [the tidyverse style guide](https://style.tidyverse.org/syntax.html#extra-spaces)

f-ritter:
Thanks for that edge case, you're right. What do you think about my new check?

Yeah, thats why I didn't add it to the default cases, but we want everyone in the team to use the same style.

AshesITR:
I was thinking more like

```r
expect_lint("\"my  =  variable\" <- 42.0", NULL, assignment_spaces)
```

AshesITR (approved):

PR #539, opened 2020-10-21
Support excluding entire directories

### LINKED ISSUE #438: Exclusions read from .lintr are depedent of the current working directory

### LINKED ISSUE #518: Allow exclusion of directories
When using `renv` in a project, `lintr::lint_dir()` will try to lint all packages installed in the library (`renv/library/...`)
Currently, one can only exclude single files, so there is no practical way to lint projects using `renv`.

A remedy would be support for excluding entire directories from linting.

AshesITR:
fixes #518 
fixes #438 

- [x] implement
- [x] test
- [x] document

The current solution is a bit inefficient since you could technically avoid listing the ignored directory.
Instead, we add full file exclusions for all files in the directories.


AshesITR:
also needed to fix #438 for the .lintr config file based test case to succeed.

AshesITR:
cc @russHyde @jimhester My hacky fix to #438 isn't quite as nice as #439 (currently not merged)
Do you have a preference regarding the approach to relative exclusions?

The approach in this PR is to temporarily setwd() in the only place where a directory prefix is necessary and set it back on.exit, namely when reading settings. #439 instead introduces a new argument to `normalize_exclusions()` to carry the path information.

We could also merge #439 first but its diff is kind of large at a quick glance.

russHyde:
Hi @AshesITR , I've been quiet on the lintr front for 6 months (job & family reasons). I never actually finished #439 ; my changeset was pretty large to begin with, and a collaborator pushed some additional changes onto my fork of lintr; so the changeset got really large. I'd like to split it into smaller chunks so that it's more easy to review, but I'm not sure where to start.

I'll try to look over your changes in the coming week (haven't looked at them yet), but would urge against using `setwd` in lintr (& any CRAN-targetted code). Thanks for looking into directory exclusions, to be able to drop directories from the config would be really valuable.

AshesITR:
A rebase to master caused some unexpected diffs...

Thanks @russHyde for the feedback so far - I've removed the call to `setwd()` instead conditionally adding `root` to the file path if it's not already an absolute path.

russHyde (changes requested):
Thanks for the PR, and for requesting my review. I like how you approached excluding directories. I've requested a couple of changes.

I started a PR that aimed to address the relative-path exclusions issue in #438. I'd like to check whether the tests that were written for that PR are all adressed in this PR before continuing.

tests/testthat/test-assignment_linter.R
```diff
@@ -1,6 +1,11 @@
 context("assignment_linter")
 options(encoding = "UTF-8")
 test_that("returns the correct linting", {
+  # unexpected { is R language dependent
+  oldlang <- Sys.getenv("LANGUAGE")
```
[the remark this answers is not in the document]
AshesITR: That sounds like a good idea. As an added bonus, this would automatically make all language-dependent tests safe, even if they are added in the future.

R/cache.R
```diff
@@ -7,9 +7,8 @@
 #' @return 0 for success, 1 for failure, invisibly.
 #' @export
 clear_cache <- function(file = NULL, path = NULL) {
```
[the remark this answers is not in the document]
AshesITR: Previously `*_cache()` messed up the test case for lint caching, because it loaded the exclusions from the package level `.lintr`.
Before fixing the relative path issue, those exclusions had no effect. Now with the correct relative paths, the bad test files are excluded from linting whenever `read_settings()` is called from somewhere within lintr, including `tests/testthat`.

R/exclude.R
```diff
@@ -71,7 +72,9 @@ parse_exclusions <- function(file, exclude = settings$exclude,
   sort(unique(c(exclusions, which(rex::re_matches(lines, exclude)))))
 }
 
-normalize_exclusions <- function(x, normalize_path=TRUE) {
+normalize_exclusions <- function(x, normalize_path = TRUE,
```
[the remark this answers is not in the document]
AshesITR: Done with `@keywords` internal to hide it from the index. Or should we even use `@noRd`?

tests/testthat/test-dir_linters.R
```diff
@@ -27,3 +27,38 @@ test_that("lint all relevant directories in a package", {
...
+  the_excluded_dir <- file.path(the_dir, "exclude-me")
+  dir.create(the_excluded_dir)
+
+  file.copy("default_linter_testcode.R", the_dir)
+  file.copy("default_linter_testcode.R", the_excluded_dir)
+  file.copy("default_linter_testcode.R", file.path(the_excluded_dir, "bad2.R"))
+
+  lints <- lint_dir(the_dir, exclusions = "exclude-me")
+  linted_files <- unique(names(lints))
+  expect_length(linted_files, 1L)
```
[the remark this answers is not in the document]
AshesITR: Added relevant test code

AshesITR:
Thank you @russHyde for taking the time to do a review.
I stole some files from your PR - I hope you don't mind.

tests/testthat/test-lint_package.R
```diff
@@ -0,0 +1,114 @@
...
+# .. should give the same results as when called from outside the package
+# with:
+# lint_package(path_to_package)
+
+# Template packages for use in testing are stored in
+# `tests/testthat/dummy_packages/<pkgName>`
+# These packages should not have a .lintr file:  Hardcoding a .lintr in a
+# dummy package throws problems during `R CMD check` (they are flagged as
+# hidden files, but can't be added to RBuildIgnore since they should be
+# available during `R CMD check` tests)
```
AshesITR, line 14: As a side note: Couldn't we override the `options(lintr.linter_file = "test_dot_lintr")` to overcome R CMD check warnings about hidden files?

AshesITR:
@russHyde do you have the time for another round of review?

russHyde:
Okay. I'll look at it today

tests/testthat/dummy_packages/assignmentLinter/DESCRIPTION
```diff
@@ -0,0 +1,9 @@
+Package: assignmentLinter
```
russHyde, line 1: To keep the `tests/testthat` folder neat, could the other dummy package (`tests/testthat/package`) be moved into `tests/testthat/dummy_packages/`. The tests that use `package` are all in `test-dir_linters.R` and they will require their filepaths to be redirected.

russHyde (changes requested):

russHyde:
In it's current form, with a dummy package in "tests/testthat/package", if I run the following from the `lintr` root, the vignettes subdirectory is not excluded from linting:

`lint_package("tests/testthat/package", exclusions = "tests/testthat/package/vignettes")`

And if I setwd to the `package` root, and run:

`lint_package(exclusions = "vignettes")` // or lint_package(".", exclusions = "vignettes")

an error comes up: "Error in normalize_exclusions(c(exclusions, settings$exclusions), root = path,  : object 'pattern' not found"

But, if I add a ".lintr" file into the "package" root with `exclusions: "vignettes"`, then the vignettes are excluded from linting.

Although the current PR does a good job of ensuring that directories are excluded when specified in the .lintr config, it doesn't appear to handle directory-exclusions that are passed via the exclusions = "..." arguments. This was an issue that meant my original PR for handling relative filepaths got really large.

I am happy for this PR to be merged, provided that you explicitly say that it handles directory exclusions that are specified in the .lintr config. If you do this, can you open a new issue about supporting directory exclusions via the "exclusions" argument to the lint_*() functions. (Maybe have a look at how @joaopmatias handled it in the unmerged https://github.com/russHyde/lintr/pull/2 )

AshesITR:
Thanks @russHyde
I've moved the package.

The current behaviour in this PR would be that `exclusions` is always relative to the project root (so your example should do `exclusions = "vignettes"` from outside as well).

It's not hard to change this. What do you prefer?

russHyde:
I'd really rather that that was done in a separate PR. We'd have to discuss whether files in `..., exclusions=..., ` are specified relative to the (possibly unspecified) project root, the (possibly absent) .lintr, the current directory, the targetted directory.

When exclusions are specified within the .lintr it seems a lot simpler: files should be specified relative to the directory containing the .lintr file or as full-paths.

AshesITR:
Sounds good. Then this PR is done as far as I can tell.

AshesITR:
ping @russHyde travis build is finally complete

R/lint.R
```diff
@@ -231,17 +241,23 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = NULL,
 #'   lint_package(
 #'     linters = with_defaults(semicolon_linter = semicolon_terminator_linter())
 #'     cache = TRUE,
-#'     exclusions = list("inst/doc/creating_linters.R" = 1, "inst/example/bad.R")
+#'     exclusions = list("inst/doc/creating_linters.R" = 1, "inst/example/bad.R", "data")
```
russHyde, line 244: This code doesn't work at present. Could you revert the comment back to it's original form.

NEWS.md
```diff
@@ -13,6 +13,8 @@
 * Fix handling zero-length variable name error (#566, #567, @renkun-ken)
 * New `missing_argument_linter()` (#563, #565, @renkun-ken)
 * New `sprintf_linter()` (#544, #578, @renkun-ken)
+* exclusions are now always relative to the location of the `.lintr` file and they support excluding entire directories
```
russHyde, line 16: Perhaps "Exclusions specified in the .lintr file are now relative to the location of that file and support excluding entire directories"

russHyde (changes requested):
Just some minor changes. Happy to merge this without CI if these are addressed.

R/lint.R
AshesITR: I'm a bit surprised that this doesn't work, but since we're probably touching this in #580 anyway, I'll revert to the old docs for now instead of fixing.

NEWS.md
AshesITR: Thanks - this formulation is way better 👍

AshesITR:
@russHyde thanks for taking all the time to review this thoroughly.
I think we can merge now.

russHyde (approved):
LGTM

russHyde:
Feel free to merge whenever. Many thanks for your hard work on this issue.

PR #546, opened 2020-10-30
Equals na comment

### LINKED ISSUE #545: equals_na_linter finds lints within comments
```
expect_lint("# x == NA", NULL, equals_na_linter)
#Error: got 1 lints instead of 0
# list(filename = "/tmp/RtmpISkLn0/file28f330767e742d", line_number = 1, column_number = 5, type = "warning", message = "Use is.na rather than == NA.", line = "# x == NA", ranges = list(c(5, 9)), linter = "equals_na_linter")
```

In fact it doesn't even need to be in a comment:

```
expect_lint("' x == NA'", NULL, equals_na_linter)
#Error: got 1 lints instead of 0
# list(filename = "/tmp/RtmpISkLn0/file28f330767e742d", line_number = 1, column_number = 5, type = "warning", message = "Use is.na rather than == NA.", line = "' x == NA'", ranges = list(c(5, 9)), linter = "equals_na_linter")
```

MichaelChirico:
Closes #545 

Beyond just skipping usages which might be found in comments, I also took the opportunity to extend the logic of the linter:

 - `NA == x` is now matched (before, it was just `x == NA`; this came for free with the move to xpath)
 - `x != NA` is also matched. It would be trivial to extend this to other comparators (`>|>=|<|<=`), but I wasn't sure that would gel with the name of the linter (not equals is similar enough I think)
 - This will also now match multiline usages like `x\n==\nNA` (again, for free with the move to xpath).

I also added a helper function for extracting `Lint`s from their `xml` nodes. This can likely be re-used in a few places.


AshesITR:
The logs complain about mising `xml_nodes_to_lint()`.
I don't see it in the commits.

Also note you'll probably need to use `@include xml_nodes_to_lint.R` in all the files that use the function in order to have roxygen compute the correct collate order.

R/equals_na_lintr.R
```diff
@@ -1,38 +1,22 @@
...
+  comparator_table <- paste0("self::", comparators, collapse = " or ")
+  NA_values <- c("NA", "NA_integer_", "NA_real_", "NA_complex_", "NA_character_")
+  NA_table <- paste("text() =", sQuote(NA_values, "'"), collapse = " or ")
+
+  xpath_fmt <- "//expr[expr[NUM_CONST[%s]]]/*[%s]"
+  xpath <- sprintf(xpath_fmt, NA_table, comparator_table)
+
+  bad_expr <- xml2::xml_find_all(xml, xpath)
+
+  lapply(bad_expr, xml_nodes_to_lint, source_file,
```
AshesITR, line 19: The implementation `xml_nodes_to_lint` is missing from this PR.
Could you please add it?

AshesITR (changes requested):

R/equals_na_lintr.R
MichaelChirico: totally forgot to include that edit to utils.R earlier 😅

AshesITR (approved):
Thanks!

MichaelChirico:
The failure on CI is due to the following:

```
lintr::expect_lint("\\", rex::rex("unexpected input"), lintr::equals_na_linter)
# Error in UseMethod("xml_find_all") : 
#   no applicable method for 'xml_find_all' applied to an object of class "NULL"
```

The issue is that `equals_na_linter` receives an object with `length(source_file$parsed_content)=1` but `!length(source_file$xml_parsed_content)`.

Is that expected behavior? Till now, I've assumed `length(source_file$parsed_content) > 0` implies the XML content would be there.

In this case, since the code doesn't in fact parse:

```
parse(text = "\\")
# Error in parse(text = "\\") : <text>:1:1: unexpected input
# 1: \
#     ^
```

I am surprised `source_file$parsed_content` is nonempty.

AshesITR:
This is due to #560.
The failing code produces an invalid XML but getParseData works on it.

MichaelChirico:
I see... that means we have to keep `parsed_content` around in order to make good error messages? Or can we `NULL` it out before passing it on to be linted?

Another workaround would be to add an `is_full_file` flag to `source_file`...

If neither, I think we'd have to do something like this in all XML-based linters:

```
# this is in seq_linter already
if (!length(source_file$parsed_content) || is.null(source_file$xml_parsed_content))
```

pretty awkward IMO!

AshesITR:
I agree that this double check looks a bit weird.
Maybe we should open a separate issue to adjust the behaviour and all current checks of this kind?

MichaelChirico:
OK moved back to WIP, I'll update it once #609 is merged.

MichaelChirico:
OK this should be ready to go now. Passing locally.

MichaelChirico:
TIL `sQuote(x, '"')` is an R 3.6 addition. fun. Will rewrite it in utils.

R/utils.R
```diff
@@ -169,6 +169,9 @@ try_silently <- function(expr) {
 viapply <- function(x, ...) vapply(x, ..., FUN.VALUE = integer(1))
 vcapply <- function(x, ...) vapply(x, ..., FUN.VALUE = character(1))
 
+# imitate sQuote(x, q) [requires R>=3.6]
+quote_wrap <- function(x, q) sprintf("%1$s%2$s%1$s", q, x, q)
```
AshesITR, line 173: The third argument to sprintf is unused, but I think `paste0(q, x, q)` would be more readable anyway.

AshesITR (changes requested):

R/utils.R
MichaelChirico: Agreed. I saw this function in the testthat helpers too, I'll just pull that into utils.

AshesITR (dismissed):
Nice, thanks!

AshesITR (approved):

PR #551, opened 2020-10-31
Add namespace_linter

### LINKED ISSUE #548: Namespace linter
#547 implements a `missing_package_linter()` which checks if all packages loaded by `library()`, `require()`, `loadNamespace()`, and `requireNamespace()` are installed.

Apart from that, we could also check the usage of `ns::obj` and `ns:::obj` as they also used to access objects in a namespace (provided by package).

We could implement a separate linter (as suggested at https://github.com/jimhester/lintr/pull/547#issuecomment-719822505) to check if 

1. The package `ns` in `ns::obj` and `ns:::obj` are installed.
2. The `obj` in `ns::obj` is exported from namespace `ns`.
3. The `obj` in `ns:::obj` exists in namespace `ns`.

While checking 1 does not load any new package, checking 2 and 3 might load the package being checked. I'm not sure if we allow loading extra packages from user code?

renkun-ken:
Closes #548 

image

* [x] Check namespace calls (`::` and `:::`)
* [x] Check exports and non-exports
* [x] Check unneeded `:::`
* [x] Test cases


R/namespace_linter.R
```diff
@@ -0,0 +1,82 @@
+#' @describeIn linter that checks for missing packages and symbols in namespace calls
```
AshesITR, line 1: Can we add a note here that this linter loads namespaces so it could potentially change the global state?

AshesITR (changes requested):
travis logs report that `devtools::document()` is needed

R/namespace_linter.R
renkun-ken: Updated

renkun-ken:
>travis logs report that devtools::document() is needed

Running `devtools::document()` will change several unrelated places as my roxygen2 has version 7.1.1 and nothing related to this PR. Does it matter if I commit changes to docs unrelated to this PR and update to `RoxygenNote: 7.1.1` in `DESCRIPTION`?

AshesITR:
It just occurred to me that a lint for unneeded ":::" would also be useful
`dplyr:::mutate(iris, a = 1)` -> `'mutate' is exported from {dplyr}. Use dplyr::mutate instead.`

Do you mind adding that case?

AshesITR (approved):

AshesITR:
It's fine for me. We could document() on master first and merge back if that helps keep the PR diff focussed?

renkun-ken:
Unneeded `:::` is checked now via 79899d3.

AshesITR:
Now `devtools::document()` should introduce minimal changes and make the travis build pass.

AshesITR (approved):

renkun-ken:
I send #553 and once it is merged, then existing PRs could merge back to keep the PR clean.

AshesITR:
https://travis-ci.org/github/jimhester/lintr/jobs/740571404#L3239-L3241

I'll check the T_and_F_symbol_linter

AshesITR:
Okay weird, `T_and_F_symbol_linter` is actually the `\name{}` of `linters.Rd`...

AshesITR (approved):

R/namespace_linter.R
```diff
@@ -0,0 +1,127 @@
...
+      vapply(expr, as.character, character(1L))
+    }
+
+    pkg_nodes <- xml2::xml_find_all(ns_nodes, "preceding-sibling::*[1]")
+    sym_nodes <- xml2::xml_find_all(ns_nodes, "following-sibling::*[1]")
+    pkgs <- get_node_text(pkg_nodes)
+    ops <- xml2::xml_text(ns_nodes)
+    syms <- get_node_text(sym_nodes)
+
+    installed_packages <- .packages(all.available = TRUE)
```
MichaelChirico, line 29: I think I would prefer if this were a linter with options (function factory), and we can supply `installed_packages` externally (defaulting to what's used here).

would make this linter more useful in a broader set of use cases

renkun-ken: OK. The most general form could be user supplied installed packages. The use case might be customized library location, or a limited set of packages to check against instead of all currently installed packages. I think this makes sense. Let me create a PR to make changes for both `namespace_linter` and `missing_package_linter`.

renkun-ken: As for `namespace_linter`, if some packages in user supplied packages are not actually installed or the library path is different from the current session that runs the linter, then `getNamespace()` and other namespace functions won't work so that the linter will think the symbols after `::` and `:::` are not available and complain. How do you think it should behave?

Issue #510, opened 2020-08-17
Clearly document which linters are `default_linters`


stufield:
It is currently somewhat difficult to go through the `?lintr::linters` and easily see which linters correspond to `default_linters`. Additionally, `?lintr::default_linters` does not actually list the names of the default linters. 

One could `names(lintr::default_linters)` and then compare those to the documentation, but this seems unnecessarily tedious. 

Suggest clearly splitting out the `default_linters` from other linters` in the documentation (with a Section?). And also listing the default linters in `?lintr::default_linters`? Perhaps here a link to `?linters` will suffice if they are split out clearly.

I can submit a PR if people agree this is useful (and no one really wants to work on docs)

russHyde:
This seems very useful. The documentation would benefit from a bit of work.

AshesITR:
At this point the list of linters is so long, it might be beneficial to order them into separate topics.
Is there a simple way to generate multiple docs for the same function?

Maybe at ?my_linter, ?default_linters and ?linters as well as maybe some ?my_linter_topic pages, all interlinked nicely?

Ideally ?my_linter would only show the help of that specific linter (so one can quickly find the available configuration options / argument documentation) and the others would be lists of linters akin to the current ?linters

russHyde:
By changing the roxygen for `default_linters` we could generate the following on calling `?default_linters`

```
default_linters             package:lintr              R Documentation

Default linters

Description:

     List of default linters for ‘lint’. Use ‘with_defaults’ to
     customize it.

     The set of default linters is as follows

     names(default_linters)

     ##  [1] "assignment_linter"                "closed_curly_linter"
     ##  [3] "commas_linter"                    "commented_code_linter"
     ##  [5] "cyclocomp_linter"                 "equals_na_linter"
     ##  [7] "function_left_parentheses_linter" "infix_spaces_linter"
     ##  [9] "line_length_linter"               "no_tab_linter"
     ## [11] "object_length_linter"             "object_name_linter"
     ## [13] "object_usage_linter"              "open_curly_linter"
     ## [15] "paren_brace_linter"               "pipe_continuation_linter"
     ## [17] "seq_linter"                       "single_quotes_linter"
     ## [19] "spaces_inside_linter"             "spaces_left_parentheses_linter"
     ## [21] "trailing_blank_lines_linter"      "trailing_whitespace_linter"

Usage:

     default_linters
```

---
Changes:
```
#### in zzz.R 

#' Default linters                                                                                                                             
[...]

PR #557, opened 2020-11-02
Functions not objects

### LINKED ISSUE #556: undesirable_function_linter() should look for function calls only
I think the `undesirable_function_linter()` should only look for function calls. The context is that I want to use lintr to keep watch for base file system functions creeping into usethis (https://github.com/r-lib/usethis/issues/531, https://github.com/r-lib/fs/issues/85). I can *almost* do what I want with the `undesirable_function_linter()`.

But it reports tons of false positives, i.e. warns when it should not.

`dir()` is one of the functions I'm linting for. But throughout the package, below both `R/` and `tests/`, the name `dir` is frequently used for the path to a directory. So the linter reports dozens of false warnings.

Here's a small reprex with current lintr vs. after the change in my PR (#557):

``` r
library(lintr)

lint_this <- function(this, ...) {
  file <- tempfile()
  on.exit(unlink(file))
  writeLines(this, con = file, sep = "\n")
  lint(file, ...)
}

linter <- undesirable_function_linter(fun = c("dir" = NA))

# warning is correct: I'm calling dir()
lint_this('dir(".")', linter)
#> /private/var/folders/pv/l2mbh7j90y5_vslh_1wyqxj00000gn/T/RtmpQbfI18/file121123553fe9e:1:1: warning: Function "dir" is undesirable.
#> dir(".")
#> ^~~

# warning is WRONG: I'm defining an object named 'dir'
lint_this("dir <- 1:3", linter)
#> /private/var/folders/pv/l2mbh7j90y5_vslh_1wyqxj00000gn/T/RtmpQbfI18/file121121d71526c:1:1: warning: Function "dir" is undesirable.
#> dir <- 1:3
#> ^~~

devtools::load_all("~/rrr/lintr")
#> ℹ Loading lintr

linter <- undesirable_function_linter(fun = c("dir" = NA))

# warning is correct: I'm calling dir()
lint_this('dir(".")', linter)
#> /private/var/folders/pv/l2mbh7j90y5_vslh_1wyqxj00000gn/T/RtmpQbfI18/file121127ada3f8d:1:1: warning: Function "dir" is undesirable.
#> dir(".")
#> ^~~

# lack of warning is correct: I'm defining an object named 'dir'
lint_this("dir <- 1:3", linter)
```

<sup>Created on 2020-11-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

jennybc:
Closes #556


tests/testthat/test-undesirable_function_linter.R
```diff
@@ -7,7 +7,7 @@ test_that("linter returns correct linting", {
 
   expect_lint("x <- options()", NULL, linter)
   expect_lint("cat(\"Try to return\")", NULL, linter)
-  expect_lint("lapply(x, log10)", list(message=msgL, line_number=1L, column_number=11L), linter)
+  #expect_lint("lapply(x, log10)", list(message=msgL, line_number=1L, column_number=11L), linter)
```
jennybc, line 10: I think it is worth this tradeoff, i.e. to let go of this use case and gain the ability to use the undesirable function names in other contexts, i.e. as a object name.

MichaelChirico: i would add an option to the undesirable_function_linter signature instead (maybe strict=TRUE by default?) which toggles whether to include SYMBOLs

jimhester: I agree having an option to toggle the existing behavior seems worthwhile.

jennybc:
I will turn this into a draft PR so my continued futzing with what prints in GHA doesn't notify people.

tests/testthat/test-undesirable_function_linter.R
jennybc: Alright, I'll split the GHA logging stuff off into a separate PR and implement this as a switch.

I am using this linter with usethis now, to enforce "fs everywhere!", so I'd love to get to the point where I can use dev lintr, instead of a branch on my fork.

R/undesirable_function_linter.R
```diff
@@ -1,12 +1,50 @@
-#' @describeIn linters  Report the use of undesirable functions, e.g. \code{return}, \code{options},
-#'                      or \code{sapply} and suggest an alternative.
-#' @param fun  Named character vector, where the names are the names of the undesirable functions,
-#'             and the values are the text for the alternative function to use (or \code{NA}).
+#' @describeIn linters Report the use of undesirable functions, e.g.
+#'   \code{return}, \code{options}, or \code{sapply} and suggest an alternative.
+#' @param fun Named character vector, where the names are the names of the
+#'   undesirable functions, and the values are the text for the alternative
+#'   function to use (or \code{NA}).
+#' @param symbol_is_undesirable Whether to consider the use of an undesirable
```
jennybc, line 6: I went with a specific name that hopefully explains itself (vs strict or lax).

R/undesirable_function_linter.R
```diff
@@ -1,12 +1,50 @@
...
+#'   function name as a symbol undesirable or not.
+#'
+#' @examples
```
jennybc, line 9: I realize there are no other examples and that this probably can't remain here, because it would be weird. But I had a hard time getting started with linter customization, due to the lack of examples.

Is there somewhere content like this could go?

Sidebar: It might be useful to have a function for example-ing that makes it easy to lint a character vector like this.

The `unclass()` calls were to make the example easier to write while inside RStudio, i.e. I actually wanted plain output. Presumably there is a better way to go about this and I just didn't discover it.

I know this section probably be deleted, but I wanted to pose these questions.

MichaelChirico: >  But I had a hard time getting started with linter customization, due to the lack of examples.

Agreed! It's part of what I had in mind when documenting `get_source_expressions` more extensively:

https://github.com/jimhester/lintr/pull/562/files

Does that go in the direction you had in mind? If so it would be great if we could build on that.

jennybc: I meant something simpler: I had a hard time figuring out how to use the undesirable function linter, how to use it with non-default undesirable functions, etc. I think I had to read the test and search GitHub to try to find actual usage?

With almost 40 linters documented together, it becomes almost impossible to give much info or examples on how to use an individual linter.

I know this problem is not going to get worked out in the PR and I don't want it to become a blocker!

jennybc: I now understand how to lint inlined source. So I've made my examples cleaner, but I still understand they probably have to be deleted.

The printing problem is covered by #590.

tests/testthat/test-undesirable_function_linter.R
```diff
@@ -16,3 +16,11 @@ test_that("linter returns correct linting", {
               ),
               linter)
 })
+
+test_that("it's possible to NOT lint symbols", {
+  linter <- undesirable_function_linter(
```
MichaelChirico, line 21: i think an illustrative example to include here would be something with `lapply`

R/undesirable_function_linter.R
AshesITR: We could probably get examples into the docs once we have a clear plan for #510.

tests/testthat/test-undesirable_function_linter.R
```diff
@@ -16,3 +16,17 @@ test_that("linter returns correct linting", {
...
+    fun = c("dir" = NA),
+    symbol_is_undesirable = FALSE
+  )
+  expect_lint("dir <- 1:3", NULL, linter)
+
+  linter <- undesirable_function_linter(
```
jennybc, line 27: Is this what you want @MichaelChirico?

jennybc:
I've removed the examples since they presumably cannot be merged and they served their purpose re: discussion.

tests/testthat/test-undesirable_function_linter.R
MichaelChirico: Yes, perfect! Thanks

R/undesirable_function_linter.R
jennybc: I've deleted the examples.

R/undesirable_function_linter.R
```diff
@@ -1,12 +1,22 @@
...
+#' @describeIn linters Report the use of undesirable functions, e.g.
+#'   \code{return}, \code{options}, or \code{sapply} and suggest an alternative.
+#' @param fun Named character vector, where the names are the names of the
+#'   undesirable functions, and the values are the text for the alternative
+#'   function to use (or \code{NA}).
+#' @param symbol_is_undesirable Whether to consider the use of an undesirable
+#'   function name as a symbol undesirable or not.
 #' @export
-undesirable_function_linter <- function(fun=default_undesirable_functions) {
+undesirable_function_linter <- function(fun=default_undesirable_functions,
```
AshesITR, line 9: Can we delint this? The rest LGTM.

AshesITR (changes requested):

AshesITR (approved):
LGTM, thanks!

PR #560, opened 2020-11-05
Allow invalid xml

### LINKED ISSUE #559: Invalid "\" crashes lintr
If the source code contains any invalid use of `\`, then `lintr::lint()` will stop working:

```r
col <- c("a", \)
```

The case could be simplified to the following:

```
> parse(text = "\\")                                                                                                        
Error in parse(text = "\\") : <text>:1:1: unexpected input
1: \
    ^
Backtrace:
1: parse(text = "\\")

> lintr::lint("\\\n")                                                                                                       
Error in read_xml.raw(charToRaw(enc2utf8(x)), "UTF-8", ..., as_html = as_html,  : 
  StartTag: invalid element name [68]
Backtrace:
1: read_xml.raw(charToRaw(enc2utf8(x)), "UTF-8", ..., as_html = as_html, 
2: read_xml.character(xmlparsedata::xml_parse_data(pc))
3: xml2::read_xml(xmlparsedata::xml_parse_data(pc))
4: FUN(X[[i]], ...)
5: lapply(X = top_level_expressions(parsed_content), FUN = get_single_source_expression, 
6: get_source_expressions(filename)
7: lintr::lint("\\\n")
```

Looks like it is caused by xmlparsedata writing invalid XML and lintr does not handle such case.

Before error occurs, `pc` got the following value:

```
  line1 col1 line2 col2 id parent token terminal text
1     1    1     1    1  1      0    \\    FALSE   \\
```

`xmlparsedata::xml_parse_data(pc)` produces the following XML with invalid tokens:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<exprlist>
  <\ line1="1" col1="1" line2="1" col2="1" start="3" end="3">\
  </\>
</exprlist>
```

renkun-ken:
Closes #559 

Since lintr might pass a partial or invalid parse data to `xmlparsedata::xml_parse_data`, it could also generate invalid XML text, which also causes `xml2::read_xml()` to fail.

This PR adds a `tryCatch()` to all `xml2::read_xml(xmlparsedata::xml_parse_data(pc))` so that invalid XML text would not cause `lintr::lint` to fail.

Note that all linters that uses the `source_file$xml_parsed_content` must check if it is `NULL` before querying it. This PR adds such checks to `seq_linter`, `missing_package_linter` and `namespace_linter`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@
 * `seq_linter()`'s lint message is clearer about the reason for linting. (#522, @michaelchirico)
 * New `missing_package_linter()` (#536, #547, @renkun-ken)
 * New `namespace_linter()` (#548, #551, @renkun-ken)
+* Fix possible error on invalid XML produced by xmlparsedata (#559, #560, @renkun-ken)
 
 # lintr 2.0.1
 
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -94,7 +94,7 @@ get_source_expressions <- function(filename) {
       file_lines = source_file$lines,
       content = source_file$lines,
       full_parsed_content = parsed_content,
-      xml_parsed_content = if (!is.null(parsed_content)) xml2::read_xml(xmlparsedata::xml_parse_data(parsed_content))
+      xml_parsed_content = if (!is.null(parsed_content)) tryCatch(xml2::read_xml(xmlparsedata::xml_parse_data(parsed_content)), error = function(e) NULL)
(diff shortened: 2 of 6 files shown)


AshesITR (approved):
I checked for other uses of `xml_parsed_content` (and `global_xml_parsed_content()`).
All other users already did NULL checks, these were:
 * object_name_linter
 * object_usage_linter
 * pipe_continuation_linter

PR #562, opened 2020-11-11
get_source_expressions detects terminal newlines

### LINKED ISSUE #542: readLines warning for incomplete final line
Found in many packages, e.g. `nsprcomp`:

```
tmp = tempfile()
download.file('https://raw.githubusercontent.com/cran/nsprcomp/master/R/asdev.R', tmp)
lintr::lint(tmp)
# incomplete final line found on '/tmp/RtmptJEp8r/file3fb841ddb96d'
```

It's because `get_source_expressions` uses `readLines` without `warn=FALSE`. This is easy enough to fix; filing the issue first, since this might be considered a lint as well.

MichaelChirico:
Closes #542 

Comment wanted:

 1. I documented `get_source_expressions` return object pretty thoroughly. Should this be filed as a separate PR perhaps? Also please have a look to be sure it makes sense, I wasn't totally clear on what all of the components are. If it's good to include in this PR I'll `document()` it.
 2. I didn't include the linter in this PR, the thinking is to file a separate issue & PR as a follow-up. Happy to write it here as well.
 3. I didn't touch the `readLine` usages in R/exclude.R and R/cache.R. I can look into that as needed.

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -2,10 +2,47 @@
 #'
 #' This object is given as input to each linter
 #' @param filename the file to be parsed.
+#' @return A `list` with three components: (1) `expressions`, a `list` of
+#'   `n+1` objects. The first `n` elements correspond to each expressions in
+#'   `filename`, and consist of a list of 9 elements:
+#'   (1) `filename` (`character`); (2) `line` (`integer`) the line in `filename`
+#'   where this expression begins; (3) `column` (`integer`) the column in
+#'   `filename` where this expression begins; (4) `lines` (named `character`)
+#'   vector of all lines spanned by this expression, named with the line number
+#'   corresponding to `filename`; (5) `parsed_content` (`data.frame`) as given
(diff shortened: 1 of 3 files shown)


R/get_source_expressions.R
```diff
@@ -2,10 +2,47 @@
 #'
 #' This object is given as input to each linter
 #' @param filename the file to be parsed.
+#' @return A `list` with three components: (1) `expressions`, a `list` of
```
AshesITR, line 5: This section is very nice! I think it would be even more useful if the hierarchical structure of the return value is supported by an itemize-ish layout of the documentation.

Maybe nested `\describe{...}`s work well here.

AshesITR (changes requested):

R/get_source_expressions.R
MichaelChirico: Went with `\item` since that's what `?lm` uses, and `\itemize` for the sub-list description. Rendered and it looks passable.

AshesITR (approved):

AshesITR:
Thanks!

PR #567, opened 2020-11-15
Fix handling zero-length variable name error

### LINKED ISSUE #566: lintr::lint fails with ``
If the code contains the following double backticks

```r
``
```

then `lintr::lint()` will fail:

```
> lintr::lint("``\n")                                                                                                     
Error in source_file$lines[[line_number]] : subscript out of bounds
Backtrace:
1: structure(list(filename = filename, line_number = as.integer(line_number), 
2: Lint(filename = source_file$filename, line_number = line_number, 
3: value[[3L]](cond)
4: tryCatchOne(expr, names, parentenv, handlers[[1L]])
5: tryCatchList(expr, classes, parentenv, handlers)
6: tryCatch(source_file$parsed_content <- parse(text = source_file$content, 
7: get_source_file(source_file, error = lint_error)
8: get_source_expressions(filename)
9: lintr::lint("``\n")
```

Parsing the text will produce the following error:

```
> parse(text = "``")                                                                                                      
Error in parse(text = "``") : attempt to use zero-length variable name
Backtrace:
1: parse(text = "``")
```

renkun-ken:
Closes #566.

This PR checks if the parse error message has a hint for the source code error and fixes the handling of `attempt to use zero-length variable name` error caused by empty variable names in the following cases:

```r
``
``()
''()
""()
fun(''=42)
fun(""=42)
fun(a=1,""=42)
```

When this error occurs, it searches the source code with a pattern that covers the above cases.

* [x] Fix handling of error
* [x] Test case

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -71,7 +71,6 @@ get_source_expressions <- function(filename) {
           anything),
         "\n")
       )
-      
 
     # an error that does not use R_ParseErrorMsg
     if (is.na(message_info$line)) {
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -71,6 +71,7 @@ get_source_expressions <- function(filename) {
           anything),
         "\n")
       )
+      
 
     # an error that does not use R_ParseErrorMsg
     if (is.na(message_info$line)) {
@@ -81,6 +82,31 @@ get_source_expressions <- function(filename) {
           double_quotes, capture(name = "starting", anything), double_quotes))
 
       line_location <- re_matches(source_file$content, rex(message_info$starting), locations = TRUE)
+      line_location <- line_location[complete.cases(line_location), ]
+
+      if (nrow(line_location) == 0L) {
+        if (grepl("attempt to use zero-length variable name", e$message)) {
+          # empty symbol: ``, ``(), ''(), ""(), fun(''=42), fun(""=42)
+          loc <- re_matches(source_file$content,
+            "``|''\\s*\\(|\"\"\\s*\\(|\\(''\\s*=|\\(\"\"\\s*=", locations = TRUE)
+          loc <- loc[complete.cases(loc), ]
+          if (nrow(loc) > 0) {
+            line_location <- loc[1, ]
+          }
+        } else {
+          return(
+            Lint(
+              filename = source_file$filename,
+              line_number = 1,
+              column_number = 1,
+              type = "error",
+              message = e$message,
+              line = "",
+              linter = "error"
+            )
+          )
+        }
+      }
 
       line_number <- find_line_fun(source_file$content)(line_location$start)
       column_number <- find_column_fun(source_file$content)(line_location$start)
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -85,7 +85,7 @@ get_source_expressions <- function(filename) {
       line_location <- line_location[complete.cases(line_location), ]
 
       if (nrow(line_location) == 0L) {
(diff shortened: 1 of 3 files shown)


R/get_source_expressions.R
```diff
@@ -81,6 +82,31 @@ get_source_expressions <- function(filename) {
           double_quotes, capture(name = "starting", anything), double_quotes))
 
       line_location <- re_matches(source_file$content, rex(message_info$starting), locations = TRUE)
+      line_location <- line_location[complete.cases(line_location), ]
+
+      if (nrow(line_location) == 0L) {
+        if (grepl("attempt to use zero-length variable name", e$message)) {
+          # empty symbol: ``, ``(), ''(), ""(), fun(''=42), fun(""=42)
+          loc <- re_matches(source_file$content,
+            "``|''\\s*\\(|\"\"\\s*\\(|\\(''\\s*=|\\(\"\"\\s*=", locations = TRUE)
```
AshesITR, line 91: I think we should stick to one type of regex specification in lintr.
Can you convert that to a `rex()` expression?

AshesITR: NB the location of the argument in a function call need not be the first, eg. `fun(a = 1, '' = 2)`.
I think a regex covering this as well should be ``` ``|(''|"")\s*\(|[(,]\s*=(''|``|"")```

renkun-ken: Yes, you are right. Thanks!

R/get_source_expressions.R
```diff
@@ -81,6 +82,34 @@ get_source_expressions <- function(filename) {
           double_quotes, capture(name = "starting", anything), double_quotes))
 
       line_location <- re_matches(source_file$content, rex(message_info$starting), locations = TRUE)
+      line_location <- line_location[complete.cases(line_location), ]
+
+      if (nrow(line_location) == 0L) {
+        if (grepl("attempt to use zero-length variable name", e$message)) {
```
MichaelChirico, line 88: fixed=TRUE

tests/testthat/test-error.R
```diff
@@ -42,4 +42,32 @@ test_that("returns the correct linting", {
...
+
+  expect_lint("``()",
+    rex("attempt to use zero-length variable name")
+  )
+
+  expect_lint("''()",
+    rex("attempt to use zero-length variable name")
+  )
+
+  expect_lint("\"\"()",
```
MichaelChirico, line 58: nit: `'""'` was used in the source (and IMO is easier on the eyes), I would use that here too (e.g. `'""()'`)

MichaelChirico:
LGTM! Nice PR

R/get_source_expressions.R
renkun-ken: Yes, it's better.

tests/testthat/test-error.R
renkun-ken: So we don't have a policy to only allow one type of quotes? Then sure it is better.

MichaelChirico: Somehow I thought the tidyverse rule was "use double quotes, unless using single quotes to avoid escaping single quotes", but I don't see it in the style guide actually.

However, `single_quotes_linter` in this package makes that distinction :smiley: 

https://github.com/jimhester/lintr/blob/master/R/single_quotes_linter.R#L8

renkun-ken: Thanks!

AshesITR:
codecov complains about the fallback to l1c1 if we fail to match anything.

That's fine by me.

renkun-ken:
Any other comments?

AshesITR (dismissed):

AshesITR (approved):

Issue #544, opened 2020-10-27
sprintf linter


renkun-ken:
I'm thinking of a `sprintf` linter that extracts all `sprintf(...)` calls and checks if the number of supplied arguments is consistent with the number of format placeholders in the format string when the format string is a string literal (rather than a symbol or call).

If the supplied argument is an atomic value, it could also check if its type is consistent with the format at the corresponding position.

AshesITR:
Just a memo if this will be implemented, `sprintf()` supports the `%1$d` notation for position selection (and repeated replacements).

renkun-ken:
Thanks for pointing it out!

I think we could transform calls like `sprintf(<string>, <call>, <symbol>, 20L, "a")` into `sprintf(<string>, 0, 0, 20L, "a")`, i.e. transform arguments that are not atomic scalar value into something like `0` that is compatible with all formats (?) and evaluate it and capture the error if any. Looks like it should work to identify an incorrect number of arguments and incompatible types that directly appear in the arguments?

MichaelChirico:
Just confirming in fact `0` works (whereas `TRUE`) doesn't:

```
fmts = c(
  "%d", "%i", "%o", "%x", "%X", "%f", "%e", "%E", "%g", "%G", "%a", "%A", "%s",
  "%02d", "%2d", "%1.2x", "1.2f", "1.2g", "%05s"
)
for (fmt in fmts) sprintf(fmt, 0)
# (no error)
```

PR #530, opened 2020-10-04
WIP: New linter to block stop(paste(...)) calls

MichaelChirico:
Closes #527

Filing now for early feedback. If you're OK to add the linter here I will complete the PR by adding tests etc.

Informally there are these test cases thus far:

```
writeLines('
# bad
stop(paste("a", "b", "c"))
warning(paste0("hey", "you"))
message(paste("there", "we", "go"))
stop(paste("a", "b", "c", sep=""))

# good
stop("a", "b", "c")
paste0("a", "b", "c")
stop("go to", paste("jail", "now"))
stop(paste("a", "b"), "c")
stop(paste("a", "b", collapse=","))
stop(paste("a", "b", sep="|"))
', tmp <- tempfile())

lintr::lint(tmp, stop_paste_linter)
/tmp/RtmpGfxjfA/file38c5e31aefde27:3:1: warning: Avoid stop(paste(...)) expressions, just use stop(...). Disable this with a comment...
stop(paste("a", "b", "c"))
^~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/RtmpGfxjfA/file38c5e31aefde27:4:1: warning: Avoid warning(paste0(...)) expressions, just use warning(...). Disable this with a comment...
warning(paste0("hey", "you"))
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/RtmpGfxjfA/file38c5e31aefde27:5:1: warning: Avoid message(paste(...)) expressions, just use message(...). Disable this with a comment...
message(paste("there", "we", "go"))
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/RtmpGfxjfA/file38c5e31aefde27:6:1: warning: Avoid stop(paste(...)) expressions, just use stop(...). Disable this with a comment...
stop(paste("a", "b", "c", sep=""))
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #575, opened 2020-11-22
Throw a lint for missing Rmd chunk ends instead of an error

### LINKED ISSUE #571: Malformed file! message for imbalanced chunk gates is not helpful
```
writeLines("```{r chunk}
lm(x ~ y)
", tmp <- tempfile(fileext = '.Rmd'))
lintr::lint(tmp)
# Error: Malformed file!
```

This error message should be improved

AshesITR:
fixes #571


R/extract.R
```diff
@@ -67,15 +72,22 @@ filter_chunk_end_positions <- function(starts, ends) {
...
-  code_ends <- positions[1 + code_start_indexes]
 
-  stopifnot(all(grepl("ends", names(code_ends))))
+  code_ends <- positions[pmin(1 + code_start_indexes, length(positions))]
+
+  bad_end_indexes <- which(grepl("starts", names(code_ends), fixed = TRUE))
+  if (length(bad_end_indexes)) {
+    bad_start_positions <- positions[code_start_indexes[bad_end_indexes]]
+    # This error message is formatted like a parse error
+    stop(sprintf(
```
MichaelChirico, line 85: I recommend using `gettextf` here -- `lintr` doesn't do translations (yet) but if it does someday, it'll make it easier.

It's ~basically a drop-in replacement given that we're not doing translation. If we truly wanted to prepare for translation, we'd do `stop(domain=NA, gettextf(..., domain="R-linter"))`.

Not crucial but I think it's an easy change, WDYT?

R/extract.R
```diff
@@ -67,15 +72,22 @@ filter_chunk_end_positions <- function(starts, ends) {
...
+      "<rmd>:%1$d:1: Missing chunk end for chunk (maybe starting at line %1$d).\n",
```
MichaelChirico, line 86: Oh wow, TIL `sprintf` accepts the `%1$d` notation! Awesome.

R/extract.R
AshesITR: Sounds like a good idea for another issue.
The codebase contains 9 relevant occurences of `stop()`. I think we should migrate all of them at the same time.

R/extract.R
AshesITR: Yep, I find it very handy. #544 would have to deal with that BTW.

R/extract.R
MichaelChirico: Sounds good. Filed #577 so as not to forget. Also related: #530

R/extract.R
MichaelChirico: cc @renkun-ken

MichaelChirico: @renkun-ken not sure if it would be helpful, but FYI Matt wrote some logic in C for handling `sprintf` fmts including handling the `%<pos>$<fmt>` notation:

https://github.com/Rdatatable/data.table/blob/master/src/snprintf.c

MichaelChirico (changes requested):
Logic looks ~basically good at a glance -- please add some tests to confirm everything's working

AshesITR:
I've added the two examples from our discussion as well as one with a trailing un-closed chunk (thats the reason for using `pmin()` in indexing)

NB that your example code, just as expected, isn't found to fail parsing because currently we don't detect all chunk begin / end tokens.
If we want to remain compatible with non-r code chunks and still catch this error, we'd have to parse all types of chunk boundaries and match them.

tests/testthat/test-knitr_formats.R
```diff
@@ -104,9 +104,59 @@ test_that("it does _not_ error with inline \\Sexpr", {
...
+test_that("it does lint with malformed input", {
+  expect_lint(
+    file = "knitr_malformed/incomplete_r_block.Rmd",
+    checks = "Missing chunk end",
+    default_linters
+  )
+
+  contents <- c(
+    paste(
+      "```{r chunk}",
```
MichaelChirico, line 116: WDYT about putting an ad hoc file into a string like this

    contents <- "```{r chunk}
    lm(x ~ y)
    
    
    # some text
    
    ```
    bash some_script.sh
    ```
    "

more readable than being cluttered with `"` IMO

MichaelChirico (approved):
LGTM, great!

tests/testthat/test-knitr_formats.R
AshesITR: Wouldn't the indentation be messed up either in the source code or in the ad hoc file?
Maybe a helper function that automatically detects the correct amount of trimming would produce nice code and nice ad-hoc files.

Something like

```r
if (lala) {
  ad_hoc <- trim_some("
    line1
    line2
      line3 indented
  ")
}
```

Which should produce the same value as
`ad_hoc <- "line1\nline2\n__line3 indented"` (double underscore instead of space to prevent github trimming)

MichaelChirico: The idea is for `trim_some` to trim relative to the first line? or I guess relative to the least-indented line. Looks good anyway, probably OK to just add such helper in `tests/testthat` only.

AshesITR: Exactly, relative to the least indented line. I guess I might as well add this in the current PR since travis had a random hiccup preventing merge.

MichaelChirico (dismissed):

R/extract.R
```diff
@@ -1,11 +1,16 @@
...
-extract_r_source <- function(filename, lines) {
+extract_r_source <- function(filename, lines, error = identity) {
   pattern <- get_knitr_pattern(filename, lines)
   if (is.null(pattern$chunk.begin) || is.null(pattern$chunk.end)) {
     return(lines)
   }
 
-  chunks <- get_chunk_positions(pattern = pattern, lines = lines)
+  chunks <- tryCatch(get_chunk_positions(pattern = pattern, lines = lines), error = error)
+  if (!inherits(chunks, "list")) {
```
MichaelChirico, line 9: why not `is.list` here?

R/extract.R
```diff
@@ -67,15 +72,22 @@ filter_chunk_end_positions <- function(starts, ends) {
...
-  }
 
   positions <- sort(c(starts = starts, ends = ends))
   code_start_indexes <- grep("starts", names(positions))
-  code_ends <- positions[1 + code_start_indexes]
 
-  stopifnot(all(grepl("ends", names(code_ends))))
+  code_ends <- positions[pmin(1 + code_start_indexes, length(positions))]
+
+  bad_end_indexes <- which(grepl("starts", names(code_ends), fixed = TRUE))
```
MichaelChirico, line 81: `which(grepl(.))` is just `grep()` right?

MichaelChirico (dismissed):
LGTM!

R/extract.R
AshesITR: just for consistency with L246 (implementation of get_source_file).
WDYT about switching both to `inherits(., "error")` instead? That seems more clear to me.

R/extract.R
AshesITR: Yep, totally overlooked. Thanks!

R/extract.R
MichaelChirico: agree, if we can make the branch `if(x)` instead of `if(!y)` that's always better IMO

MichaelChirico (approved):

PR #576, opened 2020-11-23
fix testcases for windows

AshesITR:
Windows line endings are 2 bytes (\r\n) so we need to strip two bytes off to remove the trailing newline.


tests/testthat/test-get_source_expressions.R
```diff
@@ -65,8 +65,8 @@ test_that("Terminal newlines are detected correctly", {
   writeLines("lm(y ~ x)", tmp <- tempfile())
   on.exit(unlink(tmp), add=TRUE)
   writeBin(
-    # strip the last element (\n)
-    head(readBin(tmp, raw(), file.size(tmp)), -1L),
+    # strip the last (two) element(s) (\r\n or \n)
+    head(readBin(tmp, raw(), file.size(tmp)), if (.Platform$OS.type == "windows") -2L else -1L),
```
MichaelChirico, line 69: Might it be more robust to do a `while` loop from the end of the file to skip past all the `\r` or `\n`? Probably overkill since we wrote the file ourselves with `writeLines` so I think only `\n` and `\r\n` are possible...

MichaelChirico:
Good catch! Darn `\r\n`...

tests/testthat/test-get_source_expressions.R
AshesITR: I'd actually want the test to fail if our code starts to produce spurious trailing newlines, no?

AshesITR:
The test failure on `trim_output` was caused by the same fact, so I've fixed both in this commit.
The resulting tree passes all tests on my Windows machine :-)

tests/testthat/test-get_source_expressions.R
MichaelChirico: I had mind an unknown platform where the line ending is neither `\n` nor `\r\n`. But I think you're right, it would be good to be made aware of such cases via a breaking test since they're so rare.

MichaelChirico (approved):

PR #578, opened 2020-11-23
sprintf linter

### LINKED ISSUE #544: sprintf linter

renkun-ken:
Closes #544 

This PR implements a `sprintf_linter()` that checks if the numbers of arguments are correct and types of arguments are compatible for `sprintf("string", ...)` calls.

It captures all `sprintf("string", ...)` calls and replace non-atomic-scalar arguments with `0` and try to evaluate these calls. If an error occurs, then we know there's inconsistency between the format string and the arguments, and the error message is reported as a lint result.

image

* [x] sprintf_linter
* [x] Test cases


tests/testthat/test-sprintf_linter.R
```diff
@@ -0,0 +1,62 @@
...
+test_that("returns the correct linting", {
+  expect_lint("sprintf('hello')",
+    NULL,
+    sprintf_linter)
+
+  expect_lint("sprintf('hello', 1)",
+    NULL,
+    sprintf_linter)
+
+  expect_lint("sprintf('hello %d', 1)",
```
AshesITR, line 11: IMHO this should be a warning (unused fmt argument). What do you think?

tests/testthat/test-sprintf_linter.R
```diff
@@ -0,0 +1,62 @@
...
+  expect_lint("sprintf('hello %d', 'a')",
+    list(message = rex("invalid format '%d'; use format %s for character objects")),
+    sprintf_linter)
+
+  expect_lint("sprintf('hello %d', 1.5)",
+    list(message = rex("invalid format '%d'; use format %f, %e, %g or %a for numeric objects")),
+    sprintf_linter)
+
+  expect_lint("sprintf('hello %d',)",
+    list(message = rex("argument is missing, with no default")),
```
AshesITR, line 36: Can we provide a more tailored error message?
Something like "format specification needs X more arguments" springs to mind.

AshesITR (changes requested):
Thanks for looking into this!

Could we add a test to verify that mixed positional and non-positional format strings also work?
E.g. `sprintf("%1$s %s", "a", "a")`

Maybe also raw strings as literal arguments.

WDYT about adding a warning lint if the number of arguments provided to `sprintf` is too many? Most of the time I encountered that, the format string was wrong and a warning would have been very helpful.

tests/testthat/test-sprintf_linter.R
AshesITR: I meant `sprintf('hello', 1)`

renkun-ken:
>Could we add a test to verify that mixed positional and non-positional format strings also work?
E.g. sprintf("%1$s %s", "a", "a")

In fact the second argument is not used in this case.

```r
> sprintf("%1$s %s", "a", "b")                                                                                                                                                                                                                               
[1] "a a"
```

Not sure if it is expected?

renkun-ken:
>Maybe also raw strings as literal arguments.

Regarding raw strings, are they allowed in our test cases since it requires R 4.0? If so, we need to put them into a conditional on R version >= 4.0? It seems that there's no raw strings in the existing test cases.

AshesITR:
> >Could we add a test to verify that mixed positional and non-positional format strings also work?
> 
> E.g. sprintf("%1$s %s", "a", "a")
> 
> 
> 
> In fact the second argument is not used in this case.
> 
> 
> 
> ```r
> 
> > sprintf("%1$s %s", "a", "b")                                                                                                                                                                                                                               
> 
> [1] "a a"
> 
> ```
> 
> 
> 
> Not sure if it "works"?

That was intentional to demonstrate the rules of this scenario.

AshesITR:
> >Maybe also raw strings as literal arguments.
> 
> 
> 
> Regarding raw strings, are they allowed in our test cases since it requires R 4.0? If so, we need to put them into a conditional on R version >= 4.0? It seems that there's no raw strings in the existing test cases.

I agree, we should make raw string tests for R >= 4.0 a separate issue

renkun-ken:
>WDYT about adding a warning lint if the number of arguments provided to sprintf is too many?

This is certainly good but requires a complete analysis of the format string to know how many and what types the arguments should/could be. It's quite likely that we need C/C++ code to do this. Currently, this linter uses a simplified approach that avoids understanding the format string by completely relying on evaluating simplified `sprintf()` calls and catch errors they throw.

tests/testthat/test-sprintf_linter.R
renkun-ken: https://github.com/jimhester/lintr/pull/578#issuecomment-732144405

tests/testthat/test-sprintf_linter.R
renkun-ken: A more tailored error message is certainly better. In fact, all the lint messages come from `sprintf()` errors such as

```r
> sprintf("%d", "a")                                                                                                                                                                                                          
Error in sprintf("%d", "a") : 
  invalid format '%d'; use format %s for character objects

> sprintf("%d", 1.5)                                                                                                                                                                                                          
Error in sprintf("%d", 1.5) : 
  invalid format '%d'; use format %f, %e, %g or %a for numeric objects
```

When the arguments are numerous, it is still ambiguous which causes the error since the error messages do not say anything about which argument is wrong.

To make the errors more informative, we might need a parser of the format string to understand what is expected as inputs.

renkun-ken:
I guess we could use a simple version first and enhance it on demand later?

AshesITR (approved):

PR #585, opened 2020-11-25
Fix unneeded_concatenation_linter in magrittr pipes

### LINKED ISSUE #573: unneeded_concatenation_linter in magrittr pipes?
Consider

```
writeLines('
"a" %>%
  c("b")
', tmp <- tempfile())
lintr::lint(tmp, lintr::unneeded_concatenation_linter)
# /tmp/Rtmp0PRc3M/fileacb8a79d41ebb:3:3: warning: Unneeded concatenation of a constant. Remove the "c" call.
#  c("b")
#   ^
```

Should we allow an exception for such cases?

The way to silence the linter is to use `.`, which may in fact be more readable:

```
writeLines('
"a" %>%
  c( . ,"b")
', tmp <- tempfile())
lintr::lint(tmp, lintr::unneeded_concatenation_linter)
```

michaelquinn32:
This fix tries to be lightweight, and mostly focused on how magrittr affects the arg count.

Fixes #573


AshesITR:
Thanks for looking into this.
Could you check if this works with the newly added native pipe `|>` in r-devel as well?

MichaelChirico:
as I mentioned here https://github.com/jimhester/lintr/pull/686#issuecomment-739522450 I'm a bit hesitant to rush off & make changes for a syntax that's not 100% confirmed in r-devel yet

AshesITR (approved):
Thanks for fixing this

PR #587, opened 2020-11-25
Change reporting of lints in GHA

### LINKED ISSUE #558: Refactor github_actions_log_lints()
Has anyone had lints found via GHA actually print file and line info in the log?

In my hands, it's not working so far.

jennybc:
Closes #558

Reports the linting via a GHA annotation (a warning) but also puts file and line details into the log

### DIFF
--- a/R/actions.R
+++ b/R/actions.R
@@ -5,9 +5,12 @@ in_github_actions <- function() {
 # Output logging commands for any lints found
 github_actions_log_lints <- function(lints) {
   for (x in lints) {
-    cat(
-      sprintf("::warning file=%s,line=%s,col=%s::%s\n", x$filename, x$line_number, x$column_number, x$message),
-      sep = ""
+    file_line_col <- sprintf(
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Does it make sense to extend this test to match the new expected format? Happy to merge anyway

https://github.com/jimhester/lintr/blob/8fbc489c0cd9ce0d4e37029cd3d29d12e6cd8572/tests/testthat/test-actions.R#L11

MichaelChirico:
Does it make sense to extend this test to match the new expected format? Happy to merge anyway

https://github.com/jimhester/lintr/blob/8fbc489c0cd9ce0d4e37029cd3d29d12e6cd8572/tests/testthat/test-actions.R#L11

jennybc:
I would probably leave this as is. When/if you update for testthat 3rd edition, this will become (or gain) a snapshot test anyway. So you won't have to hardwire your exact expectation, you'll just capture "good output" once and be alerted if things change.

MichaelChirico (approved):

PR #589, opened 2020-11-26
Remove duplicate definition of rstudio_source_markers()

jennybc:
Fixes #588

PR #590, opened 2020-11-26
Make inline data work in RStudio

jennybc:
~(#589 should be merged first)~ *done*

Thanks to @jimhester I now know it's possible to provide to-be-linted source directly to `lint()` by passing a string with a newline.

But printing the resulting lints doesn't work in RStudio, because there's no underlying source file. It looks something like this:

```
> library(lintr)
> linter <- undesirable_function_linter(fun = c("dir" = NA))
> lintr::lint('dir(".")\n', linter)
Error in .rs.normalizePath(marker$file, mustWork = TRUE) : 
  path[1]="<text>": No such file or directory
```

whereas this is what you see outside the IDE:

```
library(lintr)
linter <- undesirable_function_linter(fun = c("dir" = NA))
lint('dir(".")\n', linter)
#> <text>:1:1: warning: Function "dir" is undesirable.
#> dir(".")
#> ^~~
```

This PR makes lint printing work in RStudio, even when the source code was inlined (vs. came from a source file).

### DIFF
--- a/R/expect_lint.R
+++ b/R/expect_lint.R
@@ -45,7 +45,7 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
 
   if (is.null(file)) {
     file <- tempfile()
-    on.exit(unlink(file))
+    on.exit(unlink(file), add = TRUE)
     writeLines(content, con = file, sep = "\n")
   }
 
--- a/R/lint.R
+++ b/R/lint.R
@@ -378,11 +378,23 @@ rstudio_source_markers <- function(lints) {
   })
 
   # request source markers
-  rstudioapi::callFun("sourceMarkers",
-                      name = "lintr",
-                      markers = markers,
-                      basePath = package_path,
-                      autoSelect = "first")
+  out <- rstudioapi::callFun(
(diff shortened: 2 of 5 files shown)


R/actions.R
```diff
@@ -8,34 +8,3 @@ github_actions_log_lints <- function(lints) {
     cat(sprintf("::warning file=%s,line=%s,col=%s::%s\n", x$filename, x$line_number, x$column_number, x$message), sep = "")
   }
 }
-
```
jennybc, line 11: See #589

MichaelChirico:
Could you please add a test? Have a look in #630 for how to trick testing into behaving like RStudio

jennybc:
Test added.

The printing would probably be easier to test if all possible smarts were in `format.*` methods. The current use of `testthat::expect_output()` and `testthat::with_mock()` won't be allowed when/if upgrading to testthat 3rd edition. But I've made this test resemble the existing `print.*` test as closely as possible.

MichaelChirico (approved):

MichaelChirico:
Thanks @jennybc! filed #651 as a follow-up to this.

Issue #584, opened 2020-11-25
Who lints the linters?


russHyde:
Although there is a `.lintr` file in the lintr root, linting is not done as part of the current travis-CI workflow (or the soon to be introduced GHA CI workflow).

If people are keen to integrate lintr into their dev workflow, they're probably going to look at lintr as a good example of how to set this up. For those who read docs, the current docs explain how to set up lintr for running on Travis; they might need expanding to explain how to use lintr on GHA / other providers. For those who don't read docs, it would be good if `{lintr}` had exemplary lint-running code that they can copy/paste into their own projects (or perhaps a `use_lintr` hook could be constructed for `{usethis}` that makes setting up linting trivial).

Nonetheless, the current linting setup fails:

With master: f878ec99839d36289fcb14b4d811986933b93174 
if we run `Rscript -e "devtools::load_all(); lint_package()"` in the `lintr` root, we get the following:
```
Loading lintr
Error: Malformed file!
Execution halted
```

If we run `Rscript -e "devtools::load_all(); lint_dir('R')"` we get a long list of lints. (and also for `lint_dir("tests")`)

Some of these lints 

- are due to bugs in lintr (some fixed in recent but unmerged PRs: #441, #546
- are due to the "exclusions" not being updated when example/test code is added that is designed to fail the linters (@AshesITR 's recent directory-exclusion PR should ease fixing this kind of stuff)
- would have been fixed if linting had been a consistent part of the lintr PR workflow.

Should we fixup the lintr PR process so that linting is always run?
And if we do, how do we change the current code base so that linting the current code comes back green?

What we shouldn't do is a monolithic PR using `{styler}` or something that fixes hundreds of `function(arg=123)` to `function(arg = 123)`; when these wholesale-change commits introduce bugs they are hell to fix.

I'd suggest we rewrite the .lintr:

- exclude all test & doc directories for now
- use the smallest collection of the default linters that come back green on `<lintr>/R`
- apply these linters to any newly submitted PR
- then do individual PRs to gradually add further default-linters back in

An alternative would be to write up a .lintr that defines the linters that we definitely do want to apply to lintr and write up a CI workflow script that applies these linters to _just_ those files that were modified in a submitted PR; and if a lint crops up on a line that was touched by the PR author then linting should return red from CI.

AshesITR:
Here are some stats on babe9f7 (current HEAD of #539) for `lintr::lint_package()`:



<details>
<summary>By linter</summary>

|linter                           |   n|
[...]

PR #591, opened 2020-11-26
apply some linters to R/

MichaelChirico:
Part of #584 

I applied several linters across R/ here:

 - `commas_linter`
 - `trailing_blank_lines_linter`
 - `spaces_left_parentheses_linter`
 - `object_usage_linter`
 - `function_left_parentheses_linter`
 - `paren_brace_linter`

I skipped `equals_na_linter` which only gives a spurious match that would be fixed by #546. Similarly there's a spurious instance of `paren_brace_linter` that applies to it's own source where `"){"` is present. We should move that linter to match on the parse tree as well.

### DIFF
--- a/R/cyclocomp_linter.R
+++ b/R/cyclocomp_linter.R
@@ -22,7 +22,7 @@ cyclocomp_linter <- function(complexity_limit = 25) {
       type = "style",
       message = paste0(
         "functions should have cyclomatic complexity of less than ",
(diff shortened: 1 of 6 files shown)


russHyde:
Hi. I moved `paren_brace_linter` over to the parse tree in #441 (which is now pretty out  of date)

AshesITR (approved):

PR #609, opened 2020-11-28
Rename xml_parsed_content for the full file

### LINKED ISSUE #608: Need a better way to skip when XML parse fails
Split off from #546.

Right now we have to use this workaround:

```
(!length(source_file$full_parsed_content) || is.null(source_file$xml_parsed_content)) return(list())
```

This is pretty arcane & unreadable IMO -- hard to know without being pretty experienced with `lintr` internals what's going on here. And with lints, it's tough to just wrap that into a helper function because now that's something we need to export for downstreams to use as well I think.

Both checks are needed because

 1. We want to skip on the last expression (where `full_parsed_content` is present) and
 2. We can't use `!is.null(source_file$parsed_content)` because there are cases where `parsed_content` is present (e.g. in some parsing error situations, as fixed in #560) but `xml_parsed_content` is not. lastly
 3. We can't skip based on `xml_parsed_content` because that element has the same name for the full file & individual expressions.

I think the easiest solution is to rename `xml_parsed_content` to `full_xml_parsed_content` to be consistent with `parsed_content` :arrow_right: `full_parsed_content`. There is some (?) breakage potential here in the event that some downstream logic depends on `xml_parsed_content` being the name for the full file, but I think it's worth the change for consistency & readability.

MichaelChirico:
Closes #608 

Renaming entailed fixing some logic that relied on the name being `xml_parsed_content`. In every case I think the new code is cleaner.

Also tidied a few things in the midst of fixing tests that were breaking:

 - line width lint applied to `get_assignment_symbols` and `get_function_assignments` in `object_usage_linter`
 - `global_parsed_content` in `utils` is unused. Also, it's wrong (it should be `source_file$full_parsed_content`), so I just deleted it.
 - minor typo in some test files: `someComplicatedFunctionWithALongCameCaseName` -> `someComplicatedFunctionWithALongCamelCaseName`

TODO:

 - [x] Test against downstream `goodpractice`

tested both CRAN & GitHub master:

```
install.packages('goodpractice')
tools::testInstalledPackage('goodpractice')

# restart
remotes::install_github("MangoTheCat/goodpractice")
tools::testInstalledPackage('goodpractice')
```

Both passed

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -29,9 +29,8 @@
 #'           for .Rmd scripts, this is the extracted R source code (as text)}
 #'     \item{`full_parsed_content` (`data.frame`) as given by
 #'           [utils::getParseData()] for the full content}
-#'     \item{`xml_parsed_content` (`xml_document`) the XML parse tree of all
+#'     \item{`full_xml_parsed_content` (`xml_document`) the XML parse tree of all
 #'           expressions as given by [xmlparsedata::xml_parse_data()]}
-#'     \item{`is_full_file` (`TRUE`) records that this
 #'     \item{`terminal_newline` (`logical`) records whether `filename` has a terminal
 #'           newline (as determined by [readLines()] producing a corresponding warning)}
 #'   }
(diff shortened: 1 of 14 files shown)


AshesITR (approved):
Thanks, LGTM

PR #611, opened 2020-11-28
apply maximum line width of 120 characters

### LINKED ISSUE #593: Make R/ conform to `line_length_linter(120)`
Part of #584 

- [ ] set `line_length_linter` to 120 in `.lintr`
- [ ] fix all lints

MichaelChirico:
Closes #593 

> set line_length_linter to 120 in .lintr

That's already done for this linter right?


R/lint.R
```diff
@@ -245,7 +245,9 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = NULL,
 #'   )
 #' }
 #' @export
-lint_package <- function(path = ".", relative_path = TRUE, ..., exclusions = list("R/RcppExports.R"), parse_settings = TRUE) {
+lint_package <- function(
+  path = ".", relative_path = TRUE, ..., exclusions = list("R/RcppExports.R"), parse_settings = TRUE
+) {
```
MichaelChirico, line 250: LMK the preferred way to write long function signatures. I'm partial to

```
foo <- function(
  args...
) {
```

But I often see

```
foo <- function(arg1,
                arg2,
                arg2,
                ...) {
```

AshesITR:
Once the kick-off PR is merged (currently waiting for CI to pass), the line length will be 160.

R/lint.R
AshesITR: The rest of the codebase, e.g. `lint_dir` and `lint_package` use the latter style. We should be consistent I think.

R/get_source_expressions.R
```diff
@@ -169,7 +169,7 @@ get_source_expressions <- function(filename) {
       file_lines = source_file$lines,
       content = source_file$lines,
       full_parsed_content = parsed_content,
-      full_xml_parsed_content = if (!is.null(parsed_content)) tryCatch(xml2::read_xml(xmlparsedata::xml_parse_data(parsed_content)), error = function(e) NULL),
+      full_xml_parsed_content = if (is.null(parsed_content)) NULL else safe_parse_to_xml(parsed_content),
```
AshesITR, line 172: Can we move the `NULL` check to `safe_parse_to_xml` as well?

R/closed_curly_linter.R
```diff
@@ -37,7 +37,9 @@ closed_curly_linter <- function(allow_single_line = FALSE) {
              content_after <- unname(substr(line, parsed$col1 + 1L, nchar(line)))
              content_before <- unname(substr(line, 1, parsed$col1 - 1L))
 
-             double_curly <- rex::re_matches(content_after, rex::rex(start, "}")) || rex::re_matches(content_before, rex::rex("}", end))
+             double_curly <-
+               rex::re_matches(content_after, rex::rex(start, "}")) ||
```
AshesITR, line 41: Not quite sure, but I think the default tidyverse style doesn't encourace trailing `<-`.
Would one line break after `||` be sufficient for the 120-char limit?

R/open_curly_linter.R
```diff
@@ -33,7 +33,9 @@ open_curly_linter <- function(allow_single_line = FALSE) {
 
              only_comment <- rex::re_matches(content_after, rex::rex(any_spaces, "#", something, end))
 
-             double_curly <- rex::re_matches(content_after, rex::rex(start, "{")) || rex::re_matches(content_before, rex::rex("{", end))
+             double_curly <-
+               rex::re_matches(content_after, rex::rex(start, "{")) ||
```
AshesITR, line 37: Same as in `closed_curly_linter.R`.

AshesITR (changes requested):
Thanks for helping with the delinting process.

R/get_source_expressions.R
MichaelChirico: OK. I only didn't initially because the other usage doesn't have the `NULL` check (I assume because it's always non-`NULL` there, but wasn't sure enough to proceed).

R/closed_curly_linter.R
MichaelChirico: I don't see anything in the style guide:

 - long lines: https://style.tidyverse.org/syntax.html#long-lines
 - assignment: https://style.tidyverse.org/syntax.html#assignment-1
 - long lines again: https://style.tidyverse.org/functions.html#long-lines-1

Locally I have `xml2`, `dplyr`, and `roxygen2`, only `roxygen2` has one line ending in `<-`:

```
grep -Er "<-$" dplyr/R | wc -l
```

So at a glance it looks avoided; will change.

AshesITR (dismissed):
LGTM, thanks!

AshesITR:
Codecov could be made to pass by adding a new test for `expect_lint` that expects an error when giving a regex with capture groups.

R/actions.R
```diff
@@ -5,7 +5,10 @@ in_github_actions <- function() {
 # Output logging commands for any lints found
 github_actions_log_lints <- function(lints) {
   for (x in lints) {
-    cat(sprintf("::warning file=%s,line=%s,col=%s::%s\n", x$filename, x$line_number, x$column_number, x$message), sep = "")
+    cat(
```
MichaelChirico, line 8: should we nocov this?

AshesITR: Would be fine with it. Could also add a test similar to `trim_output()`, shouldn't be too hard because it only `cat()`s, so `testthat` can capture and check its output.

MichaelChirico: For some reason I assumed we wouldn't be able to get this working without actually being on GHA. But we can imitate that by messing with env & options, so I added a test.

tests/testthat/test-actions.R
```diff
@@ -0,0 +1,13 @@
+test_that("GitHub Actions functionality works", {
+  # imitate being on GHA whether or not we are
+  withr::with_envvar(c(GITHUB_ACTIONS = "true"), {
+    writeLines("x <- 1:nrow(y)", tmp <- tempfile())
+    on.exit(unlink(tmp))
+
+    old = options(lintr.rstudio_source_markers = FALSE)
```
AshesITR, line 7: `=` -> `<-`

AshesITR (approved):

Issue #597, opened 2020-11-27
Make R/ conform to `object_name_linter`


AshesITR:
Part of #584 

Requires #621 and #633

- [ ] enable `object_name_linter` in `.lintr`
- [ ] run `lintr::lint_package()` to find all lints
- [ ] fix all lints

PR #614, opened 2020-11-28
Allow .onLoad and other special functions regardless of name style

### LINKED ISSUE #500: `object_name_linter()` complains about ns-hooks (e.g. `.onLoad`)
Reprex:

```r
example_file <- tempfile()

readr::write_file(".onLoad <- function(libname, pkgname) {\n  # some action\n}\n",
                  path = example_file)

lintr::lint(filename = example_file, linters = lintr::object_name_linter())
```

R's [hooks for namespace events](https://rdrr.io/r/base/ns-hooks.html) are CamelCase and you can't change that AFAIK, so lintr complaining about

```
Variable and function name style should be snake_case.
```

is pointless.

AshesITR:
Also remove relevant # nolint in lintrs own .onLoad.

fixes #500

### DIFF
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -240,6 +240,7 @@ is_base_function <- function(x) {
   x %in% base_funs
 }
 
+# see ?".onLoad" and ?"Startup"
 special_funs <- c(
   ".onLoad",
   ".onAttach",
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -129,7 +129,8 @@ make_object_linter <- function(fun) {
     keep_indices <- which(
       !is_operator(names) &
         !is_known_generic(names) &
-        !is_base_function(names)
+        !is_base_function(names) &
+        !is_special_function(names)
     )
 
     lapply(
@@ -237,6 +238,20 @@ base_funs <- unlist(lapply(base_pkgs,
 
 is_base_function <- function(x) {
   x %in% base_funs
+}
+
+special_funs <- c(
+  ".onLoad",
+  ".onAttach",
+  ".onUnload",
+  ".onDetach",
+  ".Last.lib",
+  ".First",
+  ".Last"
+)
+
+is_special_function <- function(x) {
+  x %in% special_funs
 }
 
 object_lint <- function(source_file, token, message, type) {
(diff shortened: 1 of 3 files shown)


R/object_name_linters.R
```diff
@@ -240,6 +240,7 @@ is_base_function <- function(x) {
   x %in% base_funs
 }
 
+# see ?".onLoad" and ?"Startup"
```
AshesITR, line 243: Thanks, very good idea to note where they come from :-)

R/object_name_linters.R
```diff
@@ -129,7 +129,8 @@ make_object_linter <- function(fun) {
     keep_indices <- which(
       !is_operator(names) &
         !is_known_generic(names) &
-        !is_base_function(names)
+        !is_base_function(names) &
+        !is_special_function(names)
```
MichaelChirico, line 133: Is there a better name than `special`? maybe `special_hook`? "special" seems a bit generic. If no better option springs to mind, just go ahead.

MichaelChirico (dismissed):

R/object_name_linters.R
MichaelChirico: (also for the NEWS)

AshesITR: I thought if we ever need to add additional functions to this list for whatever reason, we might not want to blow up the conditional and instead just update the list of special functions with more comments on why they're special.

MichaelChirico: OK that makes sense. I was going to suggest improving the NEWS but I think that's done now too.

AshesITR:
@MichaelChirico can you re-approve (merging master wasn't clean)? Thanks!

MichaelChirico (approved):

R/object_name_linters.R
```diff
@@ -239,6 +240,21 @@ is_base_function <- function(x) {
...
+  ".onLoad",
+  ".onAttach",
+  ".onUnload",
+  ".onDetach",
+  ".Last.lib",
+  ".First",
+  ".Last"
+)
+
+is_special_function <- function(x) {
```
MichaelChirico, line 254: hmm I guess we need a test for this :sweat_smile: I'm seeing this on current master when trying to do #597:

```
R/zzz.R:185:1: style: Variable and function name style should be snake_case.
.onLoad <- function(libname, pkgname) {
^~~~~~~
```

MichaelChirico: The issue is that `strip_names` removes the leading `.`. So actually we need to search for `onLoad` etc.

PR #622, opened 2020-11-29
first pass at implementing a backport_linter

### LINKED ISSUE #506: R version compatibility linter
For package developers, it would be quite helpful to know if the code is compatible with the declared R version.

For example, if one uses `trimws(whitespace=)`, then it is only compatible with R >= 3.6.0. However, developers using R >= 3.6.0 might not be aware of this until some user or CI using R < 3.6.0 complains.

I don't know a simple way to find how compatible the functions I use in package development are unless I carefully go through the [R News](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html), [R News 3.x](https://cran.r-project.org/doc/manuals/r-devel/NEWS.3.html), etc.

I'm wondering if there's a way to make this process work in a more automatic manner, e.g. create a linter that checks the usage of functions are compatible with declared version requirement? Or suggest the use of `backport` instead?

MichaelChirico:
Closes #506 

For now it's just a simple version. We simply gather all* the new functions mentioned in `NEWS.3` and `NEWS` and check if those names are used at all in the script. My guess is that the chance of namespace collisions is low for base functions, so I didn't bother with any namespace qualification complexity.

I think adding a check for raw strings is doable in this PR, the only thing is I think we need to do a `skip` for related tests right? Since we rely on `parse()` to work.

My gut says checking R-level env variables which are added over time is not likely to bear much fruit -- a test of this would be to identify a few and cran-search whether any are used in CRAN packages.

Checking argument consistency I would say is beyond scope for now. In particular, note how complicated this will be for use cases involving `lapply` & co. I would wait on seeing if there's FR for that first :smiley: 

* I did this by eye so there's a chance I missed something


R/backport_linter.R
```diff
@@ -0,0 +1,83 @@
+#' @describeIn linters that checks for usage of unavailable functions. Not reliable for testing r-devel dependencies.
+#' @export
+backport_linter <- function(r_version = "4.0.3") {
+  function(source_file) {
+    if (is.null(source_file$xml_parsed_content) || r_version >= "r-devel") return(list())
```
AshesITR, line 5: Why is there a lexicographic comparison to "r-devel"?
Both "r-release" and "r-oldrel" would trigger this.

R/backport_linter.R
```diff
@@ -0,0 +1,83 @@
...
+
+    names_xpath <- "//*[self::SYMBOL or self::SYMBOL_FUNCTION_CALL]"
+    all_names_nodes <- xml2::xml_find_all(xml, names_xpath)
+    all_names <- xml2::xml_text(all_names_nodes)
+
+    # guaranteed to include 1 by early return above; which.min fails if all TRUE (handled by nomatch)
+    needs_backport_names <- backports[1:(match(FALSE, r_version < names(backports), nomatch = length(backports)) - 1L)]
+
+    # not sapply, which may over-simplify -- cbind makes sure apply works
+    needs_backport <- do.call(cbind, lapply(needs_backport_names, function(nm) all_names %in% nm))
```
AshesITR, line 19: `do.call(cbind, lapply(...))` looks like `vapply()` might be the way to go.
For atomic length `n > 1` results, `vapply` cbinds the results into a matrix.

R/backport_linter.R
```diff
@@ -0,0 +1,83 @@
...
+    bad_idx <- apply(needs_backport, 1L, any)
```
AshesITR, line 20: `rowSums(needs_backport) > 0L` is much faster

```
> microbenchmark::microbenchmark(
+   apply(matrix(c(T, F, T, F), nrow = 2), 1, any),
+   rowSums(matrix(c(T, F, T, F), nrow = 2)) > 0L
+ )
Unit: microseconds
                                           expr  min   lq   mean median    uq  max neval
 apply(matrix(c(T, F, T, F), nrow = 2), 1, any) 11.3 11.6 12.189  11.80 12.05 33.1   100
  rowSums(matrix(c(T, F, T, F), nrow = 2)) > 0L  3.3  3.5  3.971   3.75  4.10 17.7   100
```

R/backport_linter.R
```diff
@@ -0,0 +1,83 @@
+#' @describeIn linters that checks for usage of unavailable functions. Not reliable for testing r-devel dependencies.
+#' @export
+backport_linter <- function(r_version = "4.0.3") {
```
AshesITR, line 3: Can the default auto-detect, e.g. `getRversion()`?
Or even better for Packages, get the R version from the `Depends` field in `DESCRIPTION`? (this might be worth a separate PR, also suppressing lints for stuff imported from {backports})

AshesITR (changes requested):

R/backport_linter.R
renkun-ken, line 3: Should the default `r_version` be the current R version?

If a package author wants to use this linter, then one must specify the `r_version` to be consistent with that appears in the package `DESCRIPTION`?

R/backport_linter.R
renkun-ken, line 19: I think there's a tricky part where many developers do their own backports in their code if necessary. For example, they define a `isFALSE <- function(x) identical(x, FALSE)` in the package first and then use it in other places (I don't see this function in the `backports` list?)

In this case, the approach seems to always give false results?

R/backport_linter.R
MichaelChirico: Maybe I should just leave this as `"r"` -- I don't think we want to support tracking devel dependencies, I just wanted something that would be bigger than all R version numbers (at first I used `"999999"` as the name, `"r-devel"` felt better but you raise a good point)

R/backport_linter.R
MichaelChirico: The issue is with `n=1` (`length(all_names)=1`), where a vector is returned and then `apply` fails. `cbind` guarantees we have a matrix (I tried `sapply` first and got the same issue, I'll add `vapply` to the comment)

R/backport_linter.R
MichaelChirico: thanks... it's a shame that's the case, huh :)

R/backport_linter.R
MichaelChirico: good idea, and a good direction for follow-up. Actually I think we can go beyond `backports` (there may be other backport packages) and try to check namespace imports and/or internally defined functions & skip those. New argument `skip=` defaulting to `infer_skips` where we apply some logic.

MichaelChirico: Now I used `getRversion()`, I guess we could make it `guess_r_version()` & check the description first before defaulting to `getRversion()`?

R/backport_linter.R
AshesITR: Couldn't we consult `names(backports)` to find whether the selected version is supported?
The highest of those versions wouldn't cause lints anyway so we can `tail()` it to obtain a definitive list of R versions that can result in lints if I'm not mistaken.

R/backport_linter.R
MichaelChirico: good catch! I missed it because it's tucked at the bottom of another bullet point. this would have been a lot easier if the formatting of `NEWS` was more consistent :sweat_smile:

MichaelChirico: see above for catching custom defines.

AshesITR: Thanks for clarifying :)
An alternative would be `dim(needs_backport) <- c(length(all_names), length(needs_backport_names))` after `vapply`. But I don't mind since the number of columns is bounded by the number of backports - the `do.call(cbind, ...)` overhead is at least only constant and independent of user code.

R/backport_linter.R
MichaelChirico: if so we'd have to include in the list all released versions (e.g. 3.5.3).

And sorry I'm just waking up, I'm not following your second point

R/backport_linter.R
AshesITR: Curious if we should allow explicit whitelisting. A user with a `backport` lint could take appropriate measures (e.g. import {backports}) and then edit the `.lintr` file like `backports_linter(safe = "fun_that_has_a_local_fallback_implementation")`.
WDYT?

MichaelChirico: Can also just `cbind` or `rbind` the `vapply` result to get matrix, but I figured it wouldn't improve readability. Hadn't thought of scalability but yes, the number of rows comes from the code size while we control the number of columns.

MichaelChirico: that's what I'm thinking yep, same idea as my comment above about `skip=`.

R/backport_linter.R
AshesITR: Good point.
Wouldn't we never lint if `r_version` is at least `getRversion()`?
i.e. `if (r_version >= getRversion()) return(list())`.
That way if r_version is older than the current environment, backport problems can be detected.
If it isn't there are no backports that can be used from the current R version which would be a problem in `r_version`, right?

R/backport_linter.R
AshesITR: I suggest using `except` as the argument name - it's already used in `missing_argument_linter` with the same purpose:
```
#' @param except a character vector of function names as exceptions.
```

R/backport_linter.R
MichaelChirico: I think that's _basically_ right, but I do think it's possible to run this on a machine that has a different environment than required. Say I work on two machines and the one I'm using today has an old version. We wouldn't want to skip then necessarily right?

AshesITR: On the old version, the used backports would either break the code completely (which you'd notice), or the backports were already safely ported to your old version. No?

MichaelChirico: That's if we run tests, but `lintr` is static analysis so we could in principle run it on a different machine that doesn't execute code.

If that machine is running for a bigger library of R code (say a local CRAN), there may be different code requirements.

I think it's basically an edge case... but taking a step back, the goal of this is to get around/change the `r_version >= "r"` part? I'm happier with that now than I was with `r_version >= "r-devel"`

AshesITR: I agree. Let's just leave it as-is for now.
Maybe later become smart about what `r_version`s are found in `backports` and thus will potentially cause lints.

R/backport_linter.R
```diff
@@ -0,0 +1,82 @@
...
+        line = source_file$lines[[line1]],
+        ranges = list(c(col1, col2)),
+        linter = "backport_linter"
+      )
+    })
+  }
+}
+
+backports <- list(
+  `r` = c("...names", "checkRdContents", "numToBits", "numToInts", "packBits"),
```
AshesITR, line 48: I think this needs to be `r-devel`

AshesITR: (or maybe just "devel"). The tests depend on the exact name and the lint message looks best with `devel` imo:

```
...names (R devel) is not available for dependency R >= 3.0.0.
```

tests/testthat/test-backport_linter.R
```diff
@@ -0,0 +1,28 @@
+test_that("backport_linter detects backwards-incompatibility", {
+  # this test may be too fragile?
+  expect_lint("numToBits(1)", NULL, backport_linter("r-devel"))
+
+  # default should be current R version; all of these are included on our dependency
+  expect_lint(".getNamespaceInfo(dir.exists(lapply(x, toTitleCase)))", NULL, backport_linter())
+
+  expect_lint(
+    "numToBits(2)",
+    rex("numToBits (R r-devel) is not available for dependency R >= 4.0.0."),
```
AshesITR, line 10: These messages need adapting to the new name

R/backport_linter.R
```diff
@@ -0,0 +1,82 @@
+#' @describeIn linters that checks for usage of unavailable functions. Not reliable for testing r-devel dependencies.
+#' @export
+backport_linter <- function(r_version = getRversion()) {
+  function(source_file) {
+    if (inherits(r_version, "numeric_version")) r_version <- format(r_version)
+    if (is.null(source_file$xml_parsed_content) || r_version >= "r") return(list())
```
AshesITR, line 6: Taking another look, can't we just remove that condition?
It's never going to be true for any `r_version`.
Sorry for being slow.

MichaelChirico: I think you're right because here:

```
1:(match(FALSE, r_version < names(backports), nomatch = length(backports)) - 1L)
```

 1. If the condition is mixed `TRUE`/`FALSE` that's normal
 2. If it's all `TRUE`, we use the `nomatch`
 3. If it's all `FALSE` we get back `1`

So the edge case I was guarding against is irrelevant as implemented.

I should replace it with a warning about trying to depend on an R version older than we've recorded.

AshesITR: Sounds good. i.e. `numeric_version(r_version)[1L, 1L] < 3`

R/backport_linter.R
```diff
@@ -0,0 +1,86 @@
+#' @describeIn linters that checks for usage of unavailable functions. Not reliable for testing r-devel dependencies.
+#' @export
+backport_linter <- function(r_version = getRversion()) {
+  function(source_file) {
+    if (inherits(r_version, "numeric_version")) r_version <- format(r_version)
+    if (r_version < "3.0.0") {
```
AshesITR, line 6: We should use `numeric_version` comparison here, else it will bite us when R 10.0.0 rolls out ;)

AshesITR (approved):

R/backport_linter.R
MichaelChirico: haha that can be a problem for our future selves -- after 2050 at the current rate :smile_cat:

I prefer this way because the test below makes the same assumption by doing `r_version < names(backports)`

So really this assumption is backed into `names(backports)` as well. Maybe by R 10.0 R will allow objects to have objects as names like python

MichaelChirico:
BTW @AshesITR we should come up with a procedure for new linters & whether/how to include them as default linters / whether/how to include them in the GH action for linting `lintr`...

e.g. we could make that part of a separate PR, but in this PR we could add it in `.lintr` with a TODO?

AshesITR:
What do you mean?
add `backports_linter` to our `.lintr` now with a TODO for adding it to default linters?
The GH action (and travis) currently use our `.lintr` and I think that's a good way to have it because everyone can run `lintr::lint_package()` locally and expect the same results.

IMO adding a linter to the default linters should always be a separate issue.

MichaelChirico:
Do we only lint `lintr` itself with default linters?

MichaelChirico:
Hmm I want to include a test for raw strings, however, there's an issue:

The parse data for raw strings is not returned correctly in 4.0.0 (only in 4.0.1), so the test can't work on 4.0?

I'll have to think about it more; for now, let's merge this and file that as a follow-up?

AshesITR:
Sounds good to me.
WRT linting lintr with only the default linters: I don't mind having more strict linting in lintr itself than default.

renkun-ken (approved):
LGTM

AshesITR (dismissed):

AshesITR (dismissed):

AshesITR (approved):

PR #625, opened 2020-11-29
Make sprintf_linter capture warnings

### LINKED ISSUE #624: sprintf_linter causes test warning on r-devel
https://travis-ci.org/github/jimhester/lintr/jobs/746547257#L1728

```
── Warning (test-sprintf_linter.R:7:3): returns the correct linting ────────────

one argument not used by format 'hello'

── Warning (test-sprintf_linter.R:43:3): returns the correct linting ───────────

one argument not used by format 'hello %1$s %s'

── Warning (test-sprintf_linter.R:47:3): returns the correct linting ───────────

one argument not used by format 'hello %1$s %1$s'
```

Should catch the warning in the linter using `withCallingHandlers` and turn it into a lint.

AshesITR:
fixes #624

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,7 @@
 * Fix possible error on invalid XML produced by xmlparsedata (#559, #560, @renkun-ken)
 * Fix handling zero-length variable name error (#566, #567, @renkun-ken)
 * New `missing_argument_linter()` (#563, #565, @renkun-ken)
-* New `sprintf_linter()` (#544, #578, @renkun-ken)
+* New `sprintf_linter()` (#544, #578, #624, #625, @renkun-ken, @AshesITR)
 * Exclusions specified in the `.lintr` file are now relative to the location of that file 
   and support excluding entire directories (#158, #438, @AshesITR)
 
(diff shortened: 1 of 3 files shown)


renkun-ken (dismissed):
LGTM. Thanks for capturing and fixing this.

AshesITR:
As an added bonus, this should lint for unused arguments now on r-devel.
I expect the current test to fail because of that, though.

renkun-ken:
Not sure if a new issue should be raised. The `sprintf_linter` will complain about the following code:

```r
f <- function(id, ...) {
  sprintf("%d %d, %d", id, ...) ### to few arguments
}
```

since `...` is regarded as typical symbol and replaced with `0` before eval. Any idea how we should handle this with currentl implementation?

AshesITR:
Good catch - definitely worth an issue since I don't immediately see how to find out the required cardinality of `...` without writing a proper parser.

NB that a call like `f(42, fmt = "bla")` can really mess things up in the example.
I'd generally consider it a bad idea to put a fixed format string together with `...`. Maybe not worth fixing.

renkun-ken (approved):
LGTM

PR #629, opened 2020-11-29
coverage tests for zzz.R

MichaelChirico:
- `named_list` is unused in R/, so deleted
`str.lintr_function` won't be called unless we `@export` it to get it registered as S3. Of course, given that it hasn't been used till now, the other option is to just delete this function; LMK.


AshesITR:
@jimhester do you remember a purpose for `str.lintr_function()`?
If not, I'd vote for deletion.

jimhester:
You can delete it.

AshesITR (approved):

R/utils.R
```diff
@@ -134,11 +134,13 @@ logical_env <- function(x) {
 }
 
 # from ?chartr
+# nocov start used in .onLoad which is nocov
```
AshesITR, line 137: rot() could still be tested, no?
```
expect_identical(
  rot(c(letters, LETTERS, " ", "'")),
  c(letters[14L:26L], LETTERS, " ", "'", letters[1L:13L])
)
```

AshesITR (approved):

AshesITR (changes requested):
Please add a test for `rot()` instead of `# nocov`

MichaelChirico:
I don't see a way to test `rot` without exposing the private method, is all -- I was recently pointed to

https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones

so I've become more hesitant about testing things that we can't get to directly... if the intention is to get to 100% on covr then that's also fine. any thoughts?

AshesITR:
I'm thinking a test wouldn't hurt as the public interface using the code is # nocov'd (`.onLoad`)
(and if we did somehow manage to test `.onLoad`, that would also produce full coverage for `rot()` without a separate test)
So aiming for 100% coverage where practical would be my default go-to.

No strong opinon on this instance though. Do as you see fit, the rest of the diff looks good to me - approving.

AshesITR (dismissed):

AshesITR (approved):

MichaelChirico:
but rot is not exported right?

I was recently pointed to this when trying to test private functions:

https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods

I'm happy to do so if our goal is 100% coverage, or if you can think of a
way to test it obliquely

PR #630, opened 2020-11-29
Coverage tests for comments.R

MichaelChirico:
I got a decent way to being able to test all but `github_comment` (which I think we should `nocov`), however I got stuck. See the commented test -- to proceed, we'd need a way to temporarily set `settings$comment_bot` within a test. 

I might just be missing something, is there an easy way to do so? Is setting `LINTR_COMMENT_BOT=true` temporarily enough?


AshesITR:
According to the docs, `withr::with_env(LINTR_COMMENT_BOT = "true", {})` should do the trick.
If not, we should file an issue for that.

AshesITR (approved):

PR #633, opened 2020-11-30
fix namespace hooks exception in object_name_linter

### LINKED ISSUE #500: `object_name_linter()` complains about ns-hooks (e.g. `.onLoad`)

MichaelChirico:
Closes #500 

I think the issue is that the old one relies on using `object_name_linter_old`, which appears obsolete.

I also noticed that we shouldn't have found this issue yet in the first place, since `lintr:::is_base_function(".onLoad")` is `TRUE`. Again the issue is that we `strip_names` the variables found in the file. So if we apply `strip_names` to the `base_functions`, that should put the comparison on equal footing.


AshesITR (dismissed):
LGTM

AshesITR (approved):

PR #621, opened 2020-11-29
Object name operator

MichaelChirico:
Closes #615

PR #639, opened 2020-11-30
enable infix_spaces_linter, and apply it to the package

### LINKED ISSUE #594: Make R/ conform to `infix_spaces_linter`
Part of #584 

- [ ] enable `infix_spaces_linter` in `.lintr`
- [ ] fix all lints

MichaelChirico:
Closes #594


AshesITR:
Codecov complains about uncovered code in `object_name_linters.R` (seems unused - remove?) and `namespace.R` (haven't looked into it)

MichaelChirico:
I think the namespace one is actually a legit bug. `namespace_imports` is only called here:

https://github.com/jimhester/lintr/blob/fa05ceb2bf967ddad9230e25879b95084c80b0dc/R/object_name_linters.R#L57

without an argument. the default is to run `find_package()` (again without argument), but there's no default, so `normalizePath` always fails:

https://github.com/jimhester/lintr/blob/fa05ceb2bf967ddad9230e25879b95084c80b0dc/R/lint.R#L287

In `namespace_imports` this error happens within `tryCatch` so it just returns `NULL`.

I'm not sure the intended behavior but certainly the above is not it :smiley:

MichaelChirico:
For the unused code in object_name_linters.R, I removed that code in #621, so let's just merge that one first.

Hopefully then we'll pass codecov here.

AshesITR:
The intended behaviour seems to be to `find_package()` on the to-be-linted file, i.e. `namespace_imports()` needs the filename of the linted file, check for a `DESCRIPTION` and extract imports from other packages based on that information.
I'll create a follow-up issue from this.

AshesITR (approved):
Thank you :)

AshesITR:
R-devel apparently changed the error message for the lint we use in `test-error.R`.
I'll create a PR to allow the new error message ("unexpected end of line") as well as the old one ("unexpected input")

AshesITR:
Blocked by #659

MichaelChirico:
The coverage miss here is caused by #642 -- I touched a region of code that's not reachable currently. Merging anyway, that should be tested as part of the fix to that issue.

PR #640, opened 2020-11-30
enable commend_code_linter and apply it

### LINKED ISSUE #595: Make R/ conform to `commented_code_linter`
Part of #584 

- [ ] enable `commented_code_linter` in `.lintr`
- [ ] fix all lints

MichaelChirico:
Closes #595 

in utils.R it's "code-code" whereas in get_source_expressions.R it is only incidentally code (treating `->` as right assignment).

In both cases I used `-->` (which doesn't parse as code), not sure if this is ideal for the utils.R case.


AshesITR (approved):
Thanks!

PR #638, opened 2020-11-30
spaces_inside_linter ignores trailing comments

MichaelChirico:
Closes #636

PR #645, opened 2020-11-30
Spaces inside dummy

MichaelChirico:
same as #638, just hoping to skip triggering travis


MichaelChirico:
travis still showing up :face_with_head_bandage:

PR #650, opened 2020-12-01
Coverage for expect_lint_free, remove unused code

MichaelChirico:


AshesITR (approved):
LGTM

MichaelChirico:
@AshesITR just a heads up that Approvals seem to be a bit of a wild goose chase :smile: 

Whenever one PR would be approved & merged, it would invalidate all other Approvals (since all other PRs now have to be updated to `master`, which does the invalidation). A bit of a pain that this is the case when there are no merge conflicts. Unless I'm missing something?

AshesITR:
Actually I think the approvals stay if there is a clean merge.
But we often touch the same line in `NEWS.md` ;)

MichaelChirico:
I think there's something we can add in `.gitattributes` for `NEWS`.... maybe `NEWS.md merge=union`? IIUC `union` does what we want (the order of the bullets after merge doesn't really matter)

I can't find a clean answer without testing but I found:

 - https://stackoverflow.com/questions/5465122/gitattributes-individual-merge-strategy-for-a-file
 - section on three-way merge https://www.git-scm.com/docs/gitattributes

AshesITR:
@MichaelChirico Me clicking on the update branch button kept the approval status intact.

MichaelChirico:
ah that makes _some_ sense at least. ok I'll leave the update & merge to you then if you're the approver

PR #659, opened 2020-12-04
permit new error message (R-devel)

AshesITR:
Currently, CI builds on r-devel are failing because the parser error message for input `\` changed from "unexpected input" to "unexpected end of line".

### DIFF
--- a/tests/testthat/test-error.R
+++ b/tests/testthat/test-error.R
@@ -39,10 +39,25 @@ test_that("returns the correct linting", {
     equals_na_linter
   )
 
+  # Error message changed in R-devel as of 2020/12
+  old_lang <- Sys.getenv("LANG")
+  Sys.setenv(LANG = "en")
+  on.exit(Sys.setenv(LANG = old_lang), add = TRUE)
+  expected_message <- tryCatch(parse(text = "\\"), error = function(e) {
+    rex::re_substitutes(
+      data = conditionMessage(e),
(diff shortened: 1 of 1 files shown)


MichaelChirico:
more robust would be to generate the error minimally with base R code only

tests/testthat/test-error.R
```diff
@@ -39,10 +39,25 @@ test_that("returns the correct linting", {
     equals_na_linter
   )
 
+  # Error message changed in R-devel as of 2020/12
+  old_lang <- Sys.getenv("LANG")
+  Sys.setenv(LANG = "en")
+  on.exit(Sys.setenv(LANG = old_lang), add = TRUE)
```
MichaelChirico, line 45: another step further :)

tests/testthat/test-error.R
```diff
@@ -39,10 +39,25 @@ test_that("returns the correct linting", {
     equals_na_linter
   )
 
+  # Error message changed in R-devel as of 2020/12
+  old_lang <- Sys.getenv("LANG")
+  Sys.setenv(LANG = "en")
```
MichaelChirico, line 44: I believe the variable is `LANGUAGE` btw

AshesITR: Thanks, had a brain fart :D
Tests were failing locally with LANG.

MichaelChirico (approved):

Issue #62, opened 2015-03-08
Exclude only a specific linter from a line


jimhester:
E.g. https://github.com/bbatsov/rubocop/blob/master/README.md#disabling-cops-within-source-code

kenahoo:
This would indeed be a nice enhancement - besides making exclusions more targeted, it also serves as documentation for the reason a line is being `# nolint`ed.

PR #660, opened 2020-12-04
Allow turning off specific linters per line

### LINKED ISSUE #605: Feature: allow turning off specific linters per-line
It might be useful to disable only a specific linter on a per-line basis, e.g.

```r
`%||%` = function(x, y) {
  if (is.null(x)) y else x
}
```

produces two default lints: `object_name_linter` and `assignment_linter`.
One may actively decide that `%||%` is an acceptable name here and not want to generally disable `object_name_linter`.

Currently, this can only be achieved by # nolint'ing the entire line
```r
`%||%` = function(x, y) { # nolint
  if (is.null(x)) y else x
}
```
but unfortunately this also kills the `assignment_linter`

It would be very useful to have a way to only exclude some linters on a line (or in a block).
A syntax might be a space-separated list of linters to be disabled, e.g.

```r
`%||%` = function(x, y) { # nolint object_name_linter
  if (is.null(x)) y else x
}
```

I realize this would require work in the way exclusions are specified. Especially the line-based specification in `.lintr` files or the `exclusions` argument requires thought.

One solution would be one more level of (optional) nesting

```r
exclude: list(
    "my_file.R" = list(
      1,
      "2" = "object_name_linter",
      "4" = c("line_length_linter", "commented_code_linter")
    )
  )
```

AshesITR:
Fixes #605

- [x] implement exclusions in `lint()` and `normalize_exclusions()`
- [x] implement exclusions in `parse_exclusions()`
- [x] re-write old tests to new structure
- [x] add new tests
- [x] update internal documentation
- [x] update documentation for `exclude` argument
- [x] update documentation for comment based exclusions
- [x] add tests for interleaved exclusions
- [x] add vignette explaining in detail how to use exclusions


AshesITR:
This is going to be a fairly large changeset.
I'd implement the linter exclusion stlye last suggested (and partly introduced into lintr itself already by @MichaelChirico)

That would mean the defaults work by appending `:linter, linter2.` to either of `# nolint` or `# nolint start` in order to exclude the named linters `linter` and `linter2` for the marked line(s).

cc @jimhester @russHyde 
Do you approve of this feature or have any feedback?

russHyde:
Sorry so the syntax would be:
```
abc = 123 # nolint: assignment_linter

# nolint start: object_name_linter
MyClass <- function(x, ...) {
  # blah
}
# nolint end
```

Seems ok, is it obvious that the specified linters are being dropped?;
I note that pylint uses syntax of the form`some_code # pylint: disable=some_linter`

MichaelChirico:
What does the syntax look like for disabling optioned linters?

Mainly I have in mind something like

```
x <<- y # nolint: disable=undesirable_operator_linter(<<-) [??]
```

AshesITR:
I'd actually always use the (unique) name that the user gives in `.lintr`.
For optioned linters that should also be the linter name by default.

@russHyde: almost. Note we added a full stop after the list, so it would be `# nolint: assignment_linter.`
I also took a look at what pylint does, but I feel this adds redundance to the comment, as we already say **no**lint.

AshesITR:
Here are some sample codes for easier side-by-side comparison of the two suggestions:

```r
X = 42 # nolint: disable=assignment_linter,object_name_linter This is an explanation for the nolint

# nolint start: object_name_linter. This is an explanation for the nolint
MyClass <- function(x, ...) {
  # blah
}
# nolint end
```

```r
X = 42 # nolint: assignment_linter, object_name_linter. This is an explanation for the nolint

# nolint start: disable=object_name_linter This is an explanation for the nolint
MyClass <- function(x, ...) {
  # blah
}
# nolint end
```

AshesITR:
I think we need to solve #664 for usability with cusom linter settings. @MichaelChirico

AshesITR:
I've added two new regexes to settings, defaulting to the `: linter1, linter2,` style.
They are called `exclude_linter` and `exclude_linter_sep`. I'm not really fond of the names, but can't find any better names at the moment. Suggestions are welcome.

The implementation part of this PR should be mostly done with the last commit.
Missing lots and lots of tests for the new behaviour and documentation.

Review on the implementation is welcome.

R/lint.R
```diff
@@ -187,7 +187,7 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = NULL,
     seq_len(length(files)),
     function(i) {
       file <- files[i]
-      file %in% names(exclusions) && exclusions[[file]] == Inf
+      file %in% names(exclusions) && is_excluded_file(exclusions[[file]])
```
AshesITR, line 190: Any idea why that happens?
The function is defined in `R/exclude.R` which even comes first in the collate order...

russHyde:
So with the proposed syntax, there'd be no overlaps between lint-excluded blocks?

Would this be an error:
```
# nolint start: some_linter.
blah
# nolint start: some_other_linter
blah
# nolint end
```

And could we turn back on specific linters:
```
# nolint start: some_linter, some_other_linter.
blah
# nolint end: some_linter
# -- some_other_linter is still active?
```

R/exclude.R
```diff
@@ -25,36 +29,54 @@ exclude <- function(lints, exclusions = settings$exclusions, ...) {
...
   if (any(to_exclude)) {
     lints <- lints[!to_exclude]
   }
 
   lints
 }
+
+is_excluded <- function(line_number, linter, file_exclusion) {
+  excluded_lines <- unlist(file_exclusion[names2(file_exclusion) %in% c("", linter)])
+  Inf %in% excluded_lines || line_number %in% excluded_lines
```
MichaelChirico, line 51: I would use `any(is.infinite(excluded_lines))`

R/lint.R
MichaelChirico: This looks like an error from the linter. Maybe a bug in the linter.

MichaelChirico:
> So with the proposed syntax, there'd be no overlaps between lint-excluded blocks?

I think we should choose whatever will have the simplest implementation for now. I'm not sure how many users would want these interweaved exclusions. Though, we should definitely include test cases around this type of edge case.

I'll hold off on reviewing the code itself until we are settled on the API... Looks OK to me. Delimiting with `:` ... `.` sounds a bit strange on paper, it looks fine in practice.

R/exclude.R
AshesITR: The docs only permit `Inf` as a sentinel. Do we want to (silently) allow `-Inf` as well?

MichaelChirico: I see... `any(excluded_lines == Inf)` doesn't look much better and is probably inefficient. Fine as is.

AshesITR:
> 
> 
> So with the proposed syntax, there'd be no overlaps between lint-excluded blocks?
> 
> Would this be an error:
> 
> ```
> # nolint start: some_linter.
> blah
> # nolint start: some_other_linter
> blah
> # nolint end
> ```

Yes, it currently would be an error.
The implementation doesn't require `# nolint end` to repeat linter the scoping and instead matches start to end in a 1:1 fashion (as is currently the case as well).

To obtain the intended behaviour of the snippet, you would have to use two ranges and specify the linter scopes in each:

```
# nolint start: some_linter.
blah
# nolint end
# nolint start: some_linter, some_other_linter.
blah
# nolint end
```

> And could we turn back on specific linters:
> 
> ```
> # nolint start: some_linter, some_other_linter.
> blah
> # nolint end: some_linter
> # -- some_other_linter is still active?
> ```

Same for this, you'd have to start a new scope.

Apart from implementational simplicity it feels more error-prone if we supported "mixed modes".
What if the first section was actually intended to be


```
# nolint start: some_linter.
blah
# nolint end
# nolint start: some_other_linter.
blah
# nolint end
```

and the user accidentally deleted a `# nolint end` line.

MichaelChirico:
Noting here the push to document this new option well, perhaps with a vignette, following up https://github.com/jimhester/lintr/pull/673#discussion_r536904800

AshesITR:
@MichaelChirico WDYT about the new vignette?
Where should we link to it and how?

NEWS.md
```diff
@@ -30,7 +35,13 @@
   excluding matches found in comments; #441 and #545, @russHyde)
 * Lints are now marked with the name of the `linter` that caused them instead of the name of their implementation
   function (#664, #673, @AshesITR).
+* New syntax to exclude only selected linters from linting lines or passages. Use `# nolint: linter_name, linter2_name.`
+  or `# nolint start: linter_name, linter2_name.` in source files or named lists of line numbers in `.lintr`.
+  (#660, @AshesITR)
 * Fixed `spaces_left_parentheses_linter` sporadically causing warnings (#654, #674, @AshesITR)
+* New syntax to exclude only selected linters from linting lines or passages. Use `# nolint: linter_name, linter2_name.`
```
MichaelChirico, line 42: some merge issue for duplicate items?

tests/testthat/test-parse_exclusions.R
```diff
@@ -0,0 +1,179 @@
...
+    c("this",
+      "is #TeSt_NoLiNt: my_linter.",
+      "another",
+      "useful #TeSt_NoLiNt: my_linter.",
+      "test",
+      "testing #TeSt_NoLiNt: my_linter2."), t2)
+  expect_equal(parse_exclusions(t2), list(my_linter = c(2, 4), my_linter2 = 6))
+})
+
+test_that("it supports multiple linter exclusions", {
```
MichaelChirico, line 72: CMIIW but do we need a test for

```
# nolint some_linter_that_doesn't_exist.
```

(which I guess should be a `warning`)

R/exclude.R
```diff
@@ -98,17 +178,21 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
 
   # no named parameters at all
   if (is.null(names(x))) {
-    x <- structure(relist(rep(Inf, length(x)), x), names = x)
+    # x is a character vector of file names
+    # Normalize to list(<filename> = list(Inf), ...)
+    x <- structure(rep(list(Inf), length(x)), names = x)
```
MichaelChirico, line 183: I think `setNames(rep(.), x)` looks nicer & IINM is faster

tests/testthat/test-parse_exclusions.R
```diff
@@ -0,0 +1,179 @@
...
+    c("this #TeSt_NoLiNt_StArT",
+      "is",
+      "a #TeSt_NoLiNt_EnD",
+      "test"), t1)
+  expect_equal(parse_exclusions(t1), list(c(1, 2, 3)))
+
+  t2 <- tempfile()
+  on.exit(unlink(t2))
+  writeLines(
+    c("this #TeSt_NoLiNt_StArT",
```
MichaelChirico, line 130: does `trim_some` work here too?

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Using lintr}
+  %\VignetteEngine{knitr::rmarkdown}
+  \usepackage[utf8]{inputenc}
+---
+
+This vignette describes how to set up and configure `lintr` for use with projects or packages.
+
+## Running `lintr` on a project
```
MichaelChirico, line 14: maybe a better title for this section: "Running `lintr` at different granularities" or "Different levels of running `lintr` on a project"

MichaelChirico: Also it's probably useful to emphasize the sequence -- `lint_dir` calls `lint`, and `lint_package` calls `lint_dir` internally/

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+
+For R projects or R packages the most convenient method of configuration is creating a `.lintr` file in the project
+root. This file is in debian control format (`?read.dcf`) and each key is evaluated as R code by `lintr` when reading
+the settings.
+
+The settings file is searched for in the following order:
+
+ 1. If `options(lintr.linter_file)` is an absolute path, this file will be used.
+    The default for this option is `".lintr"`.
+ 2. If there is a file named like `lintr.linter_file` in the currently searched directory, this file will be used.
```
MichaelChirico, line 64: What's "the currently searched directory"? Is this just `getwd()`?

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+ 3. Parent directories of the currently searched directory are searched for the linter file. If one is found, this file
```
MichaelChirico, line 65: Better wording? "The first one found, if any, will be used"

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+    will be used.
+ 4. If a linter file exists in the users `HOME` directory, this file is used.
+
+If none of the steps yield an existing linter file, no configuration is loaded from the file.
```
MichaelChirico, line 69: "... and the default linters are applied (see `?linters`)" ?

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+
+### Using `options()`
+
+Settings from the `.lintr` file can be overridden using `options`. An option `option_name` is read from
```
MichaelChirico, line 73: This sentence doesn't fully make sense... `option_name` is the key in the `.lintr` file? or the option name (`getOption(option_name)`)? I believe it's the former, I think the phrasing could be clarified...

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,311 @@
...
+defaults_table <- data.frame(
+  row.names = names(default_linters),
+  settings = vapply(names(default_linters), make_setting_string, character(1L)),
+  stringsAsFactors = FALSE
+)
+
+knitr::kable(defaults_table)
+```
+
+## Exclusions
```
MichaelChirico, line 167: It might make sense to start narrowly (simple `# nolint`), then broaden to exclusion ranges (simple `# nolint start`), then broaden (in terms of code complexity) to specific exclusions (both `# nolint: <list>.` and `# nolint start: <list>.`). WDYT?

MichaelChirico:
IINM covr results are good, please be sure to add tests for the missed lines

vignettes/using_lintr.Rmd
AshesITR: What I mean is `option_name: xxx` in `.lintr` is sort of equivalent to `options(lintr.option_name = xxx)`.

vignettes/using_lintr.Rmd
AshesITR: Nope, it's the directory of the file passed to `lint()`.
This means in `lint_packages()` you could potentially hit different `.lintr` files for different directories.

tests/testthat/test-parse_exclusions.R
AshesITR: Yes, can rewrite (need to keep `writeLines()` though since `parse_exclusions` doesn't support inline data)

R/exclude.R
AshesITR: TIL `setNames` is pretty fast

```r
  structure(rep(list(Inf), length(letters)), names = letters),
  setNames(rep(list(Inf), length(letters)), letters)
)
## Unit: microseconds
##                                                         expr min  lq  mean median  uq  max neval
##  structure(rep(list(Inf), length(letters)), names = letters) 3.8 4.0 5.977   4.15 4.4 55.5   100
##           setNames(rep(list(Inf), length(letters)), letters) 1.3 1.4 1.832   1.60 1.7 22.1   100
```

tests/testthat/test-parse_exclusions.R
AshesITR: That sounds useful, but how would we check it?
I don't think we can access `names(linters)` from within `parse_exclusions`...

Maybe add as an argument to `exclude` and check there?

NEWS.md
AshesITR: Looks like it.

tests/testthat/test-parse_exclusions.R
MichaelChirico: I see... for initial implementation then, let's leave it silently ignored for simplicity. One thing at a time :smile:

R/exclude.R
MichaelChirico: I think rather, IME, `structure` is pretty slow

tests/testthat/test-parse_exclusions.R
MichaelChirico: keeping `writeLines` is fine :+1:

vignettes/using_lintr.Rmd
MichaelChirico: Oh, of course. I would add that as a parenthetical. Maybe either "in the currently searched directory (i.e, the directory of the file passed to `lint()`)" or "in `dir`, where `dir` is `dirname()` of the file passed to `lint()`" and use `dir` thereafter for conciseness.

vignettes/using_lintr.Rmd
MichaelChirico: How about

> Values in `options()` take precedence over those in `.lintr`. Note that the key `option_name` in the `.lintr` DCF translates to an R option `lintr.option_name`. If that... file. For example, `options(lintr.exclude = "# skip lint")` it will take precedence over `exclude: # nolint` in the `.lintr` file.

vignettes/using_lintr.Rmd
AshesITR: I think we should actually promote narrow exclusions as best practice. The less you `# nolint` the higher the chance of catching quality issues...

russHyde:
Crossref: #62

vignettes/using_lintr.Rmd
MichaelChirico: I meant more from the logical flow of the vignette -- it'll be easier to introduce one small concept at a time, no?

We can add a blurb to the top of this section mentioning the preferred approach, then build up to it.

AshesITR:
Closes #62

Thanks for finding it, @russHyde

vignettes/using_lintr.Rmd
AshesITR: Linked to the ### Defaults section instead, since that contains all default settings and not only the default linters.

vignettes/using_lintr.Rmd
AshesITR: Okay, I'll try that.

AshesITR:
@MichaelChirico Happy new year. Have you had a chance to look at the changes yet?

R/exclude.R
```diff
@@ -65,27 +87,85 @@ parse_exclusions <- function(file, exclude = settings$exclude,
...
+        excluded_linters <- strsplit(linters_string, exclude_linter_sep)[[1L]]
+        exclusions <- add_excluded_lines(exclusions, excluded_lines, excluded_linters)
+      }
     }
   }
 
-  sort(unique(c(exclusions, which(rex::re_matches(lines, exclude)))))
+  nolint_locations <- rex::re_matches(lines, exclude, locations = TRUE)[, "end"] + 1L
+  nolints <- which(!is.na(nolint_locations))
+  # Disregard nolint tags if they als match nolint start / end
```
MichaelChirico, line 107: typo here... `all` or `also`?

R/exclude.R
```diff
@@ -65,27 +87,85 @@ parse_exclusions <- function(file, exclude = settings$exclude,
...
+  nolints <- setdiff(nolints, c(starts, ends))
+
+  for (i in seq_along(nolints)) {
+    linters_string <- substring(lines[nolints[i]], nolint_locations[nolints[i]])
+    linters_string <- rex::re_matches(linters_string, exclude_linter)[, 1L]
+
+    # No match for linter list: Add to global excludes
+    if (is.na(linters_string)) {
```
MichaelChirico, line 115: this logic is repeated for the range & one-line case, could you put it to a helper function?

MichaelChirico: i guess this is a place where python's unpack-and-assign would come in handy for writing the helper...

R/exclude.R
```diff
@@ -98,26 +178,69 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
...
       if (any(bad)) {
         stop("Full file exclusions must be character vectors of length 1. items: ",
-             paste(collapse = ", ", which(bad)),
+             toString(which(bad)),
              " are not!",
              call. = FALSE)
       }
       names(x)[unnamed] <- x[unnamed]
-      x[unnamed] <- Inf
+      x[unnamed] <- rep(list(list(Inf)), sum(unnamed))
```
MichaelChirico, line 218: any reason to use `rep_len` below but not here?

R/zzz.R
```diff
@@ -193,6 +193,13 @@ settings <- NULL
     exclude = rex::rex("#", any_spaces, "nolint"),
     exclude_start = rex::rex("#", any_spaces, "nolint start"),
     exclude_end = rex::rex("#", any_spaces, "nolint end"),
+    exclude_linter = rex::rex(start, any_spaces, ":", any_spaces,
+                              capture(
+                                name = "linters",
+                                zero_or_more(one_or_more(none_of(",.")), any_spaces, ",", any_spaces),
```
MichaelChirico, line 199: note that this means downstream custom linters can't have `.` in their name

MichaelChirico: I have this regex lying around btw for a valid R object symbol:

```
[a-zA-Z][a-zA-Z0-9._]*|[.]|[.][a-zA-Z._][a-zA-Z0-9._]*
```

vignettes/using_lintr.Rmd
```diff
@@ -0,0 +1,320 @@
...
+    deepest path, moving upwards one level at a time.
+ 4. A linter file in the users `HOME` directory.
+
+If no linter file is found, only default settings take effect (see [defaults](#defaults)).
+
+### Using `options()`
+
+Values in `options()`, if they are not `NULL`, take precedence over those in the linter file (e.g. `.lintr`).
+Note that the key `option_name` in the linter file translates to an R option `lintr.option_name`.
+For example, `options(lintr.exclude = "# skip lint")` will take precedence over `exclude: # nolint` in the linter file.
```
MichaelChirico, line 77: Does timing matter for this? e.g. if I run

```
options(lintr.exclude = "# skip lint")
library(lintr)
```

is it any different from the reverse?

MichaelChirico:
@AshesITR sorry for the delay -- have been moving across the country since early December, finally getting settles with a good enough setup to do a review.

Some higher-level thoughts:

 1. Another good follow-up would be the option to _force_ exclusions to use a list, i.e. generic `# nolint` is not allowed. Specifying the linter makes "lint creep" far less likely, and makes it easier to search out all instances of a specific lint to boot.
 2. The `: <linters>.` syntax looks a bit odd for `nolint: start: <linters>.` because the `:` is repeated. I still don't have a good alternative I'm happy with thought so not making it a blocker...
 3. Somewhat related, some people might like to namespace-tag the linter exclusions, e.g. I keep my custom lints in a package `myLinter` and it makes it clearer to readers to see `# nolint: lintr::seq_linter, myLinter::power_linter`. Probably another follow-up issue to file.
 4. Another power feature I think was mentioned at some point is non-overlapping range exclusions, e.g. `1:# nolint start: linterA` ... `5:# nolint start: linterB` `6: ...` `7: # nolint end: linterA` ... `12: # nolint end: linterB`. I think the logic is partway there to supporting this, but it's not necessary to get this merged.

MichaelChirico (dismissed):

AshesITR:
Thanks for you review, @MichaelChirico.
The GHA failure is due to the regex-based parsing of lint exclusions, i've tried fixing it by tricking with the \code command.


re 2.: `# nolint: start` doesn't work by default, one might simply change the `exclude_linter` setting to a more pleasant regex as well.

The default is

    ^[[:space:]]*:[[:space:]]*(?<linters>(?:(?:[^,.])+[[:space:]]*,[[:space:]]*)*(?:[^,.])+)\.

What might work here (accepting `# nolint: start (bla_linter)`, `# nolint (bla_linter)`) might be using `(...)` instead of `:.. .`

    ^[[:space:]]*\([[:space:]]*(?<linters>(?:(?:[^,.])+[[:space:]]*,[[:space:]]*)*(?:[^,.])+)\)

vignettes/using_lintr.Rmd
AshesITR: It doesn't. All settings are read within `read_settings()` when calling a linting function.

R/zzz.R
AshesITR: That complicates the `: ... .` syntax.
I'll open up a discussion whether we should still change the default here while it's hot.

I think using `(` and `)` as opener and closer respectively should turn out okay, but I'd like to make this change in a new PR, correcting all our internal `# nolints` for this as well.

WDYT?

R/exclude.R
AshesITR: yes, that should say "also"

R/exclude.R
AshesITR: Just an oversight.

R/exclude.R
AshesITR: Done.

MichaelChirico:
> Unless we would require # nolint end to be specific as well

yes that's what i had in mind

MichaelChirico (approved):
I am happy to merge at this point. @russHyde and/or @jimhester it would be good to get your high-level input as well so we have a bit more consensus on the high-level "API" for this feature before release.

AshesITR:
Let's merge for now and discuss the high-level API w.r.t. support for scoping and out-of-the-box support for `.` in linter names.

bersbersbers:
I have spent some 20 minutes trying to figure out why
```
# nolint start: open_curly_linter
```
would also suppress other linters. In my eyes, the need to have a final dot *even when a comment is not following* makes it way too easy to do this wrong - especially since `# nolint start: open_curly_linter` *does* suppress the warning that you want to suppress. If you work from a baseline of few warnings, you will not even notice that you are suppressing all other future linters.

Something like https://github.com/jimhester/lintr/issues/690 is definitely required if the dot requirement is not removed.

PR #673, opened 2020-12-04
use linter names in lint() results

### LINKED ISSUE #664: linter names are hard-coded
Actual:
```r
as.data.frame(lintr::lint("A = 42\n", linters = list(lala = lintr::assignment_linter)))
>  filename line_number column_number  type                        message   line
>1   <text>           1             3 style Use <-, not =, for assignment. A = 42
>             linter
>1 assignment_linter
```

Expected:
```r
as.data.frame(lintr::lint("A = 42\n", linters = list(lala = lintr::assignment_linter)))
>  filename line_number column_number  type                        message   line
>1   <text>           1             3 style Use <-, not =, for assignment. A = 42
>             linter
>1 lala
```

AshesITR:
fixes #664

- [x] implement
- [x] add tests
- [x] document


AshesITR:
Already added a test for #660 in combination with custom linter names.

R/lint.R
```diff
@@ -81,6 +81,13 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
       else {
         expr_lints <- flatten_lints(linters[[linter]](expr)) # nolint
 
+        if (length(expr_lints)) {
+          expr_lints[] <- lapply(expr_lints, function(lint) {
```
MichaelChirico, line 85: how about moving this in to `flatten_list` instead?

tests/testthat/test-lint_file.R
```diff
@@ -109,3 +109,16 @@ test_that("lint() results do not depend on the position of the .lintr", {
...
   )
 })
+
+test_that("lint uses linter names", {
+  expect_lint("a = 2", list(linter = "bla"), linters = list(bla = assignment_linter), parse_settings = FALSE)
+})
+
+test_that("exclusions work with custom linter names", {
+  expect_lint(
+    "a = 2 # nolint: bla.",
```
MichaelChirico, line 119: If we're going to allow this, I hope it will be pretty clearly documented -- I can imagine it being a bit hard to track down what's happening as an outside contributor to a package that takes advantage of this...

AshesITR: I agree we should document all of #660 extremely clearly.
For configurable linters which might be used with more than one set of options (e.g. `undesirable_functions_linter`) I can't really see any alternative way that would allow a nicely formatted exclude tag.

Note that the auto names given by `with_defaults` are identical to the linter names, so most of the time this feature won't matter much.

R/lint.R
AshesITR: Thought about this too, but all other uses of `flatten_lints` / `flatten_list` don't have a linter name as context, so we'd need a conditional there which will only be fulfilled in this single call from `lint()`.

tests/testthat/test-lint_file.R
MichaelChirico: OK that makes sense. This is a good example to put in the docs/NEWS.

R/lint.R
MichaelChirico: Got it

AshesITR:
Hmm sadly GitHub hat a merge conflict in `NEWS.md` again...
I triple-checked https://www.git-scm.com/docs/gitattributes. Seems it's simply not supported.

NEWS.md
```diff
@@ -28,6 +28,8 @@
 * New `backport_linter()` for detecting mismatched R version dependencies (#506, @MichaelChirico)
 * `paren_brace_linter` and `no_tab_linter` also use more reliable matching (e.g.,
   excluding matches found in comments; #441 and #545, @russHyde)
+* Lints are now marked with the name of the `linter` that caused them instead of the name of their implementation
+  function (#664, #673, @AshesITR).
```
MichaelChirico, line 32: I think adding some more context (maybe an extra sentence) in the NEWS will be helpful -- I guess most existing users would learn about this by checking the NEWS on release, so a quick motivating example here would be good.

R/lint.R
MichaelChirico, line 85: Is there any easy way to only do this for the linters that are named differently than "default"?

tests/testthat/test-lint_file.R
```diff
@@ -109,3 +109,16 @@ test_that("lint() results do not depend on the position of the .lintr", {
     )
   )
 })
+
+test_that("lint uses linter names", {
+  expect_lint("a = 2", list(linter = "bla"), linters = list(bla = assignment_linter), parse_settings = FALSE)
```
MichaelChirico, line 114: either here in the test or in the examples (I guess that's waiting on #660 though?), it would be good to present an example like you mentioned, e.g. we could use `linters = list(bla = object_name_linter("SNAKE_CASE"))`

R/lint.R
AshesITR: Not that I'm aware of. However the linters list is auto-named so we're not breaking anything either.

tests/testthat/test-lint_file.R
AshesITR: The examples of ?exclude sound like a good place for this. Maybe even a vignette? Waiting for #660 since that's the bigger part of the feature.

R/lint.R
MichaelChirico: OK. I meant more as a way to avoid overwriting the whole list. Not crucial since the list is small.

AshesITR: Could add a check on linter != expr_lints[[1L]]$linter. WDYT?

MichaelChirico: ehhh let's leave it as is for now. That doesn't really buy us much.

NEWS.md
AshesITR: WDYT about making a bigger news all around the combination of this with #660?

The more I think about it the more I think it's worth a vignette...

MichaelChirico: Agreed. I'll write a note there.

MichaelChirico (approved):

PR #674, opened 2020-12-04
Fix object length warnings in spaces_left_parentheses_linter

### LINKED ISSUE #654: Warnings from cyclocomp_linter
As part of #653, I ran `lintr::lint_package()`:

```
lintr::lint_package()
```

A ton of warnings were thrown:

```
#Warning messages:
#1: In parent == parent[before_operator_idx] :
#   longer object length is not a multiple of shorter object length
# ...
#653 50: In parent == parent[before_operator_idx] :
#  longer object length is not a multiple of shorter object length
```

At a glance, I suspect that `cyclocomp_linter` is getting tripped up by the `with()` expression:

```
with(
  source_file$parsed_content,
  token == "expr" & parent == parent[before_operator_idx]
)
```

But needs more investigation

AshesITR:
fixes #654

<s>not 100% sure how to trigger the bug, so I didn't add a regression test yet.</s>
`inst/examples/complexity.R` triggers the warning on current master.


tests/testthat/test-spaces_left_parentheses_linter.R
```diff
@@ -77,3 +77,11 @@ test_that("returns the correct linting", {
...
   expect_lint("a <- -(b)", NULL, spaces_left_parentheses_linter)
 })
+
+test_that("doesn't produce a warning", {
+  expect_silent(
+    lint(
+      system.file("example/complexity.R", package = "lintr")
+    )
+  )
+})
```
MichaelChirico, line 87: trailing newline

(should we add that as a default linter btw?)

tests/testthat/test-spaces_left_parentheses_linter.R
```diff
@@ -77,3 +77,11 @@ test_that("returns the correct linting", {
 
   expect_lint("a <- -(b)", NULL, spaces_left_parentheses_linter)
 })
+
+test_that("doesn't produce a warning", {
+  expect_silent(
+    lint(
+      system.file("example/complexity.R", package = "lintr")
```
MichaelChirico, line 84: the test as written is not very clear, I can imagine trying to revisit this test on 2 years and have no idea what it's doing.

could you elaborate the comment and/or extract a more essential example from that script for clarity?

AshesITR: I'll try digging deeper for the cause.

tests/testthat/test-spaces_left_parentheses_linter.R
AshesITR: It's already part of the default linters?
cf. `zzz.R#L95`

tests/testthat/test-spaces_left_parentheses_linter.R
AshesITR: tmp

It appears to happen in nested `if` statements if there is a unary minus in the conditional of an inner `if`.
This means a repro is basically the same as the `complexity.R` function.

(Look at the line numbers from parsed content identifying siblings)

I'll add a comment what part of `complexity.R` triggers the bug.

tests/testthat/test-spaces_left_parentheses_linter.R
MichaelChirico: Sorry, i meant a linter for the terminal newline, which I just realized, isn't implemented yet! proceed...

tests/testthat/test-spaces_left_parentheses_linter.R
MichaelChirico: great, works for me

MichaelChirico (approved):

PR #663, opened 2020-12-04
temporarily set LANGUAGE in get_source_expressions

AshesITR:
fixes #661

Issue #670, opened 2020-12-04
object_name_linter default styles regression


AshesITR:
https://github.com/jimhester/lintr/blob/289e7935533527849bf1b5f6f1886bec4d65fadc/R/zzz.R#L87

The new default style "symbols" needs to be added here for it to take effect.
Original issue: #615

PR #676, opened 2020-12-05
simplify temporary language settings

MichaelChirico:
Originally wrote this because I noticed another place where we relied on English locale:

https://github.com/jimhester/lintr/blob/9104a004d50c66d3d77cc4e0d4a2c83067f8b43a/R/get_source_expressions.R#L89

But that's covered by #663 already.

Anyway, I think this approach looks nicer. Considered an approach like `with_lang("en", {...})` but the expressions within `{...}` would be huge, don't think it would actually be better.

Also note that before, we weren't actually restoring `LANGUAGE` in the event that it was currently unset, e.g.

```
ORIG = Sys.getenv("LANGUAGE", unset=NA)

Sys.unsetenv("LANGUAGE")

old = Sys.getenv("LANGUAGE")
# do stuff
Sys.setenv(LANGUAGE = old)

# initially, LANGUAGE was unset, now it's set equal to nothing
Sys.getenv("LANGUAGE", unset=NA)
# [1] ""

# true reset. Sys.setenv(LANGUAGE = NA) actually sets "NA"
if (is.na(ORIG)) Sys.unsetenv("LANGUAGE") else Sys.setenv(LANGUAGE = ORIG)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,6 +45,7 @@ Collate:
     'addins.R'
     'assignment_linter.R'
     'assignment_spaces_linter.R'
+    'backport_linter.R'
     'cache.R'
     'closed_curly_linter.R'
     'commas_linter.R'
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,8 @@
 * `object_name_linter()` now excludes special R hook functions such as `.onLoad` (#500, #614, @AshesITR and @michaelchirico)
 * `equals_na_linter()` now lints `x != NA` and `NA == x`, and skips usages in comments (#545, @michaelchirico)
 * Malformed Rmd files now cause a lint instead of an error (#571, #575, @AshesITR)
-* `object_name_linter()` gains a new default style, `"symbols"`, which won't lint all-symbol object names (in particular, that means operator names like `%+%` are skipped; #615, @michaelchirico)
+* `object_name_linter()` gains a new default style, `"symbols"`, which won't lint all-symbol object names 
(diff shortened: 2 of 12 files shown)


AshesITR:
Are you sure the current behaviour isn't resetting env?

```
"LANGUAGE" %in% names(Sys.getenv())
> [1] FALSE
Sys.setenv(LANGUAGE = "en")
"LANGUAGE" %in% names(Sys.getenv())
> [1] TRUE
Sys.setenv(LANGUAGE = "")
"LANGUAGE" %in% names(Sys.getenv())
> [1] FALSE
```

Tested on R 4.0.3 @ Windows 10.

Also, if we do make a helper, we should use it in all places (I remember `R/expect_lint.R` and `tests/testthat/test-error.R`)

MichaelChirico:
This might be a platform-dependent thing:

```
Sys.unsetenv("LANGUAGE")
"LANGUAGE" %in% names(Sys.getenv())
# [1] FALSE
Sys.setenv(LANGUAGE="")
"LANGUAGE" %in% names(Sys.getenv())
# [1] TRUE
```

AshesITR:
Okay then we should fix it in all places. Thanks for pointing it out!

AshesITR:
#670 causes spurious lints here. I'll whip up a quick fix PR.

MichaelChirico:
For `tests/testthat/test-error.R` we'd have to either extract with `:::` or re-define it, is it worth it?

AshesITR:
Using internal functions in tests seems okay to me.

MichaelChirico:
OK done

AshesITR (approved):
LGTM

PR #684, opened 2020-12-06
add semicolon_terminator_linter to default linters

### LINKED ISSUE #683: Don’t put ; at the end of a line, and don’t use ; to put multiple commands on one line.
[The tidyverse style guide](https://style.tidyverse.org/syntax.html#semicolons) specifies essentially `;` is forbidden.
Currently, we only don't lint semicolons.

Since our default linters strive to enforce tidyverse style, we should add (and rename?) the `semicolon_terminator_linter`.

NB this will cause lints in our tests/.

AshesITR:
fixes #683


R/utils.R
```diff
@@ -106,7 +106,8 @@ logical_env <- function(x) {
 rot <- function(ch, k = 13) {
   p0 <- function(...) paste(c(...), collapse = "")
   A <- c(letters, LETTERS, " '")
-  I <- seq_len(k); chartr(p0(A), p0(c(A[-I], A[I])), ch)
+  I <- seq_len(k)
+  chartr(p0(A), p0(c(A[-I], A[I])), ch)
```
AshesITR, line 110: didn't this already get de-linted somewhere?

russHyde:
Thanks for requesting my review. I'm somewhat nervous about adding additional linters to the `default_linters` since these constitute breaking changes: packages that use lintr's default_linters today may start failing tomorrow if this is merged (this happened when T/F linter was added to the defaults for example).

Nonetheless, if we set up an issue/project to define the changes we want to make before the next big release, and include "expand the `default_linters` to be consistent with tidyverse style-guidelines", I'm happy for this change to go through.

There are a number of other big changes happening to lintr at the moment, so it may make sense to document the plans for leading-in to lintr-v3. WDYT @jimhester

jimhester:
I think there are enough changes we can make the next release a compatibility breaking one and add some new default linters.

One thing we could do (separate from this PR) is provide a way to easily opt into the previous default linters.

AshesITR:
@jimhester That sounds like a good idea.
What about an option `lintr.defaults` maybe containing a version number or something?

MichaelChirico:
> What about an option `lintr.defaults` maybe containing a version number or something?

I would say, not a good idea. I think we want to maintain only one set of defaults in the long run -- our recommended defaults. The other behavior is just to help with back-compatibility in the short term. Eventually, people using it should migrate to using the recommended defaults, or adjusting their code to their own preferred defaults.

russHyde (approved):
LGTM

PR #687, opened 2020-12-06
fix delint conflicts

AshesITR:
Unfortunately, some newly enabled linters had hits in `tests`, so we aren't lint-free on master atm.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,6 +1,6 @@
 linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
    line_length_linter = line_length_linter(120),
-   cyclocomp_linter = cyclocomp_linter(29), # TODO reduce to 15
+   cyclocomp_linter = cyclocomp_linter(30), # TODO reduce to 15
    equals_na_linter = NULL, # TODO enable (#601)
    paren_brace_linter = NULL # TODO enable (#603)
  )
--- a/tests/testthat/test-error.R
+++ b/tests/testthat/test-error.R
@@ -65,6 +65,9 @@ test_that("returns the correct linting", {
   expected_message <- tryCatch(parse(text = "\\"), error = get_base_message)
   expect_lint("\\", rex(expected_message))
 
+  # restore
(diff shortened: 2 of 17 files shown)


.lintr
```diff
@@ -1,6 +1,6 @@
 linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
    line_length_linter = line_length_linter(120),
-   cyclocomp_linter = cyclocomp_linter(29), # TODO reduce to 15
+   cyclocomp_linter = cyclocomp_linter(30), # TODO reduce to 15
```
AshesITR, line 3: #673 added 1 to the cyclocomp of `lint()`.
To keep this separate from removing ~ 200 lints from tests/, I temporarily bump the limit here.

MichaelChirico (dismissed):

tests/testthat/test-error.R
```diff
@@ -60,11 +60,6 @@ test_that("returns the correct linting", {
   expected_message <- tryCatch(parse(text = "\\"), error = get_base_message)
   expect_lint("\\", rex(expected_message))
 
-  # also try when LANGUAGE initially unset
```
AshesITR, line 63: This causes non-english locales to fail the test.

AshesITR:
@MichaelChirico can you re-approve? I had a merge conflict and test failure on non-english locale due to `Sys.unsetenv()` in `test-error.R`.

MichaelChirico (approved):

MichaelChirico:
That `Sys.unsetenv` was for coverage -- do we know what's the issue exactly?

AshesITR:
What issue do you mean?
On non-english locales, Sys.unsetenv will create essentially unpredictable messages and the following tests check for these.
I think what would work would be something like

```
Sys.unsetenv("LANGUAGE")
get non-english expected error message
test in non-english locale
set_lang("en")
```

The old code was missing the `set_lang("en")` and the update of the expected error message.

MichaelChirico:
OK, we'll add that back in the future

PR #669, opened 2020-12-04
Enable tests

MichaelChirico:
Closes #668

PR #689, opened 2020-12-06
Bug/gha workflow lint

### LINKED ISSUE #688: lint workflows
I've been looking to update the `readme` CI section. I realised the install package step was missing from the `r-lib/actions` lint workflow (confirmed by this [issue and solution](https://github.com/r-lib/actions/pull/216) that were merged after I configured the GHA workflows).

This might be a bit meta, but if my understanding is correct, I might have misconfigured the GHA workflows. I think a package needs to be installed for `lintr::lint_package()` to do its thing. For the standard `.lint` GHA workflow - for any other package but {lintr} - the steps are simple: 
* install the CRAN version of {lintr}
* install the package
* lint the package

For {lintr} this is a bit trickier since we can only have one version of a package installed at any one time. Possible issues:
* The first workflow - `lint.yaml` - installs {lintr} from CRAN, which means it's linting that version of the package. 
* The second workflow - `lint-changed-files.yaml` - installs the dev version (`main`) and, as a consequence, that's the version that gets linted.  

I think both workflows should install the version in the branch subject of the pull request, namely with a step that installs the branch version of the package, something like

```
    - name: Install package
         run: R CMD INSTALL .
```

A side-effect would be that the `.lint` workflows configured for {lintr} should not be an inspiration for other package developers.

dragosmg:
closes #688

### DIFF
--- a/.github/workflows/lint-changed-files.yaml
+++ b/.github/workflows/lint-changed-files.yaml
@@ -35,13 +35,15 @@ jobs:
           install.packages(c("remotes"))
           remotes::install_deps(dependencies = TRUE)
           remotes::install_cran("gh")
-          remotes::install_github("jimhester/lintr")
           remotes::install_cran("purrr")
         shell: Rscript {0}
 
       - name: Add lintr options
         run: cat('\noptions(lintr.linter_file = ".lintr_new")\n', file = "~/.Rprofile", append = TRUE)
         shell: Rscript {0}
+
+      - name: Install package
+        run: R CMD INSTALL .
 
       - name: Extract and lint files changed by this PR
         run: |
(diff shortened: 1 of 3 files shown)


dragosmg:
@AshesITR I've made the changes suggested in #688, but this has now resulted in quite a few lints for the `lint` / `.lintr`  workflow - https://github.com/jimhester/lintr/actions/runs/404278311.

AshesITR:
Thanks for notifying. #687 will fix all of those (> 200) lints.
They were caused by #673 (increased cyclocomp of `lint()` to 30) and #669 (`tests` no longer ignored).

AshesITR (approved):
LGTM, thank you!

PR #694, opened 2020-12-06
doocument() backport_linter

AshesITR:
backport_linter was missing documentation.

- [x] added doc for argument `r_version`
- [x] run `document()`


MichaelChirico:
maybe we should add a GHA step to run document for us 😬

AshesITR:
You mean a step that would on-push run `devtools::document()` and commit all changes?

dragosmg:
@MichaelChirico @AshesITR `usethis:use_github_action_pr_commands()`? it adds [this](https://github.com/r-lib/actions/blob/master/examples/pr-commands.yaml) workflow which runs `roxygen2::oxygenise()`.

AshesITR:
That would require the author to remember to include `/document` (or `/style`) in the commit message.
Not sure how much benefit that provides compared to remember running `devtools::document()` and committing the resulting changes...

NB in the discussion around #584 we explicitly decided against using `styler` to do a bulk re-styling of the current codebase.

MichaelChirico:
> That would require the author to remember to include /document (or /style) in the commit message.

What I had in mind is the GHA runner doing it and pushing, no input needed. Probably just need to run this on push to `master`, otherwise there may be a permissions issue for getting PRs from forks.

MichaelChirico (approved):

dragosmg:
@AshesITR @MichaelChirico I was thinking along the same lines - i.e. not use {styler} and look into configuring the workflow to automatically run (without a key word). I don't mind looking into it if still needed.

AshesITR:
Sounds useful on PRs.
But we should be careful not to break builds accidentally.
`document()` on master without this pr (especially the `@param r_version` part) would have caused all builds to fail.

On the other hand, the original PR would have failed if the action was active.
That saying, we'll probably need to only be very carefull when rolling this out.

PR #698, opened 2020-12-09
fix symbols style

### LINKED ISSUE #495: False-positive object_name_linter for `%<-%`
This line

    `%<-%` <- zeallot::`%<-%`

raises

> Variable and function name style should be snake_case.

AshesITR:
fixes #495

the major issue was that `as.character()` doesn't unescape XML entities. `xml2::xml_text()` should have been used instead.


R/object_name_linters.R
```diff
@@ -49,7 +49,8 @@ object_name_linter <- function(styles = c("snake_case", "symbols")) {
 
     # Retrieve assigned name
     nms <- strip_names(
-      as.character(xml2::xml_find_first(assignments, "./text()")))
+      xml2::xml_text(xml2::xml_find_first(assignments, "./text()"))
```
AshesITR, line 52: "%<-%" was turned into "%&lt;-%" by `as.character()`

R/object_name_linters.R
```diff
@@ -283,7 +284,7 @@ loweralnum <- rex(one_of(lower, digit))
 upperalnum <- rex(one_of(upper, digit))
 
 style_regexes <- list(
-  "symbols"     = rex(start, maybe("."), zero_or_more(none_of(alnum)), end),
+  "symbols"     = rex(start, zero_or_more(none_of(alnum)), end),
```
AshesITR, line 287: The empty name can occur here because symbols like `<-` are stripped before passing on to the regex.

R/object_name_linters.R
MichaelChirico: those look the same :thinking: anyway, yes the right fix

R/object_name_linters.R
MichaelChirico: this feels somewhat hacky... but OK for now.

MichaelChirico (approved):

R/object_name_linters.R
AshesITR: Funny how GitHub unescaped the XML entity. as.character had a AND lt SEMICOLON in its output.

R/object_name_linters.R
AshesITR: The reason is the other styles should allow e.g. `snake_case<-` as a function name

R/object_name_linters.R
MichaelChirico: ah right. i saw that & was on the XML tree & was puzzled for a minute and moved on. glad we have better test cases for this now.

R/object_name_linters.R
MichaelChirico: As long as applying those rules in the same order always works, we're OK, I'm just not 100% sure of that.

PR #702, opened 2020-12-15
stricter test for has_description

MichaelChirico:
Just got bit by find_package landing on a DESCRIPTION in a parent directory which is a folder.

Changed the has_description test to be a bit stricter.


AshesITR:
Could you add a test, e.g. using a mini Project with file structure

```
DESCRIPTION # file
dir
  |- DESCRIPTION # directory
  |  \- dummy_file # to have git track the directory
  \- R
      \- some_file.R
```

And testing the output of `find_package(path = "dir/R/some_file.R")`?
The root `DESCRIPTION` could be made a proper DESCRPITION file in dcf should we decide to check even more thoroughly in the future.

R/lint.R
```diff
@@ -287,7 +287,8 @@ lint_package <- function(path = ".", relative_path = TRUE, ...,
 
 
 has_description <- function(path) {
-  file.exists(file.path(path, "DESCRIPTION"))
+  desc_info <- file.info(file.path(path , "DESCRIPTION"))
+  !is.na(desc_info$size) && desc_info$size > 0 && !desc_info$isdir
```
MichaelChirico, line 291: @AshesITR any idea why we would get this same lint 3 times?

MichaelChirico:
Sorry, I'm not following what we're trying to test exactly. We run `lint_package` on `dir` and it finds a lint in `some_file`, is that right? Don't we need `R` to be a sibling of the "correct" DESCRIPTION then?

AshesITR:
I would keep the test atomic and only test that `find_package` returns the correct (expected) package directory.
No lints expected for this test IINM.

R/lint.R
AshesITR: Those three are from three different check runs.
I'd have expected two (one for lint changed files with `.lintr_new` and one for lint package with `.lintr`)

dragosmg: Running the same lint 3 times was my doing. It should have run only twice (as @AshesITR correctly indicated - once with `.lintr` and once with `.lintr_new`). One extra run was residual from my early testing when I was forcing a workflow to run on every push (to any branch). I've now removed that since it's excessive.

MichaelChirico:
I was thinking to run `lint_package` to get at the behaviour through the "public" (exported) API, but of course we could directly test `find_package` as well. WDYT

AshesITR:
Ah I see what you're getting at. That's a good point, would also be okay. In that case the DESCRIPTION directory should have a sibling directory named R which contains sources with lints.

AshesITR:
@MichaelChirico Do you want to add a test for this so we can ship it with 3.0.0?
The change itself looks good to me.

AshesITR (approved):
LGTM

PR #703, opened 2020-12-27
Include "demo" directory in `lint_package`

dmurdoch:
Demos are typically R code, so should be linted.


AshesITR:
Thanks for the suggestion. Could you add a note to `NEWS.md`?

AshesITR (approved):

AshesITR:
Thank you 😄

PR #707, opened 2021-01-02
remove superfluous GHA lint workflow on push to any branch

### LINKED ISSUE #706: Configure GHA on push to only run on master branch
Currently when we do a PR from a branch in jimhester/lintr, both GHA on PR and on push are triggered.
Since the only "official" branch is master, we should deactivate the lint on push Action for all other branches since those would be linted by the lint on PR action with the same linters anyway.

@dragosmg is my understanding correct?

Here is an example from #702 where the same lint appears three times where I'd expect two times (one from linting with `.lintr` and one from linting with `.lintr_new`)
https://github.com/jimhester/lintr/pull/702/files/dcb2b8ffde7125bb629847d64320bf4b3cc4f215#diff-097ee63901f8ab79779c19b0edad2fe12ab0367507988e41526664eb789f4a72

dragosmg:
closes #706

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,6 @@ on:
     branches:
       - main
       - master
-      - "**"
   pull_request:
     branches:
       - main


AshesITR (approved):
Thanks for the quick fix 😄

Issue #474, opened 2020-03-30
inline_data in lint()


maelle:
I've just seen the code blocks related to `inline_data` in `lint()` and tried running

```r
filename <- "1+1\n2+2\n3+3"
lint(filename)
```

in RStudio which gave an error because the markers' file is "`<text>`"

https://github.com/jimhester/lintr/blob/edb3b12aac25d57759fb20dcd0171ba510d6242f/R/lint.R#L109-L114

* Is `lint()` for inline data actually used? (it's not documented as far as I could tell)
* If so, I suppose this is a bug report.

```r
Error in .rs.normalizePath(marker$file, mustWork = TRUE) : 
  path[1]="<text>": No such file or directory
Backtrace:
 1. (function (x, ...) ...
 2. lintr::print.lints(x)
 3. lintr::rstudio_source_markers(x) R/methods.R:57:6
 4. rstudioapi::callFun(...) R/lint.R:355:2
 5. f(...)
 6. base::lapply(...)
 7. FUN(X[[i]], ...)
 9. .rs.normalizePath(marker$file, mustWork = TRUE)
```

russHyde:
Are you using a modern version of lintr? I get the following (outside rstudio):
``
library(lintr)
> filename <- "1+1\n2+2\n3+3"
> lint(filename)
<text>:1:2: style: Put spaces around all infix operators.
1+1
~^~
<text>:2:2: style: Put spaces around all infix operators.
2+2
~^~
<text>:3:2: style: Put spaces around all infix operators.
3+3
```

**Comment by maelle** [comment]
oh sorry this is _inside_ RStudio only.

**Comment by russHyde** [comment]
Confirmed this in rstudio

**Comment by russHyde** [comment]
Just checked this again in Rstudio 1.1.456 with current master of lintr (3dbf24d5c7411471cd674e8db51a2b0b508b2aac) and could no longer confirm this bug.

```
library(devtools)
load_all() # load master {lintr}
filename <- "1+1\n2+2\n3+3"
lint(filename)
<text>:1:2: style: Put spaces around all infix operators.
1+1
~^~
<text>:2:2: style: Put spaces around all infix operators.
2+2
~^~
<text>:3:2: style: Put spaces around all infix operators.
3+3
~^~
```

**Comment by russHyde** [comment]
But, the filename argument for 'lint' does not indicate that inline data can be used

```
? lint
Description:

     Apply one or more linters to a file and return the lints found.

Usage:

[...]

PR #713, opened 2021-01-26
Docs: inline code for lint function

### LINKED ISSUE #712: Indicate that inline-data can be used in lintr::lint(filename = "...")
As noted in https://github.com/jimhester/lintr/issues/474 there is no indication in the docs for `lint` that inline-data can be used as input to the function (or how the inline data should be formatted).

russHyde:
Fixes #712 and #474

Add description of use of inline R code in the filename argument of `lint()`

### DIFF
--- a/R/lint.R
+++ b/R/lint.R
@@ -12,15 +12,27 @@ NULL
 #' Lint a file
 #'
 #' Apply one or more linters to a file and return the lints found.
+#'
 #' @name lint_file
-#' @param filename the given filename to lint.
-#' @param linters a named list of linter functions to apply see \code{\link{linters}}
-#' for a full list of default and available linters.
+#'
+#' @param filename either the filename for a file to lint, or a character
+#' string of inline R code for linting. If a newline character is detected,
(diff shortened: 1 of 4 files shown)


R/lint.R
```diff
@@ -12,15 +12,27 @@ NULL
...
 #'
 #' Apply one or more linters to a file and return the lints found.
+#'
 #' @name lint_file
-#' @param filename the given filename to lint.
-#' @param linters a named list of linter functions to apply see \code{\link{linters}}
-#' for a full list of default and available linters.
+#'
+#' @param filename either the filename for a file to lint, or a character
+#' string of inline R code for linting. If a newline character is detected,
```
MichaelChirico, line 19: nit: "if" should be "if and only if". I might reword it to something like

> The latter [inline data] applies whenever `filename` has a newline character (`\n`).

russHyde: Fixed

MichaelChirico (approved):
LGTM!

PR #717, opened 2021-01-29
docs: add linter-names to ?default_linters

### LINKED ISSUE #510: Clearly document which linters are `default_linters`

russHyde:
Fixes #510

Appends the names of the default-linters to the manpage for `default_linters`.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -69,7 +69,15 @@ with_defaults <- function(..., default = default_linters) {
 
 #' Default linters
 #'
-#' List of default linters for \code{\link{lint}}. Use \code{\link{with_defaults}} to customize it.
+#' @description List of default linters for \code{\link{lint}}. Use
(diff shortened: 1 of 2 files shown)


AshesITR (approved):
LGTM, thanks!

PR #721, opened 2021-01-31
Ignore `renv` and `packrat` by default in lint_dir

### LINKED ISSUE #697: Add renv and packrat to default-excluded directories
#492 
Suggest modifying the exclusions param  in lint_dir and linr_package to drop common directories (renv, packrat)

AshesITR:
fixes #697


russHyde (approved):
LGTM

PR #722, opened 2021-01-31
Skip commented_code_linter if actual parsed content was matched.

### LINKED ISSUE #451: False positive for commented code
It's possible to generate a false positive for commented code by creating an object with a name containing two hashtags separated by any alpha-numeric character. 

Linted code:
```R
c("#a#" = 1)
```

Message:
```
.../test.R style: Commented code should be removed.
c("#a#" = 1)
    ^~~~~~~~
```

This should not be flagged as commented code.

AshesITR:
fixes #451

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,7 +45,7 @@
 * Set the default `complexity_limit` in `cyclocomp_linter` to 15. This is the same complexity limit that is enforced via
   `default_linters` (#693, #695, @AshesITR).
 * `lint_package()` now lints files in the `demo` directory by default (#703, @dmurdoch).
-* `commented_code_linter()` now no longer lints parts of actual code (#451, @AshesITR)
+* `commented_code_linter()` uses the parse tree to find comments, eliminating some false positives (#451, @AshesITR)
 
 # lintr 2.0.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,6 +45,7 @@
 * Set the default `complexity_limit` in `cyclocomp_linter` to 15. This is the same complexity limit that is enforced via
   `default_linters` (#693, #695, @AshesITR).
 * `lint_package()` now lints files in the `demo` directory by default (#703, @dmurdoch).
+* `commented_code_linter()` now no longer lints parts of actual code (#451, @AshesITR)
 
 # lintr 2.0.1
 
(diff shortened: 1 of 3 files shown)


AshesITR:
TIL parsed content includes COMMENT tokens :)

MichaelChirico:
> TIL parsed content includes COMMENT tokens 

Why not drop the `regex` logic altogether? Extract `text` from `COMMENT` tokens and check parseability

AshesITR:
I'm working on an XPath version, will push something later.

AshesITR:
NB we still need the regex to extract potential code from the comments

R/comment_linters.R
```diff
@@ -27,37 +27,42 @@ ops <- list(
...
-                          anything
-                        )
-                    ),
-                    global = FALSE, locations = TRUE)
+  if (is.null(source_file$full_xml_parsed_content)) return(list())
+  all_comment_nodes <- xml2::xml_find_all(source_file$full_xml_parsed_content, "//COMMENT")
+  all_comments <- xml2::xml_text(all_comment_nodes)
+  code_candidates <- re_matches(
+    all_comments,
+    rex(some_of("#"), any_spaces,
```
MichaelChirico, line 35: this part isn't needed anymore right? ditto for `anything` in the capture?

AshesITR: I think we need to anchor on the (first) `#` of the comment text, which will always be there in order to get the `column_offset` correct, but I'm not completely sure.

It depends on how the start and end locations of capture groups are returned by `rex`. If they are relative to the input string, we can probably prune the regex to only extract the code candidate to be analyzed. Note that we will still need to somehow anchor it to just after the comment start token, or it will match the entire string including the `#` which would always return a parseable (comment-only) code.

The COMMENT token will contain the `#` character starting the comment, so the refactoring as is is more or less sure to deliver the same candidate code as before if no `#` is present in actual code. WDYT?

MichaelChirico: Sure, agree it's safer not to try & rock the boat during a refactor.

NEWS.md
```diff
@@ -45,6 +45,7 @@
 * Set the default `complexity_limit` in `cyclocomp_linter` to 15. This is the same complexity limit that is enforced via
   `default_linters` (#693, #695, @AshesITR).
 * `lint_package()` now lints files in the `demo` directory by default (#703, @dmurdoch).
+* `commented_code_linter()` uses the parse tree to find comments, eliminating some false positives (#451, @AshesITR)
```
MichaelChirico, line 48: reworded this

MichaelChirico (approved):

PR #723, opened 2021-01-31
fix #550

### LINKED ISSUE #550: Code style: extraction_operator_linter and unneeded_concatenation_linter use partial matching for lint type
grepping the codebase for `type = "w(?!arning)` returned two results of partial matching for warning lints:

https://github.com/jimhester/lintr/blob/1a618f3f7df4889fe39eff5e0ce4e5c0fc6eadfd/R/extraction_operator_linter.R#L20

and

https://github.com/jimhester/lintr/blob/1a618f3f7df4889fe39eff5e0ce4e5c0fc6eadfd/R/unneeded_concatenation_linter.R#L23

All other linters (also for `type = "stlye"` and `type = "error"`) use the full text. For consistency, those two uses should be converted to `type = "warning"` as well.

AshesITR:


MichaelChirico (approved):
testable?

AshesITR:
Not really, `match.arg` silently converted it to `"warning"` before.

PR #724, opened 2021-01-31
spaces_left_parentheses_linter uses XML tree

### LINKED ISSUE #628: Refactor spaces_left_parentheses_linter to use xpath
Following up #533 -- move the logic there to use XPath which should be simpler.

### LINKED ISSUE #780: False positives in spaces_left_parentheses_linter
All of the following generate lints currently; I don't think any of them should:

```
lintr::lint("x**(y + z)\n", lintr::spaces_left_parentheses_linter())
lintr::lint("'[[<-.data.frame'(object, y)\n", lintr::spaces_left_parentheses_linter())
lintr::lint("object@data@get('input')\n", lintr::spaces_left_parentheses_linter())
lintr::lint("x <- ~(. +y)\n", lintr::spaces_left_parentheses_linter())
lintr::lint("for (ii in 1:10) next()\n", lintr::spaces_left_parentheses_linter())
lintr::lint("for (ii in 1:10) break()\n", lintr::spaces_left_parentheses_linter())
# the internal newline is necessary here, not sure why
lintr::lint("if (x > 1)\n  x <- x[-(i)]", lintr::spaces_left_parentheses_linter())
```

All of these are handled by the refactor in #724

MichaelChirico:
Closes #628 , Closes #780


russHyde:
Yes I can do that.

russHyde:
Hi.
Running #724 found lints in these packages, that weren't found in main:
```
with(lints, pr %>% filter(!line %in% main$line) %>% count(package) %>% arrange(desc(n)))
# A tibble: 39 x 2
   package       n
   <chr>     <int>
 1 assertr      85
 2 GRANBase     71
 3 purrr        68
 4 shiny        47
 5 rlang        46
 6 knitr        23
 7 functools    20
 8 remotes      18
 9 withr        17
10 aoos         10
# … with 29 more rows
```

Conversely, the lints found in main but not #724:
```
with(lints, main %>% filter(!line %in% pr$line) %>% count(package) %>% arrange(desc(n)) %>% as.data.frame() %>% head(10))
      package  n
1    unitizer 15
2        Rcpp 11
3       aprof  5
4     assertr  5
5       lintr  4
6      import  3
7  Rd2roxygen  3
8   gWidgets2  2
9       knitr  2
10      rlang  2
```

`assertr` has unique lints in each branch.

Here's some examples

```
# called by the PR but not in main
with(lints, pr %>% filter(!line %in% main$line & package == "assertr") %>% sample_n(6)) %>% as.data.frame()
  package                         filename line_number column_number  type
1 assertr tests/testthat/test-predicates.R          94            57 style
2 assertr tests/testthat/test-predicates.R         179            32 style
3 assertr tests/testthat/test-predicates.R         158            31 style
4 assertr tests/testthat/test-predicates.R         133            31 style
5 assertr tests/testthat/test-predicates.R          77            35 style
6 assertr            vignettes/assertr.Rmd         313            33 style
                                                            message
1 Place a space before left parenthesis, except in a function call.
2 Place a space before left parenthesis, except in a function call.
3 Place a space before left parenthesis, except in a function call.
4 Place a space before left parenthesis, except in a function call.
5 Place a space before left parenthesis, except in a function call.
6 Place a space before left parenthesis, except in a function call.
                                                                     line
1   expect_equal(within_bounds(1,3, include.lower = FALSE)(c(0,1,2,3,4)),
2      expect_equal(within_n_mads(1)(test.vect3)(test.vect3[101]), FALSE)
3                                        expect_error(within_n_sds(1)(1),
4                   expect_equal(within_n_sds(2)(test.vect)(138.1), TRUE)
5                              expect_equal(within_bounds(3, 4)(3), TRUE)
6                        within_n_mads(1)(example.vector)(example.vector)
  linter
1 linter
2 linter
3 linter
4 linter
5 linter
6 linter
```

```
# Called in main but not the PR
with(lints, main %>% filter(!line %in% pr$line & package == "assertr")) %>% as.data.frame()
  package       filename line_number column_number  type
1 assertr R/predicates.R         247            23 style
2 assertr R/predicates.R         247            39 style
3 assertr R/predicates.R         321            25 style
4 assertr R/predicates.R         321            42 style
5 assertr  R/row-redux.R         112            40 style
                                                            message
1 Place a space before left parenthesis, except in a function call.
2 Place a space before left parenthesis, except in a function call.
3 Place a space before left parenthesis, except in a function call.
4 Place a space before left parenthesis, except in a function call.
5 Place a space before left parenthesis, except in a function call.
                                                       line linter
1        within_bounds((mu-(n*stdev)), (mu+(n*stdev)), ...) linter
2        within_bounds((mu-(n*stdev)), (mu+(n*stdev)), ...) linter
3      within_bounds((dmed-(n*dmad)), (dmed+(n*dmad)), ...) linter
4      within_bounds((dmed-(n*dmad)), (dmed+(n*dmad)), ...) linter
5             function(x){ sum((is.na(x)&(!(is.nan(x))))) } linter
```

russHyde:
`f(x)(y)(z)` seems to be a problem, but `g((x * y))` seems to be fixed

MichaelChirico:
Thanks! As I suspected `*(` is not enough. I see `-(` and `+(` there too. Also `&(`...

The repeated calls thing is the main work ahead. Will revisit.

MichaelChirico:
Much more confident in the current approach so removing WIP. Running some comparisons now to see if there are some more edge cases to consider.

AshesITR:
Did you see the lint annotations in unchanged files?
There seems to be a problem with immediately preceeding linebreaks, e.g. `lint("x <- a &&\n  (b_in_parentheses)")`

MichaelChirico:
Seems we'll need to use the affirmative (`@start - 1 = @end`) rather than the positive (`@start - 2 != @end`). Will fix.

MichaelChirico:
Still missing a ton of branches caught on current HEAD. Mostly because of nested arithmetic expressions; fixing that now.


There are a few strange edge cases like

```
break()
next()
```

Which pass on this branch but not master. It's a strange usage of `break`/`next`, but I think it's technically OK?

Another couple of things that shows up on master that I think are not lints:

```
input = object@data@get('input')
object <- '[[<-.data.frame'(object, y, value = value)
```

HEAD might also generate lints for files that don't parse where this branch doesn't.

The number of lints caught here and not on main is very small. All seem to be generated by files in the `demo` folder of this package:

https://github.com/cran/CVD/blob/master/demo/FarnsworthD15.R

For some reason HEAD is not generating a lint on this line:

https://github.com/cran/CVD/blob/85474eb2f075943a488bb25b19a2b4c0b8ba6812/demo/FarnsworthD15.R#L49

```
posI<-(as.numeric(strToB))
```

MichaelChirico:
There's one case that is pretty tough to fix with an XML approach:

```
x <- 5;(x+2)
```

That's because those are split into different expressions and we're using the expression-level approach here for cache-friendliness.

It's pretty rare (in 500 tested packages, it happened 3 times) -- WDYT?

AshesITR:
> 
> 
> There's one case that is pretty tough to fix with an XML approach:
> 
> ```
> x <- 5;(x+2)
> ```
> 
> That's because those are split into different expressions and we're using the expression-level approach here for cache-friendliness.
> 
> It's pretty rare (in 500 tested packages, it happened 3 times) -- WDYT?

I would favour correctness over cache-friendlyness.

MichaelChirico:
> I would favour correctness over cache-friendlyness

OK, done; it's actually pretty easy to just write a targeted version of the logic for the global case.

Here are some more things that turn up lints at HEAD that I think are false positives:

```
lintr::lint("for (i in 1:nj) {
  if (all(tpj[i] > tpj[(i+1):nj])) {
    tpj[i] <- mean(tpj[i:nj])
    tpj <- tpj[-(i+1:nj)]
    break
  }
}", lintr::spaces_left_parentheses_linter())
# ^ lints line 4, but that line alone gives no lint \/
lintr::lint("tpj <- tpj[-(i+1:nj)]\n", lintr::spaces_left_parentheses_linter())
```

`**` should be treated the same as `^`

```
lintr::lint("x^(2+y)\n", lintr::spaces_left_parentheses_linter())
lintr::lint("x**(2+y)\n", lintr::spaces_left_parentheses_linter())
# <text>:1:4: style: Place a space before left parenthesis, except in a function call.
# x**(2+y)
#    ^
```

`~` should be treated the same as `-`, `+` when used as a unary operator:

```
lintr::lint("x <- -(a)\n", lintr::spaces_left_parentheses_linter())
lintr::lint("x <- ~(. + a)\n", lintr::spaces_left_parentheses_linter())
# <text>:1:7: style: Place a space before left parenthesis, except in a function call.
# x <- ~(. + a)
#       ^
```

MichaelChirico:
Consolidated all of the apparent false positives caught by this branch in #780. Purging those from the discussion above to provide more focus.

MichaelChirico:
Not sure what's going on with coverage...

I had to add the branch in 80a01399bb1c3363fdc163383fc511a97049071e since tests were failing. Tests are passing now but codecov says that branch is missed? Not sure how that can be.

`xml_nodes_to_lint` still isn't used in any linters where it's possible for the start & end lines of the matched node to be different, so that's still unreachable (though it should be reachable eventually if we add more linters to use it)

R/spaces_left_parentheses_linter.R
```diff
@@ -3,61 +3,50 @@
...
-              message = "Place a space before left parenthesis, except in a function call.",
-              line = line
-            )
-          }
-        }
-
-      })
+
+    if (is.null(source_file$xml_parsed_content)) {
+      if (is.null(source_file$full_xml_parsed_content)) return(list())
```
AshesITR, line 8: This if might be unreachable, either `xml_parsed_content` or `full_xml_parsed_content` is filled (unless maybe parsing fails, not sure about that).

AshesITR:
Maybe we should unit test `xml_nodes_to_lint()` to get 100% coverage before merging this?

R/spaces_left_parentheses_linter.R
MichaelChirico: in fact it was this test that caused was triggering an error:

https://github.com/jimhester/lintr/blob/1aa166ef7aa3d85a667675847c277b83b4ff4908/tests/testthat/test-error.R#L3

Hence including this in 80a0139 -- maybe that same error doesn't show up on the codecov platform for some reason?

MichaelChirico:
> Maybe we should unit test xml_nodes_to_lint() to get 100% coverage before merging this?

Are you unable to just merge as is? I think it's fine to occasionally make a one-line backslide like this now that we're >99%, given that we look at the issue and decide it's OK.

MichaelChirico:
Ran on 1000 packages at current commit in this branch -- all that are left are differences in lints spotted in the presence of syntax errors, and the false positives mentioned in #780 -- good to merge IMO!

AshesITR:
> 
> 
> > Maybe we should unit test xml_nodes_to_lint() to get 100% coverage before merging this?
> 
> Are you unable to just merge as is? I think it's fine to occasionally make a one-line backslide like this now that we're >99%, given that we look at the issue and decide it's OK.

Merging is blocked by codecov.
Getting the coverage back up shouldn't be too much work IINM though.

MichaelChirico:
Just confirmed for sure that branch is being hit locally for me. Just pushed a test commit with tracing, let's see what comes out in the logs.

MichaelChirico:
Didn't really see anything in the trace log :monocle_face: 

Anyway, added a `:::` test for `xml_nodes_to_lint` -- glad we did! caught a small bug.

MichaelChirico:
Still couldn't hit the required coverage... #nocov'd that line for now to just move on.

R/utils.R
```diff
@@ -212,7 +212,8 @@ xml_nodes_to_lint <- function(xml, source_file, message,
   if (xml2::xml_attr(xml, "line2") == line1) {
     col2 <- as.integer(xml2::xml_attr(xml, "col2")) + offset
   } else {
-    col2 <- nchar(source_file[[line_elt]][line1])
+    # as.integer to strip names
+    col2 <- as.integer(nchar(source_file[[line_elt]][line1]))
```
AshesITR, line 216: Why not `unname()` to strip names? Seems clearer to me.

R/spaces_left_parentheses_linter.R
```diff
@@ -3,61 +3,50 @@
...
-          }
-        }
-
-      })
+
+    if (is.null(source_file$xml_parsed_content)) {
+      if (is.null(source_file$full_xml_parsed_content)) return(list()) # nocov (actually this is covered -- remove me!)
+      # 'x = 1;(x + 2)' can't be detected from the expression-level tree
+      xml <- source_file$full_xml_parsed_content
+      xpath <- "//OP-LEFT-PAREN[@start - 1 = ancestor::expr/preceding-sibling::OP-SEMICOLON/@end]"
```
AshesITR, line 11: Is the `ancestor` axis necessary here?
Wouldn't that trigger a false positive on `x <- 2L;my_function_call( (x + 2) )`?

R/utils.R
MichaelChirico: never been a fan of `unname()`, for mysterious reasons. more concretely `as.integer()` is also used in the other branch so there's some consistency too.

AshesITR: In the other branch, it's used to convert a character to integer.
IMO we should use `unname` if we only want to delete the names.

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,17 @@
+# TODO: delete this test when we have any lints using xml_nodes_to_lint where the
+#   target node can span >1 line (e.g. an <expr>); until then, this test is here
+#   for coverage
+test_that("xml_nodes_to_lint handles >1-line nodes correctly", {
+  xml <- xml2::read_xml('<x line1="1" col1="1" line2="2" col2="10" start="8" end="25"></x>')
+  source_file = list(filename = "test", lines = c(`1` = "abcdefg", `2` = "hijklmnopq"))
```
AshesITR, line 6: here's a lint

MichaelChirico: thanks! is there a place to see these without checking out the Changes tab (which I don't always do)...

maybe I should write a precommit hook for this

AshesITR: You can manually look at the action logs, e.g. [here](https://github.com/jimhester/lintr/runs/2093351910#step:9:14)
Other than that, locally running `devtools::load_all(); lint_package()` is probably the most convenient way.

AshesITR (approved):
LGTM

MichaelChirico:
Just realized we never included a NEWS item here :\ would have been useful. e.g. this update to the linter fixed some false positives, e.g. something we see a lot in `data.table` usage:

```
DT[, ":="(a = 1, b = 2)]
```

Would get linted because `"(` was considered a lint even though it's a legit call here.

From the discussion above it seems there were a few such edge cases that are now handled "more correctly" but I'm forgetting the specifics...

PR #726, opened 2021-01-31
delint test-normalize_exclusions.R

AshesITR:
s/;/\n/
makes lint_package() lint free @ master

### DIFF
--- a/tests/testthat/test-normalize_exclusions.R
+++ b/tests/testthat/test-normalize_exclusions.R
@@ -18,61 +18,88 @@ test_that("it merges two NULL or empty objects as an empty list", {
 })
 
 test_that("it returns the object if the other is NULL", {
-  t1 <- list(); t1[[a]] <- list(1:10)
+  t1 <- list()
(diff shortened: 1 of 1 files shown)


tests/testthat/test-normalize_exclusions.R
```diff
@@ -18,61 +18,88 @@ test_that("it merges two NULL or empty objects as an empty list", {
 })
 
 test_that("it returns the object if the other is NULL", {
-  t1 <- list(); t1[[a]] <- list(1:10)
+  t1 <- list()
```
MichaelChirico, line 21: i guess `t1 = list(a = list(1:10))` would be better?

AshesITR: NB `a <- tempfile()` so the list is named by a filename, not by literal `"a"`

MichaelChirico: oh sorry, silly mistake. yea setNames is not much different in that case.

MichaelChirico (approved):

PR #727, opened 2021-01-31
Anchor to opening brace instead of closing paren in paren_brace_linter

### LINKED ISSUE #583: curly_brace_linter column number is sub-optimal
Lints for code like

```r
fun <- function(){
  42L
}
```

Currently have the location of the curly brace lint at the closing parenthesis.
If it were moved to the opening curly brace, clicking the lint in RStudio Markers pane would put the cursor at just the right spot to insert the missing space.

AshesITR:
fixes #583

- [x] implement
- [ ] update tests to check lint location as well

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,8 @@
 * New `backport_linter()` for detecting mismatched R version dependencies (#506, @MichaelChirico)
 * `paren_brace_linter` and `no_tab_linter` also use more reliable matching (e.g.,
   excluding matches found in comments; #441 and #545, @russHyde)
+* `paren_brace_linter` now marks lints at the opening brace instead of the closing parenthesis, making fixing the lints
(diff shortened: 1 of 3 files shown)


AshesITR:
We could potentially test explicitly for the column number of the lints in tests, WDYT?

MichaelChirico (approved):
LGTM, glad it's an easy fix. XPath can be great!

MichaelChirico:
I'm not sure it's always ideal, but in this case I would say it's a good idea to test since the issue is explicitly about getting a better position from the linter

AshesITR:
Okay, I'll work on that tomorrow maybe.
Fixed the XPath reversal (I overlooked the intermediate `expr`)

MichaelChirico (approved):

PR #729, opened 2021-02-02
nchar() is vectorized, no need for vapply

MichaelChirico:
quick internal change. actually i think we should do this lint on the whole file too, but that is less trivial


russHyde (approved):
LGTM

PR #734, opened 2021-02-03
Simplify line length

MichaelChirico:


AshesITR (commented):
Just a quick thought: Wouldn't setting `column_number = length + 1` be more convenient for breaking up the line?

MichaelChirico:
I think this is a more general issue... in my mind the `ranges` is used to highlight the lint:

```
a_really_long_line_according_to_20_characters
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

which if we go to column+1 will look odd for that use case...

vs the use case you have in mind, which is to land the cursor just _after_ the lint when using it with the rstudio plugin.

Can we make these two use cases work together better?

AshesITR:
How would the highlight look if the range is left in tact (1 to nchar) and the column_number updated?

MichaelChirico:
Oh sorry, I misunderstood. IIUC the `^~~~` tag is only built from ranges. So `column_number` should be fine to update. But perhaps best to keep as a follow-up to keep the PR more focused.

AshesITR:
Thats fine by me.

AshesITR (approved):

Issue #738, opened 2021-02-07
Hit 99% code coverage


MichaelChirico:
As of this writing we're at 97.09%:

https://codecov.io/gh/jimhester/lintr/tree/master/R

99% would be 2450 of 2475 expressions on current HEAD, an increase of 47 lines covered.

Alternatively we can mark about the same number of lines as `# nocov` to get there.

I think that's a good ambitious target to hit for release, but don't let this issue be too much of a blocker.

---

quoting from #739 

Going through the files with coverage < 100%:

 - [ ] namespace.R is covered by #740, so we can leave it alone
 - [ ] namespace_linter.R has an uncovered lint that should trigger only if a package is installed, but loading its namespace fails. IDK how we would test that at all. I'd be fine with # nocov'ing ll 93--104
 - [ ] cache.R has gaps where we check for specified paths to not mess up loading settings.
 - [ ] We could add a test that self-lints lintr with caching enabled, I think. WDYT?
 - [ ] make_linter_from_regex.R could need a test with a custom linter created on-the-fly. The uncovered parts are the builder code.
 - [ ] get_source_expressions.R has a few blind spots. The first one is for parser errors without location information. We manually try to find location information for the attempt to use zero-length variable name case. Other cases aren't tested and I don't even know about them. I'd be fine with # nocov'ing ll 101 -- 109 since those are a fallback Lint if we missed a parse error without location information.
 - [ ] The next is in fix_column_numbers() and should be tested.
 - [ ] Finally fix_eq_assigns() has a blind spot I think could be reached if there are multiple eq-assignments to be fixed.
 - [ ] tree-utils.R has no tests for simplify = FALSE in children() and parents().
 - [ ] path_linters.R has in a few holes in split_path()
 - [ ] object_usage_linter.R seems to have an unreachable early return (IMO we should check full_xml_parsed_content for consistency across all file-level XPath-based linters. IINM coverage will increase if it's added to the test for #740.
 - [ ] comments.R the travis_build_info() seems bugged because "" %||% NULL will always be "", so travis_build_info() will never return NULL, even if !in_travis().
 - [ ] backport_linter.R has a blind spot when a backport lint spans multiple lines.
 - [ ] seq_linter.R: I'm unsure how to reach the uncovered line 29. IINM, the XPath expression guarantees existence of the children. Maybe 1:1 triggers it?
 - [ ] settings.R has a miss because ~/.lintr doesn't exist. I think we can safely # nocov l 97.
[...]

PR #739, opened 2021-02-07
improve code coverage

MichaelChirico:
Progress towards #738


MichaelChirico:
Windows is fun. Finally passing; PTAL to check that we agree this is good cross-platform behavior. Basically we're anchoring to `col1` of the parse data being good in the presence of multi-byte characters.

AshesITR:
Going through the files with coverage < 100%:

 * namespace.R is covered by #740, so we can leave it alone
 * namespace_linter.R has an uncovered lint that should trigger only if a package is installed, but loading its namespace fails. IDK how we would test that at all. I'd be fine with # nocov'ing ll 93--104
 * cache.R has gaps where we check for specified paths to not mess up loading settings.
We could add a test that self-lints `lintr` with caching enabled, I think. WDYT?
 * make_linter_from_regex.R could need a test with a custom linter created on-the-fly. The uncovered parts are the builder code.
 * get_source_expressions.R has a few blind spots. The first one is for parser errors without location information. We manually try to find location information for the attempt to use zero-length variable name case. Other cases aren't tested and I don't even know about them. I'd be fine with # nocov'ing ll 101 -- 109 since those are a fallback Lint if we missed a parse error without location information.  
The next is in `fix_column_numbers()` and should be tested.  
Finally `fix_eq_assigns()` has a blind spot I think could be reached if there are multiple eq-assignments to be fixed.
* tree-utils.R has no tests for `simplify = FALSE` in `children()` and `parents()`.
* path_linters.R has in a few holes in `split_path()`
* object_usage_linter.R seems to have an unreachable early return (IMO we should check `full_xml_parsed_content` for consistency across all file-level XPath-based linters. IINM coverage will increase if it's added to the test for #740.
* comments.R the `travis_build_info()` seems bugged because `"" %||% NULL` will always be `""`, so `travis_build_info()` will never return `NULL`, even if `!in_travis()`.
* backport_linter.R has a blind spot when a backport lint spans multiple lines.
* seq_linter.R: I'm unsure how to reach the uncovered line 29. IINM, the XPath expression guarantees existence of the children. Maybe `1:1` triggers it?
* settings.R has a miss because `~/.lintr` doesn't exist. I think we can safely # nocov l 97.
* comment_linters.R has a miss in `parsable()`. I don't think this is reachable, seems okay to remove it.
* lint.R has a miss for `lint(cache = TRUE)` and `checkstyle_output()` if `attr(lints, "path")` exists. I didn't even know about the attribute. Is this still relevant or should we remove it?
* expect_lint.R has no failure test case for `expect_lint_free()`.
* utils.R has no test for `xml_nodes_to_lint()` where the lint spans multiple lines.
* methods.R has no test for `rstudio_source_markers()` if there are no lints.

AshesITR:
> Windows is fun. Finally passing; PTAL to check that we agree this is good cross-platform behavior. Basically we're anchoring to `col1` of the parse data being good in the presence of multi-byte characters.

Seems off currently:

```
> lintr::lint("`\U2020`<- 1\n")
<text>:1:4: style: Put spaces around all infix operators.
`†`<- 1
  ~^
```

MichaelChirico:
That looks right to me -- or at least, the output when there's a multi-byte character is the same as when not:

```
> lintr::lint("`\U2020`<- 1\n")
<text>:1:4: style: Put spaces around all infix operators.
`†`<- 1
  ~^
> lintr::lint("x<- 1\n")
<text>:1:2: style: Put spaces around all infix operators.
x<- 1
~^
```

MichaelChirico:
> namespace_linter.R has an uncovered lint that should trigger only if a package is installed, but loading its namespace fails. IDK how we would test that at all. I'd be fine with # nocov'ing ll 93--104

In principal we could use `mockery` to force `getNamespace` to eval with error... marking nocov for now.

MichaelChirico:
Up to 98.5% now. pretty close!

MichaelChirico:
> could need a test with a custom linter created on-the-fly. The uncovered parts are the builder code.

Actually this helper is only used once currently (for `no_tab_linter`). Do we even need it? I think we are moving in the direction of minimizing regex-based lints already right? Not sure we would add any more usages for it in the future.

MichaelChirico:
> tree-utils.R has no tests for simplify = FALSE in children() and parents().

These are also unused in the source, so I simply removed that argument & those branches

MichaelChirico:
> comment_linters.R has a miss in parsable(). I don't think this is reachable, seems okay to remove it.

done

MichaelChirico:
OK I am ready to give up :smile: 

I'm not so familiar with the `cache` part of the code, @AshesITR if you're more familiar & could add a test or two that'd be great, otherwise I'd be happy to merge now.

AshesITR:
I'll try to take a look at the cache related tests this week.

AshesITR:
Here are the remaining < 100% covered functions:

 * `get_source_expressions()` ll 311 -- 324: The problem is `test-get_source_expressions.R` writes to file in native encoding. That means the current test case writes `\x86` instead of `\U2020` and thus doesn't actually test column fixing.  
 To fix this behaviour, we'd need to add a explicit encoding to `get_source_expressions()`, which might break existing code that doesn't use `UTF-8`. Filing a separate issue.
 * `split_path()` has two unreachable error conditions (fine IMO) and l. 130 is not covered. I think that line is unreachable on windows and maybe reachable by linting `"https://www.example.com"` on unix.
 * `object_usage_linter()` seems to never find the package name in l. 12 and has some early returns that could probably be tested with a bigger testing package
 * `expect_lint_free()` isn't tested with lints. That should be done.
 * `lint()` still has some cache-related holes.
 * `xml_nodes_to_lint()` has an untested (and currently publicly unreachable) case for multi-line lints.  
 The only linter using this function is the equals_na_linter which only lints the comparisons `==` and `!=` which never span multiple lines.

MichaelChirico:
great stuff!! thanks for the thorough investigation. let's go ahead & merge and revisit on a later release

AshesITR:
I'd like to wait for #751 so we can tick off `expect_lint_free()` as well.
I've already written a test (that blew up 😆)

AshesITR (approved):

AshesITR:
@MichaelChirico feel free to merge if the changes look good to you.

PR #740, opened 2021-02-07
fix and test namespace_imports() in object_name_linter

### LINKED ISSUE #642: Intended behaviour of `namespace_imports` for `object_name_linter`?
I think the namespace one is actually a legit bug. `namespace_imports` is only called here:

https://github.com/jimhester/lintr/blob/fa05ceb2bf967ddad9230e25879b95084c80b0dc/R/object_name_linters.R#L57

without an argument. the default is to run `find_package()` (again without argument), but there's no default, so `normalizePath` always fails:

https://github.com/jimhester/lintr/blob/fa05ceb2bf967ddad9230e25879b95084c80b0dc/R/lint.R#L287

In `namespace_imports` this error happens within `tryCatch` so it just returns `NULL`.

I'm not sure the intended behavior but certainly the above is not it :smiley:

_Originally posted by @MichaelChirico in https://github.com/jimhester/lintr/issues/639#issuecomment-735806874_

AshesITR:
fixes #642

also increases coverage of R/namespace.R to 100%


tests/testthat/test-dir_linters.R
```diff
@@ -25,6 +25,20 @@ test_that("lint all relevant directories in a package", {
...
+  # Code coverage is not detected for default_linters.
+  # We want to ensure that object_name_linter uses namespace_imports correctly.
+  # assignment_linter is needed to cause a lint in all vignettes.
+  linters <- list(assignment_linter, object_name_linter())
+  read_settings(NULL)
+  lints <- lint_package(the_pkg, linters = linters, parse_settings = FALSE)
+  linted_files <- unique(names(lints))
+
+  # lintr paths contain backslash on windows, list.files uses forward slash.
+  linted_files <- gsub("\\", "/", linted_files, fixed = TRUE)
```
MichaelChirico, line 38: not particularly important for a test case like this but i think `normalizePath` handles this right?

MichaelChirico (approved):

tests/testthat/test-dir_linters.R
AshesITR: We'd have to normalize both paths because that also makes them absolute.

PR #744, opened 2021-02-10
fix warnings in test-dir_linters.R

russHyde:
Fixed warnings that were produced when running test-dir_linters.
These stemmed from using `assignment_linter` rather than
`assignment_linter()`.


AshesITR (approved):
LGTM

PR #745, opened 2021-02-10
Reduce cyclocomp of lint function

russHyde:
Ref https://github.com/jimhester/lintr/issues/653

Splits lint() function up; cyclomatic complexity of the resulting functions < 15.

This is work-in-progress. I have to check that the memory footprint and the cacheing are unaffected by the changes.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,6 +1,6 @@
 linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
    line_length_linter = line_length_linter(120),
-   cyclocomp_linter = cyclocomp_linter(23) # TODO: reduce to 15
+   cyclocomp_linter = cyclocomp_linter(20) # TODO: reduce to 15
  )
 exclusions: list(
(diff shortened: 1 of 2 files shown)


russHyde:
Suspect it might be neater to introduce a lightweight `lint_runner` class that is defined within `lint()`.

russHyde:
Note to self, when rewriting this spike: all the cache-existence checking can be abstracted out; introduce Cache class (with separate NullCache and FileCache behaviour) pull all if(cache) logic down into the Cache class' methods; use `cache_lint.Cache` in lint()

russHyde:
#747 retains the cacheing logic better than this. I'll rewrite some of this PRs changes if that PR gets merged.

russHyde:
#754 contains some of the simple wins from this changeset (and should be merged first)

russHyde:
Closing, since #754 has been merged and #747 is in process, the changes here are valueless (and feel wrong to me; saving up all the things that are to be cached until having ran all the linters on all lines seems incorrect)

Issue #653, opened 2020-12-02
Make R/ conform to `cyclocomp_linter`


MichaelChirico:
Final part of #584.

The current target is to use `cyclocomp_linter(15)`. It's currently enabled at `cyclocomp_linter(29)`.

At current `master`, We get the following functions triggering the `cyclocomp_linter`:

 - [ ] CC=17 R/closed_curly_linter.R:6:1:closed_curly_linter
 - [ ] CC=16 R/commas_linter.R:5:1:commas_linter
 - [ ] CC=16 R/get_source_expressions.R:340:1:fix_tab_indentations
 - [ ] CC=18 R/get_source_expressions.R:393:1:fix_eq_assigns
 - [ ] CC=29 R/lint.R:25:1:lint
 - [ ] CC=18 R/namespace_linter.R:9:1:namespace_linter
 - [ ] CC=16 R/open_curly_linter.R:4:1:open_curly_linter
 - [ ] CC=17 R/path_linters.R:104:1:split_path
 - [ ] CC=19 R/sprintf_linter.R:4:1:sprintf_linter

Because each of these is more than a whitespace change/trivial refactor, each function should be addressed in its own PR.

PS One worry I have is that this one is more likely to be a high barrier for contributed PRs, as the linter's advice is not particularly actionable (my cyclomatic complexity is 16... OK... what exactly do I do now?). For the vast majority of people contributing PRs, any bot pestering them about the "cyclomatic complexity" of their PR is going to be confusing to say the least. Do we really want to send contributors on a chase to figure out what cyclomatic complexity is, and then how to apply it to their code?

One alternative would be to keep the threshold much lower during dev (say 29, like currently), and then have a pre-release task to address cyclomatic complexity among the maintainers before release.

Any thoughts?

AshesITR:
Do you think the error message of cyclocomp could be made more self-contained by including actionable advice?
Something like "try reducing the number of code branches by factoring out logic into separate, smaller functions"?

Otherwise it seems we'd have to continuously up / down the cyclocomp config...
Maybe we should discuss increasing the threshold of 15 in general. NB the factory default for cyclocomp is 25 which would make a lot of the current lints disappear.

MichaelChirico:
Closing in favor of #1186

PR #747, opened 2021-02-11
Use cache object

russHyde:
- [x] Add class Cache
- [x] Convert functions in `cache.R` to methods on Cache
- [x] Push logic re 'whether to cache' lints down from `lint()` to Cache-methods
- [ ] Split `cache.R` into `cache-class.R` and `cache-methods.R`
- [ ] Add integration tests on `lint` that require the use of a cache
- [ ] Change test-cache.R:
    - [ ] pass in Cache object
    - [ ] handle NullCache (nothing should be saved/retrieved if the user doesn't want to use a cache)

Adds a new S3 class `Cache` and converts some pre-existing cacheing functions to be methods on this class.

There was multiple checks of the form `if(cache_exists)` within the `lint()` function prior to this PR; this made reading the function complicated.

These  checks are now handled by the Cache class and it's methods: if called, `cache_lint`, `save_cache`, and `cache_file` will only modify the cache when relevant. That is, if the user has called `lint(some_file, cache = FALSE)` the `*_cache` and `cache_*` methods will still be called, but they will not create/modify a cache file.

This reduces the cyclomatic complexity of `lint()` from 33 to 23 (see #653 ) (further improvements to the readability of lint() could come from moving the linter definition and validation steps into helper functions and from splitting the lint() function into an input-validation function and a separate lint-runner function that works on prevalidated input)


AshesITR (changes requested):
I like the idea of using a custom S3 class to keep caching state in one place.
However, I'd recommend against defining S3 methods for cache methods unless absolutely necessary.

And even if we do define S3 methods, there shouldn't be a `.default` method. All of the current default implementations implicitly assume that `cache` is an environment so if we keep them, they should be renamed to `.environment`-methods.

IINM we only need implementations that always take a cache object.

russHyde:
This seems reasonable; I wasn't sure what the type of the 'cache' input was in the functions in cache.R and was just trying to implement the idea with the fewest changes to the code. Should I rewrite the other functions in `cache.R` to use `Cache` objects as input?

AshesITR:
That seems the best way to go.
Only working with `Cache` objects in all cache functions makes sense to me.

AshesITR:
WDYT about defining the generics for `.FileCache` and `.NullCache` instead?
`Cache(...)` would then create either a `structure(list(), class = c("NullCache", "Cache"))` or a `structure(list(...), class = c("FileCache", "Cache"))` and the `.NullCache` implementations would be no-ops whereas the `.FileCache` implementations would use the info contained in the list to do the requested operations to its underlying environment / file.

russHyde:
Thanks, yes that was the plan. The tests would need rewriting before that can be done, so it's not a straightforward refactoring.  It will come together soon.

.lintr
```diff
@@ -1,6 +1,6 @@
 linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
    line_length_linter = line_length_linter(120),
-   cyclocomp_linter = cyclocomp_linter(28) # TODO reduce to 15
+   cyclocomp_linter = cyclocomp_linter(19) # TODO reduce to 15
```
AshesITR, line 3: Object usage liner reports as 20 currently

R/cache-methods.R
```diff
@@ -0,0 +1,203 @@
...
+    read_settings(file)
+    path <- settings$cache_directory
+  }
+
+  if (!file.exists(path)) {
+    dir.create(path, recursive = TRUE)
+  }
+
+  save(
+    file = get_cache_file_path(file, path),
```
AshesITR, line 22: There should be a NULL check for path here.

AshesITR (changes requested):
Looks mostly good so far

R/cache-methods.R
russHyde: That should be handled earlier. If settings$cache_directory is null, then there should be an early return

.lintr
russHyde: Haha, well it is a bloated beast.

russHyde:
Thanks for your review of this draft PR @ashesITR. I really don't think I'll be able to get this done with any immediacy. I'll have to rethink the Cache implementation, because the unit tests in test-cache.R are difficult to rewrite.

R/cache-methods.R
AshesITR: Yeah, adding the check in the `if (is.null(path))` branch after l.14 is much better.

AshesITR:
Hi @russHyde. The PR has began to diverge from master but is very valuable I think.
Do you think you'll get the chance to work on this in the near future? If so, I can try to resolve the merge conflicts if that helps.

IndrajeetPatil:
Dear @russHyde, let us know if you want to resolve the merge conflicts here before you continue, or if you want us to take it from here, in case you can no longer work on it. Thanks :)

MichaelChirico:
Gonna close since this is so far diverged from `main` / not much apparent appetite for working on it. Feel free to tackle from scratch on current `main`.

PR #751, opened 2021-02-13
use correct index for rows of `code_candidates`

### LINKED ISSUE #742: `lintr` not working with `travis` any longer
I've been using `lintr` with `travis` since a long time, and it has always worked, but it has stopped working since a last few days, and I am not sure why.

The exact error I see is:

```r
84.56s$ Rscript -e 'lintr::lint_package()'
Calls: <Anonymous> ... FUN -> cat -> highlight_string -> fill_with -> paste0
Execution halted
Done. Your build exited with 0.
```

- `.travis.yml` config is here: https://github.com/IndrajeetPatil/ggstatsplot/blob/master/.travis.yml
- build log can be found here: https://travis-ci.org/github/IndrajeetPatil/ggstatsplot/jobs/758198656

AshesITR:
fixes #742

the XPath based code_candidates have one row per comment node, not one row per line

### DIFF
--- a/R/comment_linters.R
+++ b/R/comment_linters.R
@@ -57,13 +57,13 @@ commented_code_linter <- function() {
         Lint(
           filename = source_file$filename,
           line_number = line_number,
-          column_number = column_offset + code_candidates[line_number, "code.start"],
+          column_number = column_offset + code_candidates[code_candidate, "code.start"],
           type = "style",
           message = "Commented code should be removed.",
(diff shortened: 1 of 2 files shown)


MichaelChirico (changes requested):
any test we can write?

AshesITR:
I think we can add location info to the existing tests. Will do that.

MichaelChirico (approved):

PR #754, opened 2021-02-13
Define and validate linters

russHyde:
Following https://github.com/jimhester/lintr/commit/af6ec3078aeefe8aeeaf76958280199e03319406 the cyclomatic complexity of lint() went up to 34. Since we have two linting steps in the PR review process, this increase in cyclomatic complexity would have been easy to overlook at the time.

The current changes brings cyclomatic-complexity of lint() down to 28 (which is now the threshold set in .lintr), and cleans up the lint function a tiny bit.
The code for defining the set of linters used inside lint(), and for checking that they are all Linter objects has been extracted into helper functions.

I'm attempting to make other changes to lint(); but think the changes here should be included first.

All non-exported helper functions that are in lint.R have been pushed to the bottom of the file for readability.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,6 +1,6 @@
 linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
    line_length_linter = line_length_linter(120),
-   cyclocomp_linter = cyclocomp_linter(30) # TODO reduce to 15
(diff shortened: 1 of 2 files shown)


AshesITR (approved):
LGTM

PR #756, opened 2021-02-15
fix expectations in test-lint_file.R

AshesITR:
expect_warning and expect_error with fixed expectations need `regexp = "bla", fixed = TRUE`, not `fixed = "bla"`.
Thanks @MichaelChirico for pointing this out.


russHyde (approved):
LGTM

PR #757, opened 2021-02-15
Add cache-test fixtures

russHyde:
Much of the data used in the tests for 'retrieve_lint' was duplicated across tests.
Added a fixture within the 'test-cache.R' context to set up this data.

This will be useful if we introduce a class to handle cacheing logic (re https://github.com/jimhester/lintr/pull/747)


AshesITR (approved):
LGTM

PR #758, opened 2021-02-15
Correctly parse checkUsage when a multiline warning is encountered

### LINKED ISSUE #507: Missing object_usage_linter due to line break
```
foo <- function() {
    unknown_function(1)

    unknown_function(
        1
    )

    unknown_function(1)
}
```

I get this in VS Code, while in my opinion, the middle call should also be detected.

image
image

AshesITR:
fixes #507

 - [x] implement fix
 - [x] add regression test
 - [x] add NEWS bullet

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -61,6 +61,7 @@ object_usage_linter <- function() {
 
           location <- re_matches(line, rex(boundary, row$name, boundary), locations = TRUE)
 
+          # Handle multi-line lints where name occurs on subsequent lines (#507)
           if (is.na(location$start) && nzchar(row$line2) && row$line2 != row$line1) {
             lines <- source_file$content[org_line_num:(as.integer(row$line2) + info$line1 - 1L)]
             locations <- re_matches(lines, rex(boundary, row$name, boundary), locations = TRUE)
@@ -72,6 +73,7 @@ object_usage_linter <- function() {
             line <- lines[matching_row]
           }
 
+          # Fallback if name isn't found anywhere: lint the first line
           if (is.na(location$start)) {
             location$start <- 1L
             location$end <- nchar(line)
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -59,7 +59,23 @@ object_usage_linter <- function() {
 
           row$name <- re_substitutes(row$name, rex("<-"), "")
 
-          location <- re_matches(line, rex(row$name), locations = TRUE)
+          location <- re_matches(line, rex(boundary, row$name, boundary), locations = TRUE)
+
+          if (is.na(location$start) && row$line2 != row$line1) {
+            lines <- source_file$content[org_line_num:(as.integer(row$line2) + info$line1 - 1L)]
+            locations <- re_matches(lines, rex(boundary, row$name, boundary), locations = TRUE)
+
+            matching_row <- (which(!is.na(locations$start)) %||% 1L)[[1L]] # first matching row or 1 (as a fallback)
+
+            org_line_num <- org_line_num + matching_row - 1L
+            location <- locations[matching_row, ]
+            line <- lines[matching_row]
+          }
+
+          if (is.na(location$start)) {
+            location$start <- 1L
+            location$end <- nchar(line)
+          }
 
           Lint(
             filename = source_file$filename,
@@ -138,7 +154,8 @@ parse_check_usage <- function(expression) {
   try(codetools::checkUsage(expression, report = report))
 
   function_name <- rex(anything, ": ")
-  line_info <- rex(" ", "(", capture(name = "path", non_spaces), ":", capture(name = "line1", digits), maybe("-", capture(name = "line2", digits)), ")")
+  line_info <- rex(" ", "(", capture(name = "path", non_spaces), ":",
+                   capture(name = "line1", digits), maybe("-", capture(name = "line2", digits)), ")")
 
   res <- re_matches(
     vals,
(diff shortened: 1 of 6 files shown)


AshesITR:
Diff coverage needs a test for `declared_globals` to pass.

R/object_usage_linter.R
```diff
@@ -140,23 +139,37 @@ parse_check_usage <- function(expression) {
...
   function_name <- rex(anything, ": ")
-  line_info <- rex(" ", "(", capture(name = "path", non_spaces), ":", capture(name = "line_number", digits), ")")
-
-  res <- re_matches(vals,
-                    rex(function_name,
-                        capture(name = "message", anything,
-                                one_of(quote, "\u2018"), capture(name = "name", anything), one_of(quote, "\u2019"),
-                                anything),
-                        line_info))
+  line_info <- rex(" ", "(", capture(name = "path", non_spaces), ":", capture(name = "line1", digits), maybe("-", capture(name = "line2", digits)), ")")
```
MichaelChirico, line 142: please fix the long line lint

infotroph:
I've only tried this fix on one (very messy) file so far, but lintr is exiting with a new error before it gets to the multiline expression I wanted to test:

```
Calls: <Anonymous> ... print.lint -> cat -> highlight_string -> fill_with -> paste0
Execution halted
```
When linting with the released version of lintr the next message emitted is a "Lines should not be more than 80 characters" -- if `rep.int` doesn't immediately tip you off I'll try to narrow this down to a reprex tonight.

AshesITR:
@infotroph 

This seems vaguely related to #742.
Can you narrow down which linter produces NA column_numbers?
`as.data.frame(lint(...))` should help you find the issue.

AshesITR:
We're still getting NA locations if the info doesn't match on the first line of the checkUsage info.
I'll work on making the regex more robust.

infotroph:
Now working without error and with correct messages for the case I reported above, but for completeness I can report narrowed the offending lines down to an if-else with unbound names in both conditions:

```
fn <- function() {
  if (badname) {
  } else if (badname2) {
  }
}
```

Again, as of 9209e7a1b61 lintr is correctly warning on both lines and not erroring.

AshesITR:
Thanks for checking this out @infotroph.
@MichaelChirico PTAL, I think it's ready to merge now.

R/object_usage_linter.R
```diff
@@ -45,34 +45,49 @@ object_usage_linter <- function() {
...
+          }
+
+          org_line_num <- as.integer(row$line1) + info$line1 - 1L
+          line <- source_file$content[as.integer(org_line_num)]
+
+          row$name <- re_substitutes(row$name, rex("<-"), "")
+
+          location <- re_matches(line, rex(boundary, row$name, boundary), locations = TRUE)
+
+          if (is.na(location$start) && nzchar(row$line2) && row$line2 != row$line1) {
```
MichaelChirico, line 64: Could you add a quick comment here explaining this branch? At a minimum pointing to the issue #

R/spaces_left_parentheses_linter.R
```diff
@@ -9,12 +9,15 @@ spaces_left_parentheses_linter <- function() {
 
         parsed <- source_file$parsed_content[id, ]
 
+        # nolint start: object_usage_linter.
+        # suppress "no visible binding" warnings in with() block
```
MichaelChirico, line 13: do we also get the warning in `R CMD check`?

tests/testthat/test-object_usage_linter.R
```diff
@@ -215,3 +215,86 @@ test_that("used symbols are detected correctly", {
...
+    object_usage_linter()
+  )
+
+  # Kill regex match to enforce fallback to line 1 column 1 of the warning
+  expect_lint(
+    trim_some("
+      foo <- function(x) {
+        with(
+          x,
+          `\u2019regex_kill`
```
MichaelChirico, line 258: the test here is a symbol name that's not matched by our regex? so that the linter has to default to where the expression starts

AshesITR: The test causes the `boundary, res$name, boundary` regex to not match against the source code, yes.

R/spaces_left_parentheses_linter.R
AshesITR: `devtools::check()` was clean locally and in CI, so I guess not. Not sure why, though.

MichaelChirico: Hmm it's because here:

https://github.com/jimhester/lintr/blob/25b2e1d430ba13b1436af4afd4332f27762f4f8b/R/pipe_continuation_linter.R#L76

Should `object_usage_linter` be seeing that too though?

(side note maybe it's better to put that call in a less hidden place than the bottom of a specific linter's source)

AshesITR: We see those. Linted are `token`, `parent` and `terminator` IIRC.

You're right about moving the globalVariables call. Should be in `aaa.R` I think.

R/object_usage_linter.R
AshesITR: Done.

R/spaces_left_parentheses_linter.R
AshesITR: Moved all globalVariables to `aaa.R`. As it turns out, `"from"` is not necessary as a global variable, so I removed that and instead added all `colnames()` of parsed content with a comment to explain what's going on.

MichaelChirico: `with` is specifically skipped by `R CMD check` as near as I can tell, I believe it happens in this helper:

https://gitlab.com/luke-tierney/codetools/-/blob/master/R/codetools.R#L647-652

Anyway it's easy to write a simple package where `with` generates no notes.

MichaelChirico (approved):

PR #759, opened 2021-02-15
Use XPath for T_and_F_symbol_linter and fix false positives

### LINKED ISSUE #657: incorrectly tagged lint "Use FALSE instead of the symbol F"
Since a couple of days, I have started to get a new lint in my CI checks (https://github.com/IndrajeetPatil/statsExpressions/commit/cc00f65c4c6640c51cea165f6a557f13c097db85#commitcomment-44804914).

But I think this is unintentional, since here `F` refers to `F`-statistic

```r
R/helpers_anova.R:134:34: style: Use FALSE instead of the symbol F.

​        statistic = ez_df$ANOVA$F[2],

                                ~^
```

AshesITR:
fixes #657

 - [x] implement fix
 - [x] add regression test
 - [x] add NEWS bullet

Migrating to XPath makes adding other exceptions easier as well.
Had to add the argument `offset` to `xml_nodes_to_lint` to keep the column numbers unchanged.

### DIFF
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -41,7 +41,6 @@ T_and_F_symbol_linter <- function() { # nolint: object_name_linter.
         xml = expr,
         source_file = source_file,
         message = message,
-        linter = "T_and_F_symbol_linter",
         type = "style",
         offset = 1L
       )
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -9,22 +9,25 @@ T_and_F_symbol_linter <- function() { # nolint: object_name_linter.
     xpath <- paste0(
       "//SYMBOL[",
       "(text() = 'T' or text() = 'F')", # T or F symbol
-      " and count(preceding-sibling::OP-DOLLAR) = 0", # not part of a $-subset expression
-      " and count(parent::expr/following-sibling::LEFT_ASSIGN) = 0", # not target of left assignment
-      " and count(parent::expr/preceding-sibling::RIGHT_ASSIGN) = 0", # not target of right assignment
-      " and count(parent::expr/following-sibling::EQ_ASSIGN) = 0", # not target of equals assignment
+      " and not(preceding-sibling::OP-DOLLAR)", # not part of a $-subset expression
+      " and not(parent::expr[",
+      "  following-sibling::LEFT_ASSIGN", # not target of left assignment
+      "  or preceding-sibling::RIGHT_ASSIGN", # not target of right assignment
+      "  or following-sibling::EQ_ASSIGN", # not target of equals assignment
+      "])",
       "]"
     )
 
     xpath_assignment <- paste0(
       "//SYMBOL[",
       "(text() = 'T' or text() = 'F')", # T or F symbol
-      " and count(preceding-sibling::OP-DOLLAR) = 0", # not part of a $-subset expression
-      " and (",
-      " count(parent::expr/following-sibling::LEFT_ASSIGN) > 0", # target of left assignment
-      " or count(parent::expr/preceding-sibling::RIGHT_ASSIGN) > 0", # target of right assignment
-      " or count(parent::expr/following-sibling::EQ_ASSIGN) > 0", # target of equals assignment
-      ")]"
+      " and not(preceding-sibling::OP-DOLLAR)", # not part of a $-subset expression
+      " and parent::expr[", #, but ...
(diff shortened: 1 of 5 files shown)


AshesITR:
The `exclusions-test` also suffered from a #657 related false positive.
One of the eight lints was for the `F` in `F = T`.

MichaelChirico:
> Don't use {T/F} as a variable name, as it can break code relying on {T/F} being {TRUE/FALSE}.

I agree with this (and have run into this issue before). FWIW, using T/F as variables is against the style guide due to conflicts with `base::T`/`base::F` (and the `T` case is specifically called out):

https://style.tidyverse.org/syntax.html#object-names

R/T_and_F_symbol_linter.R
```diff
@@ -4,28 +4,55 @@
...
-            linter = "T_and_F_symbol_linter"
-          )
-        }
-      }
+    if (is.null(source_file$xml_parsed_content)) return(list())
+
+    xpath <- paste0(
+      "//SYMBOL[",
+      "(text() = 'T' or text() = 'F')", # T or F symbol
+      " and count(preceding-sibling::OP-DOLLAR) = 0", # not part of a $-subset expression
```
MichaelChirico, line 12: `not(preceding-sibling::OP-DOLLAR)` is more "canonical XPath" to me (not that I have any real expertise to call XPath canonical :sweat_smile:)

R/T_and_F_symbol_linter.R
```diff
@@ -4,28 +4,55 @@
...
+      " and count(parent::expr/following-sibling::LEFT_ASSIGN) = 0", # not target of left assignment
```
MichaelChirico, line 13: `not(parent::expr[following::sibling::LEFT_ASSIGN or following-sibling::EQ_ASSIGN or preceding-sibling::RIGHT_ASSIGN])` would be better IMO

R/T_and_F_symbol_linter.R
```diff
@@ -4,28 +4,55 @@
...
+      " and count(parent::expr/preceding-sibling::RIGHT_ASSIGN) = 0", # not target of right assignment
+      " and count(parent::expr/following-sibling::EQ_ASSIGN) = 0", # not target of equals assignment
+      "]"
+    )
+
+    xpath_assignment <- paste0(
+      "//SYMBOL[",
+      "(text() = 'T' or text() = 'F')", # T or F symbol
+      " and count(preceding-sibling::OP-DOLLAR) = 0", # not part of a $-subset expression
```
MichaelChirico, line 22: similarly here:

```
[not(preceding-sibling::OP-DOLLAR) and parent::expr[following-sibling::LEFT_ASSIGN or ...]]
```

R/utils.R
```diff
@@ -184,13 +184,14 @@ unescape <- function(str, q="`") {
 
 # convert an XML match into a Lint
 xml_nodes_to_lint <- function(xml, source_file, message, linter,
-                              type = c("style", "warning", "error")) {
+                              type = c("style", "warning", "error"),
+                              offset = 0L) {
```
MichaelChirico, line 188: nice extension! eventually i think we can refactor a lot of the XPath lints to use this helper. Basically as long as `message` is static (e.g. this isn't the case for `seq_linter`) I think this can be used.

R/T_and_F_symbol_linter.R
AshesITR: Sounds more idiomatic than `count() = 0` 👍 
I'll change this later.

R/utils.R
AshesITR: Maybe we can even vectorize on message?
The T and F symbol linter needs a wrapper at the call site that just generates the proper message.

R/T_and_F_symbol_linter.R
AshesITR: I kept the OP-DOLLAR part the same to help see the parallels between the two XPath expressions.
They only differ in `and not(parent::expr[...])` vs. `and parent::expr[...]`

R/T_and_F_symbol_linter.R
AshesITR: Done

R/T_and_F_symbol_linter.R
AshesITR: Done

AshesITR:
r-release CI failures are due to #761.

R/utils.R
MichaelChirico: Let's tuck this away to revisit for now

MichaelChirico (approved):

PR #755, opened 2021-02-13
capture specifically generic names 

AshesITR:
fixes #737 by capturing specifically names of declared generics followed by a `.` and a class specification.

PR #760, opened 2021-02-16
Add a script for comparing lint output on branches/PRs

MichaelChirico:
Basically extending @russHyde / @AshesITR script here:

https://gist.github.com/russHyde/568fd5af558c860d97e1b932ca773ff7

And tracking it here for transparency/easy versioning. Put in .dev which can hold various helpers for doing dev on the package.

Have been testing it on some sample CRAN packages but it's tough to get started. Lots of Depends to install from packages chosen at random... library is bloating quickly :smile_cat:

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -48,7 +48,7 @@ if ("branch" %in% names(args)) {
   branch <- args["branch"]
   is_branch <- TRUE
 } else if ("pr" %in% names(args)) {
-  pr <- args["pr"]
+  pr <- as.integer(args["pr"])
 } else {
   pr <- 709L
 }
@@ -78,7 +78,8 @@ if ("outfile" %in% names(args)) {
   outfile <- args["outfile"]
 } else {
   outfile <- normalizePath(
-    file.path("~", sprintf("lintr_compare_branches_%d.csv", as.integer(Sys.time())))
+    file.path("~", sprintf("lintr_compare_branches_%d.csv", as.integer(Sys.time()))),
+    mustWork = FALSE
   )
 }
 
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -106,7 +106,9 @@ lint_all_packages <- function(pkgs, linter) {
       if (!pkg_is_dir[ii]) {
         tmp <- file.path(tempdir(), pkg_names[ii])
         on.exit(unlink(tmp, recursive = TRUE))
-        utils::untar(pkgs[ii], exdir = tmp)
+        # --strip-components makes sure the output structure is
+        # /path/to/tmp/pkg/ instead of /path/to/tmp/pkg/pkg
+        utils::untar(pkgs[ii], exdir = tmp, extras="--strip-components=1")
         pkg <- tmp
       }
       # devtools::load_all() may not work for packages with Depends
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -1,26 +1,8 @@
 #!/usr/local/bin/Rscript
 
-# compare the lints obtained before/after a given PR
-# the PR is always compared against master
+# compare the lints obtained before/after a given PR/branch vs current master
 
-# arguments
-# --linters=linter1,linter2,...
-#   run the comparison for these linters
-# --branch=branch
-#   run the comparison for master vs this branch
-# --pr=pr
-#   run the comparison for master vs this pr
-# --packages=/path/to/package1,/path/to/package2,...
-#   run the comparison for these packages
-# --pkg_dir=pkg_dir
-#   run the comparison for all packages in dir
-# --sample_size
-#   select a sample of the packages found in
-#   --packages/--pkg_dir of the given size
-# --outfile
-#   a file to which to write the output
-
-library(magrittr)
+library(optparse)
 library(dplyr)
 library(purrr)
 library(tibble)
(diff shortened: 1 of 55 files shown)


russHyde (changes requested):
Thanks for adding this, mostly looks sound.
I hadn't heard of gert before, so thanks for that tip.
Would it be possible to allow the user to turn off running the load_all step; and could we use optparse for option handling, rather than rolling our own handler, please.

.dev/compare_branches.R
```diff
@@ -0,0 +1,193 @@
+#!/usr/local/bin/Rscript
+
+# compare the lints obtained before/after a given PR
+# the PR is always compared against master
+
+# arguments
```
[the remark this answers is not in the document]
MichaelChirico: yep, done! i don't have a go-to package for this in R so i usually just make it myself :sweat_smile:

.dev/compare_branches.R
```diff
@@ -0,0 +1,193 @@
...
+  pr <- 709L
+}
+
+# prioritize packages
+if ("packages" %in% names(args)) {
+  packages <- strsplit(args["packages"], ",", fixed = TRUE)[[1L]]
+} else if ("pkg_dir" %in% names(args)) {
+  packages <- list.files(normalizePath(args["pkg_dir"]), full.names = TRUE)
+} else {
+  packages <- file.path("~", "proj", "code_as_data", "data", "packages")
```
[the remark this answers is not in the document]
MichaelChirico: sure; done

MichaelChirico:
Didn't add the option for turning `load_all` on/off (it's easy enough), i got confused about it... I think we have to run `load_all` to source the `lintr` functions for each branch.

What I think should be optional is running `load_all` for the package sources, which I don't think is happening now? But then I don't know why I ran into some "package not found" issues... I guess it's because `object_usage_linter` does run it for the package code? In which case I'm not sure how to turn it off.

Anyway I'll take a closer look at it later.

russHyde:
Sorry about that. I confused myself. Please disregard my note about devtools::load_all; I mistakenly thought you were loading the analysed package prior to linting.

russHyde (approved):
LGTM

russHyde:
The failing tests in CI aren't related to this PR.

AshesITR:
Running `load_all()` on the linted package is necessary for `object_usage_linter` (AFAIK the only one needing to know global package state). So we might want to add this in the future.

But I ran into very messy issues while running the tests for packages that are needed by the script and contain binaries (e.g. purrr, rlang, dplyr, tidyr, ...) so that feature should probably come with more safety measures like running in separate R processes and maybe installing the active lintr branch into a temporary local library.

MichaelChirico:
Also needed for `object_name_linter` btw... maybe not loading the full package, but I believe `linter:::namespace_imports` won't work if the package's `Depends` are not available... certainly running into that issue now trying to run a comparison for #755

MichaelChirico:
Added some functionality for running interactively, I think it works pretty nicely. Would be happy to merge now, PTAL

.dev/compare_branches.R
```diff
@@ -0,0 +1,230 @@
...
+
+library(optparse)
+library(dplyr)
+library(purrr)
+library(tibble)
+library(usethis)
+library(gert)
+library(devtools)
+
+if(!file.exists("lintr.Rproj")) {
```
AshesITR, line 13: Please fix the lint

.dev/compare_branches.R
```diff
@@ -0,0 +1,230 @@
...
+    default = file.path("~", sprintf("lintr_compare_branches_%d.csv", as.integer(Sys.time()))),
+    help = "Destination file to which to write the output"
+  )
+)
+
+params <- optparse::parse_args(optparse::OptionParser(option_list = param_list))
+# treat any skipped arguments from the prompt as missing
+if (interactive()) {
+  for (opt in c("branch", "pr", "packages", "pkg_dir", "sample_size")) {
+    if (params[[opt]] == "") params[[opt]] = NULL
```
AshesITR, line 71: lint

.dev/compare_branches.R
```diff
@@ -0,0 +1,230 @@
...
+# prioritize "branch"
+is_branch <- FALSE
+if (!is.null(params$branch)) {
+  branch <- params$branch
+  is_branch <- TRUE
+} else if (!is.null(params$pr)) {
+  pr <- params$pr
+} else {
+  message("Please supply a branch (--branch) or a PR number (--pr)")
+  q("no")
```
AshesITR, line 86: Why not `stop()`? `q()` in interactive seems like a bad idea.

.dev/compare_branches.R
```diff
@@ -0,0 +1,230 @@
...
+}
+
+# prioritize packages
+if (!is.null(params$packages)) {
+  packages <- strsplit(params$packages, ",", fixed = TRUE)[[1L]]
+} else if (!is.null(params$pkg_dir)) {
+  packages <- list.files(normalizePath(params$pkg_dir), full.names = TRUE)
+} else {
+  message("Please supply a comma-separated list of packages (--packages) or a directory of packages (--pkg_dir)")
+  q("no")
```
AshesITR, line 96: dito

AshesITR (changes requested):
Looks mostly good to me.

.dev/compare_branches.R
MichaelChirico: done

.dev/compare_branches.R
MichaelChirico: agree

AshesITR (approved):
LGTM, thanks for polishing. This will be very useful.

PR #767, opened 2021-02-19
Extend branch comparison script

MichaelChirico:
Two main new features in this PR:

 1. Copy the repo to a temp dir & run the script from there. That way you can continue editing while running, as well as running multiple comparisons at once
 2. Given how common it is to skip a package due to installation reasons, etc, move to a flow where we try to accumulate `sample_size` successful lints, as opposed `sample_size` total iterations, some percentage of which are just skipped

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -27,7 +27,10 @@ invisible(file.copy(".", temp_repo, recursive = TRUE))
 message("Executing from copy of repo at ", temp_repo)
 old_wd <- setwd(temp_repo)
 if (!interactive()) {
-  .Last <- function() { unlink(temp_repo, recursive = TRUE) }
+  .Last <- function() {
+    setwd(old_wd)
+    unlink(temp_repo, recursive = TRUE)
+  }
 }
 
 param_list <- list(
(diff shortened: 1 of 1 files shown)


.dev/compare_branches.R
```diff
@@ -304,7 +304,17 @@ if (is_branch) {
 } else {
   message("Comparing PR#", pr, " to master")
 }
-message("Comparing output of lint_dir run for the following packages: ", toString(basename(packages)))
+if (length(packages) > 50L) {
```
AshesITR, line 307: wdyt about using `glue::glue_collapse()` instead?

```
> glue::glue_collapse(letters, ", ", last = " and ", width = 10)
a, b, c...
> glue::glue_collapse(letters, ", ", last = " and ", width = Inf)
a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y and z
```

MichaelChirico: I think probably overkill for a verbose message in an internal script like this... `toString` also has a `width` argument FWIW

AshesITR (approved):
LGTM

PR #776, opened 2021-03-11
lint_package returns early when no package is found

MichaelChirico:
On HEAD this gives me a funny error:

```
lintr::lint_package("/")
# Error in paths[rel_path] <- file.path(root, paths[rel_path]) : 
#   replacement has length zero
```

Should we issue a warning here?

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -254,6 +254,12 @@ run_lints <- function(pkgs, linter, check_depends, warn = TRUE) {
 }
 
 run_on <- function(what, pkgs, linter_name, ...) {
+  t0 <- Sys.time()
+  on.exit({
+    t1 <- Sys.time()
+    message("Completed on ", what, " in ", format(difftime(t1, t0, units = "mins")))
+  })
+
   # safe to use force=TRUE because we're in temp_repo
   switch(
     what,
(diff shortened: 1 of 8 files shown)


AshesITR:
Shouldn't this be an error condition?

MichaelChirico:
An error seems harsh, though maybe the return should just be NULL?

Ran into this trying to create a test for #702  -- easiest test is to make sure a `DESCRIPTION` directory is skipped and then no package is found at all.

AshesITR:
I would return NULL for `find_package` but `lint_package()` run non-interactively IMO should definitely error if there was no package, no?

If not, a warning ("No package was found at '/path'."?) with `NULL` instead of an empty `lints()` seems more correct.
Maybe even something like `NA` such that `length(lint_package())` will be different from the lint-free state.

MichaelChirico:
I'm still happy with warn + return `NULL`. Don't really like `NA` -- the `warning` should be enough to distinguish from lint-free right?

Also it's `NULL` whereas lint-free is `list()`.

AshesITR (changes requested):
LGTM. Please add a NEWS bullet.

AshesITR (approved):

PR #777, opened 2021-03-11
include execution timing in branch comparison

MichaelChirico:
Not an ideal benchmark, but with enough packages included this should give an idea of whether the linter is performing wildly differently between the branches

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -254,6 +254,12 @@ run_lints <- function(pkgs, linter, check_depends, warn = TRUE) {
 }
 
 run_on <- function(what, pkgs, linter_name, ...) {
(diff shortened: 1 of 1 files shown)


AshesITR:
Good idea!
Would it be possible / useful to somehow export these timings in a structured way?
This could be used to quantitavely analyse the timing differences if they are obtained carefully (i.e. on the same idle system)

MichaelChirico:
I think properly benchmarking the two branches is in scope for the script but would require a much bigger PR... filed this quickly as a first pass to give a rough idea (I don't expect it to work for, say, <30% performance difference, but something like 2x/3x should show up)

AshesITR (approved):
I agree. Let's add a follow-up issue for that.

Issue #752, opened 2021-02-13
Consistent handling of Encoding


AshesITR:
Currently, we use `readLines()` when linting, which assumes the system encoding when parsing files.

This results in uncovered lines w.r.t. fixing column numbers of UTF-8 Multibyte characters (at least on Windows, cf. #739) and in bugs when the file encoding doesn't match the system encoding (cf. #541)

lintr should

 * get a new setting `encoding` with a smart default (see below) overridable in the usual ways.
 * respect this setting in `get_source_expressions()`
 * Update the tests to force `UTF-8` encoding where necessary

The smart default for encoding should pick the first valid option of this list:

1. The `Encoding` field in a package `DESCRIPTION`
2. The `Encoding` field from the `.Rproj` file in a project directory
3. `UTF-8`

This default is system-independent and thus should be safe to use.

PR #782, opened 2021-03-16
Consistent handling of Encodings

### LINKED ISSUE #541: Issue using nchar for non-UTF-8 source files
Found in the `spatialprobit` package:

```
tmp <- tempfile()
download.file('https://raw.githubusercontent.com/cran/spatialprobit/master/R/metropolis-hastings-rho.R', tmp)
lintr::lint(tmp)
```

Results in a string of warnings/errors:

```
Error in nchar(content$text, "chars") : 
  invalid multibyte string, element 394
In addition: Warning messages:
1: In readLines(filename) :
  incomplete final line found on '/tmp/RtmptJEp8r/file3fb82b4c3025'
2: In gregexpr("\t", source_file[["lines"]], fixed = TRUE) :
  input string 40 is invalid in this locale
3: In gregexpr("\t", source_file[["lines"]], fixed = TRUE) :
  input string 42 is invalid in this locale
4: In gregexpr("\t", source_file[["lines"]], fixed = TRUE) :
  input string 43 is invalid in this locale
5: In gregexpr("\t", source_file[["lines"]], fixed = TRUE) :+1: 
```

The `gregexpr` warnings can be eliminated by using `useBytes = TRUE`.

I _think_ the `nchar` error can be eliminated with `type='bytes'`

AshesITR:
Implements #752

fixes #541 

Added TODO comments to all places that need changing to not miss anything.
Feel free to add more TODO comments if I missed some place.

 - [x] add TODO markers in all effected places
 - [x] add encoding to settings
 - [x] respect encoding in all places where necessary
 - [x] update documentation
 - [x] write tests for all encoding-specific behaviour with at least one non-UTF-8 and non-system encoding
 - [x] add NEWS bullet


AshesITR:
The test file is GPL 2 licensed. I'll replace the file by a self-made one to avoid license violations of spatialprobit.

.dev/compare_branches.R
```diff
@@ -146,7 +146,11 @@ test_encoding <- function(dir) {
   tryCatch({
     lapply(
       list.files(dir, pattern = "(?i)\\.r(?:md)?$", recursive = TRUE, full.names = TRUE),
-      function(x) nchar(readLines(x, warn = FALSE))
+      function(x) {
+        con <- file(x, encoding = lintr:::find_default_encoding(x))
+        on.exit(close(con))
+        nchar(readLines(con, warn = FALSE))
```
MichaelChirico, line 152: nice! purely out of curiosity, any sense for how much this increases our "hit rate" for parsing packages correctly?

NAMESPACE
```diff
@@ -52,6 +52,7 @@ export(object_name_linter)
 export(object_usage_linter)
 export(open_curly_linter)
 export(paren_brace_linter)
+export(pipe_call_linter)
```
MichaelChirico, line 55: my bad :sweat_smile:

R/exclude.R
```diff
@@ -79,6 +79,12 @@ parse_exclusions <- function(file, exclude = settings$exclude,
 
   exclusions <- list()
 
+  e <- tryCatch(nchar(lines), error = identity)
+  if (inherits(e, "error")) {
+    # Invalid encoding. Don't parse exclusions.
```
MichaelChirico, line 84: maybe a warning here instead?

R/expect_lint.R
```diff
@@ -45,8 +45,10 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
 
   if (is.null(file)) {
     file <- tempfile()
+    con <- base::file(file, encoding = "UTF-8")
```
MichaelChirico, line 48: I usually wouldn't write `base::`

R/get_source_expressions.R
```diff
@@ -77,6 +87,36 @@ get_source_expressions <- function(filename, lines = NULL) {
     # an error that does not use R_ParseErrorMsg
     if (is.na(message_info$line)) {
 
+      if (grepl("invalid multibyte character in parser at line", e$message, fixed = TRUE)) {
```
MichaelChirico, line 90: One way to increase robustness is what we've done in `data.table`:

https://github.com/Rdatatable/data.table/blob/ed72e398df76a0fcfd134a4ad92356690e4210ea/inst/tests/tests.Rraw#L106-L140

Basically, keep a database (maybe populated at `onAttach()`) of error messages as generated by R itself.

That way we don't need to change anything if R updates its own message.

Just a thought, no need to do so in this PR.

R/get_source_expressions.R
```diff
@@ -185,26 +225,31 @@ get_source_expressions <- function(filename, lines = NULL) {
...
-    list(
-      filename = filename,
-      file_lines = source_file$lines,
-      content = source_file$lines,
-      full_parsed_content = parsed_content,
-      full_xml_parsed_content = safe_parse_to_xml(parsed_content),
-      terminal_newline = terminal_newline
+  if (inherits(e, "lint") && !nzchar(e$line)) {
+    # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
+    expressions <- NULL
```
MichaelChirico, line 230: Would `expressions <- list()` be better? in case `typeof()=='list'` is expected downstream somewhere

R/lint.R
```diff
@@ -56,8 +56,10 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
 
   if (inline_data && no_filename) {
     filename <- tempfile()
+    con <- file(filename, open = "w", encoding = settings$encoding)
     on.exit(unlink(filename), add = TRUE)
-    writeLines(text = text, con = filename, sep = "\n")
+    writeLines(text = text, con = con, sep = "\n")
+    close(con)
```
MichaelChirico, line 62: put this in `on.exit()` with `unlink()`?

R/lint.R
```diff
@@ -376,6 +378,23 @@ find_package <- function(path) {
...
+  path <- normalizePath(path, mustWork = FALSE)
+
+  while (!has_rproj(path)) {
+    path <- dirname(path)
+    if (is_root(path)) {
+      return(NULL)
+    }
+  }
+
+  list.files(path = path, pattern = "\\.Rproj$", full.names = TRUE)[1L]
```
MichaelChirico, line 395: Nit: this runs an extra `list.files()` right? would it work for the `length()` check to happen in the `while()` loop instead?

R/settings.R
```diff
@@ -100,6 +105,49 @@ find_config <- function(filename) {
...
+  }
+
+  NULL
+}
+
+get_encoding_from_dcf <- function(file) {
+  if (is.null(file)) return(NULL)
+
+  dcf <- tryCatch(
+    read.dcf(file),
```
MichaelChirico, line 135: I would use `read.dcf(file, "Encoding")` directly here

man/get_source_expressions.Rd
```diff
@@ -52,3 +52,13 @@ newline (as determined by \code{\link[=readLines]{readLines()}} producing a corr
 \description{
 This object is given as input to each linter
 }
+\details{
+The file is read in using the \code{encoding} setting.
+This setting found by taking the first valid result from the following locations
+\enumerate{
+\item The \code{encoding} key from the usual lintr configuration settings.
```
MichaelChirico, line 59: link to `?default_settings` here

tests/testthat/dummy_packages/cp1252/R/cp1252.R
```diff
@@ -0,0 +1,4 @@
+# This file is encoded in Cp-1252
+
+# Comment containing non-ASCII �
```
MichaelChirico, line 3: non-ASCII in test files is OK right?

MichaelChirico (commented):
I'm pretty weak when it comes to encodings, but the API & testing look basically good to me. Good work!

.dev/compare_branches.R
AshesITR: I'd expect basically 100% since all Packages must provide a DESCRIPTION file and
Writing R Extensions says

> If the `DESCRIPTION' file is not entirely in ASCII it should contain an `Encoding' field specifying an encoding. (... also used for R files ...)

R/exclude.R
AshesITR: Not sure, `parse_exclusions` is automatically executed on all linted files, so we'd alert in multiple places since there will always be an encoding related lint in that case anyway.

R/expect_lint.R
AshesITR: `file` is also an argument name, so it's needed here for disambiguation. Otherwise, I agree 😄

R/get_source_expressions.R
AshesITR: Nice idea. That would also allow us to remove the `set_lang()` construct currently in use to force english error messages.

R/lint.R
AshesITR: Had that before, but `con` being blocking means you can't read the written contents before `close(con)` is run.

R/expect_lint.R
MichaelChirico: oh I see

R/exclude.R
MichaelChirico: true. troubles of reading the PR in alphabetical order 😅

R/lint.R
AshesITR: I could refactor `has_rproj()` to `find_rproj_at()` returning the file name if found, turning it into `while(is.null(res <- find_rproj_at(path)) { ... }`.
WDYT?

R/lint.R
MichaelChirico: 👍 please note that in a comment, thanks?

tests/testthat/dummy_packages/cp1252/R/cp1252.R
AshesITR: OK for? devtools::check()` didn't complain about it IINM.

NAMESPACE
AshesITR: I tend to forget `devtools::document()` as well 😆

R/lint.R
MichaelChirico: is.null() or length()? either way sounds like the right idea

man/get_source_expressions.Rd
AshesITR: `?default_settings` needs a little love I think. Currently it contains not much more than "An object of class list of length 12".
Maybe that would be a good place in general to document settings?

R/lint.R
AshesITR: Yeah, moving length() outside and using `head(character(), 1L)` inside seems like the most elegant solution. Thanks for your suggestion.

R/settings.R
AshesITR: Done. Thanks!

.dev/compare_branches.R
MichaelChirico: I did think I found a package that failed on current HEAD but didn't have an Encoding listed. I might be misremembering. No matter :)

man/get_source_expressions.Rd
MichaelChirico: I picked `?default_settings` kind of at random, more that "from the usual lintr configuration settings" won't be helpful if this is the first help page a user sees -- some link to help about what we mean there is needed.

If we don't have such a page yet then yes, let's use `?default_settings` :)

MichaelChirico (approved):

PR #789, opened 2021-03-27
[WIP] Change regex in addin_lint function to also remove comments

### LINKED ISSUE #788: addin_lint() does not handle .lintr file with comments and newlines correctly
The `Lint current file` addin option in RStudio gives an error when used with [the default `.lintr` file in this repository](https://github.com/jimhester/lintr/blob/master/.lintr):  

```
 Error in parse(text = config_linters) : 
  <text>:2:0: unexpected end of input
1: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)line_length_linter = line_length_linter(120),cyclocomp_linter = cyclocomp_linter(28) # TODO reduce to
   ^ 
```
The newlines are removed, but this results in invalid input for the parsing step.

I would suggest removing the config reading code from  `addin_lint()`. This would be similar to how `addin_lint_package()` just calls `lintr::lint_package()`. 

Removing the config code seems to work, at least when I try it locally.

huisman:
Fixes #788

This PR updates the regex in the `addin_lint()` function to also remove the comments that can be included in the linters section of a .lintr file.
This is a possible fix for #788.

### DIFF
--- a/R/addins.R
+++ b/R/addins.R
@@ -10,7 +10,7 @@ addin_lint <- function() {
     config_linters <- NULL
   } else {
     config <- read.dcf(config_file, all = TRUE)
-    config_linters <- gsub("\n", "", config[["linters"]])
+    config_linters <- gsub("(#.*?)?\n", "", config[["linters"]])
   }
   linters <- if (length(config_linters) == 0) {
     message("No configuration found. Using default linters.")
--- a/R/addins.R
+++ b/R/addins.R
@@ -10,7 +10,7 @@ addin_lint <- function() {
     config_linters <- NULL
   } else {
     config <- read.dcf(config_file, all = TRUE)
-    config_linters <- gsub("(#.*?)?\n", "", config[["linters"]])
+    config_linters <- config[["linters"]]
   }
   linters <- if (length(config_linters) == 0) {
     message("No configuration found. Using default linters.")


AshesITR:
Thanks for submitting this PR!

I think the actual issue is the `gsub("\n", "", ...)` call here https://github.com/jimhester/lintr/blob/be258a58aa61117f21b0eab5057be110bf5d5c77/R/addins.R#L13.
I'm not really comfortable changing the behaviour of `addin_lint()`, e.g. the proposed change removes the message if no `.lintr`config was found.

Could you try fixing the parsing problem instead?

huisman:
I think I've fixed the parsing problem by changing the regex to also replace/remove comments. 

Let me know what you think.

(I've updated the title and description of this PR, I hope that is okay.)

huisman:
Just realised that this regex will not work as expected when a #-symbol is part of a parameter. 
`linters: with_defaults( # The following TODOs are part of an effort to have {lintr} lint-free (#584)
   random_linter = random_linter(parameter="#")
 )`

R/addins.R
```diff
@@ -10,7 +10,7 @@ addin_lint <- function() {
     config_linters <- NULL
   } else {
     config <- read.dcf(config_file, all = TRUE)
-    config_linters <- gsub("\n", "", config[["linters"]])
+    config_linters <- gsub("(#.*?)?\n", "", config[["linters"]])
```
AshesITR, line 13: Do we need to `gsub()` at all?
The new regex will, as you correctly pointed out, break if `#` is contained in strings or symbol names.

AshesITR (changes requested):

R/addins.R
huisman: No, I've done a couple of tests and I don't think the `gsub()` is needed, so I removed it in a new commit.

Until now I thought the whole reading/parsing of the config file in this function wasn't necessary, but I found one  'scenario' for which the parsing needs to be done here. Currently, if you have a `.lintr` file without a linters section, you get the "No configuration found. Using default linters." message. To be able to show the message, the config needs to be parsed here.

AshesITR (approved):
Looks good to me now. Many thanks for fixing 😃

PR #795, opened 2021-04-29
Ignore load error in `load_cache`

### LINKED ISSUE #794: Corrupted cache file
It occasionally happens that the process is terminated when it is writing a cache file, and then the cache file is corrupted. However, if `lint(cache=TRUE)` is used, then it will stop on error reading the file and do not deliver results.

Since a cache file is only intended for better performance, it is not required to be working. I guess we should ignore the error reading that cache file and proceed as if it does not exist and let it be overwritten.

renkun-ken:
Closes #794

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -32,7 +32,7 @@ load_cache <- function(file, path = NULL) {
 
   file <- get_cache_file_path(file, path)
   if (file.exists(file)) {
-    load(file = file, envir = env)
+    tryCatch(load(file = file, envir = env), error = NULL)
   } # else nothing to do for source file that has no cache
 
   env


R/cache.R
```diff
@@ -32,7 +32,7 @@ load_cache <- function(file, path = NULL) {
 
   file <- get_cache_file_path(file, path)
   if (file.exists(file)) {
-    load(file = file, envir = env)
+    tryCatch(load(file = file, envir = env), error = NULL)
```
AshesITR, line 35: I think we don't need to stop on error here, but a warning would still be warranted.

AshesITR (changes requested):
I think we don't need to stop on error here, but a warning would still be warranted.

R/cache.R
renkun-ken: Sure. Turning the error into a warning looks reasonable.

AshesITR:
How would the warning look in a corrupted cache setting?
Is the error message clear for an unsuspecting user?
Otherwise, maybe something like `warning("Could not load cache file '", file, "':\n", e$message)` is a good idea.

renkun-ken:
Yes, it definitely looks better.

AshesITR (approved):
LGTM, thanks!

AshesITR:
Is load() of an empty file valid? That might be a way to get 100% coverage.

renkun-ken:
```r
> load("/dev/null")                                                                                                               
Error in load("/dev/null") : empty (zero-byte) input file
```

AshesITR:
> ```r
> > load("/dev/null")                                                                                                               
> Error in load("/dev/null") : empty (zero-byte) input file
> ```

Looks good, but we should avoid using `/dev/null` because it fails on windows. `file.create()` will create an empty file under any OS.

renkun-ken:
I added some test cases to check if it works with empty file and invalid file.

AshesITR (approved):

PR #801, opened 2021-05-19
New pipe_call_linter enforcing calls (not symbols) in magrittr pipes

MichaelChirico:
This enforces the following section of the tidyverse style guide:

https://style.tidyverse.org/pipes.html#no-arguments

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -78,6 +78,7 @@ Collate:
     'open_curly_linter.R'
     'paren_brace_linter.R'
     'path_linters.R'
+    'pipe_call_linter.R'
     'pipe_continuation_linter.R'
     'semicolon_terminator_linter.R'
     'seq_linter.R'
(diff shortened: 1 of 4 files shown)


AshesITR:
Looks good so far. Can we also add a negative test for

```
x %>% { foo }
```

just to be sure symbols are allowed in curly braces?

MichaelChirico:
does that code actually do anything?

AshesITR:
This minimal code returns foo from the chain.
However, if this is correctly not linted, more complex code blocks should also be covered - e.g.

```
x %>% {
  tmp <- .
  bla <- foo %>% unrelated_stuff(tmp)
  my_combination_fun(tmp, bla)
}
```

MichaelChirico:
that looks more reasonable to test as a minimal usage example. I'll test &
add it tomorrow.

MichaelChirico:
Updated now. Thanks!

AshesITR (approved):

PR #813, opened 2021-06-09
fix test warnings: use pipe_call_linter()

russHyde:
fix https://github.com/jimhester/lintr/issues/812 (the lintr-specific warnings at least)

Formerly, tests for the pipe-call linter used unevaluated function name
`pipe_call_linter` and threw warnings when running tests.

These have all been changed to use `pipe_call_linter()`


AshesITR (approved):
Thanks!

PR #816, opened 2021-06-26
Fix #815 - "error: unexpected input" from unicode character token

### LINKED ISSUE #815: lintr fails when parser generates "error: unexpected input" from unicode character token
Reproducible example:
```r
> lintr::lint(text="x <- 2 × 3")
# Error in nchar(content$text, "chars") : 
#   invalid multibyte string, element 7
```

The problem seems to be that, when the R parser finds an unexpected token that starts with or is a multibyte Unicode character, it halts the parse (fine) and returns _just the first byte of that character_ as the text for the offending token (bad). The first byte of a multibyte UTF-8 character alone is an invalid character and triggers the error.

I'm still investigating this bug. The source seems to be at least in `utils::getParseData()` or deeper, depending on what is the expected behavior. Anyway, whether or not there is an issue in base functions, I'll submit a pull request with a workaround tomorrow.

leogama:
Fixes fail obtained when a syntax error interrupts the parsing at a
Unicode character and, as a consequence, the parsed content returned
by `getParseData(source_file)` includes a truncated UTF-8 multibyte
representation of that character (which is invalid). See the GitHub
issue discussion (#815) for details.

The proposed solution merely "coerces" the `getParseData`'s `text`
column to what it is supposed to be according to its documentation.

An alternative would be to actually fix the table before use:
```r
content[!content$terminal, "text"] <- ""
```

### DIFF
--- a/tests/testthat/test-get_source_expressions.R
+++ b/tests/testthat/test-get_source_expressions.R
@@ -79,3 +79,12 @@ test_that("Multi-byte characters correct columns", {
     expect_equal(pc[[1L]]$col1[4L], pc[[1L]]$col1[2L] + 4L)
   })
 })
+
+test_that("Multi-byte character truncated by parser is ignored", {
+  tmp <- tempfile()
+  on.exit(unlink(tmp), add = TRUE)
+  writeLines("y <- x \U2013 42", tmp)  # Unicode en-dash
+  content <- expect_error(get_source_expressions(tmp), NA)
+  expect_identical(content$error$message, "unexpected input")
+  expect_identical(content$error$column_number, 8L)
+})
--- a/tests/testthat/test-get_source_expressions.R
+++ b/tests/testthat/test-get_source_expressions.R
@@ -85,6 +85,6 @@ test_that("Multi-byte character truncated by parser is ignored", {
   on.exit(unlink(tmp), add = TRUE)
   writeLines("y <- x \U2013 42", tmp)  # Unicode en-dash
   content <- expect_error(get_source_expressions(tmp), NA)
-  expect_identical(content$error$message, "unexpected input")
-  expect_identical(content$error$column_number, 8L)
+  expect_equal(content$error$message, "unexpected input")
+  expect_equal(content$error$column_number, 8L)
 })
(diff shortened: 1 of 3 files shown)


AshesITR:
Thanks for providing a fix.
Can you add a regression test for this?
I think your MWE is perfectly fine for that.

leogama:
Done. I didn't know exactly how to test that a specific error _doesn't_ occur, but it seems extensive enough like this.

Note: I had to force-push because, when I ran the unit test with `devtools::check()`, it modified some files (`NAMESPACE` and `man/linters.Rd`) and I committed them by mistake.

MichaelChirico:
>  I didn't know exactly how to test that a specific error doesn't occur

`expect_error(expr, NA)` is exactly the right approach IINM

> it modified some files (NAMESPACE and man/linters.Rd) and I committed them by mistake.

FYI that shouldn't be an issue, just `rm` the files in another commit and push to the same branch (I see you are on `master` on your fork, even on a fork you should try putting PRs in branches). When we do Squash+Merge here into main, that temporary issue will be resolved.

tests/testthat/test-get_source_expressions.R
```diff
@@ -79,3 +79,12 @@ test_that("Multi-byte characters correct columns", {
...
   })
 })
+
+test_that("Multi-byte character truncated by parser is ignored", {
+  tmp <- tempfile()
+  on.exit(unlink(tmp), add = TRUE)
+  writeLines("y <- x \U2013 42", tmp)  # Unicode en-dash
+  content <- expect_error(get_source_expressions(tmp), NA)
+  expect_equal(content$error$message, "unexpected input")
+  expect_equal(content$error$column_number, 8L)
```
AshesITR, line 89: Not sure if possible, but have you tried using `with_content_to_parse()` used in the other tests?

leogama: I took a look at it, but it seemed that the `with_content_to_parse()` itself would fail with the bug. And as it doesn't use the `expect` family of functions internally, the error would not be well formatted. I didn't verify it though.

leogama: Did some experiments with the tests just now. The error generated by `with_content_to_parse()` ends well formatted anyway by `testthat`. So in principle we could use it, but it doesn't expose the `error` item from `get_source_expressions()` result. I'm sending an alternative version of the tests with a modification to that helper function.

AshesITR:
Looks good now. Please add a `NEWS` bullet, then we can merge.

AshesITR (approved):

leogama:
Thank you for your time and attention. 😉

PR #818, opened 2021-06-29
R cmd gha update

dragosmg:
since the release of R 4.1 the workflow matrix is slightly out-of-date (not checking against R 3.6, but keeping the check for R 3.3). this PR should re-align the combinations with the latest minor R release.


AshesITR:
Thanks!

AshesITR (approved):

PR #820, opened 2021-06-30
Remove fix_column_numbers helper

### LINKED ISSUE #819: get_source_expression() column correction
When incorporating the fix #816 into my Encoding PR #782, I noticed test failures locally (they didn't pop up in CI :().

Digging deeper made me wonder if `fix_column_numbers()` is actually needed anymore:

R 4.1.0 gives basically identical parse data for text with and without multibyte characters:

```r
getParseData(parse(text = "`x` <- 1"))
#  line1 col1 line2 col2 id parent       token terminal text
#7     1    1     1    8  7      0        expr    FALSE     
#1     1    1     1    3  1      3      SYMBOL     TRUE  `x`
#3     1    1     1    3  3      7        expr    FALSE     
#2     1    5     1    6  2      7 LEFT_ASSIGN     TRUE   <-
#4     1    8     1    8  4      5   NUM_CONST     TRUE    1
#5     1    8     1    8  5      7        expr    FALSE  
   
getParseData(parse(text = "`\U2020` <- 1"))
#  line1 col1 line2 col2 id parent       token terminal text
#7     1    1     1    8  7      0        expr    FALSE     
#1     1    1     1    3  1      3      SYMBOL     TRUE  `†`
#3     1    1     1    3  3      7        expr    FALSE     
#2     1    5     1    6  2      7 LEFT_ASSIGN     TRUE   <-
#4     1    8     1    8  4      5   NUM_CONST     TRUE    1
#5     1    8     1    8  5      7        expr    FALSE     
```

Tested on Linux and Windows.

Can anyone verify from what R version these Corrections are no longer needed?
@MichaelChirico maybe?
@jimhester The original code is from 2014, written by you. Maybe you remember some context?

AshesITR:
The fix code was necessary for R < 3.4.1
Since we don't test on R 3.3 anymore, the old code should be removed.

R 3.4.1 fixed the bug in [PR#17254](https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17254)

fixes #819

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -252,7 +252,7 @@ get_source_file <- function(source_file, error = identity) {
     assign("e", e,  envir = parent.frame())
   }
 
-  fix_eq_assigns(fix_column_numbers(fix_tab_indentations(source_file)))
+  fix_eq_assigns(fix_tab_indentations(source_file))
 }
 
 find_line_fun <- function(content) {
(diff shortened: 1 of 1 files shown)


AshesITR:
I'm stumped what happens here. There seems to be a change in #782 that makes `fix_column_numbers` obsolete.
Closing this PR now.

PR #834, opened 2021-07-09
polish NEWS

AshesITR:
I've tried sorting all NEWS bullets and commenting a bit on the breaking changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,73 +1,105 @@
 # lintr (development version)
 
-* Updated R CMD GitHub Actions workflow to check for R 3.6 on Ubuntu, instead of R 3.3, and for R 4.0 on Windows, instead of R 3.6 (#803, @ dragosmg)
-* Added a secondary, more restrictive lint workflow - `lint-changed-files` - for newly written / modified code (#641, @dragosmg) 
-* Switched CI from Travis to GitHub Actions, using the full tidyverse recommended R CMD check. Code coverage and linting 
-  are implemented using separate GitHub Actions workflows (#572, @dragosmg)
-* `save_cache` will now recursively create the cache directory; this avoids errors that could arise if any parent 
-  directories do not exist (#60, @dankessler).
-* `extract_r_source` handles Rmd containing unevaluated code blocks with named
-  format specifiers (#472, @russHyde)
-* New style SNAKE_CASE for `object_name_linter()` (#494, @AshesITR)
-* RStudio source markers are cleared when there are no lints (#520, @AshesITR)
-* New `assignment_spaces()` lintr. (#538, @f-ritter)
-* `seq_linter()`'s lint message is clearer about the reason for linting. (#522, @michaelchirico)
-* New `missing_package_linter()` (#536, #547, @renkun-ken)
-* New `namespace_linter()` (#548, #551, @renkun-ken)
-* Fix possible error on invalid XML produced by xmlparsedata (#559, #560, @renkun-ken)
-* Fix handling zero-length variable name error (#566, #567, @renkun-ken)
-* New `missing_argument_linter()` (#563, #565, @renkun-ken)
-* New `sprintf_linter()` (#544, #578, #624, #625, @renkun-ken, @AshesITR)
+
+  
+# lintr 3.0.0
+
+## Breaking changes
+
 * Exclusions specified in the `.lintr` file are now relative to the location of that file 
   and support excluding entire directories (#158, #438, @AshesITR)
-* `lint_dir()` excludes the `renv` and `packrat` directories by default (#697, @AshesITR)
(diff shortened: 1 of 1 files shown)


MichaelChirico:
@AshesITR do you want to merge this into the current NEWS? i guess you have more context about the structure you had in mind here. i have a few ideas of what to do with the NEWS but let's see how it looks after your merge first.

AshesITR:
Yes, I'll update this PR

AshesITR:
@MichaelChirico PTAL. Especially if anyone else deserves credit for some of the issues subsumed under the google linters bullet.
Moved special thanks to @bersbersbers to the Internals section - not sure where to put, but definitely worth mentioning I think.

The structure is as follows:

 * Breaking changes: Changes which can cause errors in user code
 * Depreactions: Changes which should not cause errors, but may cause deprecation warnings in user code
 * Changes to defaults: Changes which may cause new lints to appear
 * New features: new linters / functions or new arguments to existing linters
 * Improvements and bug fixes: Other changes, such as fixed errors, false-positives, etc.
 * Internals: Changes internal to lintr with no (expected) implications for users.

If the summary is good, the next step is to sort all bullets alphabetically by the affected function name so changes to the same function are nearby in NEWS.

eitsupi:
It does not seem be listed the new `use_lintr` function? (#1120)

AshesITR:
Thanks for pointing this out, I added a bullet (referencing the issue #902 instead of its PR).

MichaelChirico:
I think the new structure is basically right but I think there's still a lot of copy work we can do to improve the NEWS.

I'll start filing PRs based against this branch for clarity / to minimize the diff noise & we can gradually merge them into here before merging to main to minimize the volume of commits hitting main.

NEWS.md
```diff
@@ -1,178 +1,246 @@
 # lintr (development version)
 
-## Potentially breaking changes and pending deprecations
+## Breaking changes
+
+* Exclusions specified in the `.lintr` file are now relative to the location of that file 
+  and support excluding entire directories (#158, #438, @AshesITR)
+* Consistent access to linters through a function call, even for linters without parameters
```
MichaelChirico, line 7: Slightly confused about what goes in breaking vs. deprecated

earlier when I read, I thought what should go here are backwards-incompatible changes, because we mention some function deprecations (like `absolute_paths_linter`) but not others (like `semicolon_terminator_linter`), and the difference is deprecations here are not back-compatible.

but this change (consistent access...) is backwards-compatible, in that we warn for providing linters as symbols, for now.

maybe the sections should be combined, and one bullet point is linter deprecations, with the removals coming first?

AshesITR:
@MichaelChirico can you add a NEWS bullet regarding the change to `open_curly_linter()`? #1221

MichaelChirico (approved):

MichaelChirico:
@AshesITR seems there's no way for me to request a review by you since you're the original PR author :|

Anyway, my pass-through is done, PTAL. Still room for improvement but we're getting closer.

NEWS.md
```diff
@@ -1,186 +1,296 @@
...
+   + correctly detect assignment generics like `names<-.class_name` (#843, @jonkeane)
+   + added new styles `"symbols"` and `"SNAKE_CASE"` (#494, #495, #615, #670, @michaelchirico and @AshesITR)
+* `object_usage_linter()`
+   + detect global variables if there are top-level dollar-assignments (#666, #709, @AshesITR)
+   + report usage warnings spanning multiple lines (#507, @AshesITR)
+   + detect usages inside `glue::glue()` constructs (#942, @AshesITR)
+   + detect within functions assigned with `=` instead of `<-` (#1081, @michaelchirico)
+   + detect functions exported by packages that are explicitly attached using `library()` or
+     `require()` calls (#1127, @AshesITR)
+   + improved location information in some cases (#1285, @AshesITR)
```
MichaelChirico, line 125: @AshesITR could you expand on this? someone reading the NEWS reads this and thinks "well, what cases?"

it's also not clear from the #1285 description -- someone reading this bullet would have to read the full PR diff to get an idea of what's changed exactly. better to give a specific example.

MichaelChirico:
@AshesITR if you're ok w the overall structure here let's merge so we can back to the habit of adding bullets with new PRs

AshesITR:
SGTM, let's first add the remaining bullets, though.

MichaelChirico:
Looking at the PR mentions I only see #1323 merged. Adding that item now

AshesITR:
Added a TODO to the unmerged PRs so we don't forget.

MichaelChirico (approved):

MichaelChirico:
finally done 🎉

hope we don't need such an intensive copy-edit process for future releases 😅

PR #841, opened 2021-07-23
Read namespace files in packages that aren't in directories with their name

jonkeane:
For some packages, the directory they are housed in is not simply their name (e.g. [Arrow's r package](https://github.com/apache/arrow/tree/master/r) is housed under `arrow/r`)

This change fixes that (as well as adds a test that it is fixed)


MichaelChirico:
please delete the .DS file

jonkeane:
Oops, sorry about that. Should be gone now

R/namespace.R
```diff
@@ -2,7 +2,7 @@
 namespace_imports <- function(path = find_package()) {
   imports <- tryCatch({
     pkg_name <- suppressWarnings(pkg_name(path))
-    data <- parseNamespaceFile(package = pkg_name, package.lib = file.path(path, ".."))
+    data <- parseNamespaceFile(package = "", package.lib = path)
```
AshesITR, line 5: Internally this will call `file.path` with a zero-length string, using `path//NAMESPACE` as the path instead of `path/NAMESPACE`. Using `package = basename(path)` would be more robust.

tests/testthat/dummy_packages/clean/R/clean_generics.R
```diff
@@ -25,3 +25,9 @@ drink_me.list <- function(x, ...) {
 drink_me.data.frame <- function(x, ...) {
   NULL
 }
+
+#' head on my_s3_object
+#' @export
+head.my_s3_object <- function(x, ...) {
+  NULL
+}
```
AshesITR, line 33: This seems unrelated to the PR. Please remove it.

jonkeane: Yup, sorry about that — I've removed it.

AshesITR: Wouldn't utils always be loaded and thus `head` correctly identified as a generic regardless of the NAMESPACE because utils is a base package?

MichaelChirico: it's possible to start R without utils IINM. e.g. Rscript doesn't load methods, I believe there's an option to control the other default packages as well

MichaelChirico: that's why R CMD check warns about not putting head (and other functions from default packages) in the NAMESPACE IIRC

AshesITR (approved):

jonkeane:
Is there anything else I should do for this PR?

AshesITR:
Sorry, I wanted to manually test this PR before merging to master, but forgot to do the merge. Thanks a lot!

PR #843, opened 2021-07-24
Allow assignment generics on custom classes object name lintr

jonkeane:
For assignment methods like `names<-.class_name` I'm seeing linting errors:

```
> lintr::lint(text = "`names<-.my_custom_class` <- function() 2", linters = lintr::object_name_linter())
<text>:1:1: style: Variable and function name style should be snake_case or symbols.
`names<-.my_custom_class` <- function() 2
```

This PR resolves this (by looking for `<-.` and replacing it with `.`, in `strip_names()` since the `<-` are stripped from the generics before they are used in the matching regex in the object name lintr. 

Alternatively, I could try keeping `<-` in both the generic list and the function names that are being linted if we think that's a better approach.


AshesITR:
Thanks for reporting this bug.
The actual problem here seems to be that `strip_names` strips too much.

    `names<-`

is in `.S3PrimitiveGenerics` and thus should be considered to be a generic in and of itself. I'd prefer to fix the detection rather than automatically assume all `<-`-functions are generic if and only if their non-setter versions are.

jonkeane:
I'm happy to try removing `strip_names()` from the generics detection and seeing if the tests pass / using the `compare_branches.R` script to check if there are any regressions.

AshesITR:
Hi @jonkeane, sorry for the late reply.
I've just locally tested removing `strip_names()` from the generics call and didn't notice any adverse side-effects.
So this really seems like the best way to fix the issue.

jonkeane:
Is there anything you need from me to get this finished?

AshesITR:
Just a few tiny things still need to be done:

 1. revert the change to `strip_names()`
 2. remove the `strip_names()` call from https://github.com/r-lib/lintr/blob/be7a7f2827bbabdc3e80c77929b9986244b21fe9/R/object_name_linters.R#L65
 3. Add a bullet the NEWS, crediting yourself and citing this PR number (since there is no associated issue IINM)

jonkeane:
Thanks for the follow up — I've done all three of those. I put the news at the end of the current list, but happy to move it elsewhere if that's perferable.

NEWS.md
```diff
@@ -130,6 +130,7 @@ function calls. (#850, #851, @renkun-ken)
 * * `object_length_linter()` correctly detects generics and only counts the implementation class towards the length. 
   This prevents false positive lints in the case of long generic names, e.g. 
   `very_very_very_long_generic_name.short_class` no longer produces a lint (#871, @AshesITR)
+* Assignment generics on custom classes (#843, @jonkeane)
```
AshesITR, line 133: Maybe ` * `object_name_linter()` now correctly detects assignment generics (#843, @jonkeane)`?

AshesITR (changes requested):
Thank you very much, almost ready to merge.

NEWS.md
jonkeane, line 133:
```suggestion
* `object_name_linter()` now correctly detects assignment generics (#843, @jonkeane)
```

jonkeane:
Done, and thank you, again!

PR #844, opened 2021-07-27
fix(undesirable_function_linter.R): no longer lints library and require

### LINKED ISSUE #814: New match_symbol_usage argument to undesirable_function_linter
We have a function `foo()` that we'd like to lint. However, `foo()` comes from the package `foo`, which is still OK, so we have the following unfortunate situation:

```
library(foo) # triggers a lint, but shouldn't
foo::foo()  # triggers a lint, as it should
```

The issue is that `undesirable_function_linter` by default also looks for `SYMBOL` usage of input functions, which is good for catching usage within `sapply`, `purrr::map`, etc.

So we can add a new argument to toggle whether `SYMBOL` is included, and run linting with

```
with_defaults(
  undesirable_function_linter(c("return")), # match_symbol_usage = TRUE by default
  undesirable_function_linter(c("foo"), match_symbol_usage = FALSE)
)
```

Happy to file the PR if that sounds like a good idea.

kpagacz:
* the undesirable_function_linter no longer lints library and require calls attaching a package with an undesired name, e.g. library(foo)

Closes #814

I went with the option of treating the library and require calls as false positives. As suggested by the OP of the issue, I have refactored undesirable_function_linter to use XML parsing instead of regexes. Let me know what you think.


R/undesirable_function_linter.R
```diff
@@ -9,38 +9,49 @@
...
-            ranges = list(c(start_col_num, end_col_num))
-          )
+    xpath <- paste0(
+      "//",
+      tokens,
+      "[",
+      paste("text()='", names(fun), "'", sep = "", collapse = " or "),
+      " and ",
+      "count(parent::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text()='library' or text()='require'])=0",
+      "]",
```
AshesITR, line 28: There seems to be a precedence error here.
Try linting `library(foo)` with c("foo", "bar") as undesired in that order to confirm.

Use parentheses around the `paste()` call to ensure correct precedence.

AshesITR (commented):
Thanks for the PR!

AshesITR (approved):

PR #845, opened 2021-08-02
Add path_prefix option to lint_package for GitHub Actions annotations

jonkeane:
In order to get GitHub annotations to show up on a PR inline in the code, the annotation path must be relative to the repository root rather than relative to the package root. For most R packages, the package root is the same as the GitHub root, but for packages that use a subdirectory (e.g. https://github.com/apache/arrow in the r/ subdirectory) the annotations don't display without adding a prefix like this.

This PR doesn't change the default case, but adds an argument to `lint_package()` that allows someone to specify the correct prefix.

### DIFF
--- a/R/lint.R
+++ b/R/lint.R
@@ -273,7 +273,8 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
 #' }
 #' @export
 lint_package <- function(path = ".", relative_path = TRUE, ...,
-                         exclusions = list("R/RcppExports.R"), parse_settings = TRUE) {
+                         exclusions = list("R/RcppExports.R"), parse_settings = TRUE,
+                         path_prefix = "") {
   pkg_path <- find_package(path)
 
   if (is.null(pkg_path)) {
@@ -300,6 +301,9 @@ lint_package <- function(path = ".", relative_path = TRUE, ...,
       lints,
       function(x) {
         x$filename <- re_substitutes(x$filename, rex(path, one_of("/", "\\")), "")
+        if (nzchar(path_prefix)) {
+          x$filename <- file.path(path_prefix, x$filename)
+        }
         x
       }
     )
--- a/R/lint.R
+++ b/R/lint.R
@@ -259,6 +259,8 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
 #' Apply one or more linters to all of the R files in a package.
 #' @param path the path to the base directory of the package, if \code{NULL},
 #' it will be searched in the parent directories of the current directory.
+#' @param path_prefix a prefix to add to the path of any file found, useful for
+#' packages in subdirectories on github actions.
 #' @inherit lint_file return
 #' @inheritParams lint_dir
 #' @examples
--- a/man/lint_package.Rd
+++ b/man/lint_package.Rd
@@ -9,7 +9,8 @@ lint_package(
   relative_path = TRUE,
   ...,
   exclusions = list("R/RcppExports.R"),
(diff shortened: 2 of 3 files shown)


R/lint.R
```diff
@@ -259,6 +259,8 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
 #' Apply one or more linters to all of the R files in a package.
 #' @param path the path to the base directory of the package, if \code{NULL},
 #' it will be searched in the parent directories of the current directory.
+#' @param path_prefix a prefix to add to the path of any file found, useful for
+#' packages in subdirectories on github actions.
```
MichaelChirico, line 263: whats the difference of `path_prefix` vs just `path`?

R/lint.R
```diff
@@ -259,6 +259,8 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
 #' Apply one or more linters to all of the R files in a package.
 #' @param path the path to the base directory of the package, if \code{NULL},
 #' it will be searched in the parent directories of the current directory.
+#' @param path_prefix a prefix to add to the path of any file found, useful for
```
MichaelChirico, line 262: I see the idea of what you're after & it makes sense, I'm less confident this is the best API for it. Maybe use `project_dir` instead, and then `path` is given relative to that?

cc @russHyde and @AshesITR for an API decision

jonkeane: Thanks for the feedback, I would be happy to do that if that's the direction y'all would like.

For what it's worth, when we run these in our CI, we actually run them from the subdirectory, so having `path_prefix` / `project_dir` only add on to the output of the GHA annotations is the behavior we want (of course we could change where we lint from easily, so that's not a great argument for this API). If we did keep this behavior, we might also consider naming it something like `annotation_path_prefix` to make it's clear it's only about annotations and not where to run the file.

AshesITR: IMO this is not really related to linting per se, but more to the `print.lints` method for GitHub.

jonkeane:
Sorry for the delay on this! Thanks for the comments. I've changed the UX here using an option (similar to what's already used in the `print.lints` method). I also updated the naming to reflect the `project_dir` suggestion.

README.md
```diff
@@ -208,6 +208,8 @@ location, namely in the `.github/workflows` directory of your repository. This f
 pull request with the lints found and they will also be printed as [annotations](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-status-checks#types-of-status-checks-on-github) along side the status check on GitHub.  If you want to disable the commenting you can
 set the environment variable `LINTR_COMMENT_BOT=false`.
 
+If your project is in a subdirectory, and you would like to use GitHub Actions annotations, you can set `options(lintr.github_annotation_project_dir = "path/to/project")` which will add the path specified to the beginning of filenames in the annotations.
```
AshesITR, line 211: I suggest wording it like "... which will make sure the annotations point to the correct files." or something like that, to describe the intention, not the technique. WDYT?

jonkeane: Oh, yes that's a much better phrasing. Thanks

jonkeane:
Ok, I've pushed a few more updates. The path on windows is a bit off: it uses a mixture of `\` and `////`. I tried using `normalizePath()` to standardize these, but it ended up producing absolute paths which won't work on *nix (since the annotation paths AFAIK must be relative to the git checkout). I'm not sure if there is anyone using these annotations on GHA from a windows builder, so this might be ok. Thoughts?

jonkeane:
Is there anything else I can do to get this past the finish line?

PR #848, opened 2021-08-11
Update link to Debian docs page

thisisnic:
I think Debian might have updated their page layout for those docs, as the previous link took me to the right page but not section - this link goes to the specific relevant section that's being referred to.


AshesITR:
Thanks for bringing this to our attention :-)

AshesITR (approved):

PR #851, opened 2021-08-17
Duplicate argument linter

### LINKED ISSUE #850: A new linter checking duplicate arguments
It is almost always a mistake to call a function with duplicate arguments, i.e. arguments of the same name, like the following:

```r
foo(bar1 = 1, bar1 = 2, bar3 = 3)
```

Therefore, in most cases, it might make sense to check all function calls and ensure there are no duplicate arguments.

One use case of duplicate argument is to produce some table that actually needs duplicate headers for demonstration in documents, e.g.

```r
data <- data.frame(
  mean = 1:5,
  sd = 0.5,
  mean = 2:6,
  sd = 1,
  check.names = FALSE
)
```

renkun-ken:
Closes #850 

This PR implements a `duplicate_argument_linter` to check if there are duplicate arguments supplied to function calls.


R/duplicate_argument_linter.R
```diff
@@ -0,0 +1,52 @@
...
+      args_text <- vapply(
+        parse(text = xml2::xml_text(args), keep.source = FALSE),
+        as.character, character(1L)
+      )
+      is_dup_arg <- duplicated(args_text)
+      dup_args <- args[is_dup_arg]
+      if (length(dup_args)) {
+        line1 <- as.integer(xml2::xml_attr(dup_args, "line1"))
+        col1 <- as.integer(xml2::xml_attr(dup_args, "col1"))
+        col2 <- as.integer(xml2::xml_attr(dup_args, "col2"))
```
AshesITR, line 35: Technically,

    tmp <- function(`
    ` = 42) {
    `
    `
    }

is valid R code, so `line2` is not guaranteed to be `line1` for call names.

tests/testthat/test-duplicate_argument_linter.R
```diff
@@ -0,0 +1,65 @@
...
+    list(message = rex("Duplicate arguments in function call.")),
+    duplicate_argument_linter())
+
+  expect_lint(
+    "list(
+      var = 1,
+      var = 2
+    )",
+    NULL,
+    duplicate_argument_linter(except = "list"))
```
AshesITR, line 64: Can we test the `except = ...` branch with anonymous functions, too?
This ensures the `calls_names` are detected properly for calls to anonymous functions.

AshesITR (changes requested):
Looks mostly good. Just a few suggestions to improve robustness.
Don't forget a NEWS bullet.

R/duplicate_argument_linter.R
renkun-ken: Here `line1` is only used for reporting. Do you have any suggestion what to do with `line2`? Does lintr support multi-lined lint result?

tests/testthat/test-duplicate_argument_linter.R
renkun-ken: Yes, exactly.

R/duplicate_argument_linter.R
renkun-ken: The actual range is from `line1, col1` to `line2, col2`.

AshesITR: IIRC we have some places where `col2` is automatically set to the end of the line if `line2 > line1`. I think that would be fine here, too. It's just a crude edge case anyway as noone in their right mind would use newlines in argument names other than for demonstrational purposes.

R/get_source_expressions.R
```diff
@@ -115,6 +115,28 @@ get_source_expressions <- function(filename, lines = NULL) {
             line = ""
           )
         )
+      } else if (grepl("repeated formal argument", e$message, fixed = TRUE)) {
```
AshesITR, line 118: Very nice :)

R/duplicate_argument_linter.R
renkun-ken: `col2` tweaked to line ends. Do we need test case for such "crude edge case"?

AshesITR: Sounds like a good idea to me.

renkun-ken: A crude edge test cases added.

AshesITR (approved):
LGTM

PR #249, opened 2017-06-16
Concatenation linter

fangly:
Hi Jim,
I changed the concatenation_linter to handle the "recursive" option properly, lint NULL passed as an argument, and lint when some atomic types would be coerced to another. I also renamed the function to better reflect its purpose (it has not made it to a release, so no backward compatibility issues exist).
Cheers,
Florent

PR #855, opened 2021-09-08
Improve the unneeded_concatenation_linter message

MichaelChirico:
Hoping to convey best practice with the message here... there are definitely cases where it's hard to know the type/length in advance & NULL is appropriate, but I think it's best to message about what's the best/preferred alternative.


AshesITR (approved):

AshesITR:
#249 provides fine-grained messages for `unneded_concatenation_linter()`, though far behind master.
Maybe worth bringing up to speed if it's not too much work.

PR #857, opened 2021-09-14
Use a slightly altered `isS3stdGeneric()` to detect non-standard generics

### LINKED ISSUE #846: Is `isS3stdGeneric()` the right way to detect S3 generics?
When checking if a function is an S3 generic (after getting it from the `NAMESPACE` file for example), [a call to `isS3stdGeneric()` is used](https://github.com/jimhester/lintr/blob/0397cb777e42ac25fcda8f7245edaadbf56f6fc0/R/namespace.R#L42) to tell if the function is an S3 generic. This works well for standard generics (when the first expression in the body calls `UseMethod()` [1]), but doesn't work so well if there is any sort of pre-processing in the generic (e.g.[`vctrs::vec_ptype_full`](https://github.com/r-lib/vctrs/blob/master/R/ptype-abbr-full.R#L33-L38) which runs some dots checking before calling `UseMethod`)

```
> isS3stdGeneric(vctrs::vec_ptype_full)
[1] FALSE
```

I've looked a bit and haven't yet found a function that will find both standard and non-standard S3 generics. Would it be better / ok to fall back to the prior check of `is.function(fun)`? Or have an option to do that?

[1] - https://stat.ethz.ch/R-manual/R-devel/library/utils/html/isS3stdGen.html

jonkeane:
I followed the path described in #846 and added a test. If it's preferable I'm happy to make that test be more of an integration-style test with one of the test fixture packages in the `tests` directory. I decided to not (yet?) add functionality to scan for `UseMethod()` elsewhere in the function, since that could produce false positives (if it's in an if/else and the other branch doesn't call `UseMethod()`, though that would be a very odd use of `UseMethod()` IMO). I'm happy to add it in if y'all think that's useful (or add a TODO that one could do that if there are requests to). I also still gated this on R > 3.5.0, though if we've implemented this ourselves and aren't depending on `utils`, we could probably use that for any version of R.

I also ran `.dev/compare_branches.R` (after a bit of RTFS to figure out how to call it). I couldn't find any documentation for the right way to interpret the results, but what I did was look at the output and confirm that there were no new lints in the `"branch"` rows that aren't also there in the `"main"` rows. Is there anything else I should do to use that script?

Resolves #846


R/namespace.R
```diff
@@ -37,9 +37,14 @@ imported_s3_generics <- function(ns_imports) {
 
 is_s3_generic <- function(fun) {
   if (getRversion() >= "3.5.0") {
```
AshesITR, line 39: Is the version requirement still necessary? It was introduced as a safeguard against a buggy implementation of `utils::isS3stdGeneric` in R 3.4.0

jonkeane: Yup, I can use this new function for all versions, it looks like there's a test for 3.4.0, but not before that. That's probably fine since 3.3.0 and earlier is quite old at this point.

AshesITR:
@jimhester it appears that `ubuntu-16.04 (release)` is still an expected check for this PR. Can you take a look?
Thanks!

AshesITR (approved):
Looks good to me, thanks!
Feel free to also add a NEWS bullet describing your improvement and referencing this PR.

jonkeane:
Added a bullet to News

jonkeane:
Is there anything else I can do to get this past the finish line? I'm happy to rebase/unconflict NEWS if that helps.

AshesITR:
This PR should be compared using the branch compare script to make sure all changes in linting behavior are intended compared to master.

If you want, you can try running the comparison script found in `.dev`.

jonkeane:
I did run the branch compare script on the first version of the implementation (the only additions after that were NEWS + making it use the new code in all versions of R not just those > 3.5)

> I also ran .dev/compare_branches.R (after a bit of RTFS to figure out how to call it). I couldn't find any documentation for the right way to interpret the results, but what I did was look at the output and confirm that there were no new lints in the "branch" rows that aren't also there in the "main" rows. Is there anything else I should do to use that script?

Was what I ran sufficient? Is there a better way to run or report the results from those? Happy to learn what the appropriate way to run/report those are if I've made a mistake. I can also run it again if that's helpful (though I'm running R 4.1 or 4.0 locally, so the code change shouldn't actually do anything differently).

AshesITR:
It would be nice to see a sample of the lints that disappear from the checked packages in comparison to master to ensure no false negatives occur as well as to verify that no new lints appear (which I think you did).

MichaelChirico:
Thanks @jonkeane for the PR & feedback about `.dev/compare_branches.R`. I wrote up some more documentation for it in #870 -- PTAL & LMK if things are still unclear. Further feedback/improvements to the script welcome.

R/namespace.R
```diff
@@ -36,13 +36,14 @@ imported_s3_generics <- function(ns_imports) {
...
-    # Available in 3.4.0, but bugged there in multiple ways that cause errors
-    # throughout many base functions, e.g. `-`, `as.null.default`, `dontCheck`
-    utils::isS3stdGeneric(fun)
-  } else {
-    is.function(fun) # nocov
-  }
+  # Inspired by `utils::isS3stdGeneric`, though it will detect functions that
+  # have `useMethod()` in places other than the first expression.
+  bdexpr <- body(fun)
+  while (is.call(bdexpr) && bdexpr[[1L]] == "{") bdexpr <- tail(bdexpr, 1L)[[1L]]
```
MichaelChirico, line 42: i guess `tail` isn't the right function here, just use `bdexpr[[length(bdexpr)]]`?

tests/testthat/test-namespace.R
```diff
@@ -0,0 +1,8 @@
+test_that("is_s3_generic", {
+  func <- function(x) {
```
MichaelChirico, line 2: could you add a test case for the error described in #1161, namely for a function where the last expression is namespace-qualified?

MichaelChirico:
As luck would have it this PR would fix #1161! So I've marked it for inclusion in 3.0.0.

@jonkeane will you be able to update the PR in the next few days? If not, I can take over to get it over the finish line. Thanks again!

MichaelChirico (approved):

MichaelChirico:
@AshesITR I don't see a lot of risk here for unexpected regression, so going ahead to merge.

Part of my release to-do list is to do some more extensive compare_branches runs of CRAN vs. current dev, so any concerns should be addressed by that as well.

AshesITR:
Okay with me as long as we do a before/after comparison at some point in time before release.

tests/testthat/test-namespace.R
```diff
@@ -0,0 +1,17 @@
...
+    print(x)
+    UseMethod("func")
+  }
```
AshesITR, line 5: Is there a valid use case for the other way around?
i.e. for
```
func <- function(x) {
  x <- UseMethod("func")
  common_postprocessing(x)
}
```

MichaelChirico: let's try & keep close to the base definition unless there's a compelling reason? i'm also a bit hesitant for this change if it's not being suggested upstream as well

AshesITR: The base definition checks the first call, so would find my example iinm. That's why I'm asking.

MichaelChirico: isn't the call that's found `<-` though, not UseMethod?

AshesITR: Ah, right. I wonder why base R looks for the first call instead of the last, then...

MichaelChirico: I didn't see anything about isS3stdGeneric() on r-devel in a quick search... maybe worth bringing up there.

cc @moodymudskipper who identified the bug fix at root in #1161

PR #859, opened 2021-09-14
replaced ubuntu 16.04 with 18.04 and 20.04

dragosmg:
closes #858 

summary: ubuntu-16.04 is deprecated which results in pipeline failures. this PR solves that by moving to ubuntu 18.04 and 20.04. it also makes use of {pak} to speed up dependency installations.

PR #863, opened 2021-09-20
move httr to Suggests

### LINKED ISSUE #861: Move {httr} to `Suggests:`
Seems like {httr} is only used [once](https://github.com/jimhester/lintr/blob/c2135e2e0dc75ba4b8eba38f868169f26481be64/R/comments.R#L62) to interact with the GitHub API, which I think is not core functionality of {lintr}. The hard-reverse dependency graph could be reduced by 6 packages and two system dependencies (curl and openssl).

Would you consider making {httr} a suggested package only?

``` r
imports <- c("rex", "crayon", "codetools", "cyclocomp", "testthat", "digest", "rstudioapi", "httr", "jsonlite", "knitr", "stats", "utils", "xml2" , "xmlparsedata")
current <- miniCRAN::pkgDep(imports, suggests = FALSE)
#> Warning in miniCRAN::pkgDep(imports, suggests = FALSE): Package not recognized:
#> stats, utils
withotu_httr <- miniCRAN::pkgDep(setdiff(imports, 'httr'), suggests = FALSE)
#> Warning in miniCRAN::pkgDep(setdiff(imports, "httr"), suggests = FALSE): Package
#> not recognized: stats, utils
setdiff(current, withotu_httr)
#> [1] "httr"    "curl"    "mime"    "openssl" "askpass" "sys"
```

<sup>Created on 2021-09-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Background: When running {lintr} on CI/Cd with https://github.com/lorenzwalthert/precommit, we want to bring down install time (plus avoid system dependencies).

MichaelChirico:
Closes #861

Does this warrant a NEWS item?

I first wrote this using `rlang::check_installed()` which is more concise & will suggest installation in interactive sessions, but that's not available before `rlang` v0.4.10 (late 2020), which would make our `rlang` dependency considerably more recent than any of the other packages. So stuck with the tried-and-true direct `requireNamespace()` approach.


AshesITR:
Adding a NEWS bullet seems fine.
Something like "reduced the number of dependencies to improve install time" should be fine and worth mentioning.

MichaelChirico:
For now, I moved all of `codetools`, `cyclocomp`, `httr`, `rstudioapi`, and `testthat` to Suggested status. `rstudioapi` is probably the toughest call there, but exactly for the use case Lorenz mentions, I don't think it's necessary to include for "slim" installations (OTOH, it's pretty lightweight, so not a big deal to include it). We could also work around `crayon` being absent but it's slightly more complicated to untangle that.

R/object_usage_linter.R
```diff
@@ -3,6 +3,9 @@
 #' \code{\link[base]{eval}} on the code, so do not use with untrusted code.
 #' @export
 object_usage_linter <- function() {
+  if (!requireNamespace("codetools", quietly = TRUE)) {
+    stop("'object_usage_linter' depends on the codetools package to work, but it's not installed.")
```
AshesITR, line 7: That's a default linter IINM, so voting for `codetools` to stay `Imports`ed.

MichaelChirico: ditto for `cylocomp` then

R/cyclocomp_linter.R
```diff
@@ -2,9 +2,11 @@
 #'   \code{\link[cyclocomp]{cyclocomp}}.
 #' @param complexity_limit expressions with a cyclomatic complexity higher than
 #'   this are linted, defaults to 25. See \code{\link[cyclocomp]{cyclocomp}}.
-#' @importFrom cyclocomp cyclocomp
 #' @export
 cyclocomp_linter <- function(complexity_limit = 15L) {
+  if (!requireNamespace("cyclocomp", quietly = TRUE)) {
+    stop("'cyclocomp_linter' depends on the cyclocomp package to work, but it's not installed.")
+  }
```
AshesITR: This too needs reverting, no?

AshesITR (dismissed):

MichaelChirico:
Need approval again after adding #nocovs, thanks!

AshesITR (approved):

MichaelChirico:
it looks like maybe there's a repo setting expecting 16.04 to run even though it was removed in #859 ?

@jimhester is that something you have to change?

jimhester:
Thanks for working on this @MichaelChirico and the review @AshesITR!

PR #864, opened 2021-09-21
remove deprecated linters

### LINKED ISSUE #862: Deprecated trailing_semicolons_linter doesn't work
```
lintr::lint("a <- 2;\n", lintr::trailing_semicolons_linter())
# Error in semicolon_terminator_linter(semicolon = "trailing")(source_file) : 
#   argument "source_file" is missing, with no default
```

The issue is the call tries to evaluate `semicolon_terminator_linter` instead of returning a function factory.

It's an easy fix, but it also raises the question -- should we just delete this? It's marked as deprecated since ~1.0.1 (4 years ago).

MichaelChirico:
Closes #862.

All of the linters here have been marked as deprecated for >4 years; `lintr` 3.0 will be the second major release since then.

I left the internal `lint_deprecated()` function and the skeleton of the documentation for easier retrieval/reuse in the future, but these could be deleted as well.


R/deprecated.R
```diff
@@ -1,11 +1,7 @@
...
-#' Functions that have been deprecated and replaced by newer ones. They will be removed in an
-#' upcoming version of \pkg{lintr} and should thus not be used anymore.
-#'
-#' @name lintr-deprecated
-#' @param source_file source file object from \code{\link{get_source_expressions}}
-#' @include object_name_linters.R
+# Deprecated functions
+#
+# Functions that have been deprecated and replaced by newer ones. They will be removed in an
+# upcoming version of \pkg{lintr} and should thus not be used anymore.
```
AshesITR, line 4: Maybe `@noRd` has a good use here?

AshesITR (approved):

PR #865, opened 2021-09-21
860: add line numbers in the "starts do not match ends" error message

### LINKED ISSUE #860: Error message not helpful when starts do not match ends of exclusion parts
Reprex:
- Put the following in `test.R`:
```r
#nolint start

c(1,2)

```
- Run `lintr:::addin_lint()`
- Note: It is important that 1) there is a start but not an end (or vice versa) and 2) there is an actual lintr problem (here no space after the comma)

Observation:
`[...]/test.R has 1 starts but only 0 ends!`

Problem:
This error message is not really helpful, especially when the file is large. It does not mention the line numbers of the start lines and that this is about the `#nolint` starts and ends.

Proposal:
Give more information in the error message (line numbers, mention `#nolint`). Currently the function has:
```r
      stop(file, " has ", length(starts), " starts but only ", 
        length(ends), " ends!")
```

Session Info:
```
R version 4.1.1 (2021-08-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.3 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=de_DE.UTF-8       LC_NUMERIC=C               LC_TIME=de_CH.UTF-8       
 [4] LC_COLLATE=de_DE.UTF-8     LC_MONETARY=de_CH.UTF-8    LC_MESSAGES=de_DE.UTF-8   
 [7] LC_PAPER=de_CH.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=de_CH.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] ps_1.6.0           crayon_1.4.1       withr_2.4.2        rprojroot_2.0.2    R6_2.5.1          
 [6] magrittr_2.0.1     stringi_1.7.4      lazyeval_0.2.2     remotes_2.4.0      rstudioapi_0.13   
[11] callr_3.7.0        rex_1.2.0          xml2_1.3.2         cyclocomp_1.1.0    desc_1.3.0        
[16] tools_4.1.1        stringr_1.4.0      xfun_0.25          lintr_2.0.1        compiler_4.1.1    
[21] processx_3.5.2     xmlparsedata_1.0.5 knitr_1.33   
```

Thanks for considering :-)

danielinteractive:
closes #860

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,7 +73,7 @@
 * New linter `duplicate_argument_linter()` checks that there are no duplicate arguments supplied to
 function calls. (#850, #851, @renkun-ken)
 * Several optional `Imported` packages have become `Suggested` dependencies: `httr`, `testthat`, and `rstudioapi`. This should allow snappier CI builds for usages not relying on some more "peripheral" features of the package.
-* Error message for mismatched starts and ends of exclusion ranges now includes line numbers. (#860, @danielinteractive)
+* Error message for mismatched starts and ends of exclusion ranges is now more helpful. (#571, #860, @AshesITR and @danielinteractive)
 
 # lintr 2.0.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,6 +73,7 @@
 * New linter `duplicate_argument_linter()` checks that there are no duplicate arguments supplied to
 function calls. (#850, #851, @renkun-ken)
 * Several optional `Imported` packages have become `Suggested` dependencies: `httr`, `testthat`, and `rstudioapi`. This should allow snappier CI builds for usages not relying on some more "peripheral" features of the package.
+* Error message for mismatched starts and ends of exclusion ranges now includes line numbers. (#860, @danielinteractive)
 
 # lintr 2.0.1
 
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -55,6 +55,17 @@ is_excluded_file <- function(file_exclusion) {
   Inf %in% file_exclusion[[names2(file_exclusion) == ""]]
 }
 
+line_info <- function(line_numbers) {
+  n <- length(line_numbers)
+  if (identical(n, 0L)) {
+    "no lines"
+  } else if (identical(n, 1L)) {
+    paste("line", line_numbers)
+  } else {
+    paste("lines", paste(line_numbers, collapse = ", "))
+  }
+}
+
 #' read a source file and parse all the excluded lines from it
 #'
 #' @param file R source file
@@ -86,13 +97,22 @@ parse_exclusions <- function(file, exclude = settings$exclude,
   }
 
   start_locations <- rex::re_matches(lines, exclude_start, locations = TRUE)[, "end"] + 1L
+  end_locations <- rex::re_matches(lines, exclude_end, locations = TRUE)[, "start"]
   starts <- which(!is.na(start_locations))
-  ends <- which(rex::re_matches(lines, exclude_end))
+  ends <- which(!is.na(end_locations))
 
   if (length(starts) > 0) {
     if (length(starts) != length(ends)) {
-      starts_msg <- sprintf(ngettext(length(starts), "%d range start", "%d range starts"), length(starts))
-      ends_msg <- sprintf(ngettext(length(ends), "%d range end", "%d range ends"), length(ends))
+      starts_msg <- sprintf(
+        ngettext(length(starts), "%d range start (%s)", "%d range starts (%s)"),
+        length(starts),
+        line_info(starts)
+      )
+      ends_msg <- sprintf(
+        ngettext(length(ends), "%d range end (%s)", "%d range ends (%s)"),
+        length(ends),
+        line_info(ends)
+      )
       stop(file, " has ", starts_msg, " but only ", ends_msg, " for exclusion from linting!")
     }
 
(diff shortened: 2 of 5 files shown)


danielinteractive:
@MichaelChirico thanks in advance for reviewing this

NEWS.md
```diff
@@ -73,6 +73,7 @@
 * New linter `duplicate_argument_linter()` checks that there are no duplicate arguments supplied to
 function calls. (#850, #851, @renkun-ken)
 * Several optional `Imported` packages have become `Suggested` dependencies: `httr`, `testthat`, and `rstudioapi`. This should allow snappier CI builds for usages not relying on some more "peripheral" features of the package.
+* Error message for mismatched starts and ends of exclusion ranges now includes line numbers. (#860, @danielinteractive)
```
MichaelChirico, line 76: I think you'll want to just merge this with the earlier NEWS item (still including the issue number & citing yourself of course)

R/exclude.R
```diff
@@ -55,6 +55,17 @@ is_excluded_file <- function(file_exclusion) {
...
 }
 
+line_info <- function(line_numbers) {
+  n <- length(line_numbers)
+  if (identical(n, 0L)) {
+    "no lines"
+  } else if (identical(n, 1L)) {
+    paste("line", line_numbers)
+  } else {
+    paste("lines", paste(line_numbers, collapse = ", "))
```
MichaelChirico, line 65: I prefer toString() over paste(., collapse=', ')

R/exclude.R
```diff
@@ -55,6 +55,17 @@ is_excluded_file <- function(file_exclusion) {
   Inf %in% file_exclusion[[names2(file_exclusion) == ""]]
 }
 
+line_info <- function(line_numbers) {
+  n <- length(line_numbers)
+  if (identical(n, 0L)) {
```
MichaelChirico, line 60: n==0 is fine since n is coming from length() just above

R/exclude.R
```diff
@@ -55,6 +55,17 @@ is_excluded_file <- function(file_exclusion) {
...
+    "no lines"
+  } else if (identical(n, 1L)) {
+    paste("line", line_numbers)
```
MichaelChirico, line 63: may as well use ngettext() here too since it's used below already

MichaelChirico (changes requested):
basically ok, some small changes

NEWS.md
danielinteractive: Can you help me which line number this would be? I stared at the list for 5 mins now and cannot see the matching item :-) thx!

R/exclude.R
danielinteractive: could do instead of this one comparison with 1 - since that function treats 0 and 2 and above the same, I still need the 0 comparison above. So not sure what looks nicer. ok to keep the 0 and 1 comparisons?..

NEWS.md
danielinteractive: I can see that you did some changes 1 year ago in https://github.com/jimhester/lintr/pull/524 but that is part of previous NEWS already. So I would suggest to keep this as a separate bullet here since there is no other NEWS entry here which matches?

danielinteractive:
@MichaelChirico thx for the review Michael, I had 2 questions above, would you maybe have a look before I revise the PR? thx!

R/exclude.R
MichaelChirico: I think the 0 case warrants its own handling here as much as I hate the extra branching

NEWS.md
MichaelChirico: sorry! just realized no item was added for #571 in #575. could you mention #571 and @AshesITR in an updated item?

danielinteractive: thx - done

R/exclude.R
danielinteractive: thx - first time I see that actually!

R/exclude.R
```diff
@@ -86,13 +97,22 @@ parse_exclusions <- function(file, exclude = settings$exclude,
...
   starts <- which(!is.na(start_locations))
-  ends <- which(rex::re_matches(lines, exclude_end))
+  ends <- which(!is.na(end_locations))
 
   if (length(starts) > 0) {
     if (length(starts) != length(ends)) {
-      starts_msg <- sprintf(ngettext(length(starts), "%d range start", "%d range starts"), length(starts))
-      ends_msg <- sprintf(ngettext(length(ends), "%d range end", "%d range ends"), length(ends))
+      starts_msg <- sprintf(
+        ngettext(length(starts), "%d range start (%s)", "%d range starts (%s)"),
```
MichaelChirico, line 107: looking again, let's make this whole expression be the line_info helper function, just add an argument to toggle start vs end. lots of redundancy in the big sprintf() expression here

danielinteractive: good idea - done

MichaelChirico (approved):

MichaelChirico:
Thanks!!

PR #867, opened 2021-09-23
fix undesirable_function_linter when linted expression is found late in file

### LINKED ISSUE #866: Regression in undesirable_function_linter
Reported by @jennybc at de3fb6b730c10f4ee2b222e1d58a5dcbc9604b68:

> This change broke my use of the undesirable function lintr in GHA:
> 
> https://github.com/r-lib/usethis/actions/workflows/lint-undesirable-functions.yaml
> 
> Specifically, here's what little info I get:
> 
> ```
> Error: Error in source_file$lines[[line]] : subscript out of bounds
> Calls: <Anonymous> ... assign_item -> <Anonymous> -> lapply -> FUN -> Lint -> structure
> ```
> 
> If I install from the previous commit, everything works normally.

I'm looking at this now. I found the error is triggered by:

testthat/test-git-default-branch.R

Investigating why...

MichaelChirico:
Closes #866

The issue is that `source_file$lines[[line]]` relies on `source_file$lines` having the line numbers as names, not as integers. So `source_file$lines[['111']]` (correct) is not the same as `source_file$lines[[111L]]` (on HEAD).

This only shows up if the expression at fault comes later in the file (or perhaps just at the 2nd expression of a 2-line file); that's reflected in the new test.

I also tidied up the xpath a bit, I think it's clearer to understand now.

I'll file a follow-up PR using the new helper a bit more broadly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,7 +70,7 @@
   files in your project. Override the default in the `encoding` setting of lintr. (#752, #782, @AshesITR)
 * New default linter `paren_body_linter()` checks that there is a space between right parenthesis and a body 
   expression. (#809, #830, @kpagacz)
-* `undesirable_function_linter` no longer lints `library` and `require` calls attaching a package with an undesired name, e.g. `library(foo)` (#814, @kpagacz)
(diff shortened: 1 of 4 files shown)


AshesITR (approved):
Thanks for the quick fix!

PR #873, opened 2021-10-09
Refactor object_length_linter

### LINKED ISSUE #871: `object_length_linter` should detect generics
As the `object_name_linter` already does, class suffixes for S3 generics shouldn't be part of the linted symbol.
That is, `object_length_linter(30)` (the default) shouldn't lint this example:

```
my_generic_function <- function(x, ...) {
  UseMethod("my_generic_function")
}

my_generic_function.my_cool_class <- function(x, ...) {
  cat("How cool and generic!\n")
}
```

AshesITR:
fixes #871 and makes the architecture of `object_length_linter` and `object_name_linter` coherent again.
I've moved the common XPath into the package namespace to prevent code duplication, but we could also re-implement `make_object_linter` instead - LMK if you like that better.
Also removed all unused and therefor redundant helpers that were used for generic detection in the regex and token-based approach.

 - [x] implement feature
 - [x] add tests
 - [x] add NEWS bullet


AshesITR:
@MichaelChirico I rebased this onto current master due to merge conflicts. Should be ready for review.

NEWS.md
```diff
@@ -97,6 +97,8 @@ function calls. (#850, #851, @renkun-ken)
 * `assignment_linter()` now lints right assignment (`->` and `->>`) and gains two arguments. `allow_cascading_assign` (`TRUE` by default) toggles whether to lint `<<-` and `->>`; `allow_right_assign` toggles whether to lint `->` and `->>` (#915, @michaelchirico)
 * `infix_spaces_linter()` gains argument `exclude_operators` to disable lints on selected infix operators. By default, all "low-precedence" operators throw lints; see `?infix_spaces_linter` for an enumeration of these. (#914 @michaelchirico)
 * `infix_spaces_linter()` now throws a lint on `a~b` and `function(a=1) {}` (#930, @michaelchirico)
+* `object_length_linter()` correctly detects generics and only counts the implementation class towards the length. 
+  (#871, @AshesITR)
```
MichaelChirico, line 101: An example here would be nice -- I needed to read the tests to understand what this meant

R/object_name_linters.R
```diff
@@ -1,3 +1,40 @@
...
+  " and not(ancestor::expr[",
+  "  preceding-sibling::OP-LEFT-BRACKET",
+  "  or preceding-sibling::LBB",
+  " ])",
+  "]",
+
+  " | ",
+
+  "//STR_CONST[",
+  " not(preceding-sibling::OP-DOLLAR)",
```
MichaelChirico, line 19: `OP-AT` too for S4 extractions?

R/object_name_linters.R
```diff
@@ -1,3 +1,40 @@
+object_name_xpath <- paste0(
+  # assignments
+  "//SYMBOL[",
+  " not(preceding-sibling::OP-DOLLAR)",
```
MichaelChirico, line 4: extract the repeated part as a variable and re-use it to make it easier to adjust going forward

R/object_name_linters.R
```diff
@@ -330,13 +188,38 @@ regexes_rd <- toString(paste0("\\sQuote{", names(style_regexes), "}"))
...
+    assignments <- xml2::xml_find_all(xml, object_name_xpath)
+
+    # Retrieve assigned name
+    nms <- strip_names(
+      xml2::xml_text(assignments)
+    )
+
+    generics <- strip_names(c(
+      declared_s3_generics(xml),
+      imported_s3_generics(namespace_imports(find_package(source_file$filename)))$fun,
```
MichaelChirico, line 207: make a wrapper for this? too much nesting --> not clear

R/object_name_linters.R
```diff
@@ -330,13 +188,38 @@ regexes_rd <- toString(paste0("\\sQuote{", names(style_regexes), "}"))
 #' @seealso [linters] for a complete list of linters available in lintr.
```
MichaelChirico, line 188: please explain the details in the documentation

R/object_name_linters.R
AshesITR: Their assignment targets are `SLOT`, not `SYMBOL`, so we don't need to except these.

R/object_name_linters.R
AshesITR: factored out `ns_imports` as a variable. No need for wrapping IMO, the logic isn't very complex, it's just a chain of functions.

MichaelChirico (approved):

PR #874, opened 2021-10-09
Remove redundant `_linter` suffix from auto names

### LINKED ISSUE #872: Remove redundant `_linter` suffix from auto names?
The new `# nolint: xxx.` feature makes it more likely for linter names to appear in source code.
I would therefor suggest we remove the redundant `_linter` suffix from the names to make most lint exclusions look more natural.
Compare:

```
my.bad.name <- function(a, b) { # nolint: object_name_linter.
  # Very long comment explaining why the name came to be and needs to stay # nolint: line_length_linter.
  a + b
}
```

to

```
my.bad.name <- function(a, b) { # nolint: object_name.
  # Very long comment explaining why the name came to be and needs to stay # nolint: line_length.
  a + b
}
```

wdyt?

cc @MichaelChirico @jimhester

AshesITR:
This causes the common `_linter` suffix to be removed from all automatic names, enabling a shorter (and in my opinion nicer) nolint syntax by default.

I've also refactored all lintr-internal linting configurations appropriately so we can see how the look-and-feel will be after the refactoring.
In doing this, I noticed that the auto-naming function of `with_defaults` might not be ideal - it completely ignores the linter name.

fixes #872

### DIFF
--- a/R/actions.R
+++ b/R/actions.R
@@ -3,8 +3,11 @@ in_github_actions <- function() {
 }
 
 # Output logging commands for any lints found
-github_actions_log_lints <- function(lints) {
+github_actions_log_lints <- function(lints, project_dir = "") {
   for (x in lints) {
+    if (nzchar(project_dir)) {
+      x$filename <- file.path(project_dir, x$filename)
+    }
     file_line_col <- sprintf(
       "file=%s,line=%s,col=%s", x$filename, x$line_number, x$column_number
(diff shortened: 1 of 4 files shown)


AshesITR:
Not convinced anymore that this is the way to go.
A less intrusive option would be to silently allow `# nolint: anything.` to suppress `anything_linter`.

PR #877, opened 2021-10-15
add debugging calls as undesirable functions

### LINKED ISSUE #876: Add browser() and other debugging calls as default undesirable functions
Leaving calls like `browser()` in the code after the bug is fixed is a common mistake; most "finished" code (especially in packages) shouldn't need to call this.

I believe the same goes for `debug()`, `undebug()`, `trace()`, `untrace()`, and `debugcall()`.

I think it would make sense for these to be in the default set of undesirable functions for `undesirable_function_linter`.

Any second thoughts?

MichaelChirico:
Closes #876


AshesITR:
Is `browser()` in the list? Didn't see it in the GitHub diff UI.

PR #883, opened 2021-11-08
new package_startup_linter

### LINKED ISSUE #882: onload_package_startup_linter for packageStartupMessage() inside .onLoad()?
Using `packageStartupMessage()` in the `.onLoad()` hook is an `R CMD check` `NOTE`. So I think it makes sense to offer a linter to detect that as an early warning.

MichaelChirico:
Closes #882


AshesITR:
This needs a `document()` to update the collate directive.

MichaelChirico:
@AshesITR you were indeed right -- `message()`, and `cat()`, `writeLines()`, and `print()`, are all blocked in both `.onLoad()` and `.onAttach()`. The carve-out was `packageStartupMessage()` is allowed in `.onAttach()`, but not `.onLoad()`.

Here is the minimal version of a linter that catches all of the `.onLoad()`/`.onAttach()` issues that QC.R --> `R CMD check` does.

We could fine-tune this further in a variety of ways:

 - Make the lint message more precise by matching the specific culprit function, instead of mentioning several functions at once
 - Add options to turn on/off the various checks
 - Add an option to match `base` behavior strictly (we currently take two extra liberties, namely, base accepts `...` as an argument name, and allows `require` and `library` to be used as `name`s, and hence in `sapply()`/`lapply()` style)

LMK which of these we should aim to land in this initial PR, or any other feedback.

AshesITR:
Nice work!

A more precise lint message would be cool - we do that in the unwanted function linter as well IIRC.
I don't think there is a need for turning off specific checks for this linter
Following suit from the actual NOTE generated by counting  `...` as a bad argument specification for `.onLoad` and `.onAttach` seems okay to me as well.

Also, QC.R does similar checks for `.onUnload` and  `.onDetach`. Do we want these as well?
 And if so, do we want to put them in this linter or in a separate `package_unload_linter`?
If we add them here, the name could be `package_hooks_linter`or `package_ns_hooks_linter` to clarify scope and intent.

MichaelChirico:
> Following suit from the actual NOTE generated by counting ... as a bad argument specification for .onLoad and .onAttach seems okay to me as well.

To clarify, currently `R CMD check` considers `.onLoad <- function(...) { }` as OK, but the linter as currently written does not allow this.

I think it makes sense to check for `.onUnload()` stuff too, good idea! And I don't think there's much of a use case for separating `.onLoad()` and `.onUnload()` checks, so I'll combine them here.

MichaelChirico:
Should be ready for another review

AshesITR (dismissed):
LGTM, nice work and very comprehensive tests!

AshesITR (approved):

MichaelChirico:
funnily enough .onUnload seems excluded from the detach hook checks in QC.R. only .Last.lib and .onDetach, somewhat puzzlingly

Issue #884, opened 2021-11-10
Upstreaming linters from Google's internal linting suite


MichaelChirico:
Internally at Google we use `lintr` to power our in-editor and CI-based R linting tools. Over the last year+ (in addition to working directly on `lintr`) we have been building out a set of linters that extends the suite offered by `lintr`. Currently, the package has >60 such linters, almost all of which we think are widely relevant. They have all been test on & applied to existing internal R code at Google.

We are happy to contribute any subset of these to `lintr`. Here is a brief summary of all of them; I can elaborate more as needed:

- [x] `AnyDuplicatedLinter` - Require usage of anyDuplicated() > 0 over any(duplicated(.))
- [x] `AnyIsNALinter` - Require usage of anyNA over any(is.na(.))
- [x] `ApplyAnonymousLinter` - Block usage of anonymous functions in *apply functions when unnecessary #1541 
- [x] `BaseOverwriteLinter` - Block assigning any variables whose name clashes with a 'base' R function #2307
- [x] `BooleanArithmeticLinter` #1579
- [x] `CharacterOnlyLinter` - Enforce library calls using symbols #2279 
- [x] `ClassEqualsLinter` - Block comparison of class with ==
- [x] `ComparisonNegationLinter` #2272 
- [x] `ConsecutiveMutateLinter` #2305
- [x] `ConsecutiveSuppressionLinter` #2306
- [x] `ElseSameLineLinter` - Require else to come on the same line as
- [x] `EmptyAssignmentLinter` - Block assignment of {} #1637
- [x] `ExpectComparisonLinter` - Require usage of expect_gt(x, y) over expect_true(x > y) (and similar) #955 
- [x] `ExpectIdenticalLinter` - Require usage of expect_identical(x, y) where appropriate #958 
- [x] `ExpectIsLinter` - Redirect away from deprecated testthat::expect_is() #946 
- [x] `ExpectLengthLinter` - Require usage of expect_length(x, n) over expect_equal(length(x), n) #950 
- [x] `ExpectNamedLinter` - Require usage of expect_named(x, n) over expect_equal(names(x), n) #949 
- [x] `ExpectNotLinter` - Require usage of expect_false(.) over expect_true(!.) #945 
- [x] `ExpectNullLinter` - Require usage of expect_null(x) over expect_equal(x, NULL) and similar #887 
- [x] `ExpectS3ClassLinter` - Require usage of expect_s3_class() or expect_s4_class() where appropriate. #943
[...]

PR #887, opened 2021-11-15
expect_null_linter

MichaelChirico:
Part of #884 

Hits `expect_equal(x, NULL)`, which should be `expect_null(x)`


R/google_linters.R
```diff
@@ -0,0 +1,10 @@
+#' @title Linters for the Google extension to the tidyverse style guide
+#'
+#' Google uses a fork of the tidyverse style guide, pursuant to which
```
MichaelChirico, line 3: cc @michaelquinn32 for copy here & in NEWS

R/expect_null_linter.R
```diff
@@ -0,0 +1,44 @@
+#' @describeIn google-linters Require usage of expect_null(x) over expect_equal(x, NULL) and similar
+#' [testthat::expect_null()] exists specifically for testing for `NULL` objects.
```
MichaelChirico, line 2: are we allowed to link to Suggested packages? 🤔

MichaelChirico: I _think_ we are OK:

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Cross_002dreferences

.Rbuildignore
```diff
@@ -10,7 +10,7 @@
 ^\.Rproj\.user$
 ^\.idea$
 ^\.dev$
-^\.lintr$
+\.lintr$
```
MichaelChirico, line 13: to quiet an `R CMD check` `NOTE` about hidden files -- we have a `.lintr` nested within our test suite

MichaelChirico:
Patch coverage drop is from `xp_or` being added but not used yet. It will be used in later linters, so I think it's fine to ignore in this PR.

R/google_linters.R
```diff
@@ -0,0 +1,9 @@
+#' @title Linters for the Google extension to the tidyverse style guide
```
michaelquinn32, line 1: Can we drop the @title tag? See the examples here:
https://style.tidyverse.org/documentation.html

.Rbuildignore
AshesITR: Wouldn't this break the tests when running on CRAN source tarballs?

R/expect_null_linter.R
```diff
@@ -0,0 +1,43 @@
+#' @describeIn google-linters Require usage of expect_null(x) over expect_equal(x, NULL) and similar
```
AshesITR, line 1: Is there a way to describe these both in google-linters and in linters?
IMO there should be one help page that lists all available linters.

.Rbuildignore
MichaelChirico: Agree it makes sense to have one landing page, but maybe we need

```
+ available linter suites
└-- linters
└-- google-linters
└-- ...
```

instead? ?linters is already 300 lines long

R/expect_null_linter.R
MichaelChirico: Agree it makes sense to have one landing page, but maybe we need

```
+ available linter suites
└-- linters
└-- google-linters
└-- ...
```

instead? ?linters is already 300 lines long

.Rbuildignore
MichaelChirico: Oh, right, since this will mean that doesn't get included. Hmm I guess we just have to live with that note then huh

R/expect_null_linter.R
AshesITR: Opened #888 for discussion around this.
Maybe we could also export a function that returns a tibble of all available linters?
We could add some metadata to this and would enable using R to analyze the available sets more quickly than skimming a 300 lines long man page...

MichaelChirico: sounds good! Do we want to handle that as part of this PR? or handle it later?

AshesITR: I think we should discuss this first to minimize the need for refactoring, but handle necessary changes in a separate PR.

MichaelChirico:
@AshesITR are we ready to start the stream of google linters now? if so I'll rebase this to the doc changes and start from here

MichaelChirico:
I'm immediately seeing the benefit of #916... a bit hard to know if I've integrated this new linter with the new approach correctly (IIUC the main thing is to be sure to add the linter+tags to the inst/lintr/linters.csv file?)

R/expect_null_linter.R
```diff
@@ -0,0 +1,48 @@
...
+#' Require usage of expect_null(x) over expect_equal(x, NULL) and similar
+#'   [testthat::expect_null()] exists specifically for testing for `NULL` objects.
+#'   [testthat::expect_equal()], [testthat::expect_identical()], and
+#'   [testthat::expect_true()] can also be used for such tests,
+#'   but it is better to use the tailored function instead.
+#'
+#' @evalRd rd_tags("expect_null_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+expect_null_linter <- function() {
```
MichaelChirico, line 12: as parameters here we could potentially toggle which of the cases apply:

(1) `expect_equal(x, NULL)` ➡️  `expect_null(x)`
(2) `expect_identical(x, NULL)` ➡️  `expect_null(x)`
(3) `expect_true(is.null(x))` ➡️  `expect_null(x)`

Is it worth doing so?

AshesITR: I think it's fine, or are there legitimate use-cases / differences in behavior for the other forms?

R/expect_null_linter.R
```diff
@@ -0,0 +1,48 @@
+#' expect_null Linter
+#'
+#' Require usage of expect_null(x) over expect_equal(x, NULL) and similar
```
AshesITR, line 3: Probably want to format as code.

inst/lintr/linters.csv
```diff
@@ -6,6 +6,7 @@ commas_linter,style readability default
 commented_code_linter,style readability best_practices default
 cyclocomp_linter,style readability best_practices default configurable
 equals_na_linter,robustness correctness common_mistakes default
+expect_null_linter,testing best_practices
```
AshesITR, line 9: We shouldn't introduce tags with only one linter IMO.
What about `package_development` as a more general tag which would also fit, e.g., `package_hooks_linter`?

R/expect_null_linter.R
```diff
@@ -0,0 +1,48 @@
...
+  Linter(function(source_file) {
+    if (length(source_file$parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    # for expect_{equal,identical}(x, NULL)
+    equal_expr_cond <- sprintf(
+      "SYMBOL_FUNCTION_CALL[%s] and following-sibling::expr[2][NULL_CONST]",
```
AshesITR, line 22: What about 'expect_equal(NULL, x)`?
Is there any downside to dropping the `[2]` component?
The `following-sibling::expr[NULL_CONST]` would require one argument (first or second) to be identical to the expression `NULL` I think.

R/expect_null_linter.R
```diff
@@ -0,0 +1,48 @@
...
+      xp_text_in_table(c("expect_equal", "expect_identical"))
+    )
+
+    # for expect_true(is.null(x))
+    is_null_call <- "SYMBOL_FUNCTION_CALL[text() = 'is.null']"
+    true_expr_cond <- xp_and(
+      "SYMBOL_FUNCTION_CALL[text() = 'expect_true']",
+      sprintf("following-sibling::expr[1][expr[%s]]", is_null_call)
```
AshesITR, line 30: I believe we don't need the `[1]` qualifier.

Also, `xp_and()` could be used above as well, no?

inst/lintr/linters.csv
MichaelChirico: basically agreed, but otoh there will be many more with this tag, see the Google sheet

R/expect_null_linter.R
MichaelChirico: not really. just trying to imagine a user wanting finer control. I guess we can wait for an FR.

man/linters.Rd
```diff
@@ -50,8 +52,8 @@ The following linters exist:
 \item{\code{\link{missing_argument_linter}} (tags: common_mistakes, configurable, correctness)}
 \item{\code{\link{missing_package_linter}} (tags: common_mistakes, robustness)}
 \item{\code{\link{namespace_linter}} (tags: configurable, correctness, robustness)}
-\item{\code{\link{no_tab_linter}} (tags: consistency, default, style)}
 \item{\code{\link{nonportable_path_linter}} (tags: best_practices, configurable, robustness)}
+\item{\code{\link{no_tab_linter}} (tags: consistency, default, style)}
```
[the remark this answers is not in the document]
MichaelChirico: #923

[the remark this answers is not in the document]
AshesITR: What's your locale?
I can't reproduce your sorting behaviour on any locale in linux.
I tried `withr::with_locale(c("LC_COLLATE"=...), sort(c("n", "_")))` and always got `c("_", "n")`...

Maybe we should do `withr::with_locale(c("LC_COLLATE"="C"), sort(...))` (hoping that works)

AshesITR:
> I'm immediately seeing the benefit of #916... a bit hard to know if I've integrated this new linter with the new approach correctly (IIUC the main thing is to be sure to add the linter+tags to the inst/lintr/linters.csv file?)


Looks good. One last thing necessary for new tags is creating a documentation for them in `R/linter_tag_docs.R`

R/expect_null_linter.R
AshesITR: Agreed.

inst/lintr/linters.csv
AshesITR: Okay fine then. For this specific tag though I think `package_development` would be more intuitive.
`testing` could also mean `if () {} else {}` or `ifelse()` or `if_else()`.
WDYT?

Anyway we need a new Rd for the new tag, to be placed in `R/linter_tag_docs.R`

man/linters.Rd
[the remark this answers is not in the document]
MichaelChirico: Let's handle it all in #923, immaterial to this PR

tests/testthat/test-expect_null_linter.R
```diff
@@ -0,0 +1,36 @@
...
+    expect_null_linter()
+  )
+
+  # length-0 could be NULL, but could be integer() or list(), so let it pass
+  expect_lint("expect_length(x, 0L)", NULL, expect_null_linter())
+})
+
+test_that("expect_null_linter blocks simple disallowed usages", {
+  expect_lint(
+    "expect_equal(x, NULL)",
```
AshesITR, line 18: `expect_equal(NULL, x)` should also lint I think

inst/lintr/linters.csv
MichaelChirico: SGTM, done

tests/testthat/test-expect_null_linter.R
MichaelChirico: for us I _believe_ this is covered by the `yoda_test_linter` (if not, it should be), but yes, we should generally write linters that work independently of having others on at the same time. Will do.

man/linters.Rd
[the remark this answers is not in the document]
AshesITR: WDYT about manually rearranging this instance to keep the commits clean until #923 is resolved?

inst/lintr/linters.csv
```diff
@@ -6,6 +6,7 @@ commas_linter,style readability default
 commented_code_linter,style readability best_practices default
 cyclocomp_linter,style readability best_practices default configurable
 equals_na_linter,robustness correctness common_mistakes default
+expect_null_linter,package_development best_practices
```
AshesITR, line 9: `backport_linter` and `package_hooks_linter` could also get the new tag.

R/expect_null_linter.R
MichaelChirico: re: `xp_and()`, recently I'm refactor to use a literal xpath as much as possible, I think it maximizes readability. done in the current commit.

I think the addition of `glue()` really enhances readability but we could try and stick with `sprintf()` to avoid the dep, WDYT?

MichaelChirico: (it may also be good to just write out `text() = 'expect_equal' or text() = 'expect_identical'` here too, but there will be other linters to import that face the same question in better use cases)

R/expect_null_linter.R
MichaelChirico: I think my concern here was avoiding thinking about other arguments to `expect_equal/expect_identical`, e.g. `expect_identical(x, y, tolerance=NULL)`. I haven't thought through whether it makes sense for such cases to exist but restricting to `postion() = 1 or position() = 2` should be fine.

R/expect_null_linter.R
MichaelChirico: the `[1]` qualifier is for `expect_true(x, info = is.null(y))`. Maybe unlikely but doesn't hurt much to include it.

(the more general issue of really agreeing with `match.call()` is ignored. i think that it's typically OK to throw a false positive if people are using really unexpected positional arguments)

man/linters.Rd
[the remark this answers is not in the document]
MichaelChirico: ping again once the rest of the PR is done. o/w i'll have to keep doing it whenever I `roxygenize()`.

R/expect_null_linter.R
MichaelChirico: covered `expect_equal(NULL, x)` now

R/expect_null_linter.R
```diff
@@ -0,0 +1,50 @@
...
+    #  (2) expect_true(is.null(x))
+    xpath <- glue::glue("//expr[
+      (
+        SYMBOL_FUNCTION_CALL[ {xp_text_in_table(c('expect_equal', 'expect_identical'))} ]
+        and following-sibling::expr[position() <= 2 and NULL_CONST]
+      ) or (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_true']
+        and following-sibling::expr[1][expr[SYMBOL_FUNCTION_CALL[text() = 'is.null']]]
+      )
+    ]")
```
AshesITR, line 34: I don't see the benefit of `glue::glue()` here but I agree seeing the entire XPath helps readability.

MichaelChirico: ok, let's save adding the dep for when the value add is clearer then

AshesITR:
LGTM without the dep on `glue::glue`
NB if no namespace import is desired, `usethis::use_package()` is a very convenient helper.

man/linters.Rd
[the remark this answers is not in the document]
AshesITR: We're good now I think

AshesITR (approved):

PR #893, opened 2021-12-19
Capture malformed raw string error

### LINKED ISSUE #892: malformed raw string literal not captured
The syntax error of the raw string in the following code is not captured:

```r
line1 <- 1
line2 <- 2
line3 <- 3
line4 <- R"--/hello/--"
```

where error has the number of line:

```
malformed raw string literal at line 4
```

renkun-ken:
Closes #892

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -137,6 +137,22 @@ get_source_expressions <- function(filename, lines = NULL) {
             line = source_file$lines[[l]]
           )
         )
+      } else if (grepl("malformed raw string literal at line", e$message, fixed = TRUE)) {
+        l <- as.integer(re_matches(
+          e$message,
+          rex("malformed raw string literal at line ", capture(name = "line", digits))
+        )$line)
+        # Invalid raw string in source code
+        return(
+          Lint(
+            filename = source_file$filename,
+            line_number = l,
+            column_number = 1L,
+            type = "error",
+            message = "Malformed raw string literal.",
+            line = source_file$lines[[l]]
+          )
+        )
       }
 
       message_info <- re_matches(e$message,
--- a/tests/testthat/test-get_source_expressions.R
+++ b/tests/testthat/test-get_source_expressions.R
@@ -122,3 +122,14 @@ test_that("Warns if encoding is misspecified", {
   expect_equal(the_lint$message, "Invalid multibyte string. Is the encoding correct?")
   expect_equal(the_lint$line_number, 1L)
 })
+
+test_that("Malformed raw string literal error is captured", {
+  skip_if_not(getRversion() >= "4.0.0")
+  expect_lint("line1 <- 1
+line2 <- 2
+line3 <- 3
+line4 <- R\"--/hello/--\"", checks = list(
+    message = "Malformed raw string literal.",
+    line_number = 4
+  ), default_linters)
+})


MichaelChirico (commented):
this lint_error function is getting pretty unwieldy.

let's split it off into a proper helper function (i.e. not anonymous)?

also, is there any way to know if we're being exhaustive here?

MichaelChirico:
Seems several of these errors are coming from gram.c:

https://github.com/wch/r-source/blob/trunk/src/main/gram.c

Would it make sense to just rake through the `error(` calls there and see which ones we should grep for? I count 43 at a quick glance, that may be overkill...

renkun-ken:
The following code extracts all errors from `gram.c`:

```r
lines <- readLines("https://raw.githubusercontent.com/wch/r-source/trunk/src/main/gram.c")
err_lines <- grep("error(_(", lines, fixed = TRUE, value = TRUE)
errors <- trimws(gsub("^.*error\\(_\\(\"(.+)\".+", "\\1", err_lines))
unique(errors)
```

Looks like the following are the unique error messages:

```
> unique(errors)                                                                                                                         
 [1] "EOF whilst reading MBCS char at line %d"                                                        
 [2] "invalid multibyte character in parser at line %d"                                               
 [3] "input buffer overflow at line %d"                                                               
 [4] "The pipe operator requires a function call as RHS"                                              
 [5] "RHS variable must be a symbol"                                                                  
 [6] "incorrect tag type at line %d"                                                                  
 [7] "allocation of source reference state failed"                                                    
 [8] "contextstack overflow"                                                                          
 [9] "repeated formal argument '%s' on line %d"                                                       
[10] "nul character not allowed (line %d)"                                                            
[11] "\\\\uxxxx sequences not supported inside backticks (line %d)"                                   
[12] "invalid \\\\u{xxxx} sequence (line %d)"                                                         
[13] "\\\\Uxxxxxxxx sequences not supported inside backticks (line %d)"                               
[14] "invalid \\\\U{xxxxxxxx} sequence (line %d)"                                                     
[15] "invalid \\\\U{xxxxxxxx} value %6x (line %d)"                                                    
[16] "invalid \\\\Uxxxxxxxx value %6x (line %d)"                                                      
[17] "mixing Unicode and octal/hex escapes in a string is not allowed"                                
[18] "string at line %d containing Unicode escapes not in this locale\\nis too long (max 10000 chars)"
[19] "malformed raw string literal at line %d"                                                        
[20] "contextstack overflow at line %d"                                                               
[21] "internal parser error at line %d"
```

renkun-ken:
Some errors are from <https://github.com/wch/r-source/blob/trunk/src/main/character.c>, such as [invalid multibyte string](https://github.com/wch/r-source/blob/trunk/src/main/character.c#L345).

MichaelChirico:
[apologies for splitting discussion a bit between the issue & PR]

OK, I had been thinking the failure was a bit more catastrophic than just garbling the error's metadata... in that case, I think we can handle these errors incrementally as they come along.

How about we leave a note in the error message to please file an issue if encountering one of the uncovered messages?

AshesITR:
I think it would be rather easy to convert all of these error messages to regexes with a group for the like number.
This would allow us to handle all of them with fairly little code. WDYT?

Instead of a giant if/else if chain we would have a vector of regexes and if any matches, construct a lint from it.

MichaelChirico:
I do worry about maintenance burden. The more messages we depend on, the higher the turnover rate will be...  if we can generate the regex & post-processing steps from the source code with a code snippet, that's moot, otherwise, I don't think it passes the cost/benefit smell test.

From that list of errors, most are very obscure, and some are internal errors or basically shouldn't happen (e.g. overflows).

(OTOH, I would definitely add a case for `The pipe operator requires a function call as RHS`.)

AshesITR:
I agree, the error(_( call to regex conversion needs to be automated, @renkun-ken did most of the work for that already.
What's missing is escaping and substitution of `line %d` with a line group regex and other `%` format specifiers by corresponding regexes.

renkun-ken:
Closing as superseded by #895.

PR #895, opened 2021-12-20
Improve line location information for R parse errors

### LINKED ISSUE #892: malformed raw string literal not captured

### LINKED ISSUE #894: Extract line info from all parser errors
The following code extracts all errors from `gram.c`:

```r
lines <- readLines("https://raw.githubusercontent.com/wch/r-source/trunk/src/main/gram.c")
err_lines <- grep("error(_(", lines, fixed = TRUE, value = TRUE)
errors <- trimws(gsub("^.*error\\(_\\(\"(.+)\".+", "\\1", err_lines))
unique(errors)
```

Looks like the following are the unique error messages:

```
> unique(errors)                                                                                                                         
 [1] "EOF whilst reading MBCS char at line %d"                                                        
 [2] "invalid multibyte character in parser at line %d"                                               
 [3] "input buffer overflow at line %d"                                                               
 [4] "The pipe operator requires a function call as RHS"                                              
 [5] "RHS variable must be a symbol"                                                                  
 [6] "incorrect tag type at line %d"                                                                  
 [7] "allocation of source reference state failed"                                                    
 [8] "contextstack overflow"                                                                          
 [9] "repeated formal argument '%s' on line %d"                                                       
[10] "nul character not allowed (line %d)"                                                            
[11] "\\\\uxxxx sequences not supported inside backticks (line %d)"                                   
[12] "invalid \\\\u{xxxx} sequence (line %d)"                                                         
[13] "\\\\Uxxxxxxxx sequences not supported inside backticks (line %d)"                               
[14] "invalid \\\\U{xxxxxxxx} sequence (line %d)"                                                     
[15] "invalid \\\\U{xxxxxxxx} value %6x (line %d)"                                                    
[16] "invalid \\\\Uxxxxxxxx value %6x (line %d)"                                                      
[17] "mixing Unicode and octal/hex escapes in a string is not allowed"                                
[18] "string at line %d containing Unicode escapes not in this locale\\nis too long (max 10000 chars)"
[19] "malformed raw string literal at line %d"                                                        
[20] "contextstack overflow at line %d"                                                               
[21] "internal parser error at line %d"
```

_Originally posted by @renkun-ken in https://github.com/r-lib/lintr/issues/893#issuecomment-997368264_

AshesITR:
I chose to use a script in `data-raw` to auto-generate the required parse error regexes and store them as .parse_error_regexes in `R/sysdata.rda`.
`get_source_expressions()` then uses a for-loop after the initial (manual) message checks to see if a parse error with line information was matched.

Fixes #892
Fixes #894

 * [x] implement
 * [x] add tests
 * [x] add NEWS bullet


AshesITR:
cc @MichaelChirico @renkun-ken

renkun-ken:
Thanks for the generalized solution!

I tried the following 

```
> str(lintr::lint(text = 'ok\nok\n"\\u{9999"'))                                                         
List of 1
 $ <text>:List of 8
  ..$ filename     : chr "<text>"
  ..$ line_number  : int 1
  ..$ column_number: int 1
  ..$ type         : chr "error"
  ..$ message      : chr "invalid \\u{xxxx} sequence (line 3)"
  ..$ line         : chr ""
  ..$ ranges       : NULL
  ..$ linter       : chr NA
  ..- attr(*, "class")= chr "lint"
 - attr(*, "class")= chr "lints"
```

but it seems the line number is not captured.

AshesITR:
Thanks for pointing this out. It seems we need to C-unescape the format messages before conversion to regex. I'll create a commit for that shortly.

AshesITR:
WDYT about removing the `(line xxx)` or `at line xxx` from error messages and capitalizing the first letter?
This would be more in line with the errors we handle manually.

renkun-ken:
Agree

data-raw/parse_errors.R
```diff
@@ -0,0 +1,47 @@
...
+  "(?s)(\\(\\?<line>\\[\\[:digit:\\]\\]\\+\\)(?:\\\\\\)| ))([^$]+?)\\$$",
+  "\\1(?<msg_2>\\2)$",
+  .parse_error_regexes,
+  perl = TRUE
+)
+
+# Check that no format specifiers remain
+stopifnot(!any(grepl("%", .parse_error_regexes)))
+
+usethis::use_data(.parse_error_regexes, internal = TRUE, overwrite = TRUE)
```
AshesITR, line 47: alternatively, we could put `dput(.parse_error_messages)` hard-coded in some source file, if we want to avoid the use of `sysdata.rda`.

MichaelChirico:
I still find what's happening here a bit strange...

we're sort of introducing a soft dependency on r-devel, right?

e.g., suppose the error message for one of these changes (e.g. as proposed by me here: https://bugs.r-project.org/show_bug.cgi?id=18259). Now `lintr` will only "work as intended" on R versions build after the SVN revision where the change was added.

Again, because it's not a catastrophic error to use the wrong error message (we'll just get the lint metadata wrong), I don't think we need to have a particularly global solution here...

it will, on the other hand, wreak havoc on our testing suites if we're not careful. In particular I don't think we should add tests for getting the line number right, lest we need to complicate those tests by making them particularly R-version-aware.

Lastly, can we confirm relying on an internet connection in data-raw is not an issue? I don't think it is since it is just used to generate the sysdata.rda file, but maybe I'm missing something. data-raw should be added to .Rbuildignore, too, right?

AshesITR:
You raise a good point regarding changed error messages.

A more fail-safe strategy might be to construct regexes for the part unlikely to change (i.e. for `{at,on} line xx` and `(line xx)`).
The disadvantage with this is that some error messages have %s as a format with user input, potentially breaking an improperly anchored line regex.
The only message which currently doesn't have the line information at the very end, is
`"string at line %d containing Unicode escapes not in this locale\\nis too long (max 10000 chars)"`.

So a good compromise IMHO would be to match `( at | on | \()line (?<line>\d+)\)?$`, anchored at the end of the error message.

Regarding internet connection for data-raw scripts: I think it's fine, as most development activities on lintr involve internet access.

MichaelChirico:
alternative idea:

have quick code snippets that run .onLoad() that trigger all of the parsing errors, and record the messages in some cache, referring back to it later. would have to work out how to "untemplate"/"regexify" the received messages.

MichaelChirico:
Maybe that won't work, or at least, I think it reduces to the same issue of designing a regex for extracting a line number from any of the possible types of input message 🤔

AshesITR:
I'm not a big fan of doing such dynamic reverse-engineering in `.onLoad` to be honest.

What comes to mind to be more robust regarding changes in the error strings over time, would be to gather `error_formats` from all released R versions. This would mean backward compatibility of this feature, though still requiring maintenance updates to `sysdata.rda` whenever an error message changes or a new message gets added.

As an alternative, the simple but not 100% correct solution I suggested, using a hand-crafted line number regex, would likely require next to no maintenance and still work 99% of the time.

MichaelChirico:
yes, let's go for the regex approach. it looks good enough to me.

renkun-ken:
The regex approach sounds good to me too.

R/get_source_expressions.R
```diff
@@ -139,6 +139,41 @@ get_source_expressions <- function(filename, lines = NULL) {
         )
       }
 
+      parse_error_rx <- rex(
```
MichaelChirico, line 142: add comments pointing to the parser source of these errors?

MichaelChirico, line 142: does this lint_error function need to be nested inside the function? it'll be easier to debug if it's in the package namespace instead

R/get_source_expressions.R
```diff
@@ -139,6 +139,41 @@ get_source_expressions <- function(filename, lines = NULL) {
...
+      )
+
+      if (grepl(parse_error_rx, e$message, perl = TRUE)) {
+        rx_match <- re_matches(
+          e$message,
+          parse_error_rx
+        )
+        l <- as.integer(rx_match$line)
+        # Sometimes the parser "line" runs one past the last line
+        l <- pmin(l, length(source_file$lines))
```
MichaelChirico, line 160: will Lint break if we return say Inf here? the last line may be misleading if the issue was an unmatched brace much earlier...

maybe an issue for a future PR, feel free to add a TODO to investigate

tests/testthat/test-get_source_expressions.R
```diff
@@ -122,3 +122,23 @@ test_that("Warns if encoding is misspecified", {
   expect_equal(the_lint$message, "Invalid multibyte string. Is the encoding correct?")
   expect_equal(the_lint$line_number, 1L)
 })
+
+test_that("Can extract line number from parser errors", {
+  skip_if_not(getRversion() >= "4.0")
+
+  with_content_to_parse("\"ok\"\nR\"---a---\"", {
+    expect_equal(error$message, "Malformed raw string literal.")
```
MichaelChirico, line 130: I'd prefer using ' to wrap this string for readability. using a newline over \n may be preferable as well...

tests/testthat/test-get_source_expressions.R
```diff
@@ -122,3 +122,23 @@ test_that("Warns if encoding is misspecified", {
   expect_equal(the_lint$message, "Invalid multibyte string. Is the encoding correct?")
   expect_equal(the_lint$line_number, 1L)
 })
+
+test_that("Can extract line number from parser errors", {
+  skip_if_not(getRversion() >= "4.0")
+
+  with_content_to_parse("\"ok\"\nR\"---a---\"", {
```
MichaelChirico, line 129: the regex has three cases:

or(" at ", " on ", " ("),

can we make sure we have tests of each? maybe add a comment to the test saying which is tested

tests/testthat/test-get_source_expressions.R
```diff
@@ -122,3 +122,23 @@ test_that("Warns if encoding is misspecified", {
   expect_equal(the_lint$message, "Invalid multibyte string. Is the encoding correct?")
   expect_equal(the_lint$line_number, 1L)
 })
+
+test_that("Can extract line number from parser errors", {
+  skip_if_not(getRversion() >= "4.0")
```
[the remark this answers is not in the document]
AshesITR: `getRversion()` feels much more natural to me.

R/get_source_expressions.R
AshesITR: Pretty sure this breaks all lintr integrations that use source locations (e.g. RStudio Markers and GitHub linting).
I think this was the reason we even discovered the bug.

R/get_source_expressions.R
AshesITR: It uses `source_file` from the outer environment.
Could be refactored to `parse_lint_error <- function(e, source_file) { ... }` and then changing

https://github.com/r-lib/lintr/pull/895/files/3f6b05b28b4b1ca3d090559863948b00e112590b#diff-5f02df44ebfb8ab681ebe1eee50257e672a0daa3de8e0ce72245130e260091fcR282

to `get_source_file(source_file, error = function(e) parse_lint_error(e, source_file))`.

WDYT?

MichaelChirico: let's do it. similar is done for xml_nodes_to_lint I believe.

we could also make it a function factory but that doesn't really help debugging.

AshesITR: Done.

AshesITR: Done.

tests/testthat/test-get_source_expressions.R
AshesITR: Done using `trim_some()` for better readability.

tests/testthat/test-get_source_expressions.R
AshesITR: Good point. Added a missing test case (on line x is only triggered by a repeated formal argument)

MichaelChirico (approved):
great! thanks

PR #897, opened 2022-01-01
specify file/line exclusions

wibeasley:
to distinguish it from excluding linting rules


AshesITR:
Not sure about the wording here, since you can also define linter exclusions per-file and per-line since the new nolint feature was added in #660.

PR #898, opened 2022-01-03
Create fresh R-CMD-check.yaml from use_tidy_github_actions()

### LINKED ISSUE #896: pdflatex is not available in R-CMD-check workflow
Currently all R CMD check workflows fail due to missing pdflatex:

```
Creating pdf output from LaTeX ...
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  pdflatex is not available
```

@dragosmg maybe you have an idea which step previously installed pdflatex?

AshesITR:
Attempt to fix the build failures pertaining to missing pdflatex.

hopefully fixes #896

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,9 +1,3 @@
-# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
-# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
-#
-# NOTE: This workflow is overkill for most R packages and
-# check-standard.yaml is likely a better choice.
-# usethis::use_github_action("check-standard") will install it.
 on:
   push:
     branches: [main, master]
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,3 +1,9 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+#
+# NOTE: This workflow is overkill for most R packages and
+# check-standard.yaml is likely a better choice.
+# usethis::use_github_action("check-standard") will install it.
 on:
   push:
     branches: [main, master]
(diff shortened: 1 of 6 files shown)


AshesITR:
@jimhester Do we want keep using 20.04 as our ubuntu version or do we downgrade to 18.04 as the tidyverse defaults?
Tests currently pass on 18.04 as well, but that would require changes to the GitHub settings, making ubuntu-18.04 (release) a required check instead of ubuntu-20.04 (release).

LMK what is preferred.

jimhester:
20.04 is probably fine for lintr, generally you only need to worry about testing on older versions when you are dealing with system dependencies with compiled code, so since lintr only uses R it should be safe to use the newer version.

AshesITR:
@jimhester There used to be a Cancel / Re-Run all Checks button, but it seems to have been removed during the moving of lintr to r-lib. Is there anyone who can re-run the checks manually? Or can we skip the pending check which seems stuck on the setup-r step?

jimhester:
@hadley or @jennybc, could you set up me to have access to the repo settings for lintr, so I can adjust the permissions for contributors? It looks like I currently only have contributor access to this repo, so can't access the setting tab myself.

jimhester (dismissed):

jennybc:
OK @jimhester I just switched your role on this repo from "Write" to "Admin". The other role I pondered was "Maintain". If we end up needing to do similar on other repos, it would be worth figuring out which role is the best fit.

https://github.com/r-lib/lintr/settings/role_details

jennybc:
@AshesITR FYI dev usethis is now bringing in `v2` workflows from <https://github.com/r-lib/actions>, whereas this PR is using `v1`.

AshesITR:
Thanks for the info. Are there some key advantages compared to the old template? For example, I noticed some hangs and other irreproducible runs with the R-CMD-check actions lately.

jennybc:
I too have noticed hangs (esp. on macOS) with `v1`, which has lead to me updating to `v2` on things I'm actively working on.

The best account of what's going on with `v2` (and how it compares to `v1`) is here:
https://github.com/r-lib/actions/releases/tag/v2

AshesITR:
@jimhester I've switched to deveolpment usethis templates (actions@v2) and upgraded our custom workflows to v2 as well.
I noticed the linting workflows ran on macOS-latest (which was causing unstable builds).
Since ubuntu-20.04 builds are generally faster than macOS-builds, I switched the linting workflows to ubuntu-20.04.

`use_tidy_github_actions()` also set the `informational: true` keys in `codecov.yml`.
We had a few PRs in the past where codecov blocked merging and the remedy was `#nocov`ing the affected lines, or where coverage decreased simply because legacy code with 100% coverage was removed.
Let me know if the codecov settings should be reverted to auto-block pull requests that reduce coverage.

AshesITR:
@jimhester did you have time to review the changes?

jimhester (approved):

PR #910, opened 2022-03-04
Start using testthat 3e

MichaelChirico:


MichaelChirico:
We have a bunch of migrations to do here..

MichaelChirico:
Finished the first pass here. Notes:

 - Couldn't finish all the `with_mock()` migrations. I don't understand how to use `mockery()` for some of them. I raised a comment on a similar `mockery` issue here: https://github.com/r-lib/mockery/issues/43. Any ideas?
 - It looks like the code tested `test-cache` has been refactored -- the `with_mock()` appear to be no-ops, so I just removed them. Let's see.
 - Also in test-cache, `expect_equal(e1, e2)` became the much uglier `expect_identical(as.list(e1), as.list(e2))`. Raised https://github.com/r-lib/waldo/issues/127 about it.
 - Also did some more general clean-ups of tests, e.g. preferring `withr::local_*` functions over `withr::with_*` and using withr more comprehensively
 - The tests in test-get_source_expressions.R are in incomprehensible mess IMO (and I've only made it worse with the `expect_identical()` shift) -- we should file a follow-up to refactor it to a more maintainable approach

tests/testthat/test-cache.R
```diff
@@ -435,18 +411,14 @@ test_that("lint with cache uses the provided relative cache directory", {
...
-
-    path <- tempfile(pattern = "my_cache_dir_"),
-    expect_false(dir.exists(path)),
-    expect_lint("a <- 1", NULL, linter, cache = path),
-    expect_true(dir.exists(path)),
-    expect_length(list.files(path), 1),
-    expect_lint("a <- 1", NULL, linter, cache = path),
-    expect_true(dir.exists(path))
-  )
+  mockery::stub(lintr:::find_default_encoding, "find_package", function(...) NULL)
```
AshesITR, line 414: Can we ensure this is the only call site that needs it?
Here are all usages of `find_package` in lintr:
image

tests/testthat/test-knitr_formats.R
```diff
@@ -12,9 +12,9 @@ test_that("it handles dir", {
     parse_settings = FALSE
   )
   has_lints <- length(lints) > 0
-  testthat::expect(has_lints, "There should be lints")
+  expect_true(has_lints, info="There should be lints")
 
-  testthat::expect_equivalent(length(unique(names(lints))), 6, info="For every file there should be at least 1 lint")
+  expect_identical(length(unique(names(lints))), 6L, info="For every file there should be at least 1 lint")
```
AshesITR, line 17: nit: There should be spaces around `=`. Also, wondering whether it's worth replacing `6L` by `length(list.files(...))` to make it more robust against the unlikely case that {knitr} gets a new format, we add it and forget about this test while it passes because linting the new file returns no lints...

tests/testthat/test-rstudio_markers.R
```diff
@@ -10,11 +13,8 @@ test_that("it returns markers which match lints", {
     ),
     class = "lints"
   )
-  with_mock(
-    `rstudioapi::callFun` = function(...) list(...),
-    `rstudioapi::executeCommand` = function(...) NULL,
-    marker1 <- rstudio_source_markers(lint1)
-  )
+
+  marker1 <- rstudio_source_markers(lint1)
```
AshesITR, line 17: Checking this led me to notice Commit d657e42ad95828dec06d9a147508a55b84727858 introduced a name clash in `print.lints`. Can anyone with RStudio on hand confirm `print.lints` no longer working since that commit?

AshesITR (commented):
Looks fine so far

tests/testthat/test-rstudio_markers.R
AshesITR: nvm. To my surprise, R allows such clashes, although very misleading:

```r
x <- function() cat("ok\n")
local({x <- 2; x()})
# > ok
```

MichaelChirico: yes for calls R essentially does get(mode = 'function')

tests/testthat/test-cache.R
MichaelChirico: my approach was: replace the `with_mock` function with `stop()` and look at the call stack. The public function call is `lint()` but `mockery::stub()` + `depth=` never seem to work for me, so I stubbed the private function instead. Tests are passing so that should be good?

tests/testthat/test-knitr_formats.R
MichaelChirico: I think tests overall could go for some house-keeping. I'll leave it out of scope for this PR; filing a follow-up.

MichaelChirico: https://github.com/r-lib/lintr/issues/918

tests/testthat/test-rstudio_markers.R
MichaelChirico: Agree it's better to avoid this behavior, but also out of scope for this PR; saw the follow-up, thanks.

tests/testthat/test-cache.R
AshesITR: This test passing doesn't guarantee it passes for the right reason (i.e. in this context with no package to be found) IINM.
Can you check if this particular test passes if you manually replace the actual `find_package` function by `stop`?
e.g. something like 

```
devtools::load_all()
library(testthat)
environment(find_package)$find_package <- function(...) { message("Tripped!"); NULL }

# Test code
linter <- assignment_linter()

with_mock(
  `lintr::find_package` = function(...) NULL,
  `lintr::pkg_name` = function(...) NULL,

  path <- tempfile(pattern = "my_cache_dir_"),
  expect_false(dir.exists(path)),
  expect_lint("a <- 1", NULL, linter, cache = path),
  expect_true(dir.exists(path)),
  expect_length(list.files(path), 1),
  expect_lint("a <- 1", NULL, linter, cache = path),
  expect_true(dir.exists(path))
)
# > (no output)

# Verify trap works
expect_lint("a <- 1", NULL, linter, cache = path)
# > Tripped!
```

AshesITR: I've gone through the full list now and I'm positive this was correct.

lint.R:273 is `lint_package()`- not called
lint.R:415 is `pkg_name()`, used by `namespace_linter()`, `object_name_linter()` and `object_usage_linter()` - but not by the `assignment_linter()` used for the test
namespace.R:2 is `namespace_linter()` - not called
object_name_linters.R:83 is `object_name_linter()` - not called
object_usage_linter.R:16 is `object_usage_linter()` - not called
settings.R:111 is `find_default_encoding()` - mocked correctly

tests/testthat/test-knitr_formats.R
AshesITR: Thanks for filing the follow-up. Can we delint those lines anyway?

tests/testthat/test-cache.R
```diff
@@ -53,116 +53,92 @@ test_that("clear_cache deletes the directory if no file is given", {
...
+  save_cache(cache = e1, file = f1, path = d1)
 
-    e2 <- new.env(parent = emptyenv()),
-    load(file = file.path(d1, fhash(f1)), envir = e2),
+  e2 <- new.env(parent = emptyenv())
+  load(file = file.path(d1, fhash(f1)), envir = e2)
 
-    expect_equal(e1, e2)
-  )
+  expect_identical(as.list(e1), as.list(e2))
```
hadley, line 141: If you use this approach, I think you need `all.names = TRUE` to verify the property implied by the test title.

OTOH I'm not sure what "non-hidden" means as I would have expected hidden to imply a variable starting with `.`, but that doesn't appear in the test.

Additionally, I'd use `expect_equal()` here because `expect_identical()` just implies exact numerical comparisons, which I don't think is important here.

tests/testthat/test-cache.R
```diff
@@ -53,116 +53,92 @@ test_that("clear_cache deletes the directory if no file is given", {
...
-    expect_equal(e2[["x"]], "foobar")
-  )
+  e1 <- new.env(parent = emptyenv())
+  assign("x", "foobar", envir = e1)
+  d1 <- tempfile(pattern = "lintr_cache_")
+  f1 <- "R/test.R"
+  save_cache(cache = e1, file = f1, path = d1)
+  e2 <- load_cache(file = f1, path = d1)
+
+  expect_equal(ls(e2), "x")
```
hadley, line 63: Ideally this would just be `expect_equal(e2, e1)`

AshesITR: Good to know that `expect_equal()` tolerates different environments as long as their contents match.

MichaelChirico: not sure that's true -- hence using as.list() in the other test. I'll investigate further.

tests/testthat/test-cache.R
MichaelChirico: that's not my reading of the docs (?expect_identical):

> expect_identical() is the baseline; expect_equal() relaxes the test to ignore small numeric differences.

i.e., only use expect_equal to ignore numerical differences

tests/testthat/test-knitr_formats.R
MichaelChirico: will do

tests/testthat/test-cache.R
hadley: Hmmm. I guess it missing another sentence that says something like: "Since in most cases tiny numeric differences aren't important, we recommend using `expect_equal()`."

MichaelChirico: :see_no_evil: our internal policy encourages `expect_identical()` now and we've basically switched all the `expect_equal()` tests already, except where `ignore*=` or `tolerance=` are needed. I have come to quite like it TBH. It encourages stronger tests.

MichaelChirico:
Hmm @AshesITR the new helper function uses `rlang` as suggested in the `testthat` vignette: https://testthat.r-lib.org/articles/custom-expectation.html

That means we need a Suggests for `rlang`, but `testthat` already Imports `rlang`, so our total dep tree doesn't really change. WDYT?

tests/testthat/test-cache.R
```diff
@@ -53,116 +53,93 @@ test_that("clear_cache deletes the directory if no file is given", {
...
-    expect_equal(e2[["x"]], "foobar")
-  )
+  e1 <- new.env(parent = emptyenv())
+  e1[["x"]] <- "foobar"
+  d1 <- tempfile(pattern = "lintr_cache_")
+  f1 <- "R/test.R"
+  save_cache(cache = e1, file = f1, path = d1)
+  e2 <- load_cache(file = f1, path = d1)
+
+  expect_environments_equal(e2, e1)
```
AshesITR, line 63: Why not stick with `expect_equal(as.list(e1), as.list(e2))` and add a `# TODO` once r-lib/waldo/pull/133 makes it to CRAN?

MichaelChirico: sure, done

AshesITR (approved):
Nice work, thanks!

PR #923, opened 2022-03-11
run doc generation in C locale

MichaelChirico:
Not the greatest that caps come before lowercase counterparts but I think it's the most consistent we can hope for.


AshesITR:
Is using `en_US.UTF-8` an option?
Not sure how Windows would handle it, but I'd be willing to try it out.

MichaelChirico:
UTF-8 of course will present issues for all but the UCRT windows builds (not sure if silently ignored).

OTOH as long as we get something that works for you and me it should be OK 😅

I can try `English`, but note that the answer mentions `LC_ALL` and `local_locale()` doesn't allow setting LC_ALL. I'll see if it works for plain LC_COLLATE.

MichaelChirico:
On my work computer `LC_COLLATE = 'English'` won't work:

```
48: In (function (category = "LC_ALL", locale = "")  ... :
  OS reports request to set locale to "English" cannot be honored
```

MichaelChirico:
OK, `en_US` seems to work well enough. Not sure what would happen on Windows.

AshesITR:
I will check what locale can be used on windows.
I'm assuming `en_US.UTF-8` is available on 99% of linux distros and macOS, so if we get `if (in_windows()) "Engish" else "en_US.UTF-8"` or something like that to work, I think it's worth the better sorting result.

MichaelChirico:
I do see this mentioning `en_US` on Windows, so maybe that's good (I don't think the UTF part will matter for the purposes of the linter docs):

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170

AshesITR:
No luck :(

```
> usethis::pr_fetch(923)
✔ Setting active project to '/home/alex/Documents/workspace/lintr'
✔ Checking out PR 'r-lib/lintr/#923' (@MichaelChirico): 'run doc generation in C locale'.
✔ Creating and switching to local branch 'tag-sort'.
✔ Setting 'upstream/tag-sort' as remote tracking branch.
> devtools::document()
ℹ Updating lintr documentation
ℹ Loading lintr
Writing NAMESPACE
Writing NAMESPACE
Writing linters.Rd
```

AshesITR:
Windows complains about
```
Warning in (function (category = "LC_ALL", locale = "")  :
  OS reports request to set locale to "en_US" cannot be honored
```

And also sorts `_` before `n` as linux does.
Can you provide more info on your locale? (maybe `sessionInfo()` from a fresh R session?)
That might help narrow down what could work and what wouldn't.

AshesITR:
https://stackoverflow.com/questions/13370051/ordering-differences-between-postgres-instances-on-different-machines-same-loca

Possibly related to our problem.

AshesITR:
@MichaelChirico What does `stringi::stri_sort(c("N", "n", "_"))` do on your system?
I get `c("_", "n", "N")` on windows regardless of LC_COLLATE.

MichaelChirico:
I get `c("_", "n", "N")` but I thought that was the issue?

In the PR the diff we reverted was

```
 \item{\code{\link{nonportable_path_linter}} (tags: best_practices, configurable, robustness)}
+\item{\code{\link{no_tab_linter}} (tags: consistency, default, style)}
 ```

which is the sort order `_` then `n`. `LC_COLLATE=C` gives `N/_/n`

**Comment by MichaelChirico** [comment]
Oh I think I'm reading the diff in reverse 😵‍💫 

This is really strange behavior 🤔 

```
sort(c("_", "n", "N"))
# [1] "_" "n" "N"
sort(c("no_", "non", "noN"))
# [1] "no_" "non" "noN"
sort(c("no_t", "nont", "noNt"))
# [1] "nont" "noNt" "no_t"
```

File: R/linter_tags.R [hunk 1]
```diff
@@ -157,17 +157,23 @@ rd_taglist <- function() {
...
+      tags <- platform_independent_sort(linters[["tags"]][[match(linter_name, linters[["linter"]])]])
       paste0("\\item{\\code{\\link{", linter_name, "}} (tags: ", toString(tags), ")}")
     }, character(1L)),
     "}", # itemize
     "}" # section
   )
 }
+
+platform_independent_sort <- function(x) {
+  stopifnot(`Package {withr} must be installed.` = requireNamespace("withr", quietly = TRUE))
```
MichaelChirico, line 176: ah I'm glad this doesn't cause an error  in R<4.

AshesITR:
Strange indeed. `stringi` is quite heavyweight but since we only need it in Suggests, we could go for `stri_sort()`, WDYT?

MichaelChirico:
I've had headaches before in CI because of `stringi`, it would be nice to avoid it. Let's play around with some other options first...

MichaelChirico:
It looks like this is the issue -- `_` is being ignored.

I don't know how to turn it off.

https://superuser.com/questions/226449/why-does-sort-ignore-special-characters-like-the-asterisk

On my personal machine I don't reproduce the issue in R (:monocle_face:) but I do reproduce it in the terminal, c.f.:

```
$ printf "nont\nnoNt\nno_t\n" | sort
nont
noNt
no_t
$ printf "nont\nnoNt\nno_t\nno_a\n" | sort
no_a
nont
noNt
no_t
$ printf "n\nN\n_\n" | sort
_
n
N
```

MichaelChirico:
Hacky, but we could take the order of `gsub("[^[:alphanumeric:]]", "0", x)` or simply `gsub("_", "0", x, fixed = TRUE)` instead. Should be enough for our use case.

R/linter_tags.R
```diff
@@ -157,17 +157,26 @@ rd_taglist <- function() {
...
+      tags <- platform_independent_sort(linters[["tags"]][[match(linter_name, linters[["linter"]])]])
       paste0("\\item{\\code{\\link{", linter_name, "}} (tags: ", toString(tags), ")}")
     }, character(1L)),
     "}", # itemize
     "}" # section
   )
 }
+
+platform_independent_sort <- function(x) {
+  if (!identical(sort(c("a", "B")), c("a", "B"))) {
```
MichaelChirico, line 176: I figure this is "good enough" for the purposes of what we want here -- if a dev wants to generate docs, just make sure their locale is "english enough".

AshesITR:
If we use `tolower(gsub("_", "0", x))` as the sort key, LC_COLLATE=C should work and we don't need to stop, or?

MichaelChirico:
That sounds right to me. caps shouldn't really come up for any linter besides `T_and_F_symbol_linter` anyway due to the naming conventions.

R/linter_tags.R
```diff
@@ -157,17 +157,26 @@ rd_taglist <- function() {
...
+    stop("Please run in an English-like locale, e.g. LC_COLLATE=en_US")
+  }
+  # see #923 -- some locales ignore _ when running sort(), others don't.
```
MichaelChirico, line 176: weird that this triggers `commented_code_linter`. seems like a bug.

AshesITR: Probably because `see #...` is parseable and the text after "#" contains code-ish chars (`sort()`).
`see issue #...` should suppress the lint.

AshesITR (approved):

R/linter_tags.R
MichaelChirico: still sounds like a false positive to me. filing for followup

PR #924, opened 2022-03-12
expect_type_linter

MichaelChirico:
Part of #884


tests/testthat/test-expect_type_linter.R
```diff
@@ -0,0 +1,54 @@
...
+
+
+local({
+  # test for lint errors appropriately raised for all is.<type> calls
+  is_types <- c(
+    "raw", "logical", "integer", "double", "complex", "character", "list",
+    "numeric", "function", "primitive", "environment", "pairlist", "promise",
+    "language", "call", "name", "symbol", "expression"
+  )
+  patrick::with_parameters_test_that(
```
MichaelChirico, line 44: We use [`patrick`](https://github.com/google/patrick) to handle parameterized tests. This particular case could be refactored into a `for` loop if we'd like to avoid the dep, but there will be other cases where it provides a cleaner solution.

R/expect_type_linter.R
```diff
@@ -0,0 +1,59 @@
...
+    xml <- source_file$xml_parsed_content
+
+    base_type_tests <- xp_text_in_table(paste0("is.", base_types))
+    xpath <- glue::glue("//expr[
+      (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical']
+        and following-sibling::expr[1][expr[SYMBOL_FUNCTION_CALL[text() = 'typeof']]]
+      ) or (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_true']
+        and following-sibling::expr[1][expr[SYMBOL_FUNCTION_CALL[ {base_type_tests} ]]]
```
MichaelChirico, line 27: here's a good use case of glue I think. with `sprintf()` we'll have to visually scan for `%s` in the long string, and formatting for multiline strings is odd with `sprintf`, e.g.

```
sprintf("
a string %s
", "the string")

# or
sprintf("the string", fmt = "
a string %s
")
```

both are somewhat awkward just to avoid `glue()` IMO

inst/lintr/linters.csv
```diff
@@ -1,44 +1,45 @@
...
 spaces_left_parentheses_linter,style readability default
+sprintf_linter,correctness common_mistakes
 T_and_F_symbol_linter,style readability robustness consistency best_practices default
 todo_comment_linter,style configurable
 trailing_blank_lines_linter,style default
 trailing_whitespace_linter,style default
 undesirable_function_linter,style efficiency configurable robustness best_practices
 undesirable_operator_linter,style efficiency configurable robustness best_practices
 unneeded_concatenation_linter,style readability efficiency
-assignment_spaces_linter,style readability
```
MichaelChirico, line 37: went ahead and sorted everything. will make it easier to convert linters with a script.

man/linters.Rd
```diff
@@ -52,8 +53,8 @@ The following linters exist:
 \item{\code{\link{missing_argument_linter}} (tags: common_mistakes, configurable, correctness)}
 \item{\code{\link{missing_package_linter}} (tags: common_mistakes, robustness)}
 \item{\code{\link{namespace_linter}} (tags: configurable, correctness, robustness)}
-\item{\code{\link{no_tab_linter}} (tags: consistency, default, style)}
```
MichaelChirico, line 55: same issue... will revert at the end again 😞

R/expect_type_linter.R
```diff
@@ -0,0 +1,60 @@
...
+#'
+#' [testthat::expect_type()] exists specifically for testing the storage type
+#'   of objects. [testthat::expect_equal()], [testthat::expect_identical()], and
+#'   [testthat::expect_true()] can also be used for such tests,
+#'   but it is better to use the tailored function instead.
+#'
+#' @evalRd rd_tags("expect_type_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+#' @importFrom glue glue
```
AshesITR, line 11: That's not necessary if you call `glue::glue()` with its namespace.

R/expect_type_linter.R
```diff
@@ -0,0 +1,60 @@
...
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    base_type_tests <- xp_text_in_table(paste0("is.", base_types))
+    xpath <- glue::glue("//expr[
+      (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical']
+        and following-sibling::expr[1][expr[SYMBOL_FUNCTION_CALL[text() = 'typeof']]]
```
AshesITR, line 24: Can we use `position() <= 2` here as well?

R/expect_type_linter.R
AshesITR: okay

R/expect_type_linter.R
MichaelChirico: Here I'd prefer not... I can imagine legitimate tests where you're testing the output of a function versus `typeof()`. Something like:

```
expect_equal(GeneratesARawTimestampCheck(x), typeof(Sys.time()))
```

(maybe not the best example, but hopefully conveys my point)

We might prefer `typeof(Sys.time())` instead of `"double"` directly because `"double"` is sort of an implementation detail, and what we really care about is that we match whatever `Sys.time()` is doing, even if that should change.

(maybe a better example would b testing the `typeof()` output of a function in a custom package). 

(maybe a good example is code that produces a factor under R<4 and a string under R>=4, and `typeof()` might provide a concise way of checking our output matches whatever the underlying R version is doing)

R/expect_type_linter.R
MichaelChirico: I added this hoping it would make `roxygen2()` automatically add `glue` to the DESCRIPTION which didn't work. i'll remove it

R/expect_type_linter.R
AshesITR: Good point but doesn't that mean we should check the other argument matches a `STR_CONST` for the lint?

MichaelChirico: I don't think so, `expect_type(expr)` is fine, so `expect_equal(typeof(x), expr)` should still lint.

AshesITR: Wait, I'm confused.
What exactly needs linting and what would be a false positive?

`expect_equal(typeof(x), y)`, `expect_equal(y, typeof(x))`, `expect_equal(typeof(x), typeof(y))`, `expect_equal(typeof((x), "constant")`, `expect_equal("constant, typeof((x))`

MichaelChirico: I think all of those _should_ use expect_type, but we are conservative and some would be false negatives.

we could check for string constant in the first argument to cover some more things, WDYT?

MichaelChirico: I see that's what you already suggested. brain fart 🙂

AshesITR: But what would a false positive be concretely?
I mean all of the examples could be converted to an equivalent `expect_type()` test, no?

MichaelChirico: in turn:

```
expect_equal(typeof(x), y) # lints already
expect_equal(y, typeof(x)) # could be made to lint
expect_equal(typeof(x), typeof(y)) # lints already
expect_equal(typeof(x), "constant") # lints already
expect_equal("constant", typeof(x)) # lints now
```

Not sure how worth it is to catch the `expect_equal(y, typeof(x))` case. There might be a false positive there I'm not thinking of. e.g. if someone prefers to write tests like `expected <- ...; expect_equal(expected, typeof(x))` and it falls into one of the cases mentioned about where we don't want to lint `typeof()` in the 2nd argument.

I'd prefer to leave it as it is now where we just look for `STR_CONST`.

AshesITR: I think we should preserve symmetry WRT the test arguments (i.e., either lint both forms or no forms).
If one form has false positives, so will the other.

MichaelChirico: I don't think the arguments are symmetric, though... the signature is not `x, y` but `object, expected` -- a clear asymmetry in the intended usage.

I'm OK with checking for `STR_CONST` in the first argument because it's a case where we can see a clear violation of the intended usage, without any chance for false positives. So the asymmetry in that case is an exception.

AshesITR: Allright. I think I finally understood a false positive could be `expect_equal(my_better_typeof(x), typeof(x))` where `expect_type(x, my_better_typeof(x))` is counter-intuitive.

MichaelChirico: thanks for pressing... agree it's a bit hard to think of false positives for this linter in particular. but this paradigm will also apply to other testthat linters so it's good to address here.

NAMESPACE
```diff
@@ -72,6 +73,7 @@ export(with_defaults)
 export(with_id)
 import(rex)
 importFrom(cyclocomp,cyclocomp)
+importFrom(glue,glue)
```
AshesITR, line 76: Forgot to `document()`?

tests/testthat.R
```diff
@@ -1,4 +1,5 @@
 library(testthat)
+library(patrick)
```
AshesITR, line 2: Why `library()` if used only by namespace?

MichaelChirico: make it clearer which package is needed by the test suite. I guess we could add a skip_if_not_installed to that file but the current approach is preferable to me

R/expect_type_linter.R
```diff
@@ -0,0 +1,62 @@
...
+#' @export
+expect_type_linter <- function() {
+  Linter(function(source_file) {
+    if (length(source_file$parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    base_type_tests <- xp_text_in_table(paste0("is.", base_types))
```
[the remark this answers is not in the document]
MichaelChirico: yea I think that's right... same issue will show up with glue a lot, so wondering if we should try and address it as a linter bug.

tests/testthat.R
AshesITR: OTOH `skip_if_not_installed()` would also be fine since the R CMD check workflows install all suggested packages so don't skip the tests.

R/expect_type_linter.R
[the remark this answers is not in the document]
AshesITR: Seems very hard to get right (`stringi::stri_interp()` would also need the fix, and we need to check what code delimiters are used)

BTW: wouldn't this still be reported by `R CMD check`? AFAIK that also runs `codetools::checkUsage()` on the code...

tests/testthat.R
MichaelChirico: yes, fail fast is the goal. testthat and Patrick are on basically equal footing as deps as I see it.

the package will also be available on CRAN. using the namespace is mainly a polite way to convey to readers that this is not part of testthat, otherwise we could just use library and not namespace-qualify.

so I would either leave it as is, or use a skip (with patrick tests always at the bottom of a given file)

AshesITR: Then let's go for a `skip`.

R/expect_type_linter.R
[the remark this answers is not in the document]
MichaelChirico: Per the docs I think we can ignore `str_interp()`:

https://stringr.tidyverse.org/reference/str_interp.html

> str_interp() is superseded in favour of [str_glue()](https://stringr.tidyverse.org/reference/str_glue.html).

I'll file a follow-up to handle glue and nolint for now.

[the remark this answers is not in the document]
MichaelChirico: Filed #942.

AshesITR (approved):

PR #926, opened 2022-03-13
add a skip for when the test might fail

MichaelChirico:
Alternatively we could (1) pick a more obscure function name like `__lintr_missing_fun` (2) loop over the letters and pick the first one that fails `exists()`


AshesITR:
I think using an obscure name is the most elegant solution.

AshesITR (approved):

PR #233, opened 2017-04-27
expect_lint() improvement

fangly:

PR #928, opened 2022-03-13
extend assignment_linter for other operators

### LINKED ISSUE #915: Cover other assignment operators in assignment_linter
Follow-up to 

https://github.com/AshesITR/lintr/pull/1#discussion_r820137490

I think `->` should be covered by assignment_linter, and we can include an option for whether to cover `<<-` and `->>` as well

MichaelChirico:
Closes #915


MichaelChirico:
@AshesITR PTAL at the test failure. Looks odd to me. For now I just removed the test.

At HEAD, `assignment_linter()` generates a lint with `ranges=NULL`:

```
dput(lintr::lint("a=1\n", lintr::assignment_linter()))
# structure(list(structure(list(filename = "<text>", line_number = 1L, 
#     column_number = 2L, type = "style", message = "Use <-, not =, for assignment.", 
#     line = c(`1` = "a=1"), ranges = NULL, linter = "lintr"), class = "lint")), class = "lints")
```

vs. in this PR it's `ranges = list(c(2L, 2L))`

```
dput(lintr::lint("a=1\n", lintr::assignment_linter()))
# structure(list(structure(list(filename = "<text>", line_number = 1L, 
#     column_number = 2L, type = "style", message = "Use <-, not =, for assignment.", 
#     line = c(`1` = "a=1"), ranges = list(c(2L, 2L)), linter = "lintr"), class = "lint")), class = "lints")
```

What's puzzling is the test is checking that exactly the new value of `ranges` _isn't_ produced by `assignment_linter()`:

https://github.com/r-lib/lintr/blob/6f55ef52f6112116dad9c80262c035e121e84465/tests/testthat/test-expect_lint.R#L28

So I'm second guessing if there's a reason for that? I can't think of any. And `ranges` will do well for printing the lint of `->` (and `<<-` and `->>`):

```
lintr::lint("a->1\n", lintr::assignment_linter())
# <text>:1:2: style: Use <-, not ->, for assignment.
# a->1
#  ^~
```

There's [this comment](https://github.com/r-lib/lintr/pull/233#issuecomment-299455726) in the original PR:

> I found out that it was practically impossible to check the "ranges" of a lint using `expect_lint()`.

tests/testthat/test-assignment_linter.R
```diff
@@ -21,3 +21,14 @@ test_that("returns the correct linting", {
     linter
   )
 })
+
+test_that("other operators are caught as well", {
+  expect_lint("1 -> blah", rex::rex("Use <-, not ->, for assignment."), assignment_linter())
+
+  # <<- and ->> are only blocked optionally
+  expect_lint("1 <<- blah", NULL, assignment_linter())
+  expect_lint("1 ->> blah", NULL, assignment_linter())
```
AshesITR, line 30: Seems odd to disallow `->` and allow `->>`.
Maybe better to change the configuration to

`allow_parent_assign = TRUE, allow_right_arrow = FALSE`

WDYT?

tests/testthat/test-expect_lint.R
```diff
@@ -25,7 +25,7 @@ test_that("single check", {
 
   expect_error(expect_lint("a=1", c(message = msg, lineXXX = 1L), linter), "invalid field")
 
-  expect_failure(expect_lint("a=1", list(ranges = list(c(2L, 2L))), linter))
+  expect_success(expect_lint("a=1", list(ranges = list(c(2L, 2L))), linter))
```
AshesITR, line 28: What was the original failure?

R/assignment_linter.R
```diff
@@ -2,23 +2,36 @@
 #'
 #' Check that `<-` is always used for assignment.
 #'
+#' @param block_double_assign Logical, `FALSE` by default. If `TRUE`, usage
+#'   of [`<<-`][base::assignOps] and [`->>`][base::assignOps] are also blocked.
 #' @evalRd rd_tags("assignment_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-assignment_linter <- function() {
+assignment_linter <- function(block_double_assign = FALSE) {
```
AshesITR, line 10: IIRC we usually have `allow_*` flags. Also, I'd call this `allow_parent_assign` since double assign could be

```
a <- 1
a <- 2
```

(which is also pointless but harder to prove pointless)

R/assignment_linter.R
```diff
@@ -2,23 +2,36 @@
...
+    node_conditions <- paste0("self::", c(
+      # always block = (NB: the parser differentiates EQ_ASSIGN and EQ_SUB)
+      "EQ_ASSIGN",
+      # -> and ->> are both 'RIGHT_ASSIGN'; check the text if not blocking ->>
+      paste0("RIGHT_ASSIGN", if (!block_double_assign) "[text() = '->']"),
+      # <-, :=, and <<- are all 'LEFT_ASSIGN'; check the text if blocking <<-
+      if (block_double_assign) "LEFT_ASSIGN[text() = '<<-']"
+    ))
+    # TODO: is there any performance consideration for //*[self::a or self::b] vs. //a|//b ?
+    xpath <- sprintf("//*[%s]", do.call(xp_or, as.list(node_conditions)))
```
AshesITR, line 25: `paste0(..., collapse = " or ")` with the conditions vector (and all conditions being fully qualified XPaths) should look more readable.

MichaelChirico: OK will do. The advantage (not relevant here) of `xp_or` is it always wraps the conditions in parentheses to avoid precedence issues like `A and B or c`; the `paste` version of that is uglier.

Maybe we could just patch `xp_or` to "just work" when receiving a vector? `xp_or(c('a', 'b'))` is the same as `xp_or('a', 'b')`, WDYT?

MichaelChirico:
ah poop. git is hard.

R/assignment_linter.R
AshesITR: Yeah, would be fine.
However, here `"//EQ_ASSIGN | //RIGHT_ASSIGN[...] | //LEFT_ASSIGN"` seems more elegant than `"//*[EQ_ASSIGN or ...]"`

R/assignment_linter.R
```diff
@@ -2,23 +2,36 @@
...
+
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+    lapply(bad_expr, gen_assignment_lint, source_file)
   })
 }
+
+gen_assignment_lint <- function(expr, source_file) {
+  operator <- xml2::xml_text(expr)
+  message <- sprintf("Use <-, not %s, for assignment.", operator)
+  if (operator %in% c("<<-", "->>")) message <- paste(message, "Assign to specific environments instead.")
```
AshesITR, line 35: Not sure about the lint message.
If we don't want `<<-` then it's usually because we don't want to use the operator for its unpredictable behaviour.
"Use <-, not <<- for assignment. Assign to specific environments instead." seems to miss this point.

What do you think about "Do not use <<- for assigning to parent environments, use assign() instead."?

MichaelChirico: personally I stopped using assign too 😅

I replace assign("key", val, env) with env[["key"]] <- val instead.

anyway I'll take a pass at better customizing the message

AshesITR: I like `assign()` because `[[<-` is a very common thing for lists...

R/assignment_linter.R
MichaelChirico: used `allow_cascading_assign`

R/assignment_linter.R
MichaelChirico: true; done

tests/testthat/test-expect_lint.R
MichaelChirico: > Error: check #1: ranges NULL did not match list(c(2L, 2L))

R/assignment_linter.R
```diff
@@ -2,27 +2,27 @@
...
-      "EQ_ASSIGN",
-      # -> and ->> are both 'RIGHT_ASSIGN'; check the text if not blocking ->>
-      paste0("RIGHT_ASSIGN", if (!block_double_assign) "[text() = '->']"),
+    # TODO: is there any performance consideration for //*[self::a or self::b] vs. //a|//b ?
+    xpath <- paste(collapse = " | ", c(
+      # always block = (NB: the parser differentiates EQ_ASSIGN, EQ_SUB, and EQ_FORMALS)
+      "//EQ_ASSIGN",
+      # -> and ->> are both 'RIGHT_ASSIGN'
+      if (!allow_right_assign) "//RIGHT_ASSIGN" else if (!allow_cascading_assign) "//RIGHT_ASSIGN[text() = '->>']",
       # <-, :=, and <<- are all 'LEFT_ASSIGN'; check the text if blocking <<-
```
AshesITR, line 23: Shouldn't we lint `:=` as well?

tests/testthat/test-expect_lint.R
AshesITR: So that seems like a way of documenting the bug we just fixed 😆 
Feel free to remove the test.

AshesITR:
Looks mostly good, just a few finishing touches needed.
Nice work!

R/assignment_linter.R
MichaelChirico: We'd have to write special logic to handle correct `rlang` and `data.table` usages. Don't think it's worth it (most people don't know that `:=` can be used for "regular" assignment and it's explicitly not mentioned in the R documentation).

tests/testthat/test-expect_lint.R
MichaelChirico: Despite not really knowing the purpose of the test I'm still a bit loath to remove it :laughing:

I changed it to use a different linter that gives `NULL` for `ranges`

NEWS.md
```diff
@@ -88,6 +88,7 @@ function calls. (#850, #851, @renkun-ken)
 * `lintr` now uses the 3rd edition of `testthat` (@MichaelChirico, #910)
 * `lintr` is adopting a new set of linters provided as part of Google's extension to the tidyverse style guide (#884, @michaelchirico)
    + `expect_null_linter()` Require usage of `expect_null(x)` over `expect_equal(x, NULL)` and similar
+* `assignment_linter()` now blocks right assignment (`->` and `->>`) and gains two arguments. `allow_cascading_assign` (`TRUE` by default) toggles whether to block `<<-` and `->>`; `allow_right_assign` toggles whether to block `->` and `->>` (#915, @michaelchirico)
```
AshesITR, line 91: block --> lint

R/assignment_linter.R
AshesITR: Agreed. Maybe worth mentioning here why we chose not to lint `:=`?

AshesITR (approved):

Issue #930, opened 2022-03-13
infix_spaces_linter should include tilde / formula operator and = in function definitions


MichaelChirico:
It's covered by `{styler}`:

```
styler::style_text("a~b")
# a ~ b
```

But not `infix_spaces_linter()`:

```
lintr::lint("a~b\n", infix_spaces_linter())
```

This is called out by omission in the style guide, which talks about one-sided formulas:

https://style.tidyverse.org/syntax.html#infix-operators

MichaelChirico:
`lintr` also misses `=` used in function definitions:

```
lintr::lint("foo <- function(a=1, b=2) { }\n", lintr::infix_spaces_linter())
```

MichaelChirico:
This was handled in #931

PR #931, opened 2022-03-13
extend infix_spaces_linter to be more flexible & correct

### LINKED ISSUE #914: Subsume assignment_spaces_linter as part of infix_spaces_linter
Follow-up from

https://github.com/AshesITR/lintr/pull/1#discussion_r820137602

is there any reason to treat `<-` differently than other infix operators?

is it sufficient to just toggle which operators to exclude with an argument to the linter?

MichaelChirico:
Closes #914 #930


tests/testthat/test-infix_spaces_linter.R
```diff
@@ -2,15 +2,19 @@ test_that("returns the correct linting", {
...
     "!=",
     "<=",
     ">=",
     "<-",
+    ":=",
+    "<<-",
     "<",
     ">",
     "->",
+    "->>",
```
AshesITR, line 17: I just noticed that the test suite tests `1 <- 2` etc.
IMO we should add sensible test cases for the assignment operators.

MichaelChirico: I copy-pasted the tests for the deleted `assignment_spaces_linter` here. I also removed a bunch of them that were throwing lints for using >1 space, because the style guide says that's OK _if it's used to improve alignment_ and our logic is (and I think should continue to be) not sophisticated enough to detect alignment issues:

https://style.tidyverse.org/syntax.html#extra-spaces

> Adding extra spaces is ok if it improves alignment of `=` or `<-`.

AshesITR: Since @f-ritter's original use-case was disallowing these, I think we should also support that rule via an argument (`allow_multiple_spaces = TRUE`). WDYT?

R/infix_spaces_linter.R
```diff
@@ -1,43 +1,81 @@
...
+#'   are included/excluded as a group (indicated by passing `"<-"`), as are `->` and `->>` (_viz_, `"->"`),
+#'   and that `=` for assignment and for setting arguments in calls are treated the same.
 #' @evalRd rd_tags("infix_spaces_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-infix_spaces_linter <- function() {
+infix_spaces_linter <- function(exclude_operators = NULL) {
   Linter(function(source_file) {
+    infix_tokens <- infix_metadata[
+      with(infix_metadata, low_precedence & !string_value %in% exclude_operators),
```
AshesITR, line 76: no visible binding for global variable ‘string_value’ lint (could be fixed via global variables or by defining them as `NULL` in the function scope)

R/spaces_left_parentheses_linter.R
```diff
@@ -29,9 +29,12 @@ spaces_left_parentheses_linter <- function() {
       for_cond <- "@start - 1 = parent::forcond/preceding-sibling::FOR/@end"
 
       # see infix_spaces_linter.R; preceding-sibling::* is needed for unary operators where '-(a)' is ok
-      unary_nodes <- c(names(unary_infix_tokens), "OP-TILDE")
+      unary_nodes <- infix_metadata[infix_metadata$unary, "xml_tag"]
       unary_selves <- paste0("self::", unary_nodes, "[preceding-sibling::*]", collapse = " or ")
-      binary_nodes <- c(names(binary_infix_tokens), "EQ_FORMALS", "OP-COMMA", "OP-LEFT-BRACE", "ELSE", "IN")
+      binary_nodes <- c(
+        infix_metadata[with(infix_metadata, low_precedence & !unary), "xml_tag"],
```
AshesITR, line 35: lints:

no visible binding for global variable ‘low_precedence’
no visible binding for global variable ‘unary’

tests/testthat/test-infix_spaces_linter.R
AshesITR: Okay, should be a lot easier if we convert the linter to XPath at the same time (requires something like `following-sibling::*[@col1 = preceding-sibling::*/@col2]` vs. `> preceding-sibling::*/@col2` iinm)

R/infix_spaces_linter.R
MichaelChirico: I think this is a false positive in the linter since `R CMD check` doesn't complain about it.

MichaelChirico: Filed #941 as follow-up

AshesITR (approved):

MichaelChirico:
Looks like we were affected by https://www.githubstatus.com/incidents/fpk08rxnqjz2 but it's not clear how to trigger a restart?

AshesITR (approved):

AshesITR:
I removed the minor `.gitignore` change to force a new commit. Seems to have worked.
Feel free to revert if you absolutely want the new ignore in this PR.

AshesITR:
── Failure (test-linter_tags.R:26:3): available_linters matches the set of linters available from lintr ──

MichaelChirico:
indeed... bad merge i think. restored the gitignore edit while we're at it

MichaelChirico:
Turns out the improvement caught some new lints inside lintr too 😎 fixed now

NEWS.md
```diff
@@ -10,7 +10,6 @@
   format specifiers (#472, @russHyde)
 * New style SNAKE_CASE for `object_name_linter()` (#494, @AshesITR)
 * RStudio source markers are cleared when there are no lints (#520, @AshesITR)
-* New `assignment_spaces()` lintr. (#538, @f-ritter)
```
AshesITR, line 13: Not sure what to do about attribution here...

MichaelChirico: I had the same hesitation... I'll mark down in the follow-up to be sure to thank them.

AshesITR (approved):

Issue #916, opened 2022-03-06
Create tests for completeness of documentation


AshesITR:
From https://github.com/AshesITR/lintr/pull/2#discussion_r820162915

> do we have a CRAN release script? we should add something that ensures this list is complete before release in case new tags are added

> Yeah I thought about the possibility to add a test case, but wasn't sure if testthat would be able to reliably inspect the current package help index.
>
> We should also check that all tags have at least one linter, that all tags listed in available_linters() are in fact documented, that all linters listed in available_linters() are exported (and documented) and all exported functions ending in _linter are in available_linters().
>
> Even better would be to check for the Tags & Linters sections where appropriate, to ensure newly created linters have all the necessary boilerplate.

jimhester:
I would suggest using `usethis::use_release_issue()` for tracking any extra tasks needed for a release.

You can add custom tasks by defining a `release_bullets()` function in the package namespace.

e.g. here is the one for usethis itself https://github.com/r-lib/usethis/blob/2402ea8e4eb098b5fd18770174fa86ae1a803658/R/usethis-package.R#L13-L17

AshesITR:
Will this gotcha apply for every release or just those with a modified `release_bullets()`?

PR #932, opened 2022-03-13
add a test to ensure inst/lintr/linters.csv is up to date

MichaelChirico:
Part of #916

This has been the biggest pain point thus far -- making sure to remember the .csv is in sync as we make edits/rename things. Having this test will at least make it so we fail early.

Renamed the only two `_linter` objects in our namespace that aren't linters; thankfully neither is exported so we don't have to worry much.

### DIFF
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -144,7 +144,7 @@ object_lint2 <- function(expr, source_file, message) {
   )
 }
 
-make_object_linter <- function(fun, name = linter_auto_name()) {
+object_linter_factory <- function(fun, name = linter_auto_name()) {
   force(name)
   Linter(function(source_file) {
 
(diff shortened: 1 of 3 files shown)


AshesITR:
I used `getNamespaceExports()` instead of `ls()`, which didn't return the factories and makes sure that the available linters are actually exported. WDYT?

I think it's still fine to rename the two function factories.

MichaelChirico:
my intent was to make sure the linters are exported but you're right this test doesn't do that, how about a separate test for that?

AshesITR:
I'm not sure if tests already see the new `getNamespaceExports()` when a new linter is added, but I'd say we should definitely add a test for that if we check this.

Maybe need to parse `NAMESPACE` -- shouldn't be too bad.

MichaelChirico:
> I'm not sure if tests already see the new getNamespaceExports() when a new linter is added, but I'd say we should definitely add a test for that if we check this.

Can't see why not. tests are run after the package is built so the namespace seen by tests should be what we want.

AshesITR (approved):

PR #937, opened 2022-03-13
Change ... precedence in lint(), lint_dir(), lint_package()

### LINKED ISSUE #935: ... should come 2nd in lint_dir()
Now, if we want to run `lint_dir()` for only a specific linter, we have to somewhat awkwardly pass the default argument to `relative_path`:

```
lintr::lint_dir("/path/to/dir", TRUE, lintr::specific_linter())
```

I think this goes against the tidyverse design principle where `...` is considered "data":

https://design.tidyverse.org/args-data-details.html

Since the next release is a major version bump, I think we can make the change now and try and fix whoever's broken, hopefully few if any people rely on this anyway.

MichaelChirico:
Closes #935


AshesITR:
Should probably add a safeguard check for the first release (warning if e.g. first positional argument is unnamed and `TRUE`)

MichaelChirico:
Hmm I started working on that, but I think it'll make the code a lot uglier.

It's easy to check (pseudo-code) `length(...) > 0L && is.logical(...[[1L]])`, but then harder to pluck out `...[[1L]]` and pass `...[-1L]` on to other functions. IINM it will require some `do.call()` usage.

I'll write it for `lint()` and we can decide whether to proceed for the others.

R/lint.R
```diff
@@ -174,12 +187,10 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
...
 #' }
 #' @export
-lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("renv", "packrat"),
-                     pattern = rex::rex(
-                       ".", one_of("Rr"),
-                       or("", "html", "md", "nw", "rst", "tex", "txt"),
-                       end
-                     ),
+lint_dir <- function(path = ".", ...,
+                     relative_path = TRUE,
```
AshesITR, line 191: We need a safeguard here too.
`lint_dir("...", FALSE)` would trigger the warning regarding `cache` once per file in "..." instead of alerting the actual problem that `relative_path` is no longer positional.

R/lint.R
```diff
@@ -93,7 +102,9 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
     lint_obj <- if (is.null(text)) filename else list(content = get_content(lines), TRUE)
     lints <- retrieve_file(lint_cache, lint_obj, linters)
     if (!is.null(lints)) {
-      return(exclude(lints, lines = lines, ...))
+      # TODO: once cache= is fully deprecated as 3rd positional argument (see top of body), we can restore the cleaner:
+      #   exclude(lints, lines = lines, ...)
```
AshesITR, line 106: Should suppress lint, e.g. using `#>` instead of `#` to make it unparseable.

R/lint.R
```diff
@@ -268,8 +279,10 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
 #'   )
 #' }
 #' @export
-lint_package <- function(path = ".", relative_path = TRUE, ...,
-                         exclusions = list("R/RcppExports.R"), parse_settings = TRUE) {
+lint_package <- function(path = ".", ...,
+                         relative_path = TRUE,
```
AshesITR, line 283: ditto

R/lint.R
MichaelChirico: this was intentional:

> I'll write it for lint() and we can decide whether to proceed for the others.

Wanted to check with you first that the implementation for `lint()` looks good

AshesITR:
I think it would be useful to add tests for the deprecated positional usage, especially as there is danger of `relative_path` being misinterpreted as `cache` since both were at position 2 and logical.

MichaelChirico:
Makes sense. Added just one test for `lint_dir()` for now -- given there are no existing tests for `relative_path=`, it's relatively cumbersome to write new tests.

Happy to add more if you see the need.

tests/testthat/test-dir_linters.R
```diff
@@ -88,3 +88,9 @@ test_that("lint_dir works with specific linters without specifying other argumen
   expect_length(lint_dir(the_dir, assignment_linter(), parse_settings = FALSE), 12L)
   expect_length(lint_dir(the_dir, commented_code_linter(), parse_settings = FALSE), 0L)
 })
+
+test_that("lint_dir continues to accept relative_path= in 2nd positional argument, with a warning", {
+  the_dir <- file.path("dummy_packages", "package", "vignettes")
+  expect_warning(positional_lints <- lint_dir(the_dir, FALSE))
```
AshesITR, line 94: Best practice is to specify the warning text

R/lint.R
```diff
@@ -35,8 +35,17 @@ NULL
...
+  # TODO(next release after 3.0.0): remove this deprecated workaround
+  dots <- list(...)
+  if (length(dots) > 0L && is.logical(dots[[1L]])) {
+    warning(
+      "'cache' is no longer available as a positional argument; please supply 'cache' as a named argument instead. ",
+      "This warning will be upgraded to an error in the next release."
+    )
+    cache <- dots[[1L]]
+    dots <- dots[-1L]
+  }
```
AshesITR, line 48: It just occurred to me that there might be a way to localize the deprecation code to one spot by using `Recall()`. Not 100% sure though. Either way is fine by me.

MichaelChirico: i'm also not familiar with `Recall()`... better to avoid then :)

IIUC it would still have the issue that we need to strip out `...[[1]]`, so it would look need `do.call(Recall)`, which may not work (I have in mind the "Note" in `?Recall`)

R/lint.R
```diff
@@ -268,8 +292,22 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
...
+  dots <- list(...)
+  if (length(dots) > 0L && is.logical(dots[[1L]])) {
+    warning(
+      "'relative_path' is no longer available as a positional argument; ",
+      "please supply 'relative_path' as a named argument instead. ",
+      "This warning will be upgraded to an error in the next release."
+    )
+    relative_path <- dots[[1L]]
+    dots <- dots[-1L]
+  }
```
AshesITR, line 309: Is this tested?

R/lint.R
AshesITR: Also, is this tested? I only see a test for `lint_dir`.

R/lint.R
MichaelChirico: no: https://github.com/r-lib/lintr/pull/937#issuecomment-1074703326

R/lint.R
MichaelChirico: added one

R/lint.R
AshesITR: Let's add one just to be safe.
Rest LGTM.

AshesITR (approved):

bersbersbers:
I have the impression that this change prevents the use of a trailing comma in `lint_dir`, as in `lintr::lint_dir(path=".",)`, which is possible in v2.0.1. Is that something that was intented; foreseen and accepted; or unintended and worthy of fixing?

MichaelChirico:
IINM that's undocumented behavior. we can try and add a deprecation warning for it anyway. could you open a new issue?

bersbersbers:
@MichaelChirico sure, https://github.com/r-lib/lintr/issues/1027

PR #939, opened 2022-03-14
Simplify lint() internals towards cyclomatic complexity target

MichaelChirico:
Towards #653

This pass reduced us from 28-->19. I think that's pretty good for now.

This enables us to tighten our belt on the lintr down to 20 hereafter.

### DIFF
--- a/R/lint.R
+++ b/R/lint.R
@@ -77,21 +77,11 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings = 
   linters <- define_linters(linters)
   linters <- Map(validate_linter_object, linters, names(linters))
 
-  lints <- list()
-  itr <- 0
-
-  cache_path <- if (isTRUE(cache)) {
-    settings$cache_directory
-  } else if (is.character(cache)) {
-    cache
-  } else {
-    character(0)
-  }
+  cache_path <- set_cache_path(cache, settings$cache_directory)
 
   if (length(cache_path)) {
     lint_cache <- load_cache(filename, cache_path)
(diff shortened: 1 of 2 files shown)


R/lint.R
```diff
@@ -100,32 +90,17 @@ lint <- function(filename, linters = NULL, cache = FALSE, ..., parse_settings =
...
-        }
-      }
-      lints[[itr <- itr + 1L]] <- expr_lints
-    }
-  }
+  n_linters <- length(linters)
+  n_checks <- length(source_expressions$expressions) * n_linters
+  lints <- lapply(seq_len(n_checks), function(check_i) {
+    # modular arithmetic to vary linters first, then expressions
+    expr <- source_expressions$expressions[[((check_i - 1L) %/% n_linters) + 1L]]
```
MichaelChirico, line 97: not the prettiest -- two other approaches:

 1. Use `DF <- expand.grid(seq_len(n_linters), seq_len(n_expr))`, then `DF[check_i, ]` has the right coords
 2. Use `M <- matrix(nrow = n_linters, ncol = n_expr)`, then `col(M)` and `row(M)` have the right coords

AshesITR: I think we're sacrificing readability for cyclomatic complexity here.
NB: did you take a look at @russHyde's PR for caching as a class?

MichaelChirico: Looks like a more systematic way of accomplishing what's attempted in this PR. Nice! Any hope for progress @russHyde ? :grimacing:

MichaelChirico:
Closing for now, but leaving the branch in case there's anything salvageable.

PR #943, opened 2022-03-15
expect_s3_class_linter and expect_s4_class_linter

MichaelChirico:
Part of #884 

Close cousin of `expect_type_linter()`


MichaelChirico:
BTW, one thing to consider -- with this linter active, it becomes somewhat cumbersome to test for the exact sequence of S3 classes, e.g. before you might write

```
expect_equal(class(DT), c("data.table", "data.frame"))
```

this linter wants

```
expect_s3_class(DT, "data.table")
expect_s3_class(DT, "data.frame")
```

The thinking is that it's usually preferable to check inheritance as opposed to an exact sequence of classes.

One thing we could do is offer an option (I'm indifferent to the default) `lint_identical` that toggles whether to lint `expect_identical()`. If it's turned off, that opens the avenue to run `expect_identical(class(x), k)` for the case outlined above. Or we could open it for `expect_equal()`.

At Google, we've really aligned on `expect_identical()` being the default test, which is why we still lint for `expect_identical()` -- otherwise we wouldn't really catch anything to begin with. (There's another `expect_identical_linter()` forthcoming that blocks `expect_equal()` besides exceptional cases)

jimhester:
What is wrong with `expect_s3_class(data.table::data.table(), c("data.table", "data.frame"))` it doesn't actually check for the exact order, it is just passed to `inherits()`, which just checks for membership. e.g. both these pass.

``` r
library(testthat)
expect_s3_class(data.table::data.table(), c("data.frame", "data.table"))
expect_s3_class(data.table::data.table(), c("data.table", "data.frame"))
```

<sup>Created on 2022-03-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

jimhester:
Oh actually I forgot how inherits works, it actually just returns true if _any_ of the input classes are found, not all of them, so separate calls to `expect_s3_class()` are probably what you would want anyway.

MichaelChirico:
Right... in priniciple the inheritance order could matter, and we might want to test for that. I'm not sure how often that's an issue though. It didn't come up at all in our internal packages, for example.

AshesITR:
I think minor de-lints are fine if they don't make the PR too large.

tests/testthat/test-expect_s3_class_linter.R
```diff
@@ -0,0 +1,97 @@
+test_that("expect_s3_class_linter skips allowed usages", {
+  # expect_s3_class doesn't have an inverted version
+  expect_lint("expect_true(!inherits(x, 'class'))", NULL, expect_s3_class_linter())
+  # NB: also applies to tinytest, but it's sufficient to test testthat
+  expect_lint("testthat::expect_s3_class(!inherits(x, 'class'))", NULL, expect_s3_class_linter())
```
AshesITR, line 5: Is the expextation function intentional? Looking for `expect_true` here.

tests/testthat/test-expect_s3_class_linter.R
```diff
@@ -0,0 +1,97 @@
...
+  # other is.<x> calls are not suitable for expect_s3_class in particular
+  expect_lint("expect_true(is.na(x))", NULL, expect_s3_class_linter())
+
+  # case where expect_s3_class() *could* be used but we don't enforce
+  expect_lint("expect_true(is.data.table(x))", NULL, expect_s3_class_linter())
+})
+
+test_that("expect_s3_class_linter blocks simple disallowed usages", {
+  expect_lint(
+    "expect_equal(class(x), 'data.frame')",
```
AshesITR, line 16: A two-element expectation test would be nice to have.

MichaelChirico: as in testing that `expect_equal(class(x), c("data.table", "data.frame"))` throws a lint as expected?

R/expect_s3_class_linter.R
```diff
@@ -0,0 +1,108 @@
...
+
+    # (1) expect_{equal,identical}(class(x), C)
+    # (2) expect_true(is.<class>(x)) and expect_true(inherits(x, C))
+    is_class_call <- xp_text_in_table(c(is_s3_class_calls, "inherits"))
+    xpath <- glue::glue("//expr[
+      (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical']
+        and following-sibling::expr[
+          expr[SYMBOL_FUNCTION_CALL[text() = 'class']]
+          and (position() = 1 or preceding-sibling::expr[STR_CONST])
```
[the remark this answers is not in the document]
MichaelChirico: no... is it worth doing so? we're in the territory of extended logic for "non-canonical" usages already. Would mean growing the logic from

```
expr[STR_CONST]
```

to (roughly)

```
expr[STR_CONST or expr[SYMBOL_FUNCTION_CALL[text = 'c'] and not(expr[position() > 1 and not(STR_CONST)])]]
```

substantially more complicated because we need to do double-negation in the xpath.

tests/testthat/test-expect_s3_class_linter.R
MichaelChirico: nice! fixed

tests/testthat/test-expect_s3_class_linter.R
MichaelChirico: Added such a test

R/expect_s3_class_linter.R
```diff
@@ -0,0 +1,108 @@
...
+  Linter(function(source_file) {
+    if (length(source_file$parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    # (1) expect_{equal,identical}(class(x), C)
+    # (2) expect_true(is.<class>(x)) and expect_true(inherits(x, C))
+    is_class_call <- xp_text_in_table(c(is_s3_class_calls, "inherits"))
```
AshesITR, line 21: false positive lint caused by glue: local variable ‘is_class_call’ assigned but may not be used.

R/expect_s3_class_linter.R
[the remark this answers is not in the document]
AshesITR: No, doesn't really seem worth it.

AshesITR (changes requested):
Almost gtg

AshesITR (approved):

R/expect_s3_class_linter.R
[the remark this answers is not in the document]
MichaelChirico: definitely keep these cases in mind for the PR of yoda_test_linter, which looks specifically for the reversed usage that would be caught by the extended logic.

ideally users will have yoda_test_linter activated alongside whichever other expect_*_linter, which can first guide them to switch the arguments, and then to change the expectation.

PR #957, opened 2022-03-18
New yoda_test_linter

MichaelChirico:
Part of #962


R/yoda_test_linter.R
```diff
@@ -0,0 +1,45 @@
...
+    xml <- source_file$xml_parsed_content
+
+    # catch the following types of literal in the first argument:
+    #   (1) numeric literal (e.g. TRUE, 1L, 1.0, NA) [NUM_CONST]
+    #   (2) string literal (e.g. 'str' or "str")     [STR_CONST]
+    #   (3) arithmetic literal (e.g. 1+1 or 0+1i)    [OP-PLUS or OP-MINUS...]
+    xpath <- "//expr[
+      expr[SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical' or text() = 'expect_setequal']]
+      and expr[2][NUM_CONST or STR_CONST or ((OP-PLUS or OP-MINUS) and count(expr[NUM_CONST]) = 2)]
+    ]"
```
AshesITR, line 28: Safeguard against false positives via `and expr[3][not(...)]`?
Otherwise this will always lint things like `expect_equal(1.0, 1.0 + 0i)`.

Alternatively, I'd be fine if the message in the latter case would be "do not compare literals in testthat"?

R/yoda_test_linter.R
```diff
@@ -0,0 +1,45 @@
...
+
+    return(lapply(
+      bad_expr,
+      xml_nodes_to_lint,
+      source_file = source_file,
+      message = paste(
+        "Tests should compare objects in the order 'actual', 'expected', ",
+        "not the reverse. For example, do ",
+        "expect_identical(CountObservations(x), 2L) instead of ",
+        "expect_identical(2L, CountObservations(x))."
```
AshesITR, line 40: Long message. It's probably sufficient to just say "Tests should compare objects in the order 'actual', 'expected'."

man/linters.Rd
```diff
@@ -17,15 +17,15 @@ Documentation for linters is structured into tags to allow for easier discovery.
...
 \item{\link[=common_mistakes_linters]{common_mistakes} (5 linters)}
 \item{\link[=configurable_linters]{configurable} (16 linters)}
 \item{\link[=consistency_linters]{consistency} (7 linters)}
 \item{\link[=correctness_linters]{correctness} (7 linters)}
 \item{\link[=default_linters]{default} (25 linters)}
 \item{\link[=efficiency_linters]{efficiency} (4 linters)}
-\item{\link[=package_development_linters]{package_development} (9 linters)}
-\item{\link[=readability_linters]{readability} (23 linters)}
+\item{\link[=package_development_linters]{package_development} (11 linters)}
+\item{\link[=readability_linters]{readability} (25 linters)}
```
AshesITR, line 28: Why +2? Did we forget a `document()` in an earlier PR?
Especially weird because the tag Rds only get one additional line...

tests/testthat/test-yoda_test_linter.R
```diff
@@ -0,0 +1,44 @@
...
+})
+
+# TODO(michaelchirico): Should this be extended to RUnit tests? It seems yes,
+#   but the argument names in RUnit (inherited from base all.equal()) are a bit
+#   confusing, e.g. `checkEqual(target=, current=)`. From the name, one might
+#   reasonably conclude 'expected' comes first, and 'actual' comes second.
+# TODO(michaelchirico): What sorts of combinations of literals can be included?
+#   e.g. expect_equal(c(1, 2), x) is a yoda test; is expect_equal(c(x, 1), y)?
+#   clearly it's not true for general f() besides c(). What about other
+#   constructors of literals? data.frame(), data.table(), tibble(), ...?
```
AshesITR, line 41: Maybe it's worth to craft up an "extended literal" XPath that contains a few whitelisted functions and use that throughout lintr if we want to match a "constant" expression?

First thought would be something like `not(SYMBOL_FUNCTION_CALL[not(text() = 'c' or ...)])` etc.

R/yoda_test_linter.R
MichaelChirico: >  expect_equal(1.0, 1.0 + 0i)

seems like a bad test besides for R itself :)

we did encounter a few of these being used as placeholder tests but (1) `expect_true(TRUE)` is better and (2) similar to how we lint for `debug()` & friends, I think it's fine to lint placeholder code that shouldn't really make it "to prod"

R/yoda_test_linter.R
MichaelChirico: SGTM. We tend to be OK with long messages to maximize actionability when users reach a lint. Nothing worse than staring at a lint message and not having any idea how to make it go away. 

anyway, I'll take another pass at parsimony here.

man/linters.Rd
MichaelChirico: 😬 might be weird merging, either in this PR or in a recent one. I'll re-document() before next push.

tests/testthat/test-yoda_test_linter.R
MichaelChirico: sounds great, I was thinking something similar. We have something like that in `StringsAsFactorsLinter` as well (which also has some logic about mixing in `rep()`), and it can be re-used in the `expect_equal($LITERAL, typeof(x))` family of lints too.

MichaelChirico: Let's mark this as a TODO for now. we have a messy XPath that works for a limited number of cases, I'd like to generalize it if possible.

For reference here's the XPath for `StringsAsFactorsLinter` (so we only look for `STR_CONST`):

```
//expr[
  expr[SYMBOL_FUNCTION_CALL[text() = 'data.frame']]
  and expr[
    (
      STR_CONST
      or (
        expr[SYMBOL_FUNCTION_CALL[text() = 'c']]
        and expr[STR_CONST]
        and not(expr[SYMBOL or expr])
      )
      or expr[
        SYMBOL_FUNCTION_CALL[text() = 'rep']
        and (
          following-sibling::expr[1][
            STR_CONST
            or (
              expr[SYMBOL_FUNCTION_CALL[text() = 'c']]
              and expr[STR_CONST]
              and not(expr[SYMBOL or expr])
            )
          ]
        )
      ] or expr[SYMBOL_FUNCTION_CALL[
        text() = 'character'
        or text() = 'as.character'
        or text() = 'paste'
        or text() = 'sprintf'
        or text() = 'format'
        or text() = 'formatC'
        or text() = 'prettyNum'
        or text() = 'toString'
        or text() = 'encodeString'
      ]]
    )
    and not(preceding-sibling::SYMBOL_SUB[1][text() = 'row.names'])
  ]
  and not(SYMBOL_SUB[text() = 'stringsAsFactors'])
]
```

R/yoda_test_linter.R
MichaelChirico: I swapped `CountObservations()` with the shorter `foo()`; I still think having a simple example is good. 166 characters seems not so bad... do we have any guidance in mind for what's long/short?

man/linters.Rd
MichaelChirico: nothing changed... here's the `package_development` list, which looks right. HEAD has 10 listed: https://github.com/r-lib/lintr/blob/585e7cd3e0e4a3aa8f78fdf0242a908ff4cfd65e/man/package_development_linters.Rd#L15-L24

R/yoda_test_linter.R
AshesITR: I agree this should be linted, but the message will be very misleading IMO.

R/yoda_test_linter.R
AshesITR: LGTM.

R/yoda_test_linter.R
MichaelChirico: That's true... still let's table it for now, by making a note to check if this actually comes up in practice before release. If so, we can customize the error message.

inst/lintr/linters.csv
```diff
@@ -62,3 +62,5 @@ undesirable_function_linter,style efficiency configurable robustness best_practi
 undesirable_operator_linter,style efficiency configurable robustness best_practices
 unneeded_concatenation_linter,style readability efficiency
 vector_logic_linter,default efficiency best_practices
+yoda_test_linter,package_development best_practices readability
+
```
AshesITR, line 66: Is this an extra newline? Can't see this in GitHub.

MichaelChirico: yes! nice catch

AshesITR:
LGTM, needs a manual merge though, so not approving for now.

AshesITR (approved):

PR #958, opened 2022-03-18
New expect_identical_linter

MichaelChirico:
Part of #884 

This one has become slightly controversial owing to Hadley's comments here: https://github.com/r-lib/lintr/pull/910#discussion_r823140324

Meaning some divergence between how the `testthat` maintainers view `expect_equal()` vs. `expect_identical()` and how we do internally at Google, where we do want to keep encouraging `expect_identical()` as the default because we find stricter tests are a better default.

I think that means we want to be careful in wording/documentation not to imply that this linter is enacting the advice of `testthat`. Hope that makes sense; not sure if I achieved it in the current iteration.


R/expect_identical_linter.R
```diff
@@ -0,0 +1,80 @@
...
+      source_file = source_file,
+      message = paste(
+        "expect_identical(x, y) is the default way of comparing two objects in",
+        "testthat unit tests. Only use expect_equal() if testing numeric",
+        "equality (and specifying tolerance= explicitly) or there's a need to",
+        "ignore some attributes, e.g. with ignore_attr = 'names' for the 3rd",
+        "edition or check.attributes = FALSE for the second. For testing",
+        "approximate equality to a whole number, you can also avoid setting",
+        "tolerance= by including an explicit decimal point, e.g.,",
+        "expect_equal(x, 1.0), not expect_equal(x, 1)."
```
AshesITR, line 75: That's a rather long message compared to other linters.
Maybe shorten to "only use expect_equal() if non-default arguments, e.g. tolerance, are needed"?

tests/testthat/test-expect_identical_linter.R
```diff
@@ -0,0 +1,67 @@
...
+    expect_identical_linter()
+  )
+  expect_lint(
+    "expect_equal(x, 1)",
+    rex::rex("expect_identical(x, y) is the default way of comparing two objects"),
+    expect_identical_linter()
+  )
+  expect_lint(
+    "expect_equal(x, TRUE)",
+    rex::rex("expect_identical(x, y) is the default way of comparing two objects"),
```
AshesITR, line 54: NB this would send a lint after refactoring for not using `expect_true(x)`.

R/expect_identical_linter.R
```diff
@@ -0,0 +1,80 @@
...
+    #   2. conditions for expect_true()
+    xpath <- glue::glue("//expr[
+      (
+        SYMBOL_FUNCTION_CALL[text() = 'expect_equal']
+        and not(
+          following-sibling::SYMBOL_SUB[text() = 'tolerance']
+          or following-sibling::SYMBOL_SUB[
+            (
+              text() = 'check.attributes'
+              and following-sibling::expr[1][NUM_CONST[text() = 'FALSE']]
```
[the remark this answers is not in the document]
MichaelChirico: That's probably right. I'll simplify.

tests/testthat/test-expect_identical_linter.R
MichaelChirico: I'll note that in a comment... this test is to make sure we get each of the different flavors of `NUM_CONST`

R/expect_identical_linter.R
```diff
@@ -0,0 +1,80 @@
...
+            )
+            or starts-with(text(), 'ignore_')
+          ]
+          or following-sibling::expr[
+            expr[SYMBOL_FUNCTION_CALL[text() = 'c']]
+            and expr[NUM_CONST[contains(text(), '.')]]
+          ]
+          or following-sibling::expr[NUM_CONST[contains(text(), '.')]]
```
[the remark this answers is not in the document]
MichaelChirico: Let's target making this an argument for the #963 solution

R/expect_identical_linter.R
MichaelChirico: put some of the details in the man page, should be more digestible now

R/expect_identical_linter.R
```diff
@@ -0,0 +1,74 @@
...
+#'   is an equivalent but unadvised method of the same test. Further,
+#'   [testthat::expect_equal()] should only be used when `expect_identical()`
+#'   is inappropriate, i.e., when `x` and `y` need only be *numerically
+#'   equivalent* instead of fully identical (in which case, provide the
+#'   `tolerance=` argument to `expect_equal()` explicitly). This also applies
+#'   when it's inconvenient to check full equality (e.g., names can be ignored,
+#'   in which case `ignore_attr = "names"` should be supplied to
+#'   `expect_equal()` (or, for 2nd edition, `check.attributes = FALSE`).
+#'
+#' NB: The linter allows `expect_equal()` in three circumstances:
```
AshesITR, line 14: Not a fan of NB in documentation.
I'd a) drop the NB and leave the rest as-is or b) make an `@section Exceptions` to highlight that there are exceptions to the general rule stated above.

AshesITR (approved):

PR #961, opened 2022-03-19
refactor xp_and/xp_or to match common usage

MichaelChirico:


MichaelChirico:
Or maybe, we just change it to wrap `paste(x, collapse = sep)`. Current usages don't really need `...`. Only one is in upcoming `yoda_test_linter()` and that can easily swap `xp_or(...)` for `xp_or(c(...))`

AshesITR (approved):

PR #967, opened 2022-03-21
Silence testthat suite

### LINKED ISSUE #965: Messy testthat output due to 3e
From `?testthat::expect_warning`:

> In the 3rd edition, these functions match (at most) a single condition. All additional and non-matching (if regexp or class are used) conditions will bubble up outside the expectation. If these additional conditions are important you'll need to catch them with additional expect_message()/expect_warning() calls; if they're unimportant you can ignore with suppressMessages()/suppressWarnings().

This causes a lot of noise in the output of `devtools::test()` on master atm.

### LINKED ISSUE #966: `backport_linter()` warns multiple times
When working on fixing #965, I ran across the reason: The tested warning was generated twice.
I'm working on a fix for all test suite related problems, this will be fixed together with it, by moving the check outside of the `Linter()` and into the factory.

AshesITR:
Fixes #965
Fixes #966

- [x] test-backport_linter.R:9 Warning bubbles #966
- [x] test-cache.R produces output
- [x] test-cache.R:91 Two warnings bubble
- [x] test-ci.R:39 Deprecation warning for `with_mock()`
- [x] test-ci.R:52 Deprecation warning for `with_mock()`
- [x] test-dir_linters.R produces output
- [x] test-expect_lint.R produces output
- [x] test-expect_lint.R:59 calls RStudio API
- [x] test-knitr_formats.R produces output
- [x] test-lint_package.R produces output
- [x] test-methods.R:117 Deprecation warning for `with_mock()`
- [x] test-methods.R:129 calls RStudio API
- [x] test-object_usage_linter.R produces output
- [x] test-rstudio_markers.R:97 Deprecation warning for `with_mock()`
- [x] corrected minor typo in `rstudio_source_markers()` error message.


R/lint.R
```diff
@@ -218,14 +218,14 @@ lint_dir <- function(path = ".", relative_path = TRUE, ..., exclusions = list("r
   lints <- flatten_lints(lapply(
     files,
     function(file) {
-      if (interactive()) {
+      if (interactive() && !identical(Sys.getenv("TESTTHAT"), "true")) {
```
AshesITR, line 221: I couldn't get `mockery::stub()` to work here.
Trying `mockery::stub(lint_dir, "interactive", FALSE)` didn't work as intended.
Trying `mockery::stub(lint_package, "interactive", FALSE, depth = 2)` worked, but messed up the global state so much that following tests couldn't even find `rex::re_matches` etc.

In the end, I thought it's probably useful to silence `lint_dir()` during testthat runs automatically.

tests/testthat/test-cache.R
```diff
@@ -86,9 +86,16 @@ test_that("load_cache returns an empty environment if reading cache file fails",
   cache_f1 <- file.path(d1, fhash(f1))
   writeLines(character(), cache_f1)
 
-  expect_warning(e2 <- load_cache(file = f1, path = d1))
+  expect_warning(e2 <- load_cache(file = f1, path = d1), "Could not load cache file")
   saveRDS(e1, cache_f1)
-  expect_warning(e3 <- load_cache(file = f1, path = d1))
+  expect_warning(
+    expect_warning(
```
AshesITR, line 92: Alternatively, we could `suppressWarnings()` the two R warnings unrelated to the test. LMK what you prefer.

tests/testthat/test-rstudio_markers.R
```diff
@@ -94,14 +94,12 @@ test_that("rstudio_source_markers apply to print within rstudio", {
   empty <- withr::local_tempfile()
   file.create(empty)
 
-  with_mock(
-    `rstudioapi::hasFun` = function(x, ...) TRUE,
-    `rstudioapi::callFun` = function(...) cat("matched\n"), {
-    l <- lint(tmp, seq_linter())
-    expect_output(print(l), "matched", fixed = TRUE)
+  mockery::stub(print.lints, "rstudioapi::hasFun", function(x, ...) TRUE)
+  mockery::stub(print.lints, "rstudio_source_markers", function(x) cat("matched\n"))
```
AshesITR, line 98: I couldn't get `mockery::stub(rstudio_source_markers, ...)` calls to do the correct thing but figured we can resort to testing that `rstudio_source_markers()` is called here.

R/backport_linter.R
```diff
@@ -7,12 +7,13 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 backport_linter <- function(r_version = getRversion()) {
+  if (inherits(r_version, "numeric_version")) r_version <- format(r_version)
```
MichaelChirico, line 10: shouldn't we be doing the opposite? instead of comparing strings. it hasn't happened for any R versions but in principle `4.10.0` would break the logic here.

(doesn't need to be addressed in this PR)

tests/testthat/test-cache.R
```diff
@@ -86,9 +86,13 @@ test_that("load_cache returns an empty environment if reading cache file fails",
   cache_f1 <- file.path(d1, fhash(f1))
   writeLines(character(), cache_f1)
 
-  expect_warning(e2 <- load_cache(file = f1, path = d1))
+  expect_warning(e2 <- load_cache(file = f1, path = d1), "Could not load cache file")
   saveRDS(e1, cache_f1)
-  expect_warning(e3 <- load_cache(file = f1, path = d1))
```
MichaelChirico, line 91: > The concrete warnings are version-dependent.

a bit too vague as a comment (I'm imagining reading this comment in 4 years' time).

maybe a `skip()` for older versions here is appropriate?

tests/testthat/test-methods.R
```diff
@@ -114,16 +114,15 @@ test_that("print.lint works for inline data, even in RStudio", {
...
   )
 
+  mockery::stub(print.lints, "rstudioapi::hasFun", function(...) FALSE)
   withr::with_options(
     list("lintr.rstudio_source_markers" = TRUE),
-    with_mock(
-      `rstudioapi::hasFun` = function(...) TRUE,
-      expect_output(print(l), "not =")
-    )
+    expect_output(print(l), "not =")
```
MichaelChirico, line 120: test is too vague for me... `not =` doesn't convey all that much. what's the rest of the message here? not clear to me what's being tested exactly.

tests/testthat/test-rstudio_markers.R
MichaelChirico: add that as a comment, maybe with a TODO you we think it's preferable to eventually test `rstudio_source_markers()` instead?

R/lint.R
AshesITR: Filed #972 as a follow-up. I think it's worth investing more thought into what and how to control output.
The "." output is quite daunting at times especially when linting a large codebase with many files.
`cli` could do a lot regarding output, or `progressr`.

tests/testthat/test-rstudio_markers.R
AshesITR: I like the idea of unit testing `rstudio_source_markers()` separately, but I'm not sure what exactly to test other than that the calls made via `rstudioapi` fire and get the same input across time.

R/backport_linter.R
AshesITR: It's used later vs. `names(backports)` and contains non-version strings ("devel").
Not sure if that would break. Swapping the format and comparison would fix half of the problem at no cost, but I just noticed we need `4.1.0` backports in the data anyway, so lets create a follow-up.

tests/testthat/test-ci.R
```diff
@@ -36,9 +36,8 @@ test_that("Printing works for Travis", {
 
   l <- lint(tmp)
 
-  with_mock(github_comment = function(x, ...) cat(x, "\n"), .env = asNamespace("lintr"), {
-    expect_output(print(l), "*warning:*", fixed = TRUE)
-  })
+  mockery::stub(print.lints, "github_comment", function(x, ...) cat(x, "\n"))
```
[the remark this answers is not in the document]
AshesITR: Not in the slightest, no.
I also feel `mockery` leaves a few things to be desired as of now, e.g. namespaced calls can't be overridden at `depth > 1`, necessitating us to `stub()` the actual private implementations.

Also, I tried `mockery::stub(lint_package, "interactive", FALSE, depth = 2L)` and it corrupts the entire environment ☹️

tests/testthat/test-methods.R
AshesITR: That's part of the lint message printed to console.
A better way may be to mock `print.lint()`, which does the heavy lifting.

tests/testthat/test-ci.R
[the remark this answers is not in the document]
MichaelChirico: a shame `with_mock()` is deprecated without a perfect successor.

tests/testthat/test-methods.R
MichaelChirico: if the rest of the message is messy, add a comment here like "full message is like '... not= ...'" to give a hint of what should be expected

R/backport_linter.R
AshesITR: Fixed by #973

tests/testthat/test-cache.R
AshesITR: I've never worked with snapshots before. Do they support different R versions giving different output?

AshesITR: Different suggestion to keep the code somewhat clean: Wrap it in `withCallingHandlers()` and explicitly check the bubbled warning messages for those that we know of, discarding them silently and bubbling up any additional warnings. WDYT?

MichaelChirico: yes, via `variant=`.

See testthat's suite e.g. https://github.com/r-lib/testthat/tree/main/tests/testthat/_snaps.

Personally I think tidyverse started over-relying on them -- they're very fragile (e.g. an upstream like `crayon` can make a small change in how output is displayed and now your snapshot test is broken) and have been hell to maintain inside Google, and are turned off on CRAN by default.

But it could work to use them in limited cases. Something to consider at least.

MichaelChirico: > bubbling up any additional warnings

I'll have to look at the implementation for what this means exactly. But SGTM anyway.

AshesITR: Something like that: `withCallingHandlers({ warning("ok"); warning("oops") }, warning = function(w) {if (conditionMessage(w) == "ok") invokeRestart("muffleWarning") else w})`

MichaelChirico (approved):

PR #969, opened 2022-03-21
Make `object_usage_linter` glue aware

### LINKED ISSUE #942: object_usage_linter could me made glue-aware
See discussion here: https://github.com/r-lib/lintr/pull/924#r826189886

And Jim's recommendation here: https://github.com/MichaelChirico/potools/issues/252#issuecomment-963099061

We could use this to cover objects which are only used inside glue interpolation.

AshesITR:
Fixes #942

LMK if the tests are sufficient.

There are two blacklisted arguments, `.envir` and `.transformer` to ensure they are what glue uses by default.
Also, only calls with only STR_CONST expressions are evaluated.


R/object_usage_linter.R
```diff
@@ -98,6 +107,51 @@ object_usage_linter <- function() {
   })
 }
 
+extract_glued_symbols <- function(expr) {
+  glue_calls <- xml2::xml_find_all(
+    expr,
+    xpath = paste0(
+      "descendant::SYMBOL_FUNCTION_CALL[text() = 'glue']/", # a glue() call
```
MichaelChirico, line 114: maybe as a TODO: there's also `glue_safe()`, `glue_sql()`, and `glue_col()`, which we could run through the same logic as `glue()`, and `glue_data()` (and similar safe/sql/col variants)

MichaelChirico:
Also remove the `#nolint`s that this solves since there are only a few

R/object_usage_linter.R
AshesITR: `stringr::str_interp()` could also be supported. Adding note.

MichaelChirico (approved):

PR #974, opened 2022-03-21
adapt backport linter to R 4.1.0 and make versioning more robust

### LINKED ISSUE #973: Update `backport_linter()` to include 4.1.0 backports
Also should take a look at robustness, cf.

shouldn't we be doing the opposite? instead of comparing strings. it hasn't happened for any R versions but in principle `4.10.0` would break the logic here.

(doesn't need to be addressed in this PR)

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/967#discussion_r831290652_

AshesITR:
Fixes #973


R/backport_linter.R
```diff
@@ -7,13 +7,37 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 backport_linter <- function(r_version = getRversion()) {
-  Linter(function(source_file) {
-    if (inherits(r_version, "numeric_version")) r_version <- format(r_version)
-    if (r_version < "3.0.0") {
-      warning("It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.")
-      r_version <- "3.0.0"
+  if (is.character(r_version) &&
+    re_matches(r_version, rex(start, "release" %or% list("oldrel", maybe("-", digits)) %or% "devel", end))) {
```
MichaelChirico, line 11: extract this to a helper for readability

R/backport_linter.R
```diff
@@ -7,13 +7,37 @@
...
+    # Support devel, release, oldrel, oldrel-1, ...
```
MichaelChirico, line 12: are these "flavors" documented somewhere? if so maybe link that in a comment

R/backport_linter.R
```diff
@@ -7,13 +7,37 @@
...
+    all_versions <- names(backports)
+    minor_versions <- unique(re_substitutes(all_versions, rex(".", digits, end), ""))
+    version_names <- c("devel", "release", "oldrel", paste0("oldrel-", seq_len(length(minor_versions) - 3L)))
+    if (!r_version %in% version_names) {
+      # This can only trip if e.g. oldrel-99 is requested
+      stop("`r_version` must be a version number or one of ", toString(sQuote(version_names)))
     }
+    requested_version <- minor_versions[match(r_version, table = version_names)]
+    available_patches <- all_versions[startsWith(all_versions, requested_version)]
+    selected_patch <- which.max(as.integer(substr(
```
MichaelChirico, line 23: nit: i think it's more readable as

```
which.max(as.integer(
  substr(...)
))
```

R/backport_linter.R
```diff
@@ -7,13 +7,37 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 backport_linter <- function(r_version = getRversion()) {
-  Linter(function(source_file) {
-    if (inherits(r_version, "numeric_version")) r_version <- format(r_version)
-    if (r_version < "3.0.0") {
-      warning("It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.")
-      r_version <- "3.0.0"
+  if (is.character(r_version) &&
```
MichaelChirico, line 10: extract all this as a helper and write

`r_version <- normalize_r_version(...)`?

R/backport_linter.R
```diff
@@ -22,7 +46,10 @@ backport_linter <- function(r_version = getRversion()) {
     all_names <- xml2::xml_text(all_names_nodes)
 
     # guaranteed to include 1 by early return above; which.min fails if all TRUE (handled by nomatch)
-    needs_backport_names <- backports[1:(match(FALSE, r_version < names(backports), nomatch = length(backports)) - 1L)]
+    needs_backport_names <- head(
+      backports,
+      match(FALSE, r_version < R_system_version(names(backports)), nomatch = length(backports)) - 1L
```
MichaelChirico, line 51: tricky code... extract as a helper with doc/comments explaining what it's doing / why not use something more obvious

tests/testthat/test-backport_linter.R
```diff
@@ -11,22 +13,36 @@ test_that("backport_linter detects backwards-incompatibility", {
...
     "trimws(...names())",
     list(
       rex("trimws (R 3.2.0) is not available for dependency R >= 3.0.0."),
-      rex("...names (R devel) is not available for dependency R >= 3.0.0.")
+      rex("...names (R 4.1.0) is not available for dependency R >= 3.0.0.")
     ),
     backport_linter("3.0.0")
   )
+
+  # oldrel specification
```
MichaelChirico, line 35: add an error case like you pointed out in the comments: `oldrel-99`

R/backport_linter.R
AshesITR: Not really. I found them in `r-lib/actions` which uses an API from r-hub to get them.

R/backport_linter.R
AshesITR: I've refactored this now to use logical indexing and changed the variable name. Does that make it more understandable?

MichaelChirico: Didn't see the `match()` thing was just copied from earlier (in fact IIRC I am the author of it 😃), but was talking about that, not so much `head()`.

definitely much cleaner now... i wonder why we didn't do that before 🤔

MichaelChirico (approved):

PR #975, opened 2022-03-21
New any_is_na_linter

MichaelChirico:
Part of #962

PR #976, opened 2022-03-21
fix lints

MichaelChirico:
There are also:

 - A false-positive `expect_length_linter()` where `info=` is used
 - A false-positive `expect_comparison_linter()` where `info=` is used
 - Dozens and dozens of `expect_identical()` lints

Happy to switch our default to `expect_identical()`, but wanted to check in first -- WDYT?


AshesITR:
I don't really see the benefit in doing s/expect_identical/expect_equal/ over our testing codebase at the moment.

MichaelChirico:
follow-up https://github.com/r-lib/lintr/pull/975#issuecomment-1075579641

After merge, I still see approval here

image

PR #397, opened 2019-08-21
lint repeated forwarding of dots-args

fabian-s:
Handing over `...` more than once inside a function body is dangerous
and prone to cause errrors, warnings or unexpected behavior.

Ignores list(...), function(...) and repeated calls to the same
function.

Content of `~/tmp/dotdotdot.R`:
  ``` r
double <- function(...) {
  # dangerous
  y <- plot(...)
  x <- lm(...)
}
single <- function(x, z, ...) {
  # ok
  dots <- list(...)
  y <- plot(...)
}
formal <- function(x, z, ...) {
  # ok
  y <- plot(...)
  f <- function(...) return()
}
repeated <- function(x, z, ...) {
  # ok
  plot(...)
  y <- z * x
  if (TRUE) plot(...)
}
```
```
devtools::load_all("~/lehre/lintr")
#> Loading lintr

lint("~/tmp/dotdotdot.R", multiple_dot_args_linter)
#> ~/tmp/dotdotdot.R:3:13: style: functions should not forward ... multiple times.
#>   y <- plot(...)
#>             ^~~
#> ~/tmp/dotdotdot.R:4:11: style: functions should not forward ... multiple times.
#>   x <- lm(...)
#>           ^~~
```

PR #982, opened 2022-03-22
suggestions for compare_branches

fabian-s:
I had to fiddle around a little bit with this to get it to work for #397, maybe you'll find these useful/sensible as well:  

- 105: make sure `if(is.na(...`  always evaluates to `logical(1)`
- 153: actually use parameter "sample_size"
- 349f: more fault-tolerant loop over linters
- 370: flush warnings to terminal at the end

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -102,7 +102,7 @@ params <- optparse::parse_args(optparse::OptionParser(option_list = param_list))
 if (interactive()) {
   for (opt in c("branch", "pr", "packages", "pkg_dir", "sample_size")) {
     # typed arguments get cast even when missing, probably to NA
-    if (is.na(params[[opt]]) || params[[opt]] == "") params[[opt]] <- NULL
+    if (isTRUE(is.na(params[[opt]]) || params[[opt]] == "")) params[[opt]] <- NULL
   }
 }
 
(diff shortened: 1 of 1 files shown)


fabian-s:
Also: If you ask pull request makers to run this to evaluate a new linter they propose, I think they also need to define a "dummy" linter of the same name in `master`, like:

```
multiple_dot_args_linter <- function(source_file) {
   Linter(function(source_file) return(list())))
}
```

MichaelChirico:
> I think they also need to define a "dummy" linter of the same name in master

I'd rather refactor the code not to run anything, even with a decoy linter it'll take a long time to run if nothing's accomplished

.dev/compare_branches.R
```diff
@@ -346,10 +346,18 @@ if (length(packages) > 50L) {
   )
 }
 
+df_otherwise <- tibble::tibble(
```
MichaelChirico, line 349: could you use tidyverse style here?

fabian-s: sorry. see https://github.com/r-lib/lintr/pull/982/commits/1798de0810868a5c7c09c5e84d2962661dc022c9

MichaelChirico:
Thanks for the improvements!!

.dev/compare_branches.R
MichaelChirico: np. understand styler::style is the easiest approach for such a request. and we should probably comply with that globally anyway.

Issue #962, opened 2022-03-20
Google linters for 3.0.0 release


MichaelChirico:
Splitting off a subset of #884 as a TODO list for 3.0.0 release; the rest of the linters can be done after that.

- [x] `AnyDuplicatedLinter` - Require usage of anyDuplicated() > 0 over any(duplicated(.)) #986
- [x] `AnyIsNALinter` - Require usage of anyNA over any(is.na(.)) #975
- [x] `ClassEqualsLinter` - Block comparison of class with == #989
- [x] ~`StopPasteLinter`~ `condition_message_linter` - Block usage of paste() and paste0() with messaging functions using ... #999
- [x] ~`MultipleStopifnotLinter`~ `ConsecutiveStopifnotLinter` - Force consecutive stopifnot() calls into just one #1005 
- [x] `ElseSameLineLinter` - Require else to come on the same line as the preceding brace #1002
- [x] `ExpectComparisonLinter` - Require usage of expect_gt(x, y) over expect_true(x > y) (and similar) #955 
- [x] `ExpectIdenticalLinter` - Require usage of expect_identical(x, y) where appropriate #958 
- [x] `ExpectIsLinter` - Redirect away from deprecated testthat::expect_is() #946 
- [x] `ExpectLengthLinter` - Require usage of expect_length(x, n) over expect_equal(length(x), n) #950 
- [x] `ExpectNamedLinter` - Require usage of expect_named(x, n) over expect_equal(names(x), n) #949 
- [x] `ExpectNotLinter` - Require usage of expect_false(.) over expect_true(!.) #945 
- [x] `ExpectNullLinter` - Require usage of expect_null(x) over expect_equal(x, NULL) and similar #887 
- [x] `ExpectS3ClassLinter` - Require usage of expect_s3_class() or expect_s4_class() where appropriate. #943
- [x] `ExpectS4ClassLinter` - Require usage of expect_s4_class(x, k) over expect_true(is(x, k)) #943 
- [x] ~`ExpectTrueFalseAndConditionLinter`~ `conjunct_test_linter` - Force && conditions in expect_true(), expect_false() to be written separately #948 
- [x] `ExpectTrueFalseLinter` - Require usage of expect_true(x) over expect_equal(x, TRUE) #947 
- [x] `ExpectTypeLinter` - Require usage of expect_type(x, type) over expect_equal(typeof(x), type) #924 
[...]

PR #983, opened 2022-03-22
New if_else_match_braces_linter

MichaelChirico:
Part of #962

We might consider this one to be a default linter btw, it's basically part of the style guide.


AshesITR:
In favor of making this a default linter, although I don't see explicit mention of this rule in https://style.tidyverse.org/syntax.html#control-flow
@jimhester WDYT? Should we lint

```r
if (test) a else {
  long_code
}

if (test) {
  long_code
} else a
```

by default?

MichaelChirico:
FWIW styler agrees:

```
styler::style_text('if (test) a else {
  long_code
}

if (test) {
  long_code
} else a
')
# if (test) {
#   a
# } else {
#   long_code
# }

# if (test) {
#   long_code
# } else {
#   a
# }
```

hadley:
Yeah this is definitely required by tidyverse style.

MichaelChirico:
OK, done

tests/testthat/test-if_else_match_braces_linter.R
```diff
@@ -0,0 +1,64 @@
...
+  expect_lint("if (TRUE) 1", NULL, if_else_match_braces_linter())
+
+  lines_brace <- trim_some("
+    if (TRUE) {
+      1
+    } else {
+      2
+    }
+  ")
+  expect_lint(lines_brace, NULL, if_else_match_braces_linter())
```
AshesITR, line 12: Just curious, why extract the `trim_some()` into a variable?

AshesITR (approved):
LGTM

tests/testthat/test-if_else_match_braces_linter.R
MichaelChirico: we do so internally with a preference erring on the side of assigning variables, especially in tests (where if the test fails, we can look at the intermediate output more easily).

i might not write it the same from scratch, but avoiding extra diffs from our internal source where possible

AshesITR (approved):

PR #987, opened 2022-03-23
New function_brace_linter

MichaelChirico:
Part of #962 

Shall we mark this one as default? I don't see it called out specifically in the style guide.


MichaelChirico:
We were not totally sure what to do with a case like we see in `lintr`:

https://github.com/r-lib/lintr/blob/b91dc57099ae218198ac3afec222fd86ec91fba9/R/package_hooks_linter.R#L82-L85

In the end we decided to enforce that having braces. But we might skip linting this type of usage optionally.

AshesITR:
https://style.tidyverse.org/syntax.html#inline-statements

It's written explicitly for `if ()` but I think it can be generalized to functions. Happy to have this as a default.

cc @jimhester

R/function_brace_linter.R
```diff
@@ -0,0 +1,28 @@
...
+#'
+#' This linter catches function definitions spanning multiple lines of code
+#'   that aren't wrapped in braces
+#'
+#' @evalRd rd_tags("function_brace_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+function_brace_linter <- function() {
+  Linter(function(source_file) {
+    if (length(source_file$parsed_content) == 0L) {
```
AshesITR, line 11: xml_parsed_content may still be NULL

MichaelChirico: nice catch!

AshesITR:
So, do we make it a default?

MichaelChirico:
> So, do we make it a default?

Yes. Any thoughts on the edge case-y example in the above comment? https://github.com/r-lib/lintr/pull/987#issuecomment-1075918853

AshesITR:
I'd lint it. It's not super easy to see at first glance that there is no closing parenthesis on the function definition line.

AshesITR (dismissed):

AshesITR (approved):

PR #989, opened 2022-03-23
New class_equals_linter

MichaelChirico:
Part of #962


AshesITR:
I've also seen `switch(class(x), ...)` in the wild; sometimes `switch(mode(x), ...)` is appropriate and sometimes an `if ... else if ...` chain.
IINM, `mode(x)` is at least guaranteed to be a scalar.

MichaelChirico:
awesome, great idea!

let's mark extensions like this as new issues for now -- once we have all the linters imported for #962, we can do another round of triage to see what's worth implementing in time for 3.0.0

R/class_equals_linter.R
```diff
@@ -0,0 +1,41 @@
...
+#'   for example [is.character()] or [is.data.frame()].
+#'
+#' Similar reasoning applies for `class(x) %in% "character"`
+#'
+#' @evalRd rd_tags("class_equals_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+class_equals_linter <- function() {
+  Linter(function(source_file) {
+    if (length(source_file$parsed_content) == 0L) {
```
AshesITR, line 15: Should check for xml_parsed_content since it requires XML

AshesITR (approved):

PR #996, opened 2022-03-25
New nested_ifelse_linter

MichaelChirico:
Part of #962


R/nested_ifelse_linter.R
```diff
@@ -0,0 +1,49 @@
...
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+
+    return(lapply(
+      bad_expr,
+      xml_nodes_to_lint,
+      source_file = source_file,
+      lint_message = function(expr) {
+        matched_call <- xml2::xml_text(xml2::xml_find_first(expr, "expr/SYMBOL_FUNCTION_CALL"))
+        lint_message <- sprintf("Don't use nested %s() calls; instead, try", matched_call)
+        paste(lint_message, "(1) data.table::fcase; (2) dplyr::case_when; or (3) merging to a lookup table.")
```
AshesITR, line 41: (3) may be better "using a lookup table". If the lookup is just ok one column, `new[match(Col, old)]` in general or `unname(new[old])` for charactes is the optimal code in my testing.

Or maybe "merging or matching to a lookup table"?

MichaelChirico: good idea. i like "using a lookup table". The merging approach is pretty ugly in general, if it's needed. The simple `named_vector[[lookup_keys]]` approach is nicer when it's possible.

AshesITR (approved):
Very nice. I'd consider this a candidate for default linters.
Also, common mistake? At least I've seen it many times in beginner code.
Happy to merge either way. LMK what you think.

MichaelChirico:
my hesitation for adding it as a default is that the "good" alternatives might require new dependencies in general. the base-only alternatives can get pretty ugly... I'll leave it out of defaults for now. we can always revisit

PR #997, opened 2022-03-25
New paste_sep_linter

MichaelChirico:
Part of #962


R/paste_sep_linter.R
```diff
@@ -0,0 +1,39 @@
...
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    paste_cond <- "expr[SYMBOL_FUNCTION_CALL[text() = 'paste']]"
+    sep_cond <- file.path(
+      "SYMBOL_SUB[text() = 'sep']",
+      # NB: length-2 means '' or "" (*not* 'ab' or 'cd' which have length 4)
+      "following-sibling::expr[1][STR_CONST[string-length(text()) = 2]]"
```
AshesITR, line 20: Edge case: What about `r"()"`?
Probably easier to understand with explicit comparisons.

NEWS.md
```diff
@@ -102,6 +102,7 @@ function calls. (#850, #851, @renkun-ken)
    * `any_is_na_linter()` Require usage of `anyNA(x)` over `any(is.na(x))`
    * `outer_negation_linter()` Require usage of `!any(x)` over `all(!x)` and `!all(x)` over `any(!x)`
    * `numeric_leading_zero_linter()` Require a leading `0` in fractional numeric constants, e.g. `0.1` instead of `.1`
+   * `paste_sep_linter()` Require usage of `paste0()` over `paste(sep = "")`
```
AshesITR, line 105: What about linting `paste0(..., sep = "<anything>")`?

R/paste_sep_linter.R
```diff
@@ -0,0 +1,39 @@
...
+    )
+    xpath_fmt <- "//expr[%s and %s]"
+    xpath <- sprintf(xpath_fmt, paste_cond, sep_cond)
```
AshesITR, line 23: Unused.

R/paste_sep_linter.R
MichaelChirico: yea... this is gonna come up a few times. we're still on R 3.6.3 so haven't had to lint for any raw strings stuff yet.

in the case of this linter, I think it's fine as is... `paste(sep = R'()')` seems unlikely.

this linter had mainly been useful for tidying up very old code written before paste0 existed, FWIW

PS we may have to do a thing where we merge first and tidy up later, for any case where there's something unique to do for R>4.0. it'll be tough to do such changes in my work environment.

NEWS.md
MichaelChirico: let's think about it in an issue. there are a few paste-related linters incoming. as I was writing this PR I thought it might make sense to merge them all into just one paste_linter.

another thing I see sometimes that I'd like to lint is paste0(x, collapse=). if there's only one input it's the same as the paste version, but it'd be nice to be consistent and the paste0 looks a bit distracting/misleading to me

R/paste_sep_linter.R
AshesITR: I can chip in with 4.x stuff.
It should also be safe and future proof to instead check `identical(eval(xml_text(sep_str_const)), "")`. WDYT?

R/paste_sep_linter.R
MichaelChirico: thanks!

AshesITR (approved):

PR #999, opened 2022-03-25
New stop_paste_linter

MichaelChirico:
Part of #962 

This one could probably go for an improved name as well, but I am consistently stumped here.


R/stop_paste_linter.R
```diff
@@ -0,0 +1,52 @@
...
+    xml <- source_file$xml_parsed_content
+
+    translators <- c("packageStartupMessage", "message", "warning", "stop")
+    xpath <- glue::glue("//expr[
+      expr[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(translators)} ]]
+      and expr[
+        expr[SYMBOL_FUNCTION_CALL[text() = 'paste' or text() = 'paste0']]
+        and not(SYMBOL_SUB[
+          text() = 'collapse'
+          or (
```
AshesITR, line 24: Better unnest the double negation.

R/stop_paste_linter.R
```diff
@@ -0,0 +1,52 @@
...
+            text() = 'sep'
+            and not(following-sibling::expr[1][STR_CONST[text() = '\"\"' or text() = '\" \"']])
```
AshesITR, line 26: This won't capture all string literals representing the empty string or a space.

MichaelChirico: left as a TODO for now

AshesITR:
Maybe call it `redundant_paste_linter()`? Could also add `cat()` to the list IINM.

MichaelChirico:
`cat()` is slighlty different in that its default is `' '` not `''`. but maybe right that using `cat`'s own `sep` is always preferable to building the string with `paste()`/`paste0()`.

However I have some gut feeling about why we didn't do that in the first place... can't place exactly why. I will run an updated linter that catches `cat()` against our internal codebase first to make sure nothing sticks out to me.

AshesITR:
`cat()` also auto-collapses using `sep`. Maybe that's why?

```r
cat(">", letters, "<")
# > a b c d e f g h i j k l m n o p q r s t u v w x y z <
```

MichaelChirico:
Good point... so I guess it might be a bit muddled to include that in the same linter.

What about `singal_construction_linter`? Maybe too broad?

AshesITR:
Sounds too unfamiliar I think. What about `message_linter()` or `message_construction_linter()`?
It should be clear what a message is, plus `message()` is also linted.

man/condition_message_linter.Rd
```diff
@@ -0,0 +1,20 @@
...
+\code{stop(paste0(...))} is strictly redundant -- \code{stop(...)} is equivalent.
+\code{stop(...)} is also preferable to \code{stop(paste(...))}. The same applies to
+all default condition functions, i.e., \code{\link[=stop]{stop()}}, \code{\link[=warning]{warning()}}, \code{\link[=message]{message()}},
+and \code{\link[=packageStartupMessage]{packageStartupMessage()}}.
+}
+\seealso{
+\link{linters} for a complete list of linters available in lintr.
+}
+\section{Tags}{
+No tags are given.
```
AshesITR, line 19: needs a document()

R/condition_message_linter.R
```diff
@@ -0,0 +1,56 @@
+#' Block usage of paste() and paste0() with messaging functions using ...
+#'
+#' `stop(paste0(...))` is strictly redundant -- `stop(...)` is equivalent.
+#'   `stop(...)` is also preferable to `stop(paste(...))`. The same applies to
+#'   all default condition functions, i.e., [stop()], [warning()], [message()],
+#'   and [packageStartupMessage()].
+#'
+#' @evalRd rd_tags("stop_paste_linter")
```
AshesITR, line 8: old name needs to be replaced here for proper docs.

AshesITR (changes requested):
LGTM apart from the doc problem.

R/condition_message_linter.R
MichaelChirico: ah good catch. this might be an issue elsewhere too...

AshesITR: Found two: expect_s4_class_linter and conjunct_expectation_linter
image

MichaelChirico: I'll add the other one to #1014

AshesITR:
Should we `stop()` or at least issue a `warning()` if tags are empty? The "No tags were given" message seems to be too quiet...

AshesITR (approved):

PR #1000, opened 2022-03-26
New redundant_ifelse_linter

MichaelChirico:
Part of #962


R/redundant_ifelse_linter.R
```diff
@@ -0,0 +1,62 @@
+#' Prevent ifelse() from being used to produce TRUE/FALSE or 1/0
+#'
+#' Expressions like `ifelse(x, TRUE, FALSE)` and `ifelse(x, FALSE, TRUE)` are
```
AshesITR, line 3: I think `ifelse(cond, 1.0, 0.0)` is fine actually, because it doesn't require the reader to know that `as.numeric(cond)` has the same result.

WDYT about making the numeric coercion lint configurable?

tests/testthat/test-redundant_ifelse_linter.R
```diff
@@ -0,0 +1,79 @@
...
+  )
+  expect_lint(
+    "ifelse(x > 5, 0L, 1L)",
+    rex::rex("Prefer as.integer(x) to ifelse(x, 0L, 1L)"),
+    redundant_ifelse_linter()
+  )
+
+  expect_lint(
+    "ifelse(x > 5, 1, 0)",
+    rex::rex("Prefer as.numeric(x) to ifelse(x, 1, 0)"),
```
AshesITR, line 36: Nit: IINM 1 and 0 are implicitly integers.

MichaelChirico:
```
typeof(ifelse(runif(10) > .5, 1, 0))
# [1] "double"
```

?

R/redundant_ifelse_linter.R
MichaelChirico: SGTM; done.

Leaving linting on by default. I do think `ifelse(cond, 1, 0)` is still usually the wrong thing to do.

AshesITR:
Needs a document().

AshesITR:
`@param allow10` is missing

AshesITR (approved):

PR #1002, opened 2022-03-26
New else_same_line_linter

MichaelChirico:
Part of #962

This one is also a candidate for `default_linters`


AshesITR:
Do we already have a linter that ensures the opening braces are ob the same line?

Name suggestion: `else_braces_linter`?

MichaelChirico:
I was thinking maybe we want to combine this with `if_else_match_braces_linter` into just an `if_else_linter` that identifies any style issues with `if`/`else` clauses, WDYT?

AshesITR:
Both of these lint based on braces. What about `if_else_braces_linter`?
Or just `braces_linter` subsuming `braces_left_parenthesis_linter` as well and also (if not already present) checking `{` appears exactly one space after and on the same line as `function() {`, `while (cond) {`, `for () {`, `repeat {` etc.

MichaelChirico:
`braces_linter` sounds great. Agree it's good to have a one-stop shop for enforcing all of the style guide rules around bracing.

IINM it'll require some more deprecation though, right? If so, maybe the right approach is

 1. Merge this as is
 2. File a follow-up to merge the linters
 3. Merge the linters
 4. Mark any merged linters appearing in previous releases as deprecated

WDYT?

AshesITR:
SGTM. Maybe make the new one configurable to allow C style bracing etc.
deprecation could then be made to return a properly configured `braces_linter()` maybe.

Approving. Will you open a follow-up issue? Thanks!

AshesITR (approved):

AshesITR (approved):

PR #1003, opened 2022-03-27
New unreachable_code_linter

MichaelChirico:
Part of #962 

The current version of this works best alongside our `ExplicitReturnLinter()`, where terminal statements will already be `return()`. The logic of nailing down the last expression in xpath is a bit of a mess so we avoided replicating it inside this linter. Now I'm wondering how useful this linter is without that added logic  🤔


R/unreachable_code_linter.R
```diff
@@ -0,0 +1,45 @@
...
+
+    # NB:
+    #  - * returns all children, including the terminal }, so the position
+    #    is not last(), but last()-1. If there's no }, this linter doesn't apply.
+    #  - land on the culprit expression
+    xpath <- "
+    //FUNCTION
+    /following-sibling::expr
+    /*[
+      self::expr
```
AshesITR, line 27: This is equivalent to using `/expr` instead of `/*`

R/unreachable_code_linter.R
```diff
@@ -0,0 +1,45 @@
...
+      and expr[SYMBOL_FUNCTION_CALL[text() = 'return' or text() = 'stop']]
+      and (position() != last() - 1 or not(following-sibling::OP-RIGHT-BRACE))
```
AshesITR, line 29: Curious: Does this cope properly with return in switch?

AshesITR (commented):
This one might benefit from running on some packages in the wild. WDYT?

MichaelChirico:
> This one might benefit from running on some packages in the wild. WDYT?

I recommend covering all the new linters together.

FWIW, the linters have all been vetted on our internal code base (O(10k) files) so I'm reasonably confident about false positives. False negatives are something we can always improve iteratively.

R/unreachable_code_linter.R
MichaelChirico: not quite, because `position()` needs to include all nodes, not just `expr`

R/unreachable_code_linter.R
MichaelChirico: Indeed it doesn't -- just added an example test (commented out) to make sure I captured what you had in mind.

I think `tryCatch()` is equally affected.

AshesITR (approved):

PR #1004, opened 2022-03-28
New regex_subset_linter

MichaelChirico:
Part of #962


R/regex_subset_linter.R
```diff
@@ -0,0 +1,88 @@
...
+      "not(parent::*[LEFT_ASSIGN or EQ_ASSIGN or RIGHT_ASSIGN])"
+    )
+    # However the equality operator in XPath works for expr[.] = expr/expr[.],
+    #   i.e., node1 = node2, suits our purpose -- we need that the
+    #   xml2::xml_text() that would come out from the matched nodes is the same,
+    #   i.e. that whatever expression comes in <expr>[grepl(pattern, <expr>)]
+    #   matches exactly, e.g. names(x)[grepl(pattern, names(x))].
+    subset_cond_fmt <- xp_and(
+      "expr[SYMBOL_FUNCTION_CALL[%s]]",
+      "expr[position() = %d] = parent::expr/expr[1]"
```
AshesITR, line 46: Cool! What is the XPath definition of `=` in this context?
It must ignore attributes I see (col1 e.g. will differ), anything else?

R/regex_subset_linter.R
```diff
@@ -0,0 +1,88 @@
...
+regex_subset_linter <- function() {
+  Linter(function(source_file) {
+    if (length(source_file$xml_parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    parent_expr_cond <- xp_and(
+      "OP-LEFT-BRACKET",
```
AshesITR, line 34: Is `LBB` also worth linting?

AshesITR (dismissed):

R/regex_subset_linter.R
MichaelChirico: I genuinely don't know. would be great to find documentation on this... thus far it's "just worked" wherever needed though (also tested in a few other linters, not sure if any are included yet)

MichaelChirico: Thanks for the spur to check again... found some docs!

https://www.w3.org/TR/1999/REC-xpath-19991116/#booleans

Courtesy this SO Q&A: https://stackoverflow.com/a/4629660/3576984

MichaelChirico: Added that link to the comment

AshesITR (approved):

PR #1006, opened 2022-03-28
New system_file_linter

MichaelChirico:
Part of #962


AshesITR:
Related: Prevent literals with `/` or `\` in them for `file.path` and `system.file`.
Should check path linters if they know `system.file`.

AshesITR (approved):

PR #1011, opened 2022-03-28
Extend conjunct_expectation_linter to include stopifnot(), rename to conjunct_test_linter

MichaelChirico:
Part of #962 

Similar naming issue to our `TrueFalseAndConditionLinter` which became `conjunct_expectation_linter`. We might consider merging them into just one linter.


MichaelChirico:
> I think the old name still fits.

To me `expectation` is pretty related to `testthat` though, `conjunct_assertion` would be more appropriate here for `stopifnot()`...

`conjunct_test_linter` maybe?

R/conjunct_stopifnot_linter.R
```diff
@@ -0,0 +1,39 @@
...
+conjunct_stopifnot_linter <- function() {
+  Linter(function(source_file) {
+    # need the full file to also catch usages at the top level
+    if (length(source_file$full_xml_parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$full_xml_parsed_content
+
+    xpath <- "
```
MichaelChirico, line 19: I also need to add a TODO for R>=4.0 not to lint in this case:

```
stopifnot(`Descriptive reason` = A && B)
```

possibly optionally

AshesITR: Hmm. Could still be written as
```
stopifnot(
  `Descriptive reason, A failed` = A,
  `Descriptive reason, B failed` = B
)
```

AshesITR:
Fine with that as well.

> > I think the old name still fits.
> 
> To me `expectation` is pretty related to `testthat` though, `conjunct_assertion` would be more appropriate here for `stopifnot()`...
> 
> `conjunct_test_linter` maybe?

Fine by me. `assertthat::assert_that` could be added to the list of linted functions.

AshesITR:
Nope, `assertthat::assert_that(A && B, msg = x)` is equivalent to `assertthat::assert_that(A, B, msg = x)`, very similar to testthat.

R/conjunct_stopifnot_linter.R
MichaelChirico: Right, but it can be tedious... consider

```
stopifnot(
  `x is a logical scalar` = length(x) == 1 && is.logical(x) && !is.na(x),
  `y is a logical scalar` = length(y) == 1 && is.logical(y) && !is.na(y), 
)
```

So I think it should be optional.

note that this is equivalent to allowing

```
assertthat::assert_that(length(x) == 1, is.logical(x), !is.na(x), msg="x is a logical scalar")
```

instead of requiring several `assert_that()` calls with a unique `msg`

R/conjunct_test_linter.R
```diff
@@ -17,9 +17,11 @@ conjunct_expectation_linter <- function() {
 
     xml <- source_file$full_xml_parsed_content
 
+    # TODO(#1016): include assert_that() for consideration here too
```
AshesITR, line 20: That would be just ` or text() = 'assert_that'` in the XPath

MichaelChirico: right... plus tests. better not to clutter this PR with that

AshesITR (approved):

Issue #998, opened 2022-03-25
Combine all paste linters


AshesITR:
let's think about it in an issue. there are a few paste-related linters incoming. as I was writing this PR I thought it might make sense to merge them all into just one paste_linter.

another thing I see sometimes that I'd like to lint is paste0(x, collapse=). if there's only one input it's the same as the paste version, but it'd be nice to be consistent and the paste0 looks a bit distracting/misleading to me

__Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/997#discussion_r835576040__

AshesITR:
The `toString()` lint could also point to `glue::glue_collapse()` in its message in case the actual intention is to create a human friendly enumeration.

MichaelChirico:
Closing here; tracking the rest in #1108 after release

PR #1013, opened 2022-03-28
New paste_to_string_linter

MichaelChirico:
Part of #962 

Also part of #998 -- for now, leaving them unmerged for simplicity (will require a fair amount of manual work relative to my automated script for importing our linters). I'll also take a pass at tidying up the xpath while addressing #998.


R/paste_to_string_linter.R
```diff
@@ -0,0 +1,37 @@
...
+      sprintf(collapse_cond_fmt, xp_text_in_table(c("', '", '", "'))),
+      sprintf(paste_cond_fmt, xp_text_in_table(c("paste", "paste0")))
+    )
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+
+    return(lapply(
+      bad_expr,
+      xml_nodes_to_lint,
+      source_file = source_file,
+      lint_message = 'toString(.) is more expressive than paste(., collapse = ", ")',
```
AshesITR, line 33: Can we hint at `glue::glue_collapse()` for human-friendly formatting?
```
glue::glue_collapse(letters[1:5], sep = ", ", last = " and ")
#> a, b, c, d and e
```

R/paste_to_string_linter.R
```diff
@@ -0,0 +1,37 @@
...
+    paste_cond_fmt <- "expr[SYMBOL_FUNCTION_CALL[%s]]"
+    collapse_cond_fmt <- "SYMBOL_SUB[text() = 'collapse']/following-sibling::expr[1][STR_CONST[%s]]"
+
+    # 3 expr: the function call, the argument, and collapse=
+    # TODO(michaelchirico): catch raw-string equivalents
+    xpath <- sprintf(
+      "//expr[count(expr) = 3 and %s and %s]",
+      sprintf(collapse_cond_fmt, xp_text_in_table(c("', '", '", "'))),
+      sprintf(paste_cond_fmt, xp_text_in_table(c("paste", "paste0")))
+    )
```
AshesITR, line 26: `glue()` this?

AshesITR:
Build failure seems to be caused by `knitr` again.

R/paste_to_string_linter.R
MichaelChirico: > the issue is with the mixed `'` and `"`, at least one of which will have to be escaped in `glue` --> less readable. I'll take another look though.

R/xp_utils.R
```diff
@@ -1,7 +1,15 @@
...
 
 # like `text() %in% table`, translated to XPath 1.0
-xp_text_in_table <- function(table) paste("text() = ", quote_wrap(table, "'"), collapse = " or ")
+xp_text_in_table <- function(table) {
+  # xpath doesn't seem to have a standard way of escaping quotes, so attempt
+  #   to use "" whenever the string has ' (not a perfect solution). info on
+  #   escaping from https://stackoverflow.com/questions/14822153
+  single_quoted <- grepl("'", table, fixed = TRUE)
+  table[single_quoted] <- dQuote(table[single_quoted], '"')
+  table[!single_quoted] <- sQuote(table[!single_quoted], "'")
```
AshesITR, line 10: This is R >= 3.6 only.

> * sQuote() and dQuote() get an explicit q argument with obvious default instead of using getOption("fancyQuotes") implicitly and unconditionally.

R/paste_to_string_linter.R
AshesITR: Could factor that out into a variable (`comma_space_strings`?)
Too bad we can't use raw strings in the package source code.

AshesITR:
ERROR: unable to collate and parse R files for package ‘lintr’

document() should fix that.

MichaelChirico:
> ERROR: unable to collate and parse R files for package ‘lintr’

I think because of the issue fixed in [9cf6f61](https://github.com/r-lib/lintr/pull/1013/commits/9cf6f61a0a3a334ddb17d6814566671f37eefc61). but document() did catch an update that wasn't captured yet.

man/linters.Rd
```diff
@@ -84,6 +84,7 @@ The following linters exist:
 \item{\code{\link{paren_body_linter}} (tags: default, readability, style)}
 \item{\code{\link{paren_brace_linter}} (tags: default, readability, style)}
 \item{\code{\link{paste_sep_linter}} (tags: best_practices, consistency)}
+\item{\code{\link{paste_to_string_linter}} (tags: best_practices, consistency)}
```
AshesITR, line 87: why didn't the tag counts change as well?

MichaelChirico: funky merging... ran again and it updated. confirmed the new counts are right:

```
grep "consistency" inst/lintr/linters.csv | wc -l
# 16
grep "best_practices" inst/lintr/linters.csv | wc -l
# 35
```

AshesITR (approved):

PR #1014, opened 2022-03-28
Fix tags for a linter that was renamed in transit

MichaelChirico:
Same issue as was caught in #999


AshesITR:
Ditto for `expect_s4_class_linter()`, which has the _s3_ version in `@evalRd`.

MichaelChirico:
Since `conjunct_expectation_linter` is renamed in #1011, I'll revert here (it's fixed in that PR already), and just fix `expect_s4_class_linter()`

AshesITR (approved):

AshesITR (approved):

PR #1015, opened 2022-03-28
extend `available_linters(tags = )`, add `linters_with_tags()`

### LINKED ISSUE #980: tags= argument for available_linters()
As well as filtering by `package`, it would be helpful to run `available_linters(tags = "package_development")` and receive all the rows associated with `package_development` linters

AshesITR:
Fixes #980

I moved the `with_defaults()` test from `test-settings.R` to `test-with.R` and moved the implementation from `zzz.R` to `with.R`.

`with_tags()` also allows us to test that `default_linters` is synchronized to `available_linters(tags = "default")` so I did that.


AshesITR:
Maybe adding a crossref from `available_linters()` to  `with_tags()` is a good idea?
And maybe also from `?linters` to help discover the new feature.
Since `with_tags()` can be used as a drop-in replacement for `with_defaults()` as long as `default =` is missing, I think `with_tags()` could even be documented to become the standard way of configuring lintr. WDYT?

AshesITR:
Not sure what is up with `codecov()` erring out for `all.equal()` when `ubuntu-release` doesn't and both run R 4.1.3...

AshesITR:
```
2022-03-28T20:45:06.8497300Z ##[error]Error: Failure in `/tmp/RtmpjHIBqn/R_LIBS2488420bd9ed/lintr/lintr-tests/testthat.Rout.fail`
2022-03-28T20:45:06.8506140Z ol_linter\": target, current do not match when deparsed"
2022-03-28T20:45:06.8506879Z - "Component \"trailing_blank_lines_linter\": target, current do not match when deparsed"
2022-03-28T20:45:06.8507591Z - "Component \"trailing_whitespace_linter\": target, current do not match when deparsed"
2022-03-28T20:45:06.8508204Z - "Component \"vector_logic_linter\": target, current do not match when deparsed"
2022-03-28T20:45:06.8508735Z + "Component \"line_length_linter\": Component \"length\": Mean relative difference: 0.5"
```

Sounds like the `all.equal()` check in codecov compares `deparse(fun)`...

MichaelChirico:
> EDIT: Found the problem. covr::report() modifies functions to count line hits. These modifications cause inconsistent deparse() results in our case it seems.

seems like a `covr` bug... will that block submission here?

AshesITR:
Nope. Added a skip nevertheless.

NEWS.md
```diff
@@ -93,6 +93,7 @@ function calls. (#850, #851, @renkun-ken)
   * Each linter can have multiple tags
   * New function `available_linters()` to list available linters and their tags 
     This feature is extensible by package authors providing add-on linters for {lintr}.
+  * New function `with_tags()` to help build a list of linters using tags.
```
MichaelChirico, line 96: I think the name here only makes sense if you're very used to `with_defaults()`. but since we basically want to supersede that usage, we should start from scratch thinking about the name here.

```
linters_with_tags()
select_linters_with_tags()
get_linters_by_tag()
```

something along those lines?

tests/testthat/test-with.R
```diff
@@ -0,0 +1,29 @@
...
+  expect_named(with_defaults(assignment_linter), names(with_defaults()))
+})
+
+test_that("all default linters are tagged default", {
+  expect_named(with_defaults(), available_linters(tags = "default")$linter)
+
+  # TODO(michaelchirico): use plain expect_equal after waldo#133 makes it into a CRAN release
+  # Here, the environment()s are different because factories use them.
+  skip_if_not(getRversion() >= "4.1") # Desired all.equal behaviour only available in >= 4.1
+  # covr modifies package functions causing differing deparse() results even for identical anonymous functions.
```
MichaelChirico, line 12: let's file a bug & reference it here to eventually remove it

NEWS.md
AshesITR: Bear in mind the most common place for this is in a `.lintr` file:

```
linters: with_tags(tags = c("default", "package_development"))
```

Maybe `linters_with_tags()` would be fine, but neither get nor select correctly describe the function, since it actively constructs linters from a factory.

MichaelChirico: I am liking `linters_with_tags()` the most

tests/testthat/test-with.R
AshesITR: Not sure if covr can do much here because the `default_linters` list is prebuilt. If my suspicion is correct, `with_tags()[["assignment_linter"]]` should be equal to `assignment_lintr()` even in covr, because both are constructed at runtime.

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  is.na(vals) <- nms == vals
+  default[nms] <- vals
+
+  res <- default[!vapply(default, is.null, logical(1L))]
+
+  res[] <- lapply(res, function(x) {
+    prev_class <- class(x)
+    if (inherits(x, "function") && !inherits(x, "lintr_function")) {
```
MichaelChirico, line 66: let's use `is.function(x)` instead?

```
f = lm
class(f) = 'boo'
inherits(f, 'function')
# [1] FALSE
is.function(f)
# [1] TRUE
```

AshesITR: Let's refactor `with_defaults()` in a separate PR to keep this focused.

tests/testthat/test-with.R
AshesITR: Elaborated the comment in hopes of making the dilemma clear. Not a covr bug in my view -- simply an artifact of how we define `default_linters`.

NEWS.md
```diff
@@ -93,6 +93,7 @@ function calls. (#850, #851, @renkun-ken)
   * Each linter can have multiple tags
   * New function `available_linters()` to list available linters and their tags 
     This feature is extensible by package authors providing add-on linters for {lintr}.
+  * New function `linters_with_tags()` to help build a list of linters using tags.
```
MichaelChirico, line 96: The NEWS should mention the supersession plan? Or we save that until #1029 is done?

R/linter_tags.R
```diff
@@ -3,6 +3,8 @@
 #' Obtain a tagged list of all Linters available in a package.
 #'
 #' @param packages A character vector of packages to search for linters.
+#' @param tags Optional character vector of tags to search. Only linters with at least one matching tag will be
+#' returned. If `tags` is `NULL`, all linters will be returned.
```
MichaelChirico, line 7: Should we export an `available_tags()` too to help downstreams align to our terminology? Just a simple wrapper like

```
available_tags <- function(packages = "lintr") {
  unique(unlist(available_linters(packages = packages)[["tags"]]))
}
```

But saves them the effort of knowing how to do this & raises the prominence of it.

Feel free to mark as a follow-up.

R/with.R
```diff
@@ -0,0 +1,125 @@
+#' Modify lintr defaults
```
MichaelChirico, line 1: really wish github made it clearer whether this is just copy-pasted from the old file :|

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+    if (inherits(x, "function") && !inherits(x, "lintr_function")) {
+      class(x) <- c(prev_class, "lintr_function")
+    }
+    x
+  })
+}
+
+#' Create a tag-based linter configuration
+#'
+#' Make a new list based on all linters tagged with `tags` from provided by `packages`.
```
MichaelChirico, line 75: > Make a new list based on all linters tagged with `tags` from provided by `packages`.

grammar?

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+#' The result of this function is meant to be passed to the `linters` argument of `lint()`, or put in your
+#' configuration file.
+#'
+#' @param ... Arguments of elements to change. If unnamed, the argument is automatically named.
+#' If the named argument already exists in the list of linters, it is replaced by the new element.
+#' If it does not exist, it is added. If the value is `NULL`, the linter is removed.
+#' @inheritParams available_linters
+#'
+#' @return A modified list of linters.
+#' @seealso [with_defaults] for basing off lintrs set of default linters.
```
MichaelChirico: nit:

does

```
#' @seealso
#'  [with_defaults] for basing off lintrs set of default linters.
#'  [available_linters] to get a data frame of available linters.
#'  [linters] for a complete list of linters available in lintr.
```

produce the same Rd content? If so that would look cleaner to me

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+#' @export
+linters_with_tags <- function(..., tags = "default", packages = "lintr") {
+  tagged_linters <- list()
+
+  for (package in packages) {
+    pkg_ns <- loadNamespace(package)
+    available <- available_linters(packages = package, tags = tags)
+    if (nrow(available) > 0L) {
+      linter_factories <- mget(available$linter, envir = pkg_ns)
+      linters <- mapply(function(linter_factory, linter_name) {
```
MichaelChirico, line 109: looks like too much nesting here. maybe like this instead?

```
build_linter <- function(linter_factory, linter_name) {
  linter <- tryCatch(linter_factory(), error = identity)
  if (inherits(linter, "error")) {
    stop("Could not create linter with ", package, "::", linter_name, "(): ", conditionMessage(linter))
  }
  # Otherwise, all linters would be called "linter_factory".
  attr(linter, "name") <- linter_name
  linter
}

linters <- mapply(build_linter, linter_factories, names(linter_factories))
```

?

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+  })
+}
+
+#' Create a tag-based linter configuration
+#'
+#' Make a new list based on all linters tagged with `tags` from provided by `packages`.
+#' The result of this function is meant to be passed to the `linters` argument of `lint()`, or put in your
+#' configuration file.
+#'
+#' @param ... Arguments of elements to change. If unnamed, the argument is automatically named.
```
MichaelChirico, line 79: > Arguments of elements to change.

Not clear what this means. Maybe just "Passed on to [with_defaults()]"? Or `@inheritParams with_defaults`?

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+#' # Get all linters useful for package development
+#' linters_with_tags(tags = "package_development")
+#'
+#' # Get all linters provided by lintr
+#' linters_with_tags(tags = NULL)
+#'
+#' # Get all linters tagged as "default" from lintr and mypkg
+#' \dontrun{linters_with_tags(packages = c("lintr", "mypkg"))}
+#' @export
+linters_with_tags <- function(..., tags = "default", packages = "lintr") {
```
MichaelChirico, line 101: my sense is there should be no default for `tags=`. Maybe the signature should be switched, too, to allow `tags=` supplied positionally?

```
linters_with_tags <- function(tags, ..., packages = "lintr") 
```

Seeing

```
linters: linters_with_tags("default")
```

in config looks more natural to me.

R/with.R
```diff
@@ -0,0 +1,125 @@
...
+  tagged_linters <- list()
+
+  for (package in packages) {
+    pkg_ns <- loadNamespace(package)
+    available <- available_linters(packages = package, tags = tags)
+    if (nrow(available) > 0L) {
+      linter_factories <- mget(available$linter, envir = pkg_ns)
```
MichaelChirico, line 108: maybe `ifnotfound = rep(list(NULL), nrow(available))` and then check `lengths(linter_factories)` as an extra safety check, in case downstreams' linter DB is out of sync?

tests/testthat/test-with.R
MichaelChirico: i see, makes sense, thanks. agree with the skip.

MichaelChirico (commented):
Getting closer, thanks!

R/with.R
AshesITR: I opted to copy the params but sure, can inherit. Not sure if worth the effort because this will be touched soon.
Maybe the naming part is best described in `@return`?

R/with.R
AshesITR: I'd rather `stop()` in that case since it's a bug in {package}. Maybe craft an informative message if `mget` doesn't already do that. WEYT?

NEWS.md
AshesITR: Best done in tandem with #1029 I think, when we know what the final outcome of the refactoring will be.

R/linter_tags.R
AshesITR: Sounds good to me.

R/with.R
AshesITR: Happy to include packages twice, looked less elegant to me though

R/with.R
AshesITR: Modulo a few empty lines, it does. Changed.

R/with.R
AshesITR: Factored out to top-level for easier debugging.

R/with.R
AshesITR: Added a `getNamespaceExports()` based check to create a nice error message. `mget()` produces `Error: 'nonexistent_linter' not found` with no clue as to where it wasn't found and why we tried to get it.

MichaelChirico: right, I was thinking to stop() after ifnotfound... getNamespaceExports works too!

AshesITR:
PR fails seem to originate elsewhere.
Also, the test warnings() are back it seems...

https://github.com/r-lib/lintr/runs/5749209064?check_suite_focus=true#step:6:177

R/with.R
MichaelChirico: just a grammar question "from provided by packages" doesn't parse to me

R/with.R
MichaelChirico: ok let's revisit in future PRs

tests/testthat/test-linter_tags.R
```diff
@@ -13,6 +13,17 @@ test_that("available_linters returns a data frame", {
...
 
+test_that("available_tags returns a character vector", {
+  tags <- available_tags()
+  tags2 <- available_tags(c("lintr", "not-a-package"))
+  empty <- available_tags("not-a-package")
+
+  expect_type(tags, "character")
+  expect_identical(tags, tags2)
+  expect_length(empty, 0L)
+  expect_true(all(available_linters()$tags[[1L]] %in% tags))
```
MichaelChirico, line 24: why just the first row?

AshesITR: Just a smell test.
Since the implementation is so simple, I didn't feel the need to loop all rows. LMK if you want to see this added.

R/with.R
AshesITR: 🤦🏻  removing "from"...

R/with.R
AshesITR: Yeah, detecting file splits in addition to file renames would be nice.

tests/testthat/test-linter_tags.R
MichaelChirico: unlist(db$tags) instead?

AshesITR: That's a fancy way of testing `unlist(db$tags) == unlist(db$tags)` 🤷🏻

MichaelChirico: but then the current test is just a fancy way of testing 

db$tags[[1]] %in% unlist(db$tags)

no?

AshesITR: ... which at least ensures that `db$tags[[1]]` isn't a list.
What bothers me with `unlist()` is that that's precisely the implementation, so we don't test anything at all about the aggregation.

But we can test for uniqueness, via `anyDuplicated()`.

MichaelChirico: I see, but that wasn't communicated by the test as written. maybe unlist(resursive=FALSE) would make that clearer?

anyway current tests look good enough, no sense blocking the PR any further 

>

MichaelChirico (approved):

Issue #963, opened 2022-03-20
Write an XPath for extended known-literal expressions


MichaelChirico:
Follow-up to https://github.com/r-lib/lintr/pull/957#discussion_r830463257

There are a few places that would benefit from some shared logic about what constitutes a "literal expression" along the lines of these:

```
const_exprs <- c(
  '"a"',
  '1',
  '1+1i',
  'c("a")',
  'c("a", 1)',
  'c("a", 1+1i)',
  'rep("a", 2)',
  'rep("a", n)',
  'rep(c("a", "b"), 5)',
  'c(rep("a", 10), rep("b", 10))'
)
```

but excluding non-constant expressions like:

```
non_constant_exprs <- c(
  'a',
  'a + 1',
  'foo(1)',
  'foo(c(1, 2))'
)
```

I think `c()` and `rep()` are the most common ways to build constant expressions, so we can limit it to combinations of those. Note that I consider `rep(1:3, n)` as being "constant" here, as the variable part is "constant enough" for the places where we'd use this:

 - `strings_as_factors_linter` which looks for `data.frame`s declared with known-character columns
 - `expect_{type,length,named,s3_class}_linter` which look for literal exprs in the 1st argument
 - `yoda_test_linter` which does the same

AshesITR:
I'm not sure whether it's good to call the `rep(1:3, n)` expression a constant (but willing to call it "almost constant").
Regarding functions used: `1:3 == seq_len(3)` is what I default to due to `1:n` being bad for `n == 0`.
This means we could whitelist `seq`, `seq_len`, `seq_along` and `seq.int` as well.

PR #1020, opened 2022-03-28
New strings_as_factors_linter

MichaelChirico:
Part of #962 

Can surely be tidied up a bit as part of #963, but filing as is for now to finish of #962.


R/strings_as_factors_linter.R
```diff
@@ -0,0 +1,77 @@
...
+    )
+    bad_expr <- xml2::xml_find_all(xml, sprintf("//expr[%s]", expr_cond))
+
+    return(lapply(
+      bad_expr,
+      xml_nodes_to_lint,
+      source_file = source_file,
+      lint_message = paste(
+        "This code relies on stringsAsFactors=TRUE, which will",
+        "change in the future. Please supply this argument explicitly,",
```
AshesITR, line 71: The future as in roughly a year ago 😄 
Better say changes in R >= 4.0.0?
Doesn't this lint any `data.frame()` call with a literal character argument and without explicit `stringsAsFactors`?
In that case the message should be "provide stringsAsFactors or don't use character constants", no?

MichaelChirico: good point! took another pass at it.

tests/testthat/test-strings_as_factors_linter.R
```diff
@@ -0,0 +1,148 @@
+test_that("strings_as_factors_linter skips allowed usages", {
+  expect_lint("data.frame(1:3)", NULL, strings_as_factors_linter())
+  expect_lint("data.frame(x = 1:3)", NULL, strings_as_factors_linter())
+
+  expect_lint("data.frame(x = 'a', stringsAsFactors = TRUE)", NULL, strings_as_factors_linter())
+  expect_lint("data.frame(x = 'a', stringsAsFactors = FALSE)", NULL, strings_as_factors_linter())
+  expect_lint("data.frame(x = c('a', 'b'), stringsAsFactors = FALSE)", NULL, strings_as_factors_linter())
+
+  # strings in argument names to c() don't get linted
+  expect_lint("data.frame(x = c('a b' = 1L, 'b c' = 2L), stringsAsFactors = FALSE)", NULL, strings_as_factors_linter())
```
AshesITR, line 10: Why provide `stringsAsFactors` in this test case?

R/strings_as_factors_linter.R
```diff
@@ -0,0 +1,77 @@
...
+      'not(SYMBOL_SUB[text() = "stringsAsFactors"])'
+    )
+    bad_expr <- xml2::xml_find_all(xml, sprintf("//expr[%s]", expr_cond))
+
+    return(lapply(
+      bad_expr,
+      xml_nodes_to_lint,
+      source_file = source_file,
+      lint_message = paste(
+        "This code relies on stringsAsFactors=TRUE, which changed",
```
AshesITR, line 70: Incorrect message if code is for R 3.6. Maybe "relies on the value of stringsAsFactors, whose default changed in R 4.0"?

tests/testthat/test-strings_as_factors_linter.R
MichaelChirico: Hmm maybe just sloppy-pasted. But turns out there's a real false positive there!

Probably not picked up on our side because quoted keyword args aren't allowed if not needed, and passing literally-named vectors to `data.frame()` seems uncommon. So let's file as a follow-up bug for now.

R/strings_as_factors_linter.R
```diff
@@ -67,9 +67,10 @@ strings_as_factors_linter <- function() {
       xml_nodes_to_lint,
       source_file = source_file,
       lint_message = paste(
-        "This code relies on stringsAsFactors=TRUE, which changed",
-        "in version R 4.0. Please supply an explicit value for",
-        "stringsAsFactors for this code to work with earlier versions of R."
+        "This code relies on the default value of stringsAsFactors,",
+        "which changed in version R 4.0. Please supply an explicit value for",
+        "stringsAsFactors for this code to work versions of R both before",
```
AshesITR, line 72: grammar (to work _with_ versions of R)

tests/testthat/test-strings_as_factors_linter.R
```diff
@@ -20,58 +21,58 @@ test_that("strings_as_factors_linter skips allowed usages", {
 test_that("strings_as_factors_linter blocks simple disallowed usages", {
   expect_lint(
     "data.frame('a')",
-    rex::rex("This code relies on stringsAsFactors=TRUE"),
+    "This code relies on the default value of stringsAsFactors",
```
AshesITR, line 24: We factor the message out to a variable in some cases. May be worth here too, to keep the diff small upon change in wording.

R/strings_as_factors_linter.R
```diff
@@ -67,9 +67,10 @@ strings_as_factors_linter <- function() {
...
+        "and after this split."
```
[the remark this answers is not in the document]
MichaelChirico: redundant with usage of `change` in previous sentence :)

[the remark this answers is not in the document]
AshesITR: True :) So we can shorten to
"This code relies on the default value of stringsAsFactors, which changed in version R 4.0. Please supply an explicit value for stringsAsFactors."

tests/testthat/test-strings_as_factors_linter.R
MichaelChirico: i think there's a tradeoff with readability:

https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests

R/strings_as_factors_linter.R
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    atomic_const_cond <- "STR_CONST"
+    # like data.frame('a') and data.frame(c('a', 'b'))
+    c_cond <- xp_and(
+      "expr[SYMBOL_FUNCTION_CALL[text() = 'c']]",
+      "expr[STR_CONST]",
+      "not(expr[SYMBOL or expr])"
```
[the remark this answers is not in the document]
MichaelChirico: that didn't quite work (3 new test failures). Let's save for follow-up.

tests/testthat/test-strings_as_factors_linter.R
AshesITR: Not sure I follow here, it would reduce the line count by half and highlight the relevant part, namely the input that should lint.
Readability is fine imo because the negative test cases are separate from the positive cases.

AshesITR: Ah, I see. Keeping the variables local to the `test_that()` scope seems like a good trade-off to me. My IDE allows jumping to a definition and back quickly, so I favor less duplication.

NEWS.md
```diff
@@ -127,6 +127,7 @@ function calls. (#850, #851, @renkun-ken)
    * `consecutive_stopifnot_linter()` Require consecutive calls to `stopifnot()` to be unified into one
    * `ifelse_censor_linter()` Require usage of `pmax()` / `pmin()` where appropriate, e.g. `ifelse(x > y, x, y)` is `pmax(x, y)`
    * `system_file_linter()` Require file paths to be constructed by `system.file()` instead of calling `file.path()` directly
+   * `strings_as_factors_linter()` Check code written for R < 4.0 for code that might be affected by the new `stringsAsFactors = FALSE` default
```
AshesITR, line 130: Needs an update

AshesITR (commented):
Rest looks good. Thank you!

AshesITR (approved):

PR #1023, opened 2022-03-29
add a series of tests for help completeness

### LINKED ISSUE #916: Create tests for completeness of documentation

MichaelChirico:
Closes #916 

The tests themselves are a bit of an ugly/unreadable mess. Hopefully the liberal use of `info=` helps out future readers/devs.


tests/testthat/test-linter_tags.R
```diff
@@ -27,3 +27,69 @@ test_that("available_linters matches the set of linters available from lintr", {
...
+  # some of the following might seem redundant, so note: we want to establish
+  #   that entries exist in the help _if and only if_ they exist in the csv in inst/.
+  #   so e.g. checking that all linters in the DB match a line in the Rd is not enough,
+  #   e.g. if a deleted/renamed linter's entry still persists in the help page, and vice versa
+
+  expect_true(exists("linters", envir = help_env))
+  # objects in help_env are class Rd, see ?as.character.Rd (part of 'tools')
+  linter_help_text <- paste(as.character(help_env$linters), collapse = "")
+
+  # Rd markup for items looks like \item{\code{\link{...}} (tags: ...)}
```
AshesITR, line 51: maybe reference the relevant `rd_*` function too?

tests/testthat/test-linter_tags.R
```diff
@@ -27,3 +27,69 @@ test_that("available_linters matches the set of linters available from lintr", {
...
+  linter_item_regex <- "[\\]item[{][\\]code[{][\\]link[{][a-zA-Z0-9._]+[}][}] [(]tags: [a-z_, ]+[)][}]"
+  expect_identical(
+    length(gregexpr(linter_item_regex, linter_help_text)[[1L]]),
+    nrow(lintr_db),
+    info = "Count of linter entries in ?linters matches inst/lintr/linters.csv row count"
+  )
```
AshesITR, line 57: Both are sorted, so if we add a capture group for the linter function name, we could use `expect_equal()` which would provide us with the offending linters. WDYT?

tests/testthat/test-linter_tags.R
```diff
@@ -27,3 +27,69 @@ test_that("available_linters matches the set of linters available from lintr", {
...
+
+    tag_help <- help_env[[paste0(tag, "_linters")]]
+    tag_help_text[[tag]] <- paste(as.character(tag_help), collapse = "")
+    # Rd markup for items looks like \item{\code{\link{...}}}
+    tag_item_regex <- "[\\]item[{][\\]code[{][\\]link[{][a-zA-Z0-9._]+[}][}][}]"
+    expect_identical(
+      length(gregexpr(tag_item_regex, tag_help_text[[tag]])[[1L]]),
+      tag_table[[tag]],
+      info = paste0("Count of linter entries in ?", tag, "_linters matches tags given in inst/lintr/linters.csv")
+    )
```
AshesITR, line 75: ibid.

tests/testthat/test-linter_tags.R
```diff
@@ -27,3 +27,69 @@ test_that("available_linters matches the set of linters available from lintr", {
...
+  }
+
+  for (ii in seq_len(nrow(lintr_db))) {
+    # NB: matching tags requires reproducing the "platform-independent sort" we do; instead, rely on the other tests
+    #   for matching the tags
```
AshesITR, line 80: WDYT about sorting the tags in `available_linters()`? That would guarantee sortedness (is that even a word?) here and allow `rd_tags()` to rely on sorted tags.

tests/testthat/test-linter_tags.R
MichaelChirico: Seems a bit fragile to me since the sorting is coming from different origins.

How about capture group + `sort()` here?

tests/testthat/test-linter_tags.R
MichaelChirico: I have kind-of been writing the tags in relevance order, which makes more sense to me than alphabetical. Don't think it makes much sense to over-orient ourselves to making the test easier to write.

capture group + sort in the test would be better.

tests/testthat/test-linter_tags.R
MichaelChirico: PTAL. hopefully easier to understand with the new approach. And moreover, it already worked -- it caught that the count for `consistency_linters` is out of date :)

tests/testthat/test-linter_tags.R
```diff
@@ -27,3 +27,101 @@ test_that("available_linters matches the set of linters available from lintr", {
...
+
+  # (1) from above
+  expect_identical(
+    help_linters[order(help_linters$linter), ],
+    lintr_db[order(lintr_db$linter), ],
+    info = "Database implied by ?linters is the same as is available_linters()"
+  )
+
+  db_tag_table <- as.data.frame(
+    table(tag = unlist(lintr_db$tags)),
```
AshesITR, line 77: Curious: What will happen if we accidentally provide a tag twice in `linters.csv`?

AshesITR (approved):

PR #1025, opened 2022-03-29
dont use expect_setequal

### LINKED ISSUE #1018: Don't use expect_setequal() in linter tags test
```
lintr_db$linter[10] absent from `all_linters`
── Failure (test-linter_tags.R:26:3): available_linters matches the set of linters available from lintr ──
`all_linters`[11] absent from lintr_db$linter
```

The failure note is pretty unhelpful

MichaelChirico:
Closes #1018


tests/testthat/test-linter_tags.R
```diff
@@ -16,14 +16,15 @@ test_that("available_linters returns a data frame", {
 test_that("default_linters and default tag match up", {
   avail <- available_linters()
   tagged_default <- avail[["linter"]][vapply(avail[["tags"]], function(tags) "default" %in% tags, logical(1L))]
-  expect_setequal(tagged_default, names(default_linters))
+  expect_identical(sort(tagged_default), sort(names(default_linters)))
```
AshesITR, line 19: `tagged_default` is guaranteed to be platform-independently sorted. imo, we could require the sortin in `default_linters` as well, i.e. requiring an alphabetically sorted list of `default_linters`. This would allow dropping the `sort()` on both sides.

AshesITR: No more breaking than new default linters iinm. But yes, on its own it could be a breaking change for edge-case usage depending on the positions of linters in `default_linters` AFAIK there is no nice way to positionally override a linter, though - so probably a non-issue.

AshesITR: Looks fine to me, the sorting should be documented in `@return` I think, just to be sure.

MichaelChirico: OK, done

AshesITR (commented):
needs a document()

AshesITR (approved):

PR #1031, opened 2022-03-30
Merge paste_sep_linter and paste_to_string_linter into paste_linter

MichaelChirico:
Part of #998


R/paste_linter.R
```diff
@@ -0,0 +1,74 @@
...
+    if (length(source_file$xml_parsed_content) == 0L) {
+      return(list())
+    }
+
+    xml <- source_file$xml_parsed_content
+
+    if (allow_empty_sep) {
+      empty_sep_lints <- NULL
+    } else {
+      # NB: string-length()=2 means '' or "" (*not* 'ab' or 'cd' which have length 4)
```
AshesITR, line 28: maybe a helper `string_to_code` is in order? I think we don't really need to support `r"-----(asd)-----"`, so the code could be rather simple:
1. use `deparse(str)` to generate the correct escapes for double quotes
2. substitute \" by " and ' by \' and change the quotes to single quotes
3. use `paste0('r"(', str, ')"')` as a reasonable approximation to the raw string literal (bugged if str contains `)"`, bug fixable should we actually need it)

MichaelChirico: filed #1034, I think it makes most sense to address this globally

AshesITR (approved):

PR #1032, opened 2022-03-30
implement as plain R + regex

AshesITR:
NB merge target is the original fixed_regex PR #1021 so we can separately review the implementation of `is_not_regex` and the PR in full.


AshesITR:
Changed target to master temporarily to get GHA goodness.

AshesITR:
@MichaelChirico LMK if you would like me to adapt `.dev/compare_branches.R` to include an argument (`--base-branch`?)
Allowing a local branch from r-lib as reference should give enough flexibility I think.

MichaelChirico:
yes please do. agree about scope.

MichaelChirico:
Some patience here... my mirror has gathered some dust that's slow to remove :)

AshesITR:
No worries, thanks for the update 😊

MichaelChirico:
On a sample of 2,000 packages, there are 8 false positives where the R version says `fixed` but the C version does not:

[diff.csv](https://github.com/r-lib/lintr/files/8405308/diff.csv)

They are all the same expression:

```
gsub("\\[|\\]", "", s, perl = TRUE)
```

(also FWIW the two branches ran in almost indistinguishable time -- 13.18 (C) vs 13.49 minutes)

AshesITR:
Ah I see that must be because the special regex skips if a `[` is before it. That needs another not-escaped check before it.

Great news on the performance side!

MichaelChirico:
Assuming we can fix the issue, let's (1) merge #1021 then quickly (2) merge this to `master` as a follow-up. That way we can (1) keep the original C version in the repo history and (2) make it easier to give you credit for the great improvement!

AshesITR:
Found a fix and added it. Also added a test case for that to make sure it works.
LMK how you want to proceed.

MichaelChirico:
first I'll run again to make sure we didn't whack-a-mole any new issues.

I think we do want to squash some commits but not others which will be a pain... easier to merge the two PRs with squash

AshesITR:
Allright. I pre-approved the C implementation. Merge at your discretion; LMK if I need to take another look.

MichaelChirico:
NP. I have it running on my personal machine on all 18k+ CRAN packages so it should be done when I'm home. I'll hold off merging the other until this is ready to go.

thanks again!!

MichaelChirico:
just dawned on me why this iteration is running so slowly! (after about 10 hours running it was done ~10K of eventually 36K packages)

before I was running on the .tgz in the bin directory. I believe that means only R files in the inst directory would actually be included.

this time I'm running in the src directory so the full packages are being linted.

(mostly as a note to future self... I should probably add some instructions about using the CRAN mirror correctly)

MichaelChirico:
Another major problem:

https://github.com/r-lib/lintr/blob/49fd3468a5ec8fb14d5986809b3ca22511ab5ef7/.dev/compare_branches.R#L231

Doesn't actually clean up the temp package until _all_ packages are finished! :see_no_evil: not ideal for when running on all of CRAN. Seems to have clogged my disk and may be responsible for a large chunk of the slowdown. Unfortunately while trying to dig myself out of that hole I messed things up and the 2nd branch failed to run :cry: restarted now.

It's unfortunately a bit awkward to work around this for now. I think we need a bit of a refactor -- I also notice the order of the loops (outer: linters, inner: packages) is probably bad.

AshesITR:
It could also run in parallel for each branch (i.e. run on several packages).
Also, I always expected giving packages by name and having the script use `download.packages()` automatically to fetch what should be tested.

MichaelChirico:
> It could also run in parallel for each branch (i.e. run on several packages).

thought of that as well but not sure if it'll introduce new headaches due to current writes (I don't think it should -- shying away conservatively for now)

MichaelChirico:
Took a long time to figure it out because I thought it was the comparison script at fault.

Look at https://github.com/ibecav/CGPfunctions/blob/999a2c12831c268e94aa5921f4de662f1943fb65/R/chaid_table.R#L139-L142

The issue is that

```
eval(parse(text = "c(\"%in%\"=\"is\",\"c\\\\('\"=\"'\",\"'\\\\)\"=\"'\")"))
#  %in% c\\('  '\\) 
#  "is"   "'"   "'" 
```

has length>1, breaking the `vapply()` usage.

MichaelChirico:
Added a test to the `fixed_regex` branch

MichaelChirico:
Fixed now. Tidying up then starting another run.

MichaelChirico:
@AshesITR here's a fun one:

https://github.com/ggrothendieck/gsubfn/blob/66d52d9c0db277d39d3502d1db74911e8257fb64/R/gsubfn.R#L163

```
strsplit(z, "\1")
```

it seems we've got an `xmlparsedata::xml_parse_data()` issue here -- here's what's on the parse tree:

```
'strsplit(z, "\\1")' %>%
  parse(text = .) %>%
  xmlparsedata::xml_parse_data() %>%
  xml2::read_xml() %>%
  as.character() %>%
  writeLines()
# <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
# <exprlist>
#   <expr line1="1" col1="1" line2="1" col2="17" start="19" end="35">
#     <expr line1="1" col1="1" line2="1" col2="8" start="19" end="26">
#       <SYMBOL_FUNCTION_CALL line1="1" col1="1" line2="1" col2="8" start="19" end="26">strsplit</SYMBOL_FUNCTION_CALL>
#     </expr>
#     <OP-LEFT-PAREN line1="1" col1="9" line2="1" col2="9" start="27" end="27">(</OP-LEFT-PAREN>
#     <expr line1="1" col1="10" line2="1" col2="10" start="28" end="28">
#       <SYMBOL line1="1" col1="10" line2="1" col2="10" start="28" end="28">z</SYMBOL>
#     </expr>
#     <OP-COMMA line1="1" col1="11" line2="1" col2="11" start="29" end="29">,</OP-COMMA>
#     <expr line1="1" col1="13" line2="1" col2="16" start="31" end="34">
#       <STR_CONST line1="1" col1="13" line2="1" col2="16" start="31" end="34">"\"</STR_CONST>
#     </expr>
#     <OP-RIGHT-PAREN line1="1" col1="17" line2="1" col2="17" start="35" end="35">)</OP-RIGHT-PAREN>
#   </expr>
# </exprlist>
```

The other branch works, but I'm kind of surprised it does TBH.

AshesITR:
> There's a rare bug when `stringr::str_replace_all()` uses the multi-replacement feature. Took a long time to figure it out because I thought it was the comparison script at fault.
> 
> 
> 
> Look at https://github.com/ibecav/CGPfunctions/blob/999a2c12831c268e94aa5921f4de662f1943fb65/R/chaid_table.R#L139-L142
> 
> 
> 
> The issue is that
> 
> 
> 
> ```
> 
> eval(parse(text = "c(\"%in%\"=\"is\",\"c\\\\('\"=\"'\",\"'\\\\)\"=\"'\")"))
> 
> #  %in% c\\('  '\\) 
> 
> #  "is"   "'"   "'" 
> 
> ```

Why is that even a `STR_CONST`?
Easy enough to fix though, can add a safeguard to return `NA_character_` and silently ignore these.

> 
> 
> has length>1, breaking the `vapply()` usage.
> 
> 
> 
> I have a vague recollection of coming across this when writing the linter... we should always return no lint here because there's no way to use `fixed()` inside the name of the character vector. The `fixed_regex` branch version succeeds here.

MichaelChirico:
> Why is that even a STR_CONST?

It's the argument keyword that shows up as `STR_CONST`, e.g. `c("a" = 1L)` would also get a `STR_CONST`. This makes sense -- the logic is just tripped up because it shows up at the same parse level as the `STR_CONST` in `c("a")`. This is similar to `EQ_SUB`, and makes sense, because the LHS of `EQ_SUB` can't be an expression, so no need to add another parse level.

> Easy enough to fix though, can add a safeguard to return NA_character_ and silently ignore these.

Check the latest commit, it's fixed already (kind of... thinking again I think it needs to be more precise)

AshesITR:
> > Why is that even a STR_CONST?
> 
> It's the argument keyword that shows up as `STR_CONST`, e.g. `c("a" = 1L)` would also get a `STR_CONST`. This makes sense -- the logic is just tripped up because it shows up at the same parse level as the `STR_CONST` in `c("a")`. This is similar to `EQ_SUB`, and makes sense, because the LHS of `EQ_SUB` can't be an expression, so no need to add another parse level.

Ah, I was expecting the `expr` to contain a `SYMBOL_FUNCTION_CALL`. Maybe `*[self::STR_CONST and position() = 1]` instead of `STR_CONST` as selector condition for the argument expr is more robust?

> > Easy enough to fix though, can add a safeguard to return NA_character_ and silently ignore these.
> 
> Check the latest commit, it's fixed already (kind of... thinking again I think it needs to be more precise)

Sorry, missed that.

MichaelChirico:
@AshesITR have a fix for the issue in https://github.com/r-lib/xmlparsedata/issues/25 -- should we just port that to our side?

AshesITR:
https://stackoverflow.com/questions/19533804/is-there-a-regular-language-to-represent-regular-expressions

I just noticed that our current regex-based approach might be problematic (are `\\\\[.]` and `\[\[]` correctly linted as fixed?)
Instead of our negated regex detection, the language of fixed regexes could be directly detected.

MichaelChirico:
IINM `\[\[]` is caught by both branches, `\\\\[.]` is only caught by the C branch.

Not totally sure I go the escaping right -- code I used

```
lintr::lint(text = R'(gsub("\\\\\\\\[.]", "", x))', linters = lintr::fixed_regex_linter())
lintr::lint(text = R'(gsub("\\[\\[]", "", x))', linters = lintr::fixed_regex_linter())
```

AshesITR:
Thanks for checking. I'll think about a fix.

MichaelChirico:
barring any new crash-inducing bugs i'll also be able to share how relevant this is -- currently about 25% done a run of 4000 packages

MichaelChirico:
Both `xpath`s need some fine-tuning though, so I'll push those then run again.

MichaelChirico:
Mixed results -- some things caught by R only, some things caught by C only

Mainly it looks like on the C branch, I am too strict for the `[$CHAR]` case (e.g. `$CHAR` can be an escaped character, or a `\u`-escaped string), while the R branch is too loose (namely, the default regex allows `[]...]` to be a single character class including `]` and `...`).

### False positives in the R-only branch:

 - [`strsplit(rangeStr, "[][ ]")`](https://github.com/cran/FuzzyR/blob/9cc8853d055c08356d073a3572e0a2f24840f6b1/R/FuzzyInferenceSystem.R#L649)
 - [`gsub("[][]", "", line)`](https://github.com/cran/forrel/blob/c9fdf8fd1f41d880dfcb631829a447f6caaa2204/R/readFam.R#L441)
 - [`strsplit(colnames(obj)[-1], "[],[]")`](https://github.com/cran/ptycho/blob/a589616bbce0693c0a418f4858e037e0fd4aab65/R/proc.R#L72)
 - [`gsub("[]}]", ")", transformed)`](https://github.com/cran/renv/blob/79d9fab19ce79741d9d1a02001d2278923dbcfea/R/json-read.R#L38)
 - [`gsub("[]:]","",betainfo)`](https://github.com/cran/CpGassoc/blob/a00f76379370cd5e15ed45ad5d930baab9d0231e/R/cpg.combine.R#L53)

### False negatives in the C-only branch:

 - [`stringr::str_replace_all(lines, "[\u0451]", "\u0435")`](https://github.com/Dementiy/vkR/blob/7fa6189617191e4690acfe536a53df79d205209b/R/utils.R#L107)
 - [`gsub("\u{A0}", " ", out, useBytes = TRUE)`](https://github.com/r-lib/roxygen2/blob/03e4a644b3f63924645f866acaea4ca4b170c77a/R/rd-usage.R#L145) (one other identical hit in `roxygen2`)
 - [`grep("[]]", x)`](https://github.com/cran/ips/blob/d0ec95655c0369b9aeba3783019149b16c9e011c/R/read.nex.R#L14) (23 total hits like this or `[]]`)
 - [`gsub( '[\\]', '/', dirname( chname))`](https://github.com/cran/mvbutils/blob/0de12969abaf828b5106290951318c79f0741a18/R/mvbutils.R#L4788) (10 total hits like this)
 - [`gsub("[\r]", "", config_char)`](https://github.com/cran/Rsagacmd/blob/badbb5355f360a66c024ab1caf4b7abcb36d48af/tests/testthat/test-saga-environment.R#L71) (5 total hits like this or `[\n]`)
 - [`strsplit(ls[1],"[\"]")`](https://github.com/cran/bnstruct/blob/b9c8305c7f4c6e4074ddffce64665480411e9347/R/io-methods.R#L22) (2 total hits like this)

MichaelChirico:
OK just pushed some fixes on the C side for the false negatives. Not the prettiest code but we can touch it up if we abandon the regex approach.

MichaelChirico:
FWIW I think a rough regex equivalent for char classes that I just handled on the C side would be:

```
\[.\]
|
\[[\].\]
|
\[[\]x[0-9a-fA-F]{1,2}\]
|
\[[\][0-7]{1,3}\]
|
\[[\]u[0-9a-fA-F]{1,4}\]
|
\[[\]u\{[0-9a-fA-F]{1,4}\}\]
|
\[[\]U[0-9a-fA-F]{1,8}\]
|
\[[\]U\{[0-9a-fA-F]{1,8}\}\]
```

AshesITR:
@MichaelChirico I got the tests to pass by double escaping the inputs. Can you check what the intention was?
The way the original tests were written, the actual regex simply got unescaped UTF-8 characters directly; was that intentional?
If it was, I don't know why GHA builds kept failing on `[\xA7]` whereas my local system passed all tests.

MichaelChirico:
> was that intentional?

no :) writing tests through several layers of escapes is hard

MichaelChirico:
OK I'll try and get a sample run going tomorrow for one last comparison

MichaelChirico:
Added a bunch of these as test cases... looks mostly related to handling of `\n` / `\r`

AshesITR:
Looks like we need the global option too.

AshesITR:
@MichaelChirico can you double-check the tests?
After setting `(?s)` instead of `(?m)` as the regex option, all tests pass except for ones I think are actually wrong. I'll wait with the push until after you've confirmed the tests need fixing.

MichaelChirico:
> That's a false negative on the C side it seems.

true, agreed

tests/testthat/test-fixed_regex_linter.R
```diff
@@ -150,38 +150,38 @@ test_that("1- or 2-width octal escape sequences are handled", {
   linter <- fixed_regex_linter()
   msg <- rex::rex("For static regular expression patterns, set `fixed = TRUE`.")
 
-  expect_lint('strsplit(x, "\\1")', msg, linter)
+  expect_lint('strsplit(x, "\\\\1")', msg, linter)
```
MichaelChirico, line 153: sorry I missed this -- I don't think this is right. What we want is

```
strsplit(x, "\1")
```

i.e., the string literal for the octal escape 1:

```
identical("\1", rawToChar(as.raw(1)))
# [1] TRUE
```

This adjusted test is actually a regex (well, not really for `strsplit()` _per se_), being the group reference:

```
grepl("(abc)def\\1", "abcdefabc")
```

IINM the `()` will always make sure it gets recognized as regex.

The version here is actually an error

```
strsplit(letters, "\\1")
# Error in strsplit("a\\1b", "\\1") : 
#   invalid regular expression '\1', reason 'Invalid back reference'
```

tests/testthat/test-fixed_regex_linter.R
```diff
@@ -150,38 +150,38 @@ test_that("1- or 2-width octal escape sequences are handled", {
...
-  expect_lint("str_split(x, '[\\uF]')", msg, linter)
-  expect_lint("str_split(x, '[\\u01]')", msg, linter)
-  expect_lint("str_split(x, '[\\u012]')", msg, linter)
-  expect_lint("str_split(x, '[\\u0123]')", msg, linter)
-  expect_lint("str_split(x, '[\\U8]')", msg, linter)
-  expect_lint("str_split(x, '[\\U1d4d7]')", msg, linter)
-  expect_lint("str_split(x, '[\\u{1}]')", msg, linter)
-  expect_lint("str_split(x, '[\\U{F7D5}]')", msg, linter)
-  expect_lint("str_split(x, '[\\U{1D4D7}]')", msg, linter)
+  expect_lint("gsub('[\\\\n]', '', x)", msg, linter)
```
MichaelChirico, line 160: ditto for all of these

tests/testthat/test-fixed_regex_linter.R
```diff
@@ -150,38 +150,38 @@ test_that("1- or 2-width octal escape sequences are handled", {
...
 })
 
 test_that("bracketed unicode escapes are caught", {
   linter <- fixed_regex_linter()
   msg <- rex::rex("For static regular expression patterns, set `fixed = TRUE`.")
 
-  expect_lint('gsub("\\u{A0}", " ", out, useBytes = TRUE)', msg, linter)
-  expect_lint('gsub("abc\\U{A0DEF}ghi", " ", out, useBytes = TRUE)', msg, linter)
-  expect_lint('gsub("\\u{A0}\\U{0001d4d7}", " ", out, useBytes = TRUE)', msg, linter)
+  expect_lint('gsub("\\\\u{A0}", " ", out, useBytes = TRUE)', msg, linter)
```
MichaelChirico, line 182: ditto

tests/testthat/test-fixed_regex_linter.R
```diff
@@ -5,336 +5,194 @@
...
 })
 
 test_that("escaped characters are handled correctly", {
-  expect_lint("gsub('\\n+', '', sql)", NULL, fixed_regex_linter())
-  expect_lint('gsub("\\n{2,}", "\n", D)', NULL, fixed_regex_linter())
-  expect_lint('gsub("[\\r\\n]", "", x)', NULL, fixed_regex_linter())
-  expect_lint('gsub("\\n $", "", y)', NULL, fixed_regex_linter())
-  expect_lint('gsub("```\\n*```r*\\n*", "", x)', NULL, fixed_regex_linter())
-  expect_lint('strsplit(x, "(;|\n)")', NULL, fixed_regex_linter())
-  expect_lint('str_count(a$x$html, "\\\\<li\\\\>")', NULL, fixed_regex_linter())
```
MichaelChirico: rather than remove this, we should add it to the true positive list

MichaelChirico:
>  The R string is "\\<li\\>" so the regex is \<li\> which is equivalent to <li> with fixed = TRUE.

Actually, sorry, I was ~kind of wrong. `\<` and `\>` are actual special characters in regex.

Personally I've _never_ used them before, but I made the note to myself in the C comments:

https://github.com/r-lib/lintr/pull/1021/files#diff-7b254c143caf4ac4b411f716e8cea7be6168bd5259b69506361cf3d17571c0dbR133

From [`?regex`](https://rdrr.io/r/base/regex.html):

> The symbols `\<` and `\>` match the empty string at the beginning and end of a word.

But also

> All the regular expressions described for extended regular expressions are accepted [under `perl=TRUE`] except `\<` and `\>`: in Perl all backslashed metacharacters are alphanumeric and backslashed symbols always are interpreted as a literal character.

And indeed:

```
grepl("\\<li\\>", " li ")
# [1] TRUE
grepl("\\<li\\>", " li ", perl=TRUE)
# [1] FALSE
```

So... we're kind of both right, both wrong here. The original package author's usage is correct, because `str_count` :arrow_right: `perl` REs.

I think the only reliable fix is to detect whether `perl=TRUE` should be applied...

which I think is _well_ beyond the scope of this PR.

For now, I'd rather err on the side of being conservative and always assume `\<` is a regex (i.e, what the C branch currently does) -- better to have false negatives than false positives. We can revisit later if there's user demand.

tests/testthat/test-fixed_regex_linter.R
MichaelChirico: NVM, let's leave this untested, per comment below: https://github.com/r-lib/lintr/pull/1032#issuecomment-1112794556

tests/testthat/test-fixed_regex_linter.R
AshesITR: Can we then use printable examples?
e.g. `[\xA7]` fails because `\xA7` dosen't match `.`

Also, the escaped versions should be added for some, e.g. `\u` escapes are also valid in regex.

Otherwise I don't know how to match "literally any unicode codepoint that exists".

MichaelChirico: maybe it got lost in the comment stack, but here's what to match:

https://github.com/r-lib/lintr/pull/1032#issuecomment-1094218015

```
\[.\]
|
\[[\].\]
|
\[[\]x[0-9a-fA-F]{1,2}\]
|
\[[\][0-7]{1,3}\]
|
\[[\]u[0-9a-fA-F]{1,4}\]
|
\[[\]u\{[0-9a-fA-F]{1,4}\}\]
|
\[[\]U[0-9a-fA-F]{1,8}\]
|
\[[\]U\{[0-9a-fA-F]{1,8}\}\]
```

AshesITR: That's what I do match, but _after_ R escaping, i.e. as literal regex characters.

AshesITR: Exactly. `as.character(parse())` to be precise. I'm at least aware of `\uXXXX` being a valid escape for regex as well, so some of the `\\\\` variants should lint too. It seems there are currently too many allowed escapes (since `\1` is a backreference and not an octal escape in regex).

MichaelChirico: > I'm at least aware of \uXXXX being a valid escape for regex as well

Could you give an example? In `gsub("\u0001", "", x)`, the regex engine only sees the byte value -- it has no idea `\u` was ever written since that is handled by the R parser.

MichaelChirico: Compare

```
gsub(".*([abc]).*", "\1", "0123abcdef")
# "\001"
gsub(".*([abc]).*", "\\1", "0123abcdef")
# "c"
```

AshesITR:
https://www.regular-expressions.info/refcharacters.html

Found a reference explaining some R regex escapes. Looks like we need `\o` and `\c`, and change the octal escape regex to `\\\d{2,3}|\\0\d{1,3}`.

tests/testthat/test-fixed_regex_linter.R
AshesITR: Hm the docs seem not to be correct regarding some escapes, but e.g. hex escapes do work:

```
> gsub("\\x80", "", "\x80")
[1] ""
```

AshesITR:
@MichaelChirico I found the problem with `\xA7` or any non-utf8 string.
How do we want to deal with that?

```
> tst <- `Encoding<-`("\xA7", "UTF-8")
> grepl(".", tst)
[1] TRUE
> grepl(".", tst, perl = TRUE)
Warning in grepl(".", tst, perl = TRUE) :
  input string 1 is invalid UTF-8
[1] FALSE
```

AshesITR:
Behaviour regarding non-printable characters and regex seems highly platform-dependent: build failures / successes depend on whether `grepl(".", "\xA7") || grepl(".", "\xA7", perl = TRUE)` returns TRUE or not.
I suggest we change our tests to exclude invalid UTF-8 sequences, such as `\xA7`. WDYT, @MichaelChirico?

MichaelChirico:
> I suggest we change our tests to exclude invalid UTF-8 sequences, such as \xA7. WDYT, @MichaelChirico?

SGTM... I was only hoping to test behavior for `\x` sequences... let's switch to `\x32` == `'2'`?

MichaelChirico:
> https://www.regular-expressions.info/refcharacters.html

nice find!

> Looks like we need \o and \c, and change the octal escape regex to \\\d{2,3}|\\0\d{1,3}.

i'm happy to punt on `\o` and `\c` for now :)

AshesITR:
During some manual testing, I noticed that at least the most recent R versions don't have all of the features mentioned at the link I provided.

AshesITR:
@MichaelChirico All tests succeed on all platforms now 🥳

MichaelChirico:
awesome!! thanks again for your patience/diligence here!

I'll start another run tomorrow evening and hopefully we can (finally) merge

AshesITR:
Sounds good, thanks a lot for the extensive testing and feedback!

MichaelChirico:
Some issues... getting warnings running the linter on some packages, e.g. `g3viz`:

```
lintr::lint_package(linters=lintr::fixed_regex_linter())
.................
Warning messages:
1: In grepl(paste0("(?s)", rx_static_regex), str, perl = TRUE) || grepl(rx_static_regex,  :
  'length(x) = 2 > 1' in coercion to 'logical(1)'
2: In grepl(paste0("(?s)", rx_static_regex), str, perl = TRUE) || grepl(rx_static_regex,  :
  'length(x) = 2 > 1' in coercion to 'logical(1)'
```

(IINM that'll be an error in r-devel)

Packages that don't seem to want to lint on this branch:

```
addinsOutline, alignfigR, anyLib, ASIP, ast2ast, BaBooN, baseflow, bbw, Biocomb, caretEnsemble, cgam, cgmanalysis, compareODM, copBasic, crossword.r, DistPlotter, dscore, ecm, envDocument, ergmito, errorlocate, exp2flux, fail, fcr, g3viz, geno2proteo, GFE, ggpattern, ggpol, gllm, gmpoly, gsignal, huito, japanstat, JuliaCall, knotR, loo, lsa, marelac, mldr, MM4LMM, mma, mmaqshiny, mmm2, MRFcov, mwcsr, nat.utils, paletteer, plumberDeploy, polite, pxR, r2symbols, rcbayes, RcmdrPlugin.DCE, readmoRe, riskclustr, RLogicalOps, roloc, Rsgf, RWmisc, SCINA, semnova, shinybrms, spex, stpm, SuperLearner, tci, terrainr, thinkr, tidyvpc, TiPS, waterfalls, whereami
```

MichaelChirico:
Also still getting some false positives here on the R side IINM:

```
                                                                                                                line
   1:                 sites=sort(sample(c(0, gregexpr("[KR]", protein, perl=T)[[1]], nchar(protein)), 2, replace=F))
   2:                 sites=sort(sample(c(0, gregexpr("[KR]", protein, perl=T)[[1]], nchar(protein)), 2, replace=F))
   3:                                                             if (TRUE %in% grepl("(Intercept)", names(data))) {
   4: old_userpwd <- stringr::str_extract(environ_lines[userpwd_line_index], "(?<=AURIN_API_USERPWD=)\\\\w+:\\\\w+")
   5:                                                            url = gsub("https://www.bitmex.com", "", prep_url),
  ---                                                                                                               
2803:                               if(length(grep("delta.mean",names(PPI)))!=0) Ydelta.meanM <- log10(Ydelta.meanM)
2804:                                                                  if(length(grep("delta.mean",names(PPI)))!=0){
2805:                                       y <- gsub('\\\\^([0-9])+','^\\\\1.',y) # canonical, all pos num follow ^
2806:            y <- gsub('([a-zA-Z])([-0-9]+)', '\\\\1^\\\\2.',y) # non-canonical, unsigned or neg num follow char
2807:                                                                       y <- if (regexpr("%[qQ]", format) > 0) {
```

haven't pored through everything but definitely see some that shouldn't be linting

MichaelChirico:
I think that may be throwing things off for the rest of the lints. The current results have a ton of false positives that I don't reproduce if I try and run the expressions individually.

MichaelChirico:
simple enough fix :sweat_smile:

AshesITR:
I was somehow convinced that str was a single string and not a vector 😅

MichaelChirico:
OK another report. Ran on 1300 unique packages. Still a few cases where C & R disagree.

 1. `str_replace_all()` false positives. This is a known false positive when `stringr` functions are used in `magrittr` pipelines. examples:

```
  Package: threeBrain, file: R/class_brainatlas.R
self$atlas_type <- stringr::str_replace_all(atlas_type, '[\\W]', '_')

  Package: threeBrain, file: R/fs_brain2.R
atlas_t_alt <- stringr::str_replace_all(atlas_t, '[\\W]', '_')

  Package: campfin, file: R/normal-address.R
stringr::str_replace_all("^([:digit:]+)([:alpha:]+)", "\\1 \\2") %>%

  Package: campfin, file: R/normal-address.R
stringr::str_replace_all("([:alpha:]+)([:digit:]+)$", "\\1 \\2")

  Package: doseminer, file: R/extract.R
str_replace_all('([0-9]+)([x/])([a-z]+)', '\\1 \\2 \\3') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('([a-z]+)([0-9]+)', '\\1 \\2') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('([0-9]+)([a-z]+)', '\\1 \\2') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('([0-9]+) ?(-|(?:up )?to|or) ?([0-9]+)', '\\1 - \\3') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('(\\bq) ([1-8]) ([dh])', '\\1\\2\\3') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('(\\w+)(bd|[qt]ds)\\b', '\\1 \\2') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('(?<!take )([0-9]+) (?:times daily|a day)', '\\1 / day') %>%

  Package: doseminer, file: R/extract.R
str_replace_all('(\\d+[.]?\\d*) (\\d+[.]?\\d* ml spoon)', '\\1 x \\2') %>%

  Package: torchaudio, file: R/temp.R
stringr::str_replace_all("(:[^,=\n]+)(,|( =)|\n)", "\\2") %>%

  Package: torchaudio, file: R/temp.R
stringr::str_replace_all("\n(.+)(\\-=)", "\n\\1 = \\1 -") %>%

  Package: torchaudio, file: R/temp.R
stringr::str_replace_all("\n(.+)(\\/=)", "\n\\1 = \\1 /") %>%
```

This looks like the C side is recognizing the `\\1` substitutions as being regex-y and skipping while R side is not.

 2. others:

   + Package: pander, file: R/pandoc.R **R is right, false negative for C, because duplicates in char class don't matter**
    ```r
    x <- gsub('[\\\\]', '', x) # backslashes
    ```
   + Package: SGP, file: R/courseProgressionSGP.R **C is right, false positive for R, this uses `\<` special**
    ```r
    setattr(sgp_object_subset[["GRADE_CHAR"]], "levels", gsub("\\<CT\\>", "EOCT", levels(sgp_object_subset[["GRADE_CHAR"]])))
    ```
   + Package: stpm, file: R/spm_time-dependent.R **C is right, false positive for R, _because `perl=FALSE`_**
    ```r
    p.temp.coeff <- trim(unlist(strsplit(p.temp[i],"[\\*]",fixed=F)))
    ```
   + Package: FastRWeb, file: R/parse-multipart.R. **R is right, false negative for C, _because `perl=TRUE`_**
    ```r
    data$filename <- strsplit(filename,'[\\/]',perl=TRUE)[[1L]]
    ```

The ones that hing on the value of `perl` I think we can punt on for now. So I believe we should fix the `\<` case and we can be done. `\1` can't show up in the `pattern` argument of a regex without also using `()`, so I think we can safely ignore it. Better to focus effort on improving the XPath to skip the pipeline matches in the first place.

AshesITR:
> self$atlas_type <- stringr::str_replace_all(atlas_type, '[\\W]', '_')

That's not a FP related to pipes, no?
It is a false positive, though, since that is a regex as evident by this experiment:

```
> stringr::str_replace_all("\\W-", '[\\W]', '_')
[1] "_W_"
> gsub("[\\W]", "_", "\\W-")
[1] "__-"
```

`\<` seems to be fixed iff `perl = TRUE`, so that's also dependent on the value of `perl`

```
> gsub("\\<", "#", "\\<a<s<d")
[1] "\\<#a<#s<#d"
> gsub("\\<", "#", "\\<a<s<d", perl = TRUE)
[1] "\\#a#s#d"
```

So I'd suggest we assume `perl = TRUE` for the moment, i.e. let `\<` lint but make sure `[\W]` doesn't.
That way we'd at least be consistent in the flavor we try to detect and could at a later point extend to `perl = FALSE`, which affects the base regex functions. WDYT?

The necessary fix would be to disallow some characters after `\\` within `[`, namely the character class shorthands.

MichaelChirico:
thanks!! starting the merge 🚀🚀🚀🚀

MichaelChirico (approved):

MichaelChirico:
thanks again!!

AshesITR:
Thank you too!

AshesITR:
I think converting the regex to a whitelist of to be linted regexes should eliminate the false positives.

PR #1033, opened 2022-03-30
extend compare_branches to compare two branches

AshesITR:
Also did a few ergonomic improvements:

 - Don't litter console output with package startup messages. They make `--help` very hard to read.
 - Removed the historical default linter (object_usage_linter) because it makes to sense.
 - Added `--base-branch=master` option to modify the reference branch.
 - Stopped littering `~` with result files, instead write to the `.dev` directory in lintr and `.gitignore` the generated csv files.
   This makes working with the results from an IDE much easier.

Could still use a safeguard against Ctrl+C leaving interactive sessions in the `tempdir()` with the need to manually `unlink()` because subsequent runs of the script in the same session fail due to `Rtmp.../lintr_repo` being non-empty.

### DIFF
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -123,6 +123,9 @@ if (interactive()) {
 linter_names <- strsplit(params$linters, ",", fixed = TRUE)[[1L]]
 
 base_branch <- params$base_branch
+if (is.null(base_branch) || is.na(base_branch) || !nzchar(base_branch)) {
+  stop("Please supply a base branch (--base-branch)")
+}
 
 # prioritize "branch"
 is_branch <- FALSE
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -121,6 +121,9 @@ if (interactive()) {
 }
 
 linter_names <- strsplit(params$linters, ",", fixed = TRUE)[[1L]]
+if (length(linter_names) == 0L) {
+  stop("Please supply linters (--linters)")
+}
 
 base_branch <- params$base_branch
 if (is.null(base_branch) || is.na(base_branch) || !nzchar(base_branch)) {
--- a/.dev/compare_branches.R
+++ b/.dev/compare_branches.R
@@ -73,7 +73,7 @@ param_list <- list(
     help = "Run the comparison for base vs. this PR"
   ),
   optparse::make_option(
-    "--base-branch",
+    "--base_branch",
     default = if (interactive()) {
       readline("Name a branch to use as base (skip to use master): ")
     } else {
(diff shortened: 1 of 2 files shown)


AshesITR:
@MichaelChirico I don't have a local CRAN mirror at my disposal, so can't properly run this atm. PTAL and see if it works.

.dev/compare_branches.R
```diff
@@ -49,24 +51,35 @@ if (!interactive()) {
...
+    default = if (interactive()) {
+      readline("Provide a comma-separated list of linters to compare: ")
+    },
+    help = "Run the comparison for these linter(s) (comma-separated)"
   ),
   optparse::make_option(
     "--branch",
     default = if (interactive()) {
-      readline("Name a branch to compare to master (or skip to enter a PR#): ")
+      readline("Name a branch to compare to the base branch (or skip to enter a PR#): ")
```
MichaelChirico, line 62: maybe default to assignment_linter in batch mode?

MichaelChirico: or maybe just error if not provided in batch mode? is that what happens when default=NULL?

.dev/compare_branches.R
```diff
@@ -104,10 +117,13 @@ if (interactive()) {
     # typed arguments get cast even when missing, probably to NA
     if (isTRUE(is.na(params[[opt]]) || params[[opt]] == "")) params[[opt]] <- NULL
   }
+  if (isTRUE(is.na(params$base_branch) || params$base_branch == "")) params$base_branch <- "master"
```
MichaelChirico, line 120: use | here? are we sure length 1?

MichaelChirico:
thanks for this! basically looks fine. small tweaks

.dev/compare_branches.R
AshesITR: Comes from `optparse`. I'd be surprised to see anything but a length-1 character vector.

.dev/compare_branches.R
AshesITR: Crude error messages I think.
`stop()` is probably the best thing to do.

.dev/compare_branches.R
AshesITR: NB we lint `if (a | b)` 😉

MichaelChirico: not if (isTRUE(a | b)) though 😎

AshesITR: Anyway. I don't see any optparse documentation that hints at the ability to parse more than length-1 arguments, and we know our argument list only contains length-1 arguments. I reckon even the `isTRUE()` isn't strictly necessary.

MichaelChirico: ok anyway this is a dev script... doesn't need to be perfectly robust.

MichaelChirico (approved):

PR #1038, opened 2022-04-01
Fix lints

MichaelChirico:
Now that we're basically done adding new linters, I took a pass at examining all lints at HEAD; fixed the main ones.


tests/testthat/test-missing_argument_linter.R
```diff
@@ -72,11 +72,11 @@ test_that("returns the correct linting", {
 
   expect_lint("switch(a =, b = 1, 0)",
     list(message = rex("Missing argument in function call.")),
-    missing_argument_linter(c()))
+    missing_argument_linter(NULL))
 
   expect_lint("alist(a =)",
     list(message = rex("Missing argument in function call.")),
-    missing_argument_linter(c()))
+    missing_argument_linter(NULL))
```
AshesITR, line 80: That should be a false positive, right? `alist` is special in the sense that it's used explicitly to craft argument lists with or without defaults.

tests/testthat/test-missing_argument_linter.R
```diff
@@ -72,11 +72,11 @@ test_that("returns the correct linting", {
 
   expect_lint("switch(a =, b = 1, 0)",
     list(message = rex("Missing argument in function call.")),
-    missing_argument_linter(c()))
+    missing_argument_linter(NULL))
```
AshesITR, line 75: These should be `character()`

AshesITR (approved):

AshesITR:
Thanks!

PR #1039, opened 2022-04-02
add style guide links for default linters

### LINKED ISSUE #911: Man pages for linters implementing tidyverse style rules should link to the section in the guide
As a follow-up to https://github.com/r-lib/lintr/pull/908

https://github.com/AshesITR/lintr/pull/1/files#diff-ae410c8bea944744ba98dad960334f7fa8d5303ae2ea4f2d6ebcdc0d8063f765

makes me think that we should link the style guide wherever relevant. Much simpler now that each linter will have its own page (before it might have felt a bit cluttered to do so).

also not a huge priority so not fixing it in that PR chain.

MichaelChirico:
Closes #911

Any idea how fragile these links are? I guess the CRAN release checklist does URL checking, so we should be able to update stale links at release?

Is it better to add the links in the details (as done here), or to put the link in `@seealso`?

### DIFF
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -3,7 +3,9 @@
 #' Avoid the symbols `T` and `F` (for `TRUE` and `FALSE`).
 #'
 #' @evalRd rd_tags("T_and_F_symbol_linter")
-#' @seealso [linters] for a complete list of linters available in lintr.
+#' @seealso
+#'   [linters] for a complete list of linters available in lintr. \cr
+#'   <https://style.tidyverse.org/syntax.html#logical-vectors>
 #' @export
 T_and_F_symbol_linter <- function() { # nolint: object_name_linter.
   Linter(function(source_file) {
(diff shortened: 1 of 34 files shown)


AshesITR:
`@seealso` seems like the intended place for external references: https://r-pkgs.org/man.html#roxygen-comments

MichaelChirico:
> Also: The URL checker can't detect broken anchors IINM. Do we want to test for these? Should be relatively easy to do with XPath and given that we already test the help DB.

Not sure I follow... doesn't this require a web connection? I would avoid adding tests that require web connection unless strictly necessary.

To me a FR for the URL checking package seems better.

AshesITR:
Okay. Not sure if a FR of this size for the base R package tools is likely to be considered, though:
 https://github.com/wch/r-source/blob/c17766c2d89d9e5724f5353b37415c40971beb77/src/library/tools/R/urltools.R#L392-L637

That likely can't use `xml2` to do the heavy lifting, so would be much more work.
Let's hope link rot won't be a big problem :)

AshesITR (approved):

MichaelChirico:
> Not sure if a FR of this size for the base R package tools

I meant for https://github.com/r-lib/urlchecker to add the option

AshesITR:
Filed https://github.com/r-lib/urlchecker/issues/22.

PR #1049, opened 2022-04-04
warn if some NULLs are not in default

AshesITR:


R/with.R
```diff
@@ -57,6 +57,16 @@ with_defaults <- function(..., default = default_linters) {
     )
   }
 
+  to_null <- vapply(vals, is.null, logical(1L))
+  if (!all(nms[to_null] %in% names(default))) {
+    bad_nms <- setdiff(nms[to_null], names(default))
+    is_are <- if (length(bad_nms) > 1L) "are" else "is"
+    warning(
+      "Trying to remove ", glue::glue_collapse(sQuote(bad_nms), sep = ", ", last = " and "),
+      ", which ", is_are, " not in `default`."
```
MichaelChirico, line 66: does ?with_defaults describe `default`? otherwise this message could be confusing

MichaelChirico (approved):
Approving in case there's nothing to add to the warning to help orient the user what 'default' is

R/with.R
AshesITR: It has to for R CMD check to succeed. Here's the roxy:

`#' @param default list of elements to modify.`

MichaelChirico: 👍 silly misunderstanding on my part, thanks

Issue #1034, opened 2022-03-30
Ensure linter logic allows for R>=4 raw strings


MichaelChirico:
Some linters (e.g. current `paste_sep_linter` and `paste_to_string_linter`) rely on assumptions about `STR_CONST` nodes that are designed with R<4 in mind, where `R"(a string)"` is not available.

We should take a pass over linters to ensure robustness to this new style of literal strings.

AshesITR:
Note: `as.character(parse(text = 'r"(a)"', keep.source = FALSE)) == "a"` seems like a base R way to extract the string without `eval()`.

MichaelChirico:
Here's all the linters that might need to be addressed:

```
grep -Flr "STR_CONST" R | awk -F/ '{print $2}'

object_usage_linter.R
extraction_operator_linter.R
expect_type_linter.R
literal_coercion_linter.R
strings_as_factors_linter.R
condition_message_linter.R
path_linters.R
object_name_linters.R
single_quotes_linter.R
paste_linter.R
expect_s3_class_linter.R
get_source_expressions.R
expect_named_linter.R
make_linter_from_regex.R
unneeded_concatenation_linter.R
missing_package_linter.R
sprintf_linter.R
yoda_test_linter.R
```

MichaelChirico:
One more TODO here: I think `unquote()` can be entirely replaced by the new `as.character(parse())` approach.

PR #1056, opened 2022-04-09
fix-up R parser bug in get_source_expressions

MichaelChirico:
Split off from #1032 because in principle it could affect the logic whenever `STR_CONST` is involved (especially, it's probably needed as part of #1034)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -143,6 +143,7 @@ function calls. (#850, #851, @renkun-ken)
 * `object_name_linter()` now correctly detects assignment generics (#843, @jonkeane)
 * `trailing_whitespace_linter()` now also lints completely blank lines by default. This can be disabled by setting the
   new argument `allow_empty_lines = TRUE` (#1044, @AshesITR)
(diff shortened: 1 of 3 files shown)


AshesITR (approved):

MichaelChirico:
Adding a NEWS bullet and tests...

AshesITR (approved):

AshesITR:
Will this fix the xml as well? I suppose no, just wanted to confirm.

MichaelChirico:
Yes it will, because `xmlparsedata` builds from the parsed data.frame if we pass one.

Here we pass the data.frame:

https://github.com/r-lib/lintr/blob/49fd3468a5ec8fb14d5986809b3ca22511ab5ef7/R/utils.R#L99

Here `xmlparsedata` builds off the data.frame it receives instead of running `utils::getParseData()` again:

https://github.com/r-lib/xmlparsedata/blob/ee6a15f0b9162e2d215635d9771dc5c09d3eacbf/R/package.R#L65

(this is why the other touch-ups, i.e. for tabs and equals signs, are also reflected on the XML tree we get back from `xml_parse_data()`)

PR #1057, opened 2022-04-10
compare_branches update: refactor loop order

MichaelChirico:
Bunch of touch-ups here; the main thing is to re-order the execution so that we loop over packages, then linters. This also facilitates deleting packages after each iteration, instead of accumulating them by the 1000s which was done before.

Also, instead of accumulating lints in one giant list, each package is written to a disk first. This makes things easier in the case that the linter crashes mid-way (was happening too much to me, and if I want to run all linters on all packages, e.g. before release, this will almost surely happen again).

Finally, edited the `utils::untar()` invocation to only extract the files that `lintr::lint_package()` looks at, which should make it a fair amount faster (e.g. it skips man/)


.dev/compare_branches.R
```diff
@@ -15,14 +17,19 @@
 #     ./dev/compare_branches --branch=master --outfile=old.csv ...
 #   And then compare the results found in new.csv & old.csv.
 
+# TODO
+#  - make sure this works for comparing tags to facilitate release testing
+#  - handle the case when working directory is not the lintr directory
+
 suppressPackageStartupMessages({
   library(optparse)
+  library(data.table, include.only = "fwrite")
```
AshesITR, line 26: I didn't know about this, nice to know 👍🏻

.dev/compare_branches.R
```diff
@@ -207,164 +223,127 @@ get_deps <- function(pkg) {
...
-    ii <- ii + 1L
+lint_one_package <- function(package, linters, out_dir, check_deps) {
+  package_is_dir <- file.info(package)$isdir
+  package_name <- gsub("_.*", "", basename(package))
+
+  if (!package_is_dir) {
+    tmp <- file.path(tempdir(), package_name)
+    # only extract files that lintr::lint_package() cares about
+    package_files <- utils::untar(package, list = TRUE)
+    lint_files <- grep(file.path(package_name, "(R|tests|inst|vignettes|data-raw|demo)"), package_files, value = TRUE)
```
AshesITR, line 234: Shouldn't this regex be anchored just in case?

.dev/compare_branches.R
```diff
@@ -207,164 +223,127 @@ get_deps <- function(pkg) {
...
+    # exclude directories because untar() gets confused when extracting path/to and then path/to/file
+    lint_files <- lint_files[!endsWith(lint_files, "/")]
+    # --strip-components makes sure the output structure is
+    # /path/to/tmp/pkg/ instead of /path/to/tmp/pkg/pkg
+    utils::untar(package, exdir = tmp, extras = "--strip-components=1")
```
AshesITR, line 239: Is this argument portable? Otherwise, maybe better to accept the sub-optimal package structure.

.dev/compare_branches.R
AshesITR: OTOH: `fwrite()` is always namespaced below.
WDYT about `loadNamespace("data.table")` instead?
Will keep the search path clean while also erroring if the dependency is missing.

.dev/compare_branches.R
MichaelChirico: My read of the man page suggests it should be.

More generally I am a bit loath to build too much robustness here if there's not a need for it from package devs since it's not part of the "prod" package.

.dev/compare_branches.R
MichaelChirico: Hmm I actually don't think this is used. I guess I never solved how to extract only certain files here. Switched to a TODO instead.

.dev/compare_branches.R
MichaelChirico: I think `library()` is simpler for a dev script like this -- the search path concern is only relevant for interactive use which should only be used for debugging really. This script is designed for running in batch mode (hence the `#!/usr/local/bin/Rscript` at the top)

.dev/compare_branches.R
AshesITR: Okay. Though this should at least in principle be able to run on all OSes :)
We'll see if it does when release time comes.
WDYT about adding a release bullet?

MichaelChirico: makes sense -- though I'm pretty confident it will work everywhere. ok windows the right untar should be bundled with Rtools, which is  a requirement for package dev anyway

AshesITR (approved):

PR #1061, opened 2022-04-12
Rename semicolon_terminator_linter

### LINKED ISSUE #1040: Refactor semicolon_terminator_linter for consistency?
`semicolon_terminator_linter` strikes me as a bit "off" w.r.t. the style elsewhere in the package

 1. The name `semicolon_linter` seems better to me
 2. The signature `function(semicolon = c("compound", "trailing"))` looks odd. I think we are aligning more on toggling these types of things with `allow*`-style arguments, e.g. `function(allow_compound = FALSE, allow_trailing = FALSE)` would be equivalent to the current signature.

Are either of these worth doing?

Of course, we can use a deprecation cycle to make the changes more gradually.

MichaelChirico:
Closes #1040


AshesITR:
Do we want to test that a deprecation warning is issued?

MichaelChirico:
Good idea... caught a bug.

Also changed `with_defaults()` not to use default `sQuote()` since it's hard to test against on fancy quote machines, and `sQuote(., "'")` isn't available to all of our R dep versions.

tests/testthat/test-semicolon_linter.R
```diff
@@ -69,9 +69,46 @@ test_that("Compound semicolons only", {
...
+
+test_that("deprecation notices for semicolon_terminator_linter succeed, and the deprecated version works", {
+  expect_warning(
+    linter <- semicolon_terminator_linter(),
+    "Function semicolon_terminator_linter was deprecated",
+    fixed = TRUE
+  )
+  expect_lint("a <- 1", NULL, linter)
+  expect_lint("a <- 1;", rex::rex("Trailing semicolons are not needed."), linter)
+  expect_lint("a <- 1; b <- 2", rex::rex("Compound semicolons are not needed."), linter)
```
AshesITR, line 87: The lint message seems a bit off. Not needed -> discouraged?
They aren't redundant after all.

tests/testthat/test-semicolon_linter.R
```diff
@@ -69,9 +69,46 @@ test_that("Compound semicolons only", {
...
+    fixed = TRUE
+  )
+  expect_lint("a <- 1", NULL, linter)
+  expect_lint("a <- 1;", rex::rex("Trailing semicolons are not needed."), linter)
+  expect_lint("a <- 1; b <- 2", NULL, linter)
+
+  # with_defaults warns about now-absent semicolon_terminator_linter
+  expect_warning(
+    d <- with_defaults(semicolon_terminator_linter = NULL),
+    "Trying to remove 'semicolon_terminator_linter', which is not in `default`\\."
```
AshesITR, line 111: Why not `rex::rex()` here?

R/with.R
```diff
@@ -62,7 +62,7 @@ with_defaults <- function(..., default = default_linters) {
     bad_nms <- setdiff(nms[to_null], names(default))
     is_are <- if (length(bad_nms) > 1L) "are" else "is"
     warning(
-      "Trying to remove ", glue::glue_collapse(sQuote(bad_nms), sep = ", ", last = " and "),
+      "Trying to remove ", glue::glue_collapse(quote_wrap(bad_nms, "'"), sep = ", ", last = " and "),
```
AshesITR, line 65: I deliberately chose `sQuote()` here to allow use of fancy quotes.
Testing for the warning could still be done by using `rex::rex("...", anything, "quoted", anything)` if we really want to test the whole message.

NEWS.md
```diff
@@ -8,6 +8,7 @@
...
 * Removed deprecated functions `absolute_paths_linter`, `camel_case_linter`, `multiple_dots_linter`, `snake_case_linter`, and `trailing_semicolons_linter`. They have been marked as deprecated since v1.0.1, which was released in 2017.
+* Rename `semicolon_terminator_linter` to `semicolon_linter` for better consistency. `semicolon_terminator_linter` survives but is marked for deprecation. The new linter also has a new signature, taking arguments `allow_compound` and `allow_trailing` to replace the old single argument `semicolon=`, again for signature consistency with other linters.
 * The `...` arguments for `lint()`, `lint_dir()`, and `lint_package()` have promoted to an earlier position to better match the [Tidyverse design principal](https://design.tidyverse.org/args-data-details.html) of data->descriptor->details. This change enables passing objects to `...` without needing to specify non-required arguments, e.g. `lint_dir("/path/to/dir", linter())` now works without the need to specify `relative_path`. This affects some code that uses positional arguments. (#935, @michaelchirico)
```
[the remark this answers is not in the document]
MichaelChirico: Yes :)

adding it to the NEWS cleanup PR so we don't forget...

R/with.R
MichaelChirico: I see. The issue was the fancy quotes are multi-byte, so regex `.` doesn't capture them. I'll revert.

AshesITR (approved):

Issue #979, opened 2022-03-22
Check relevance of false positive in yoda_test_linter


MichaelChirico:
This is a known false positive in `yoda_test_linter()`:

```
expect_equal(1, 1)
```

This should be linted (it's a placeholder that shouldn't really be checked in "to prod"), but currently the message is confusing.

We should check the practical relevance of this concern, though, before investing in a workaround.

MichaelChirico:
Found usages like this in 3 (out of 1005 sampled) packages:

https://github.com/DanChaltiel/crosstable/blob/bb50afa7abb52a9195836f06fc067827a9c1affa/tests/testthat/todo_test-1-crosstable-v2.R#L396

https://github.com/SoerenPannier/emdi/blob/ddbb55e75345170193c5f629cb455b7c61f4eade/tests/testthat/test_fh_std_spatial_robust.R#L320

https://github.com/jonocarroll/ggghost/blob/888ca70248bc501d3af4ca318e61a46d9c159286/tests/testthat/test_Ops.R#L10

The `ggghost` usages look like false positives -- the test is explicitly ensuring an overloaded `+` method works as expected for arithmetic input. That seems like a prime candidate for `# nolint` to me, so I am fine to ignore it.

As to the others, yes, let's invest in customizing the message to be better for this case.

MichaelChirico:
closed by #1065

PR #1064, opened 2022-04-12
More improvements to compare_branch

MichaelChirico:
These fixes are to facilitate #979 -- there, we don't really need to compare two branches, just to run the linter on a bunch of packages. So patched to allow no branch to be passed.

Also, since it's tightly related to `testthat`, it would be wasteful to include packages not using `testthat` in a sample. So it's nice to be able to supply a sample of packages manually if needed. That's facilitated by re-interpreting `--pkg_dir` and `--packages` not to be an either/or, but as two sides of the same coin. Now, `--pkg_dir` tells where to find the packages (defaulting to `.`, which allows the old usage of `--packages` to keep working), then `--packages` tells which ones to include.

Lastly, I got tired of copy/pasting the convoluted path to my CRAN mirror, so I exported it to env and let the script read that automatically.

NB: this PR is currently based against #1057.


.dev/compare_branches.R
```diff
@@ -141,22 +144,24 @@ if (is.null(base_branch) || is.na(base_branch) || !nzchar(base_branch)) {
...
 } else if (!is.null(params$pr)) {
   pr <- params$pr
 } else {
-  stop("Please supply a branch (--branch) or a PR number (--pr)")
+  has_target <- FALSE
 }
 
-# prioritize packages
+if (is.null(params$pkg_dir)) {
+  params$pkg_dir <- "."
```
AshesITR, line 158: Does this default make any sense since it should be run from the lintr project root?

.dev/compare_branches.R
```diff
@@ -338,12 +343,14 @@ run_workflow <- function(what, packages, linter_names, branch, number) {
 }
 
 message("Comparing the output of the following linters: ", toString(linter_names))
```
AshesITR, line 340: Only makes sense if `has_target`.

MichaelChirico:
@AshesITR #1057 needs to be merged before this one btw :)

.dev/compare_branches.R
MichaelChirico: Good point... I am not sure this is working as intended at HEAD anyway... just made it an error for now not to provide `pkg_dir` with a TODO & note.

AshesITR:
Thanks for reminding me. I took a look at #1057 now.

.dev/compare_branches.R
AshesITR: By the way: It would be very user-friendly to auto-download packages if `pkg_dir` is missing or doesn't contain any / the requested packages. That's how I initially thought this script worked.

MichaelChirico: Makes sense... added a TODO as an exercise for future readers for now :)

AshesITR (approved):

MichaelChirico:
@AshesITR looks like maybe UCRT is breaking some tests?

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test-get_source_expressions.R:118:3): Warns if encoding is misspecified ──
the_lint$message (`actual`) not equal to `msg` (`expected`).

`actual`:   "Invalid multibyte character in parser. Is the encoding correct?"
`expected`: "unexpected '<'"                                                 
```

AshesITR:
I made #1090 to use the correct expected error message for R >= 4.2.0.

PR #1070, opened 2022-04-12
customize yoda_test_linter message for placeholder tests

### LINKED ISSUE #979: Check relevance of false positive in yoda_test_linter

MichaelChirico:
Closes #979 

Based against #1069 for now.


AshesITR:
All other parts of the PR chain look fine. Merging top-to-bottom is fine once this is ready.

PR #1072, opened 2022-04-13
new allow_named_stopifnot argument for conjunct_test_linter

MichaelChirico:
Closes #1011

Decided to _allow_ the named conditions by default -- that's kind of the point of using the named approach to begin with. It can be abused, but I think this is the better default.

I especially had in mind helper functions, where the argument names might not mean anything to the end user, so the test failure `length(x) > 1 is not TRUE` is not particularly helpful if the user didn't call a function with argument `x`.


tests/testthat/test-conjunct_test_linter.R
```diff
@@ -95,3 +95,13 @@ test_that("conjunct_test_linter blocks simple disallowed usages of stopifnot() a
     conjunct_test_linter()
   )
 })
+
+test_that("conjunct_test_linter's allow_named_stopifnot argument works", {
+  # allowed by default
+  expect_lint("stopifnot('x is a logical scalar' = length(x) == 1 && is.logical(x) && !is.na(x))", NULL, conjunct_test_linter())
```
AshesITR, line 101: nit: IINM the error is taken directly from the argument name, so in production code, the name should be an error message like "`x` must be a logical scalar".

AshesITR (approved):

Issue #921, opened 2022-03-11
Add/export a readable wrapper for testing whether a linter is on an expression or the whole file


MichaelChirico:
We've got (minor perturbations of) this little idiomatic piece at the top of basically every linter:

https://github.com/r-lib/lintr/blob/23b06bdd4d8a28d94fa42eab4aebb58f8b108238/R/cyclocomp_linter.R#L13-L15

https://github.com/r-lib/lintr/blob/23b06bdd4d8a28d94fa42eab4aebb58f8b108238/R/line_length_linter.R#L13

https://github.com/r-lib/lintr/blob/23b06bdd4d8a28d94fa42eab4aebb58f8b108238/R/missing_argument_linter.R#L11

This is too obscure to a first-time reader. We should add a wrapper of this expression that makes it clear what's being done/why, and also export it for use in custom downstream linters.

AshesITR:
Good idea. `is_complete_file(source_file, require_xml = FALSE)`?
I remember the `is.null()` check on xml_parsed_content being used in some places specifically because it can happen that `xmlparsedata` fails or returns invalid XML.

Impl would be something like

```
if (require_xml) {
  !is.null(source_file[["full_xml_parsed_content"]])
} else {
  !is.null(source_file[["file_lines"]])
}
```

MichaelChirico:
Twice now I've been bitten by the `require_xml` part so this helper will do wonders for developers who don't understand the arcana of `xml` being sometimes present, sometimes not.

OTOH, maybe we could add a "dummy tree" when parsing fails and we can't add a real AST? Something as simple as

```
<xml>
<exprlist>
  <COMMENT>Parsing this file failed to generate a valid AST</COMMENT>
</exprlist>
```

Currently the failure mode is highly obscure:

```
══ Failed tests ══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
── Error (test-error.R:46:3): returns the correct linting ───────────────────────────
Error in `UseMethod("xml_find_all")`: no applicable method for 'xml_find_all' applied to an object of class "NULL"
Backtrace:
    ▆
 1. └─lintr::expect_lint("\\", rex(expected_message)) at test-error.R:46:2
 2.   └─lintr::lint(file, ...) at lintr/R/expect_lint.R:56:2
[...]

PR #1073, opened 2022-04-13
exported helper for determining if a linter is run at the expression/file level

MichaelChirico:
Towards #921. Pausing here for discussion before copying it over to more linters.

Still not very sure on the naming/signature here. There's a bit of a tension...

It's a bit weird to me to use `!is.null()` in the function and then `!is_tree_level()` -- the double negation feels odd.

But OTOH, using

```
if (is_tree_level(source_file, "complete", require_xml = TRUE))
```

instead feels worse -- (1) it feels better to declare "we want the expression-level tree; return if not" than to say "we don't want the complete-file tree (it's up to the reader's knowledge to remember the alternative is the expression-level tree)" (2) that approach would also require a tighter implementation, because `is.null(full_xml)` leaves three options for what's _non_-null. So we really want to filter based on what _is_ present, rather than what's not.

Another option would be to re-name the function, e.g. `not_tree_level()`, then we'd get

```
if (not_tree_level(source_file, "expression", require_xml = TRUE)) return(list())
```

but it doesn't stick with me right away.


AshesITR:
tree doesn't seem to self-explanatory to me. What about `is_lint_level(level = c("expression", "file"), require_xml = FALSE)`?

MichaelChirico:
> What about is_lint_level(level = c("expression", "file"), require_xml = FALSE)

I guess "lint level" is not immediately clear either (does it mean loose/strict linting maybe?), but is probably clear enough in actual usage.

Now I'm also questioning whether using `source_file` as the argument is also misleading -- `expression_metadata`?

AshesITR:
The object isn't really expression metadata, is it?
I'm thinking of something with code or source in its name, but I agree `source_file` isn't ideal. Renaming is fine because the argument is positional anyway.

Maybe `source_code`? Reads fine to me, e.g.: `source_code$lines`, `source_code$file_lines`.

MichaelChirico:
source_code sounds good to me! nice! source_expressions could also work because it ties nicely to the object's source (get_source_expressions)

AshesITR:
Good catch, `source_expression` seems perfect.

MichaelChirico:
@AshesITR ready for design review again. I think the main question is whether we want

```
if (!is_lint_level(source_expression, "expression"))
```

or

```
if (not_lint_level(source_expression, "expression"))
```

AshesITR:
I prefer the `!` variant.
Would also rename `level = "complete"` to `level = "file"` in case we ever add static analysis capabilities across files (e.g. required for checking DESCRIPTION dependencies in packages).

Also, can we add the valid levels as a character to the default of level?
I'm fine with stopping if level isn't explicitly given, but I absolutely hate interfaces that don't tell me in the signature what valid string constants are (a recent example being rex' regex options).

Looks good with these changes I think.

MichaelChirico:
> Also, can we add the valid levels as a character to the default of level?

The reason not to have it as a default is to force `!missing(level)`, and the valid values are in the first line of the function body...

OTOH I suppose we could use one of the levels as a default and then `stopifnot(!missing(level))`, WDYT?

AshesITR:
Yeah that's what I meant.

MichaelChirico:
OK will start applying this to the rest of the linters

MichaelChirico:
This is ready for full review

AshesITR:
```
Undocumented arguments in documentation object 'is_lint_level'
    'source_expression'
```

MichaelChirico:
fixed the typo & a few other things, PTAL.

R/is_lint_level.R
```diff
@@ -0,0 +1,21 @@
+#' Is this an expression- or a file-level source object?
+#'
+#' Helper for determining whether the current `source_expression` contains
+#'   the full file tree, or is just a single expression.
```
AshesITR, line 4: Full file tree isn't really clear to me. Maybe write "contains all expressions in the current file, or just a single expression." instead?

R/is_lint_level.R
```diff
@@ -0,0 +1,21 @@
...
+#'
+#' @param source_expression A parsed expression object, i.e., an element
+#'   of the object returned by [get_source_expressions()].
+#' @param level Which level of expression is being tested? `"expression"`
+#'   means an individual expression, while `"file"` means the full
+#'   source tree for this file.
+#' @param require_xml If `TRUE`, the XML equivalent of the parse tree
+#'   is required. Use this whenever writing a linter based on logic
+#'   expressed via Xpath.
```
AshesITR, line 13: Typo: XPath has a capital P

R/is_lint_level.R
```diff
@@ -0,0 +1,21 @@
+#' Is this an expression- or a file-level source object?
+#'
+#' Helper for determining whether the current `source_expression` contains
+#'   the full file tree, or is just a single expression.
+#'
+#' @param source_expression A parsed expression object, i.e., an element
+#'   of the object returned by [get_source_expressions()].
+#' @param level Which level of expression is being tested? `"expression"`
+#'   means an individual expression, while `"file"` means the full
+#'   source tree for this file.
```
AshesITR, line 10: ditto.

AshesITR:
The code changes look good to me. Just a few documentation suggestions.

NEWS.md
```diff
@@ -174,6 +174,7 @@ function calls. (#850, #851, @renkun-ken)
 * `trailing_whitespace_linter()` ignores trailing whitespace in strings by default. 
   This can be disabled using `allow_in_strings = FALSE` (#1045, @AshesITR)
 * Moved the default lintr cache directory from `~/.R/lintr_cache` to `R_user_dir("lintr", "cache")`. Note that this major version update invalidated the old cache anyway, so it can be safely deleted. (#1062, @AshesITR)
+* New `is_lint_level()` helper for readably explaining which type of expression is required for a custom linter. Some linters are written to either require the full file's parse tree (for example, `single_quotes_linter()`. Others only need single expressions, which is more cache-friendly (most linters are written this way to leverage caching). (#921, @michaelchirico)
```
AshesITR, line 177:
```suggestion
* New `is_lint_level()` helper for readably explaining which type of expression is required for a custom linter. Some linters are written to either require the full file's parse tree (for example, `single_quotes_linter()`). Others only need single expressions, which is more cache-friendly (most linters are written this way to leverage caching). (#921, @michaelchirico)
```

AshesITR (commented):
mini typo. rest LGTM. Thanks!

AshesITR (approved):

PR #1076, opened 2022-04-14
Helper for extracting string values from literals and application thereof

### LINKED ISSUE #1034: Ensure linter logic allows for R>=4 raw strings

MichaelChirico:
Closes #1034


R/condition_message_linter.R
```diff
@@ -23,20 +23,14 @@ condition_message_linter <- function() {
...
-            and following-sibling::expr[1]/STR_CONST[text() = '\"\"' or text() = '\" \"']
-          ]
-        )
       ]
     ]")
 
     bad_expr <- xml2::xml_find_all(xml, xpath)
+    sep_value <- get_r_string(xml2::xml_find_first(bad_expr, "./expr/SYMBOL_SUB[text() = 'sep']/following-sibling::expr/STR_CONST"))
 
     return(lapply(
```
AshesITR, line 38: Unneeded `return()`

tests/testthat/test-absolute_path_linter.R
```diff
@@ -200,3 +174,13 @@ test_that("returns the correct linting", {
     expect_lint(double_quote(path), NULL, linter)
   }
 })
+
+test_that("raw strings are handled correctly", {
+  skip_if_not_installed("base", "4.0.0")
```
AshesITR, line 179: For consistency I think we should stick to a `getRVersion()` based skip.

tests/testthat/test-single_quotes_linter.R
```diff
@@ -26,3 +26,10 @@ test_that("returns the correct linting", {
     linter
   )
 })
+
+test_that("handles R>=4.0.0 strings", {
+  skip_if_not_installed("base", "4.0.0")
+  expect_lint('R"(hello \'\' there)"', NULL, single_quotes_linter())
```
AshesITR, line 32: Just to be sure, `r'(")'` is also a lint, right?

tests/testthat/test-absolute_path_linter.R
MichaelChirico: I wrote a helper for this instead... not a fan of the skip message `getRversion() < '4.0.0' is not TRUE`

tests/testthat/test-single_quotes_linter.R
MichaelChirico: you mean not a lint right?

AshesITR: No it should lint I think because we only want double quotes unless it requires escapes.

`r"(")"` is fine so no need for `'` at all.

MichaelChirico: Hmm I see what you mean. Not sure I agree though... `r"(")"` may be valid but it's not so readable. `r'(")'` is still clearer IMO

MichaelChirico: I got as far as writing the regex before changing my mind though, I think this works for posterity:

```
rex(
  list(start, single_quote, any_non_double_quotes, single_quote, end) %or%
  list(start, character_class('Rr'), single_quote)
)
```

MichaelChirico: maybe it's best to leave this as an option? if so let's handle in follow-up

AshesITR (approved):

PR #1078, opened 2022-04-16
Don't define backports if not needed

MichaelChirico:
Plus other minor cleanup of the helpers in R/utils.R

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -204,13 +204,11 @@ get_function_assignments <- function(xml) {
     setmethod_assignment_functions
   )
 
-  get_attr <- function(x, attr) as.integer(xml2::xml_attr(x, attr))
-
   res <- data.frame(
-    line1 = viapply(funs, get_attr, "line1"),
-    line2 = viapply(funs, get_attr, "line2"),
-    col1 = viapply(funs, get_attr, "col1"),
-    col2 = viapply(funs, get_attr, "col2"),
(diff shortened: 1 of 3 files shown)


R/object_usage_linter.R
```diff
@@ -204,13 +204,11 @@ get_function_assignments <- function(xml) {
...
   )
 
-  get_attr <- function(x, attr) as.integer(xml2::xml_attr(x, attr))
-
   res <- data.frame(
-    line1 = viapply(funs, get_attr, "line1"),
-    line2 = viapply(funs, get_attr, "line2"),
-    col1 = viapply(funs, get_attr, "col1"),
-    col2 = viapply(funs, get_attr, "col2"),
+    line1 = as.integer(vapply(funs, xml2::xml_attr, attr = "line1", "")),
```
AshesITR, line 208: Is there a performance trade-off between calling `as.integer()` outside the `vapply` (as it is now) vs. inside (as it was before)?

MichaelChirico: Small on absolute scale, but yes, it's basically `as.integer(x)` vs. `vapply(x, as.integer, 1L)`, bound to be slower. A quick benchmark gave me 4-10x slower (isolating the `as.integer()` comparison). But e.g. 1 vs 6 microseconds for length-10 input.

MichaelChirico: Including the xml2::xml_attr step on a test, I found 33% improvement (27 down to 18 us).

AshesITR (approved):

Issue #833, opened 2021-07-09
Polish NEWS


AshesITR:

MichaelChirico:
TODO: put an item about changes to `default_linters` in the "breaking changes" section at the top

MichaelChirico:
https://github.com/r-lib/lintr/pull/1061#discussion_r851636461

https://github.com/r-lib/lintr/blob/fc6cb377839ffb6862601e72c0e1b86f7ef00015/NEWS.md?plain=1#L11

AshesITR:
TODO explain what flexibility was lost to #1041 (unification of brace related linters)

MichaelChirico:
Leaving this as open for now to force us to take one more pass before release

PR #1085, opened 2022-04-19
explain programmatic access to linters

### LINKED ISSUE #1058: Document using call() over get() for new functional-style linters
Compare:

```
library(lintr)
linter <- "assignment_linter"

as.data.frame(lintr::lint(text = "a = 1", linters = get(linter)()))$linter
# [1] "get"
as.data.frame(lintr::lint(text = "a = 1", linters = eval(call(linter))))$linter
# [1] "assignment_linter"
```

Or, perhaps this is a bug?

Noticed during recent use of .dev/compare_branches:

https://github.com/r-lib/lintr/blob/f1643b7430a769947a862c4cd37d0e930bd32f50/.dev/compare_branches.R#L329

MichaelChirico:
Closes #1058

This may not be the best approach here -- open to ideas here. But it does seem like a natural fit. Maybe we can add something to the NEWS too -- maybe as part of the broader push to clean up NEWS (#833). Or maybe just put it in the 3.0.0 blog post (that's a thing right?)

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -7,14 +7,14 @@ NULL
 #' @name linters
 #'
 #' @description A variety of linters is available in \pkg{lintr}. The most popular ones are readily
-#' accessible through [default_linters()], though there are additional ones you may want
-#' to use.
+#' accessible through [default_linters()].
 #'
 #' Within a [lint()] function call, the linters in use are initialized with the provided
 #' arguments and fed with the source file (provided by [get_source_expressions()]).
 #'
 #' A data frame of all available linters can be retrieved using [available_linters()].
-#' Documentation for linters is structured into tags to allow for easier discovery.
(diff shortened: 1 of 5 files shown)


AshesITR:
Seems more elegant to mention the name attribute.

You can manually name a linter using

```
library(lintr)
linter_name <- "assignment_linter"
linter <- get(linter_name)()
attr(linter, "name") <- "my_linter"
```

Which is exactly what we do in `linters_with_tags()`.

MichaelChirico:
> Seems more elegant to mention the name attribute.

not sure i agree it's more elegant :)

How about just mentioning both approaches? See latest

AshesITR:
It's more flexible at least since it allows renaming to arbitrary names.

AshesITR (approved):

PR #1089, opened 2022-04-23
Rename source_file -> source_expression

MichaelChirico:
As suggested in #1073; splitting off for clarity since this change is independent of that.

Marked as a breaking change because of the external signature change in `with_id()` and `ids_with_token()`.

Also realized that the reason it's named `source_file` to begin with is that the object starts as the result of `srcfile`, to which we add more decoration gradually:

https://github.com/r-lib/lintr/blob/d916a21b79eb46d74ca6521c951d073eeb6456df/R/get_source_expressions.R#L57

Still I think `source_expression` is a better name for the object we end up passing along as `get_source_expressions()` output (hence the function's own name).


MichaelChirico:
I suspect the CI failure may be related to the R 4.2 release...

AshesITR:
Can we deprecate the old `source_file` argument just in case someone didn't supply it positionally, to ease the transition?

Linter extension packages will have a lot of work migrating from 2.0.1 to 3.0.0 anyway already.

MichaelChirico:
Good idea. Didn't add any tests since there are no tests for `with_id` or `ids_with_token` yet, and this PR is bloated enough as is :)

R/yoda_test_linter.R
```diff
@@ -41,10 +41,18 @@ yoda_test_linter <- function() {
...
+        second_const <- xml2::xml_find_first(expr, glue::glue("expr[position() = 3 and ({const_condition})]"))
+        if (is.na(second_const)) {
+          paste(
+            "Tests should compare objects in the order 'actual', 'expected', not the reverse.",
+            sprintf("For example, do %1$s(foo(x), 2L) instead of %1$s(2L, foo(x)).", matched_call)
+          )
+        } else {
+          sprintf("Avoid storing placeholder tests like %s(1, 1)", matched_call)
+        }
+      },
```
AshesITR, line 55: Based against a different PR?

R/ids_with_token.R
```diff
@@ -15,11 +15,16 @@
 #'
 #' @param fun For additional flexibility, a function to search for in
 #' the `token` column of `parsed_content`. Typically `==` or `%in%`.
+#' @param source_file [DEPRECATED] Same as `source_expression`. Will be removed.
 #' @return `ids_with_token`: The indices of the `parsed_content` data frame
 #' entry of the list of source expressions. Indices correspond to the
 #' *rows* where `fun` evaluates to `TRUE` for the `value` in the *token* column.
 #' @export
-ids_with_token <- function(source_expression, value, fun = `==`) {
+ids_with_token <- function(source_expression, value, fun = `==`, source_file) {
```
AshesITR, line 23: Can we add tests for the warning, like we did for `semicolon_terminator_linter()`?

R/with_id.R
```diff
@@ -9,7 +9,11 @@
 #' of `parsed_content`.
 #' @return `with_id`: A data frame corresponding to the row(s) specified in `id`.
 #' @export
-with_id <- function(source_expression, id) {
+with_id <- function(source_expression, id, source_file) {
```
AshesITR, line 12: ditto

R/yoda_test_linter.R
MichaelChirico: hmm maybe a bad merge. i'll rebase.

R/ids_with_token.R
MichaelChirico: would prefer as a follow-up -- will get buried in this huge PR.

R/yoda_test_linter.R
MichaelChirico: should be ok now

AshesITR (approved):

PR #1090, opened 2022-04-25
fix windows test for R 4.2.0

AshesITR:


tests/testthat/test-get_source_expressions.R
```diff
@@ -109,8 +109,9 @@ test_that("Warns if encoding is misspecified", {
   expect_s3_class(the_lint, "lint")
 
   msg <- "Invalid multibyte character in parser. Is the encoding correct?"
-  if (.Platform$OS.type == "windows") {
-    # Windows parser throws a different error message because the source code is converted to native encoding
+  if (.Platform$OS.type == "windows" && getRversion() < "4.2.0") {
```
MichaelChirico, line 112: maybe we can just use l10n_info()? would rather test for capabilities than system properties

also worth giving a quick check on the rest of the codebase for similar tests...

AshesITR: How would this test go with `l10n_info()` I don't have 4.2.0 on windows at the moment.

AshesITR: Is `if (!l10n_info()[["UTF-8"]]) { ... }` correct? I just installed R 4.2.0 to see what windows UCRT reports:

```
> l10n_info()
$MBCS
[1] TRUE

$`UTF-8`
[1] TRUE

$`Latin-1`
[1] FALSE

$codepage
[1] 65001

$system.codepage
[1] 65001
```

4.1.2 reports
```
> l10n_info()
$MBCS
[1] FALSE

$`UTF-8`
[1] FALSE

$`Latin-1`
[1] TRUE

$codepage
[1] 1252

$system.codepage
[1] 1252
```

MichaelChirico: that looks right to me

Issue #1041, opened 2022-04-02
Unify bracing-related linters / parentheses / if-else clauses?


MichaelChirico:
We currently have a variegated list of linters related to bracing and parentheses:

 - `closed_curly_linter()`
 - `else_same_line_linter()`
 - `function_brace_linter()`
 - `function_left_parentheses_linter()`
 - `if_else_match_braces_linter()`
 - `open_curly_linter()`
 - `paren_body_linter()`
 - `paren_brace_linter()`
 - `spaces_inside_linter()`
 - `spaces_left_parentheses_linter()`

Should we consolidate any of these?

AshesITR:
Conceptuall, I can see linters related to the style of function definitions and linters related to flow control statements.
The many stylistic difference between the two is the spacing of the control condition (requires a space after the keyword) vs. the spacing of the argument list (forbids space after the `function` keyword).

The `spaces_inside_linter()` doesn't really fit this description imo.
`paren_body_linter()` could be included in a `function_style_linter()` but isn't related to braces.

A `brace_linter()` could unify
 * `closed_curly_linter()`, keeping the `allow_single_line = FALSE` option
 * `else_same_line_linter()`
 * `function_brace_linter()` -- btw: the lint `type` should probably be "style".
 * The `FUNCTION` part of `function_left_parentheses_linter()`, but not the `SYMBOL_FUNCTION_CALL` part.
 * `if_else_match_braces_linter()
 * `open_curly_linter()`, keeping the `allow_single_line = FALSE` option and adding an option to allow C-style (i.e. `{` must always be on its own line -- NB though that `else` can't be on a different line than `}` in R, so this should probably only apply to `function()`)
 * `paren_brace_linter()`, which should also lint if line breaks are used as spacing unless aforementioned C-style option is on.
 * `spaces_left_parentheses_linter()`

Alternatively, we could unify to two linters 

1.`function_style_linter()` with options `allow_single_line = FALSE`, `c_style_braces = FALSE`, `allow_space_before_args = FALSE` and `allow_braceless_multiline = FALSE`
2. `flow_style_linter()` with options `allow_single_line = FALSE`, `c_style_braces = FALSE` and `allow_braceless_multiline = FALSE`

NB that since many of these linters are default in 2.0.1, custom linter configs with, e.g. `with_defaults(open_curly_linter = NULL)` will currently just stop working as expected without a warning.
[...]

PR #1096, opened 2022-04-25
deprecate open_curly_linter and merge it into brace_linter

AshesITR:
Based against #1095, part of #1041

 - remove open_curly_linter from defaults
 - refactor to XPath based approach
 - no longer lint trailing whitespace (there's a separate linter for that)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 * Rename `semicolon_terminator_linter` to `semicolon_linter` for better consistency. `semicolon_terminator_linter` survives but is marked for deprecation. The new linter also has a new signature, taking arguments `allow_compound` and `allow_trailing` to replace the old single argument `semicolon=`, again for signature consistency with other linters.
 * Combined several curly brace related linters into a new `brace_linter` (#1041, @AshesITR):
   + `closed_curly_linter()`
+  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace
   + Require `else` to come on the same line as the preceding `}`, if present (#884, @michaelchirico)
   + Require functions spanning multiple lines to use curly braces (@michaelchirico)
   + Require balanced usage of `{}` in `if`/`else` conditions (@michaelchirico)
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@
 * Combined several curly brace related linters into a new `brace_linter` (#1041, @AshesITR):
   + `closed_curly_linter()`
   + `open_curly_linter()`, no longer linting unnecessary trailing whitespace
+  + `paren_brace_linter()`, also linting `if`/`else` and `repeat` with missing whitespace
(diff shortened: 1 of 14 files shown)


NEWS.md
```diff
@@ -11,6 +11,7 @@
 * Rename `semicolon_terminator_linter` to `semicolon_linter` for better consistency. `semicolon_terminator_linter` survives but is marked for deprecation. The new linter also has a new signature, taking arguments `allow_compound` and `allow_trailing` to replace the old single argument `semicolon=`, again for signature consistency with other linters.
 * Combined several curly brace related linters into a new `brace_linter` (#1041, @AshesITR):
   + `closed_curly_linter()`
+  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace
```
MichaelChirico, line 14: let's add a test around this? it looks like `lintr::lint(text = "function(x) {   \n  x\n}", linters = list(open_curly_linter(), trailing_whitespace_linter()))` returns two lints before this fix. Let's test to ensure there's only one.

R/open_curly_linter.R
```diff
@@ -9,6 +9,7 @@
 #'   <https://style.tidyverse.org/syntax.html#indenting>
 #' @export
 open_curly_linter <- function(allow_single_line = FALSE) {
```
MichaelChirico, line 11: Could you explain why not call `brace_linter()` from here? I have an idea but it would be good to have it written out.

I am slightly worried about the logic between the two drifting apart. Ditto for `closed_curly_linter()`. Is there any tests we could set up to prevent that?

NEWS.md
AshesITR: The problem still persists (because I left `open_curly_linter()` as-is, but we can test that `brace_linter()` doesn't lint said expression.

R/open_curly_linter.R
AshesITR: I've left those alone to ensure only the specific lints are produced by the old implementations.
The only behavioural change is `open_curly_linter()` linting superfluous whitespace where `brace_linter()` does not.

To make this sure, I copied all tests from the merged linters over to `tests/testthat/test-brace_linter.R`

Enabling `brace_linter()` to only produce selected old lints seems like unnecessary complication of the new linter to me. WDYT?

MichaelChirico: SGTM. it is technically a sort of breaking change (no longer able to toggle the linters individually), though I'm not sure anyone actually needed that. let's be sure it's adequately conveyed in the NEWS -- perhaps as part of the NEWS issue

NEWS.md
MichaelChirico: yep, a test with brace_linter is what I had in mind

R/open_curly_linter.R
AshesITR: Yeah, let's consolidate all the breaking changes and migration instructions in a central place to keep them consistent.

PR #1097, opened 2022-04-25
merge paren_brace_linter into brace_linter

AshesITR:
Based against #1096, part of #1041

 * added lints for `else{` and `repeat{` while I was at it.

### DIFF
--- a/R/brace_linter.R
+++ b/R/brace_linter.R
@@ -3,6 +3,7 @@
 #' Perform various style checks related to placement and spacing of curly braces:
 #'
 #'  - Opening curly braces are never on their own line and are always followed by a newline.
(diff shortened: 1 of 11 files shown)


R/brace_linter.R
```diff
@@ -54,6 +55,21 @@ brace_linter <- function(allow_single_line = FALSE) {
       )
     ))
 
+    xp_open_preceding <- "parent::expr/preceding-sibling::*[1][self::OP-RIGHT-PAREN or self::ELSE or self::REPEAT]"
```
MichaelChirico, line 58: nice catch!

MichaelChirico (approved):

PR #1099, opened 2022-04-25
support automatic exclusion of deprecated linters

### LINKED ISSUE #1098: Add `deprecated` as a linter tag and remove these from `linters_with_tags(NULL)`
Since we're planning to deprecate some linters starting with 3.0.0, these shouldn't be automatically instantiated by `linters_with_tags(NULL)`.
I think the cleanest way to do this is add a deprecated tag and change `linters_with_tags()` to automatically exclude linters tagged `deprecated`.

AshesITR:
Fixes #1098 

An alternative approach would be to make the argument `except_tags = "deprecated"`, which would allow, e.g. `linters_with_tags(tags = NULL, except_tags = "package_development")` too.

LMK if you'd prefer this more flexible approach.


MichaelChirico:
> make the argument `except_tags = "deprecated"`

yea, that sounds better to me. `include_deprecated` sounds too specific, it's easier to imagine `except_tags` being more generally useful.

R/linter_tags.R
```diff
@@ -40,17 +42,20 @@
...
-available_linters <- function(packages = "lintr", tags = NULL) {
+available_linters <- function(packages = "lintr", tags = NULL, include_deprecated = FALSE) {
   if (!is.character(packages)) {
     stop("`packages` must be a character vector.")
   }
   if (!is.null(tags) && !is.character(tags)) {
     stop("`tags` must be a character vector.")
   }
+  if (!is.logical(include_deprecated) || length(include_deprecated) != 1L || is.na(include_deprecated)) {
+    stop("`include_deprecated` must be a flag.")
```
MichaelChirico, line 53: I think `flag` is not a universal term, "logical scalar" or just "TRUE or FALSE" would be better to me

AshesITR:
I also added a test that deprecation warnings are issued for and only for linters tagged "deprecated" to make sure we keep the two in sync. LMK if you come up with a more elegant way to test for the expected warnings (including the correct cardinality)

AshesITR:
also obsoletes #1109

R/linter_tag_docs.R
```diff
@@ -80,3 +80,12 @@ NULL
...
 NULL
+
+#' Deprecated linters
+#' @name deprecated_linters
+#' @description
+#' Linters that are deprecated and provided for backwards compatiblility only.
+#' These linters will be excluded from `linters_with_tags()` by default.
+#' @evalRd rd_linters("deprecated")
+#' @seealso [linters] for a complete list of linters available in lintr.
+NULL
```
MichaelChirico, line 91: now i see why `?deprecated_linters` wasn't being generated in #1109 -- didn't have the `NULL` anchor! 🤦

tests/testthat/test-linter_tags.R
```diff
@@ -33,8 +33,30 @@ test_that("default_linters and default tag match up", {
...
+      warning = function(w) {
+        if (warns_expected >= 1L && grepl("was deprecated", conditionMessage(w))) {
+          warns_expected <<- warns_expected - 1L
+          invokeRestart("muffleWarning")
+        } else {
+          w
+        }
+      }
+    )
+    if (warns_expected != 0L) {
```
MichaelChirico, line 52: I think `expect_identical(warns_expected, 0L)` is more testthat-ish... in which case we may want to take an increment (not decrement) approach, then failure would be more readably `expect_identical(n_deprecation_warnings, num_deprecated_linters)`

MichaelChirico (commented):
basically LGTM

tests/testthat/test-linter_tags.R
AshesITR: I agree, but was afraid of the failure info "warns_expected is not equal to 0" being misleading.
Counting from 0 and up should fix this problem though.

MichaelChirico (approved):

Issue #1087, opened 2022-04-20
Support Box


latot:
Hi, I'm using lintr in code and github, I think would be great support the box module, because actually lintr don't detect the functions contained in the modules.

https://github.com/klmr/box

Thx.

AshesITR:
Hi, can you describe what kind of support you'd like?
Note that the `object_usage_linter` can see functions defined in your globalenv, so it may help to import your boxes before calling `lint()` if you see too many false positives from that particular linter.

latot:
mm, I have a code, but seems the problem is in other place, i'll check it deeper, for now, I think would be great a fix for this:

```
b.R:1:11: style: Put spaces around all infix operators.
box::use(./a)
```

Thx.

MichaelChirico:
seems like an unfortunate syntax choice by box to save 2 characters IMO

AshesITR:
If you are using the development version of `lintr`, you could wrap your `box::use()` statements in a nolint block like this:

```
# nolint start: infix_spaces_linter.
box::use(./a)
# nolint end
```

AshesITR:
In lintr 2.0.1, you can use `# nolint start` instead of `# nolint start: infix_spaces_linter.` as a workaround (though this will suppress all lints in the surrounded region, so use with care).

AshesITR:
Yes, @klmr, the `infix_spaces_linter` uses the XML parse tree, so can do arbitrary XPath based rules to lint or not lint something.
To avoid false negatives, it's probably best to exactly match `box::use(<expr>/<expr>)` as a negative in this case.

Feel free to create a PR adding this exception to infix_spaces_linter.

PR #1105, opened 2022-04-26
Add an exception for `box::use()` to infix spaces linter

klmr:
As discussed in #1087, this is a draft to extend the `infix_spaces_linter` with an exception for `lhs/rhs` expressions inside a `box::use()` call, which represent import declarations that should resemble file paths, and thus don’t call for spaces around `/`.

My XPath-fu is fairly rusty so I’d appreciate a review of the implementation. I’ve opted not to modify the rule’s current XPath, instead filtering the resulting lint list by excluding all occurrences of `/` inside `box::use()`. The XML structure of the parsed expressions makes this somewhat awkward, because rather than being nested (as I naïvely expected), the surrounding `box::use()` call expression is *adjacent* to the parent node of the `/` expression.

I’m not 100% sure whether my way of testing for a `box::use()` call is maximally specific or whether it could lead to false positives … however, I can’t see how this would happen.


tests/testthat/test-infix_spaces_linter.R
```diff
@@ -135,3 +135,20 @@ test_that("infix_spaces_linter can allow >1 spaces optionally", {
...
+      a,
+      a/b,
+      ../a,
+      alias = a/b/c[xyz = abc, ...],
+    )"
+  )
+
+  for (decl in use_declarations) {
+    expect_lint(decl, NULL, infix_spaces_linter())
+  }
```
AshesITR, line 153: This will produce unhelpful test failure messages if the tests ever start failing.
Better to duplicate `expect_lint()` calls, so the reported line number immediately leads to the failing test case.

R/infix_spaces_linter.R
```diff
@@ -112,6 +112,22 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces
...
+    bad_expr <- Filter(
+      function (expr) {
+        is.na(xml2::xml_find_first(
+          expr,
+          "self::OP-SLASH[
+            ancestor::expr/preceding-sibling::OP-LEFT-PAREN/preceding-sibling::expr[
+              ./SYMBOL_PACKAGE[text() = 'box'] and
+              ./SYMBOL_FUNCTION_CALL[text() = 'use']
+            ]
+          ]"
```
AshesITR, line 125: The XPath looks sound.
I think it would be preferrable to include it as a `not()` condition in the bad_expr XPath though.
Is there any infix operator that can occur in `box::use()` expressions and that should be linted?
Otherwise I think this can simply be appended to the conditions via `and not(...)`.

AshesITR:
Thanks for submitting this PR!

klmr:
Thanks, I’ve made adjustments in response to your two comments. 👍

AshesITR (approved):
LGTM, thanks again!

PR #1112, opened 2022-04-28
ignore $symbol in undesirable_function_linter

### LINKED ISSUE #1050: False-positive `undesirable_function_linter` for `df$source <- 1`
image

I thought this had beed addressed in https://github.com/r-lib/lintr/issues/556, but maybe not completely.

AshesITR:
fixes #1050
added a regression test and did some mild XPath refactoring

### DIFF
--- a/R/undesirable_function_linter.R
+++ b/R/undesirable_function_linter.R
@@ -22,12 +22,15 @@ undesirable_function_linter <- function(fun = default_undesirable_functions,
       tokens <- "SYMBOL_FUNCTION_CALL"
     }
 
-    xpath <- sprintf(
-      "//*[(%s) and (%s) and not(parent::expr/preceding-sibling::expr[SYMBOL_FUNCTION_CALL[%s]])]",
+    xpath <- paste0("//*[", xp_and(
       paste0("self::", tokens, collapse = " or "),
       xp_text_in_table(names(fun)),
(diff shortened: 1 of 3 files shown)


R/undesirable_function_linter.R
```diff
@@ -22,12 +22,15 @@ undesirable_function_linter <- function(fun = default_undesirable_functions,
       tokens <- "SYMBOL_FUNCTION_CALL"
     }
 
-    xpath <- sprintf(
-      "//*[(%s) and (%s) and not(parent::expr/preceding-sibling::expr[SYMBOL_FUNCTION_CALL[%s]])]",
+    xpath <- paste0("//*[", xp_and(
       paste0("self::", tokens, collapse = " or "),
```
AshesITR, line 26: Curious what is faster, `//SYMBOL[...] | //SYMBOL_FUNCTION_CALL[...]` or `//*[[(self::SYMBOL or self::SYMBOL_FUNCTION_CALL) and ...]`.

I'm guessing the first XPath might be faster because it doesn't have to enter all nodes, but not sure.
We have this idiom in a couple of places so may be worth investigating.

MichaelChirico: if there is a "smart" compilation stage, the execution could be the same for both. agree it's worth investigating

Issue #699, opened 2020-12-09
Add `implicit_integer_linter` to our linters?


AshesITR:
Currently 508 lints in the codebase.
I don't think this should be a default linter, because it is very restrictive.
However, we could use this more strict linter on our code. WDYT?

The fix is to intentionally use `1L` or `1.0` depending on what the data type should be.
Most of our code probably uses integers.

MichaelChirico:
FWIW i have such linter written here already:

https://github.com/Rdatatable/data.table/pull/4278

That said, I don't know what the LICENSE implications are here, maybe best to implement from scratch on `lintr`?

AshesITR:
It already exists?

MichaelChirico:
not as a `Lint`, but the logic to search the XML parse tree & find integer-as-numeric

BTW, I don't know how much it matters for this linter (it can just be turned off if this is a concern?), but long vector support means `x[1]` is probably fine? since the index in general could be double.

AshesITR:
I'm confused.

R/implicit_integer_linter.R exists in lintr already and works, no?

Re `x[1]`: that expression is syntactically equivalent to `x[1L]`. That's why we call it implicit integers. If you truly wanted to index with a double, you'd need to write `x[1.0]`.

MichaelChirico:
Your confusion is well-founded -- I didn't know of that function in `lintr`!

`x[2147483648]` is not syntactically equivalent to `x[2147483648L]`, however (since the latter is `NA_integer_`). That said, I did just check and `x[2147483648]` does not produce a lint, so ignore me :smiley:

AshesITR:
Imo that should lint implicit doubles should be avoided 😄

MichaelChirico:
(BTW, yes, let's add this!)

AshesITR:
This has been done.

PR #1113, opened 2022-04-28
de-lint master

AshesITR:
remove all easy to clean lints from master.
Notable exception: `lint()` has reached a cyclocomp of 37, mainly due to compatibility layer and the outstanding PR to factor out the cache complexity.

I also added `implicit_integer_linter()` to `.lintr_new` in line with #699.
However, master currently has 449 implicit integer lints, so not enabling it yet.


MichaelChirico:
The PR is showing some implicit integer lints on changed files... up to you whether to fix now

MichaelChirico (dismissed):

AshesITR:
Sounds like a good idea. Will do.

MichaelChirico (dismissed):

MichaelChirico (approved):

PR #1117, opened 2022-05-02
deprecate with_defaults -> linters_with_defaults

### LINKED ISSUE #1029: Refactor `with_defaults()`
https://github.com/r-lib/lintr/blob/b4747a305b7a9c8756d626b2d23a78d43f38f226/R/zzz.R#L65

let's use `is.function(x)` instead?

```
f = lm
class(f) = 'boo'
inherits(f, 'function')
# [1] FALSE
is.function(f)
# [1] TRUE
```

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/1015#discussion_r837765986_

AshesITR:
Fixes #1029 

 - rename the core function to modify_defaults
 - add deprecation test for with_defaults()
 - update documentation to refer to linters_with_defaults() instead of with_defaults() where appropriate.


tests/testthat/test-with.R
```diff
@@ -39,3 +40,22 @@ test_that("can instantiate all linters without arguments", {
...
   expect_length(really_all_linters, nrow(available_linters(exclude_tags = NULL)))
 })
+
+test_that("with_defaults is supported with a deprecation warning", {
+  defaults <- linters_with_defaults()
+  expect_warning(
+    old_defaults <- with_defaults(),
+    rex::rex("Use linters_with_defaults instead.")
+  )
+  expect_equal(defaults, old_defaults)
```
MichaelChirico, line 50: identical?

MichaelChirico:
I wonder if `linters_with_defaults(default = defaults, ...)` is a bit tough to read? too redundant? maybe i've just been starting at it too long. Is there a better name for that argument that gets at what it's for? `default_set`? `baseline`?

tests/testthat/test-with.R
AshesITR: These are currently identical, but not e.g. if using `linters_with_tags("default")` as a reference, because this constructs new functions with their own environments.
That's why I decided to only test equality.

AshesITR:
To be fair, the argument is usually not needed, and it would read `linters_with_defaults(default = default_linters)` compare
`linters_with_tags(tags = )`.

We could pluralize the argument for better consistency across the two?

MichaelChirico:
I think that's indeed a little better. Let's go with that for now. I am worried I'm overthinking this.

MichaelChirico (dismissed):

MichaelChirico (approved):

PR #1118, opened 2022-05-02
Change location of default cache directory

### LINKED ISSUE #1062: Location of the lintr cache
@jimhester and I got this email from Kurt Hornik / CRAN.

> This has
>
>   ./lintr/R/zzz.R:    cache_directory = "~/.R/lintr_cache",   
>
> in violation of the CRAN Policy: can you please change to use
> tools::R_user_dir("lintr", "cache") instead?
>
> CC'ing Jennifer for devtools: this has
>
> ./devtools/man/lint.Rd:The lintr cache is by default stored in
>  \verb{~/.R/lintr_cache/}
>
> which will need to be changed accordingly ...

I'm relaying since I'm not sure how often Jim is checking in on lintr. But it sounds like it may be time to change the default location of lintr's cache. boxr recently did something similar and there's a discussion of the new location [starting around here in this issue thread](https://github.com/r-box/boxr/issues/219#issuecomment-1048274673).

AshesITR:
Using `R_user_dir()`, from backports for R < 4.0.0 and from tools on newer R versions.


MichaelChirico:
Could you improve the PR title? make it easier for future searchers to understand the context quickly without needing to click around

NEWS.md
```diff
@@ -168,6 +168,7 @@ function calls. (#850, #851, @renkun-ken)
 * Several linters tightened internal logic to allow for raw strings like `R"( a\string )"` (#1034, @michaelchirico)
 * `object_usage_linter()` correctly detects functions assigned with `=` instead of `<-` (#1081, @michaelchirico)
 * `undesirable_function_linter()` no longer lints undesirable symbols if they are used as names (#1050, @AshesITR)
+* Moved the default lintr cache directory from `~/.R/lintr_cache` to `R_user_dir("lintr", "cache")` (#1062, @AshesITR)
```
MichaelChirico, line 171: this is technically a breaking change, right?

AshesITR: Not really, why would it?

MichaelChirico: it essentially invalidates any existing cache right? unless user copies it over to the new directory

AshesITR: The cache needs to be invalidated anyway due to implementation changes. We touched many default linters I think.

MichaelChirico: makes sense. let's note that in the NEWS item then? and we can fuss more about where it comes in order later

AshesITR: What should we note? "The old directory can be removed without side effects."?

MichaelChirico (approved):

NEWS.md
MichaelChirico: Added a sentence... feel free to merge if you agree it's useful.

PR #1119, opened 2022-05-02
Ignore trailing whitespace in strings by default

### LINKED ISSUE #1045: Ignore trailing whitespace inside strings (optionally?)
Currently this lints:

```
lintr::lint("a <- '\n  \n'", lintr::trailing_whitespace_linter())
# <text>:2:1: style: Trailing whitespace is superfluous.
#   
# ^~
```

But there could be any number of reasons to include trailing whitespace inside a string.

AshesITR:
Fixes #1045

Works by excluding all lines which are within a STR_CONST (`@line1` to `@line2 - 1`).
The last line doesn't need to be excluded because STR_CONST is guaranteed to end in a non-blank, so trailing whitespace after it should lint.

### DIFF
--- a/tests/testthat/test-trailing_whitespace_linter.R
+++ b/tests/testthat/test-trailing_whitespace_linter.R
@@ -49,9 +49,24 @@ test_that("also handles trailing whitespace in string constants", {
   msg <- rex::rex("Trailing whitespace is superfluous.")
 
   expect_lint("blah <- '  \n  \n'", NULL, linter)
+  # Don't exclude past the end of string
+  expect_lint(
+    "blah <- '  \n  \n'  ",
+    list(message = msg, line_number = 3L),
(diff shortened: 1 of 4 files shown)


tests/testthat/test-trailing_whitespace_linter.R
```diff
@@ -43,3 +43,30 @@ test_that("also handles completely empty lines per allow_empty_lines argument",
     trailing_whitespace_linter(allow_empty_lines = TRUE)
   )
 })
+
```
MichaelChirico, line 46: fix the `implicit_integer_linter` hits while we're here in this file? (see above in UI)

AshesITR:
#1126 is causing the r-devel build failure.

MichaelChirico (approved):

PR #1120, opened 2022-05-02
add use_lintr()

### LINKED ISSUE #902: [feature request] generate a ".lintr" file function
New users who are not familiar with the DCF format will have a hard time trying to set up a `.lintr` file.
I think it would be useful to have a function to generate a basic `.lintr` file.

Related to #886.

AshesITR:
Fixes #902 

Currenty, one of two minimal configuration files can be generated.
LMK what you think.


AshesITR:
Based against #1117 for now, so the documentation links won't need updating.

MichaelChirico:
should we add this to usethis? or at least offer to

AshesITR:
They don't want it: https://github.com/r-lib/usethis/issues/1580

R/use_lintr.R
```diff
@@ -0,0 +1,45 @@
+#' Use lintr in your project
+#'
+#' Create a minimal lintr config file as a starting point for customization
+#'
+#' @param path Path to project root, where a `.lintr` file should be created.
+#' If it already exists, an error will be thrown.
```
MichaelChirico, line 6: "it" slightly ambiguous here. "If the .lintr file already exists..."?

R/use_lintr.R
```diff
@@ -0,0 +1,45 @@
...
+#' @param type What kind of configuration to create?
+#'
+#'  * `minimal` creates a minimal lintr config, based off the default linters ([linters_with_defaults()]).
```
MichaelChirico, line 9: "based on" or "based off of"

MichaelChirico: nvm... "based on":

https://books.google.com/ngrams/graph?content=based+on+the%2C+based+off+the%2C+based+off+of+the&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Cbased%20on%20the%3B%2Cc0%3B.t1%3B%2Cbased%20off%20the%3B%2Cc0%3B.t1%3B%2Cbased%20off%20of%20the%3B%2Cc0#t1%3B%2Cbased%20on%20the%3B%2Cc0%3B.t1%3B%2Cbased%20off%20the%3B%2Cc0%3B.t1%3B%2Cbased%20off%20of%20the%3B%2Cc0

MichaelChirico, line 9: maybe "minimal" should be "tidyverse"? e.g. we might add `type = "google"` or `type = "bioconductor"` here eventually, then "tidyverse" would make more sense by contrast

R/use_lintr.R
```diff
@@ -0,0 +1,45 @@
...
+#'  * `full` creates a lintr config using all available linters using [linters_with_tags()].
```
MichaelChirico, line 10: "using" twice... "all available linters in [linters_with_tags()]"?

R/use_lintr.R
```diff
@@ -0,0 +1,45 @@
...
+#' # or try all linters
+#' lintr::use_lintr(type = "full")
+#'
+#' # then
+#' lintr::lint_dir()
+#' }
+use_lintr <- function(path = ".", type = c("minimal", "full")) {
+  config_file <- normalizePath(file.path(path, getOption("lintr.linter_file")), mustWork = FALSE)
+  if (file.exists(config_file)) {
+    stop("File '", config_file, "' already exists.")
```
MichaelChirico, line 28: Maybe not the most helpful error, since we are generating that path from the user's input.

`"Found an existing configuration file at '", config_file, "'"`?

tests/testthat/test-use_lintr.R
```diff
@@ -0,0 +1,26 @@
+test_that("use_lintr works as expected", {
+  tmp <- tempfile()
+  dir.create(tmp)
+  on.exit(unlink(tmp, recursive = TRUE), add = TRUE)
```
MichaelChirico, line 4: handle this with `withr` instead?

MichaelChirico (commented):
thanks! mostly minor stuff

R/use_lintr.R
AshesITR: maybe via?

MichaelChirico: works for me

AshesITR:
The failure on test-coverage is something I saw locally as well, but the message "produced warnings" is super unhelpful.
I couldn't find out what warning was being generated. Any idea?

Failure on devel is related to updates to parser error messages. Will create an issue for that.

AshesITR:
#1126 is causing the devel build failure.

AshesITR:
Found the test-coverage problem: `write.dcf()` introduces line breaks at arbitrary places if `width` is reached.
Thus, it put a line break between `# see` and `vignette("using_lintr")`, causing the latter to be evaluated and produce the warning (vignette not found) during `read_settings()`.

MichaelChirico:
oh, just noticed this merged into a branch? was that intentional @AshesITR 🤔

AshesITR:
> Based against #1117 for now, so the documentation links won't need updating.

😃

Issue #487, opened 2020-04-30
Allowing lone curly braces in some cases


MichaelChirico:
I think the curly brace alone is the most readable in some cases. Particularly I came across this writing a `switch` where the default is multiline:

```
# file: brace.R
switch(x,
  "a" = do_something(x),
  "b" = do_another(x),
  { # default
    do_first(x)
    do_second(x)
  }
)
```

```
lintr::lint('brace.R')
/private/tmp/brace.R:4:3: style: Opening curly braces should never go on their own line and should always be followed by a new line.
  { # default
  ^
```

Is there any option to turn off this failure for this case? This alternative I think is not very readable:

```
switch(x,
  "a" = do_something(x),
  "b" = do_another(x), {
    do_first(x)
    do_second(x)
  }
)
```

There was also a request for exceptions in a case like `with` (though I would still use EOL-`{` in that case personally):

https://github.com/jimhester/lintr/issues/188#issuecomment-275643515

russHyde:
The simplest fix would be to make your .lintr more lenient:
```
linters: with_defaults(
  line_length_linter(100),
  multiple_dots_linter = NULL,
  object_name_linter = NULL,
  camel_case_linter = NULL,
  open_curly_linter = NULL,
  closed_curly_linter(allow_single_line = TRUE),
  object_usage_linter = NULL,
  cyclocomp_linter = NULL)
exclusions: list("inst/profile/general.R" = 1, "inst/profile/shell.R")
```

As an aside, `camel_case_linter` is no longer one of the defaults, so NULL-ing it out of your config doesn't do anything (assuming lintr >= 2.0.0). `object_name_linter` has taken over it's duties. To only admit snake-case or [c|C]amel-case you could use `object_name_linter = object_name_linter(c("snake_case", "camelCase", "CamelCase"))`

MichaelChirico:
this would be the other extreme though right, where that linting rule is simply shut off in all cases?

russHyde:
Yes. To achieve your use-case, I believe we'd have to modify `open_curly_linter` to separate analysis of the preceding / following newlines.

bersbersbers:
Another use case of this is passing a multi-line code block to a function:

```
fun <- function(text, code) {
    message(text)
    code
}

fun(
    "This is very very very long text.",
    {
        message("This is the code.")
        message("It's stupid, but proves my point.")
    }
)
```

Or is there a better way to format this code?

bersbersbers:
[...]

PR #1121, opened 2022-05-02
allow (\n{ in brace_linter

AshesITR:
quick fix to #487 (also allow `(\n{`)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@
 * Rename `semicolon_terminator_linter` to `semicolon_linter` for better consistency. `semicolon_terminator_linter` survives but is marked for deprecation. The new linter also has a new signature, taking arguments `allow_compound` and `allow_trailing` to replace the old single argument `semicolon=`, again for signature consistency with other linters.
 * Combined several curly brace related linters into a new `brace_linter` (#1041, @AshesITR):
   + `closed_curly_linter()`, also allowing `}]` in addition to `})` and `},` as exceptions.
-  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace and also allowing `,` and `%>%` on
+  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace and also allowing `(`, `,` and `%>%` on
     preceding lines as exceptions. (#487, #1028)
   + `paren_brace_linter()`, also linting `if`/`else` and `repeat` with missing whitespace
   + Require `else` to come on the same line as the preceding `}`, if present (#884, @michaelchirico)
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -11,7 +11,7 @@
 * Rename `semicolon_terminator_linter` to `semicolon_linter` for better consistency. `semicolon_terminator_linter` survives but is marked for deprecation. The new linter also has a new signature, taking arguments `allow_compound` and `allow_trailing` to replace the old single argument `semicolon=`, again for signature consistency with other linters.
 * Combined several curly brace related linters into a new `brace_linter` (#1041, @AshesITR):
   + `closed_curly_linter()`, also allowing `}]` in addition to `})` and `},` as exceptions.
-  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace and also allowing `,` and `%>%` on
+  + `open_curly_linter()`, no longer linting unnecessary trailing whitespace and also allowing `(`, `,` and `%>%` on
```
MichaelChirico, line 14: `open_curly_linter` doesn't throw a lint on your test case below on the CRAN version -- IINM this was a regression introduced by the move to XPath in `brace_linter`. Can you confirm that and if so, we don't need a NEWS item right?

AshesITR: Can you verify using the `open_curly_linter()` implementation? It's deprecated but still there and the old code should lint due to `!some_before`...
Or was there another change to open_curly_linter in the mean time?

MichaelChirico: I still don't have a computer set up to test. maybe this weekend.

AshesITR:
```
> packageVersion("lintr")
[1] ‘2.0.1.9000’
> lintr::lint("tryCatch(\n  {\n    print(1)\n  },\n  error = function(err) {\n  }\n)", linters = list(lintr::open_curly_linter()))
Warning: Linter open_curly_linter was deprecated in lintr version 2.0.1.9001. Use brace_linter instead.
<text>:2:3: style: Opening curly braces should never go on their own line and should always be followed by a new line.
  {
  ^
```

AshesITR: I just noticed I tested on the dev version (oops).
CRAN has the same behaviour, though:

```
> packageVersion("lintr")
[1] ‘2.0.1’
> lintr::lint("tryCatch(\n  {\n    print(1)\n  },\n  error = function(err) {\n  }\n)", linters = list(lintr::open_curly_linter()))
<text>:2:3: style: Opening curly braces should never go on their own line and should always be followed by a new line.
  {
  ^
```

MichaelChirico: thanks for confirming

MichaelChirico (approved):

PR #1129, opened 2022-05-09
support R-devel parseError metadata

### LINKED ISSUE #1126: r-devel changed parser error location info
Causes test failures at the moment. See here https://github.com/r-lib/lintr/runs/6339234285?check_suite_focus=true

It seems the new messages generally have `(/path/to/file.R:line:col)` to mark the location.

AshesITR:
Fixes #1126.
One test message had to be shortened to match both old and new R versions, the rest works by using the `parseError` metadata provided by R-devel instead of working with the message strings in `lint_parse_error()`.


AshesITR:
Already merged

PR #1136, opened 2022-05-16
Auto-detect imported objects from packages in `object_usage_linter()`

### LINKED ISSUE #1127: no visible global variables (object-usage-linter) is still a matter
`packageVersion('lintr')`: `‘2.0.1.9000’`
I installed `lintr` via github and I am at commit `4ca7e484`

The following example will yield the object-usage-linter error:

```R
library(tidyverse)

dog = function() {
    a = rnorm(5) %>% sum()
    a
}

```

yields `no visible global function definition for ‘%>%’ object_usage_linter [4, 1]`

`R --version`

```
R version 4.1.2 (2021-11-01) -- "Bird Hippie"
Copyright (C) 2021 The R Foundation for Statistical Computing
Platform: aarch64-apple-darwin20 (64-bit)
```

AshesITR:
Partially fixes #1127.
The example provided in the original issue is more complex because `library(tidyverse)` doesn't export the functions used.
Instead, it attaches other packages as a side-effect. I don't have a good idea how to (without acutally running `library(...)`) detect these.

IIRC, the packages aren't even registered as `Depends`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -173,6 +173,7 @@ function calls. (#850, #851, @renkun-ken)
 * `trailing_whitespace_linter()` ignores trailing whitespace in strings by default. 
   This can be disabled using `allow_in_strings = FALSE` (#1045, @AshesITR)
 * Moved the default lintr cache directory from `~/.R/lintr_cache` to `R_user_dir("lintr", "cache")`. Note that this major version update invalidated the old cache anyway, so it can be safely deleted. (#1062, @AshesITR)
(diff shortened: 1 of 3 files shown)


MichaelChirico:
IMO it's a tidyverse bug that its dependencies aren't so declared, but maybe we're missing something... added a comment to the existing bug requesting clarification:

https://github.com/tidyverse/tidyverse/issues/40#issuecomment-1127999243

MichaelChirico:
LGTM here. I'm surprised this wasn't done previously 😳

MichaelChirico (approved):

PR #1138, opened 2022-05-16
Use partial matching for linter names in exclusion directives

### LINKED ISSUE #872: Remove redundant `_linter` suffix from auto names?

AshesITR:
fixes #872 

- transitioned our internal `# nolint`s to the short-hand notation
- we raise a warning if no linter could be matched by an exclusion
- documented the short-hand notation
- added an example in the `using_lintr` vignette
- added tests
- extended NEWS bullet for #660

### DIFF
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -15,7 +15,7 @@
 #'  3. exclusions parameter, a named list of files with named lists of linters and lines to exclude them on, a named
 #'     list of the files and lines to exclude, or just the filenames if you want to exclude the entire file, or the
 #'     directory names if you want to exclude all files in a directory.
-exclude <- function(lints, exclusions = settings$exclusions, ...) {
+exclude <- function(lints, exclusions = settings$exclusions, linter_names = NULL, ...) {
   if (length(lints) <= 0) {
     return(lints)
   }
@@ -23,7 +23,7 @@ exclude <- function(lints, exclusions = settings$exclusions, ...) {
   df <- as.data.frame(lints)
 
   filenames <- unique(df$filename)
-  source_exclusions <- lapply(filenames, parse_exclusions, ...)
+  source_exclusions <- lapply(filenames, parse_exclusions, linter_names = linter_names, ...)
   names(source_exclusions) <- filenames
 
 
@@ -77,6 +77,7 @@ line_info <- function(line_numbers, type = c("start", "end")) {
 #' `exclude` or `exclude_start` marker.
 #' @param exclude_linter_sep regular expression used to split a linter list into indivdual linter names for exclusion.
 #' @param lines a character vector of the content lines of `file`
+#' @param linter_names Names of active linters
 #'
 #' @return A possibly named list of excluded lines, possibly for specific linters.
 parse_exclusions <- function(file, exclude = settings$exclude,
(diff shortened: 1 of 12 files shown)


R/exclude.R
```diff
@@ -154,13 +157,25 @@ add_excluded_lines <- function(exclusions, excluded_lines, excluded_linters) {
...
   # No match for linter list: Add to global excludes
   if (is.na(linters_string)) {
     exclusions <- add_excluded_lines(exclusions, lines, "")
   } else {
     # Matched a linter list: only add excluded lines for the listed linters.
     excluded_linters <- strsplit(linters_string, exclude_linter_sep)[[1L]]
+    if (!is.null(linter_names)) {
+      idxs <- pmatch(excluded_linters, linter_names, duplicates.ok = TRUE)
+      if (anyNA(idxs)) {
+        bad <- excluded_linters[is.na(idxs)]
```
MichaelChirico, line 170: is.na(idxs) is basically run three times here... save & reuse

tests/testthat/dummy_projects/project/partially_matched_exclusions.R
```diff
@@ -0,0 +1,13 @@
...
+
+# nolint start: s. warn (and lint) because of non-unique identifier
+x <- 42; y <- 2
+# nolint end
+
+# nolint start: bogus_linter. warn because of nonexistent identifier
+
+# nolint end
+
+# nolint: hokus_pokus, bogus.
```
MichaelChirico, line 13: nit: we spell it hocus pocus :)

tests/testthat/test-exclusions.R
```diff
@@ -68,3 +73,22 @@ test_that("it gives the expected error message when there is mismatch between mu
     fixed = TRUE
   )
 })
+
+test_that("partial matching works for exclusions but warns if no linter found", {
+  read_settings(NULL)
+
+  expect_warning(
```
MichaelChirico, line 80: is this really the way to do multiple warning tests in testthat? oh boy 😂

the withCallingHandlers approach is not particularly nicer... maybe we need a helper. i've definitely written `expect_warnings()` before.

otoh, it seems like an intentional design decision that testthat has limited matching exactly one warning at a time:

https://testthat.r-lib.org/reference/expect_error.html

MichaelChirico:
LGTM after small fix to reduce redundancy

tests/testthat/test-exclusions.R
AshesITR: Filed https://github.com/r-lib/testthat/issues/1625 as a suggestion

MichaelChirico (approved):

Issue #1133, opened 2022-05-11
Improve documentation for default_undesirable_functions


MichaelChirico:
Currently I don't see any reasoning behind why the various default undesirables are considered undesirable.

Inspired by this SO Q:

https://stackoverflow.com/questions/65466324/why-are-sapply-and-options-undesirable

AshesITR:
We could document those either in the lint itself (e.g. `use lapply() or vapply() because they are type-stable.`) or document in `?default_undesirable_functions`. WDYT?

MichaelChirico:
I think both would be good. In the lint is better for users, in the docs for developers.

AshesITR:
The message is currently hard-coded to be "Function {fun} is undesirable. As an alternative, {explanation}".
So explaining the reasoning in the lints would require refactoring the generated message.
Any idea for a good "format" for the new messages?

Maybe simply "Function {fun} is undesirable. {explanation}"?

MichaelChirico:
It can be similar to our deprecation note, right? old / new / explanation?

AshesITR:
That would make the old interface (named character vector) impossible to use for customization.

MichaelChirico:
right... in that case your suggestion is the best we can do I think. (I always found the old pattern to restrictive anyway)

AshesITR:
Fixed in #1149

PR #1139, opened 2022-05-16
Better documentation of undesirable functions

AshesITR:
Part of #1133 

Initial draft for feedback on the content.
I've added an `@details` section to `?default_undesirable_functions` that tries to explain why some functions might be undesirable.

NB I think there is a difference between packages (where e.g. `source()` is a no-no) and general projects (where `source()` is the go-to for localized code reuse).

LMK what you think about this.


R/zzz.R
```diff
@@ -44,6 +44,44 @@ default_linters <- modify_defaults(
 #' There is a list for the default elements and another that contains all available elements.
 #' Use [modify_defaults()] to produce a custom list.
 #'
+#' @details
+#' The following functions can be regarded as undesirable:
```
MichaelChirico, line 48: are considered undesirable by default? not sure we currently convey that here

AshesITR: The catch is we have `default_undesirable_functions`, which is a proper subset of `all_undesirable_functions`.
Documentation currently contains remarks on all, not just on the defaults.

R/zzz.R
```diff
@@ -44,6 +44,44 @@ default_linters <- modify_defaults(
...
+#'  * `.libPaths()` permanently modifies the library location. Use `withr::with_libpaths()` for a temporary change
+#'    instead.
+#'  * `library()` modifies the global search path. Use roxygen2's @importFrom statement in packages, or `::` in scripts.
+#'  * `loadNamespace()` doesn't provide an easy way to signal failures. Use the return value of `requireNamespace()`
+#'    instead.
+#'  * `mapply()` isn't type stable. Use `Map()` to guarantee a `list()` is returned and simplify accordingly.
+#'  * `options()` permanently modifies the session options. Use `withr::with_options()` for a temporary change instead.
+#'  * `par()` permanently modifies the graphics device parameters. Use `withr::with_par()` for a temporary change
+#'    instead.
+#'  * `require()` modifies the global search path. Use roxygen2's @importFrom statement in packages, or `::` in scripts.
```
MichaelChirico, line 68: slightly odd for `library()` and `require()` to have the same diagnosis -- isn't part of this to steer towards `library()` in scripts?

R/zzz.R
MichaelChirico: Got it... maybe "The following functions are often regarded as undesirable:" ?

R/zzz.R
```diff
@@ -44,6 +44,44 @@ default_linters <- modify_defaults(
...
+#'  * `return()` is not necessary as the last value of a function is automatically returned.
+#'  * `sapply()` isn't type stable. Use `vapply()` with an appropriate `FUN.VALUE=` argument to obtain type stable
+#'    simplification.
+#'  * `setwd()` modifies the global working directory. Use `withr::with_dir()` for a temporary change instead.
+#'  * `sink()` permanently redirects output. Use `withr::with_sink()` for a temporary redirection instead.
+#'  * `source()` loads code into the global environment unless `local = TRUE` is used, which can cause unexpected
+#'    behaviour.
+#'  * `substring()` was superseded by `substr()`.
```
MichaelChirico, line 76: is that so? i haven't seen that reference before.

here are key differences between the two from what i can tell:

 - `substring` has a default `last` value, `substr` requires `start` and `stop`
 - `substring` vectorizes across `first`/`last` on scalar input: `substring("abcd", 1, 1:3)` vs `substr("abcd", 1, 1:3)`

maybe some other subtle difference too. but they both call the same `.Internal(substr())` pretty quickly.

R/zzz.R
AshesITR: That's true. For scripts, `library()` shouldn't be undesirable and `require()` should be avoided.

R/zzz.R
AshesITR: We even use `attach()` occasionally, to attach some helper functions for interactive use in a `.Rprofile`.

MichaelChirico: `return()` is one that definitely doesn't make sense to me -- I'd be fine with dropping it as undesirable. probably the better approach is to delegate that to the upcoming `explicit_return_linter`, which will use tidyverse style by default (throw a lint on a terminal `return()` statement)

for attach, i agree there's definitely valid use cases, but they are rare (?) compared to the "beginner" mistake of using it to try and manage the search path

AshesITR: Yep, I was suggesting to keep these. I'd still weaken the statement due to the fact that `library()` is ubiquitous in non-package projects.

R/zzz.R
AshesITR: I wasn't able to find any explicit statement other than that the docs state `substring()` is S-compatible (i.e. very old).

AshesITR:
A similar problem is for the three undesirable operators. I reckon we can do this in one PR, WDYT?

R/zzz.R
```diff
@@ -44,6 +44,45 @@ default_linters <- modify_defaults(
 #' There is a list for the default elements and another that contains all available elements.
 #' Use [modify_defaults()] to produce a custom list.
 #'
+#' @details
+#' The following functions are sometimes regarded as undesirable:
+#'
+#'  * `attach()` modifies the global search path. Use roxygen2's @importFrom statement in packages, or `::` in scripts.
```
MichaelChirico, line 50: should these all be links? `[attach()]`, etc?

AshesITR: I thought about that too, wasn't sure because `withr` e.g. is just Suggested and not Imported, therefor not all links are guaranteed to work.

R/zzz.R
```diff
@@ -44,6 +44,45 @@ default_linters <- modify_defaults(
...
+#'  * `browser()` pauses execution when run and is likely a leftover from debugging. It should be removed.
+#'  * `debug()` traps a function and causes execution to pause when that function is run. It should be removed.
+#'  * `debugcall()` works similarly to `debug()`, causing execution to pause. It should be removed.
+#'  * `debugonce()` is only useful for interactive debugging. It should be removed.
+#'  * `detach()` modifies the global search path. Detaching environments from the search path is rarely necessary in
+#'    production code.
+#'  * `ifelse()` isn't type stable. Use an `if`/`else` block for scalar logic, or use
+#'    `dplyr::if_else()`/`data.table::fifelse()` for type stable vectorized logic.
+#'  * `.libPaths()` permanently modifies the library location. Use `withr::with_libpaths()` for a temporary change
```
MichaelChirico, line 59: are we allowed to use `[withr::with_libpaths()]`? i can't tell from WRE whether Suggests are allowed as documentation links

R/zzz.R
```diff
@@ -44,6 +44,45 @@ default_linters <- modify_defaults(
...
+#'  * `library()` modifies the global search path. Use roxygen2's @importFrom statement in packages, or `::` in scripts.
+#'  * `loadNamespace()` doesn't provide an easy way to signal failures. Use the return value of `requireNamespace()`
+#'    instead.
+#'  * `mapply()` isn't type stable. Use `Map()` to guarantee a `list()` is returned and simplify accordingly.
+#'  * `options()` permanently modifies the session options. Use `withr::with_options()` for a temporary change instead.
+#'  * `par()` permanently modifies the graphics device parameters. Use `withr::with_par()` for a temporary change
+#'    instead.
+#'  * `require()` modifies the global search path. Use roxygen2's @importFrom statement in packages, and `library()`
+#'    or `::` in scripts.
+#'  * `return()` is not necessary as the last value of a function is automatically returned. # TODO remove this
```
MichaelChirico, line 70: this `TODO` will also show up in the documentation, is that intentional?

R/zzz.R
MichaelChirico: I _think_ we're allowed to link Suggests. Since worst case the link is just dead -- doesn't affect the code running or anything. But I can't see any documentation as of now.

R/zzz.R
AshesITR: Yes, in the sense that that TODO should be completed before release.

MichaelChirico: OK, filed #1146

R/zzz.R
MichaelChirico: I would leave those unlinked (so just `\code{}` is fine), and add links for everything in our dependency net

AshesITR:
Locally, I could get `devtools::check()` to pass with links to all functions. Let's see if GHA lets us pass as well.

MichaelChirico:
Perfect. LGTM pending the GHA passes.

MichaelChirico (approved):

AshesITR:
Do we want to edit the lint messages here as well, or do that in tandem with #1146?

MichaelChirico:
let's do it in another PR since this already has a bunch of history

PR #1140, opened 2022-05-17
xml_nodes_to_lint vectorized & exported

### LINKED ISSUE #1124: Export `xml_nodes_to_lint()`?
Used in most linters by now, because it's a very useful helper for XPath-based linting.
Since we already export `with_id` and `ids_with_token` as helpers for the parse data based approach, I think we could also export `xml_nodes_to_lint()`.

Also thinking about the function name: Shouldn't this be either vectorized (on nodesets or lists of nodes) or singular (`xml_node_to_lint()`)?

cc @MichaelChirico

MichaelChirico:
Closes #1124

Also dropped `global=` as an argument -- IIRC that's a relic of a time when the `names()` of the `source_expression` couldn't reliably be used to distinguish whether the expression was at the expression or file level. Now we check for expression level (the most common case) and if unavailable, use the global object.


R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
...
+#'
+#' Convenience function for converting nodes matched by XPath-based
+#'   linter logic into a [Lint()] object to return.
+#'
+#' @inheritParams Lint
+#' @param xml An `xml_node` object (to generate one `Lint`) or an
+#'   `xml_nodeset` object (to generate several `Lint`s), e.g. as returned by
+#'   [xml2::xml_find_all()] or [xml2::xml_find_first()].
+#' @param source_expression A source expression object, e.g. as
+#'   returned by [get_source_expressions()].
```
AshesITR, line 24: , or passed to the linting function by [lint()]?

R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
...
+#' @param lint_message The message to be included as the `message`
+#'   to the `Lint` object. If `lint_message` is a `function`,
+#'   this function is first applied to `xml` (so it should be a
+#'   function taking an `xml_node` as input and must produce a
+#'   length-1 character as output).
+#' @param offset Integer, default 0. The amount by which to offset the
```
AshesITR, line 30: Currently, the implementation only offsets col1.
That's not very intuitive IMO.

WDYT about `expand = 0L` instead, recycled to length 2 and used as `ranges = c(col1 - expand[1L], col2 + expand[2L])`?
If I remember correctly, it's only used once in out codebase anyway.

R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
...
 xml_nodes_to_lint <- function(xml, source_expression, lint_message,
                               type = c("style", "warning", "error"),
-                              offset = 0L,
-                              global = FALSE) {
+                              offset = 0L) {
+  if (length(xml) == 0L) {
+    return(list())
+  }
+  if (inherits(xml, "xml_nodeset")) {
+    return(lapply(xml, xml_nodes_to_lint, source_expression, lint_message, type, offset))
```
AshesITR, line 41: Maybe class it as `Lints` for nicer printing?

R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
...
+#' @param lint_message The message to be included as the `message`
+#'   to the `Lint` object. If `lint_message` is a `function`,
+#'   this function is first applied to `xml` (so it should be a
+#'   function taking an `xml_node` as input and must produce a
+#'   length-1 character as output).
+#' @param offset Integer, default 0. The amount by which to offset the
+#'   `col1` and `col2` values taken from `xml` when producing
+#'   the `ranges` value in the `Lint` object.
+#' @export
 xml_nodes_to_lint <- function(xml, source_expression, lint_message,
```
AshesITR, line 34: Should be plural now.

R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
...
+#'   returned by [get_source_expressions()].
+#' @param lint_message The message to be included as the `message`
+#'   to the `Lint` object. If `lint_message` is a `function`,
+#'   this function is first applied to `xml` (so it should be a
+#'   function taking an `xml_node` as input and must produce a
+#'   length-1 character as output).
+#' @param offset Integer, default 0. The amount by which to offset the
+#'   `col1` and `col2` values taken from `xml` when producing
+#'   the `ranges` value in the `Lint` object.
+#' @export
```
AshesITR, line 33: We should document the `@return` since it can be one of `list<Lint>` or `Lint`.

R/xp_utils.R
```diff
@@ -11,32 +11,59 @@ xp_text_in_table <- function(table) {
   return(paste0("text() = ", table, collapse = " or "))
 }
 
-# convert an XML match into a Lint
+#' Convert an XML match into a Lint
```
AshesITR, line 14: XML node / nodeset?

R/xp_utils.R
MichaelChirico: doesn't `lint.R` handle that?

IIUC this matches the behavior at HEAD of `lapply(expr, xml_nodes_to_lint)`, right?

R/xp_utils.R
MichaelChirico: I also found this parameter a bit confusing and yes, it's only used for `T_and_F_symbol_linter`... my original thought was to remove it.

`expand` isn't sticking for me but the idea sounds good... `range_expand` maybe? `adjust_range`? `range_offset`?

AshesITR: range_offset sounds good. T_and_F would use `c(-1L, 0L)` then.

R/xp_utils.R
AshesITR: Yes, but if we export it, I see how it might be used for interactive development of the XPaths, where the nicer print() might be cool.

R/xp_utils.R
AshesITR: Sorry, I mean the function name, since it now produces a list of lints.

R/xp_utils.R
MichaelChirico: actually the behavior at HEAD for `T_and_F_symbol_linter` looks wrong to me:

```r
lintr::lint("for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}\n", lintr::T_and_F_symbol_linter())
<text>:1:26: style: Use TRUE instead of the symbol T.
for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}
                        ~^
<text>:1:35: style: Use FALSE instead of the symbol F.
for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}
```

Using `range_offset = c(-1L, 0L)` looks right:

```r
lintr::lint("for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}\n", lintr::T_and_F_symbol_linter())
<text>:1:25: style: Use TRUE instead of the symbol T.
for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}
                        ^ 
<text>:1:34: style: Use FALSE instead of the symbol F.
for (i in 1:10) {x <- c(T, TRUE, F, FALSE)}
```

so I'll continue but need to edit some tests

MichaelChirico: unless the intention is to always have `~^`? but it's weird for the arrow to point at the wrong place

MichaelChirico: Are we sure we need `range_offset` actually? the printing looks fine to me with `range_offset=0`

AshesITR: Hmm maybe `xmlparsedata` changed its output?

MichaelChirico: A bit hard to tell since none of the tests recorded the expected `ranges`:

https://github.com/r-lib/lintr/blob/1ae01287f9c97536e097715beb08e88a25632067/tests/testthat/test-T_and_F_symbol_linter.R#L11-L22

MichaelChirico: just installed `xmlparsedata` 1.0.0 and it has the same output as the CRAN version & current devel

AshesITR: https://github.com/r-lib/lintr/pull/759

I dug it out, the reason for `offset` was to preserve the original location info.
Can't really tell which is best usability wise, but I'm open to changing the locations if the usability is more or less the same.

One thing that comes to mind is if the source location would be more convenient for a jump-to-lint action (i.e. clicking on the lint in the RStudio Markers pane or equivalent)

MichaelChirico: Oh, good call. Indeed, without an offset we are landed to the left of `T` (in `TRUE`), whereas at HEAD we're landed to the right. But this is governed by `column_number`, not `ranges`.

So should we make it so `offset` nudges only `column_number`, but not `ranges`?

AshesITR: Not sure if it isn't required for column_number to be within the ranges.
Feel free to test.
Otherwise, maybe introduce `column_offset` and `range_offset`?

R/T_and_F_symbol_linter.R
```diff
@@ -55,14 +55,8 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
...
     c(
-      lapply(
-        bad_exprs, make_lint,
-        fmt = "Use %s instead of the symbol %s."
-      ),
-      lapply(
-        bad_assigns, make_lint,
-        fmt = "Don't use %2$s as a variable name, as it can break code relying on %2$s being %1$s."
-      )
+      make_lint(bad_exprs, "Use %s instead of the symbol %s."),
```
AshesITR, line 58: nit: should be called make_lints now.

AshesITR (approved):
LGTM, nice work!

AshesITR (approved):

R/xp_utils.R
MichaelChirico: Well, turns out we definitely don't enforce `range[1] <= column_number <= range[2]` elsewhere :)

https://github.com/r-lib/lintr/blob/5328347a2c44413f225da516c824115b9abac460/R/implicit_integer_linter.R#L21-L26

PR #1142, opened 2022-05-17
pass at modernizing creating-linters vignette

### LINKED ISSUE #1122: Update vignettes
Make sure the vignettes `using_lintr.Rmd` and `creating_linters.Rmd` reflect the current APIs and best practices.

MichaelChirico:
Closes #1122 

Probably needs a few more passes to get it right, but this should be a good start.


vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
 ```r
-Linter(function(source_file) {
+Linter(function(source_expression) {
 ```
-This is the closure. It will be called on the `source_file` variable that
+
+This is the closure. It will be called on the `source_expression` variable that
 contains the top level expressions in the file to be linted. The call to
-`Linter()` simply gives this closure the class 'linter'.
+`Linter()` gives this closure the class 'linter' and guesses its name.
```
AshesITR, line 86: Not 100% sure if the "and guesses its name" is clear.

vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
+as they are tokenized from the `R` parser. These tokens are obtained
+from `parse()` and `utils::getParseData()` calls done prior to calling the new linter.
 `getParseData()` returns a `data.frame` with information from the source parse
 tree of the file being linted. A list of tokens is available from
-[r-source/src/main/gram.y](https://github.com/wch/r-source/blob/ff1bca2f21aba271d428474f00494eece5c36dd3/src/main/gram.y#L293-L307).
+[r-source/src/main/gram.y](https://github.com/r-devel/r-svn/blob/master/src/main/gram.y#L395-L412).
+
+`source_file$xml_parsed_content` uses `xmlparsedata::xml_parse_data()` to convert the `getParseData()` output into an XML tree,
+which enables writing linter logic in [XPath](https://www.w3schools.com/xml/xpath_intro.asp), a powerful language
+for expressing paths within the nested XML data structure. Most linters in `lintr` are built using XPath because it is
```
AshesITR, line 105: WDYT about "...a powerful language for computation on the abstract syntax tree / parse tree"?

vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
+Here, we return early if `source_expression` is not the expression-level object. `get_source_expression()` returns an
+object that parses the input file in two ways -- once is done expression-by-expression, the other contains all of the
+expressions in the file. This is done to facilitate caching. Suppose your package has a long source file (e.g., 100s of expressions) --
+rather than run linters on every expression every time the file is updated, when caching is activated `lintr` will
+only run the linter again on expressions that have changed.
+
+Therefore, it is preferable to write expression-level linters whenever possible. Two types of exceptions observed in `lintr` are
+(1) when several or all expressions are _required_ to ensure the linter logic applies (e.g., `conjunct_test_linter` looks for
+consecutive calls to `stopifnot()`, which will typically appear on consecutive expressions) or (2) when the linter logic
+is very simple & fast to compute, so that the overhead of re-running the linter is low (e.g., `single_quotes_linter`).
```
AshesITR, line 123: Perhaps add a paragraph explaining the purpose of `require_xml = TRUE` and note that it's only required if XPath is used instead of parsed_content directly?

vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
+
+Here is the core linter logic. `source_expression$xml_parsed_content` is copied to a local variable
+(this is not strictly necessary but facilitates debugging). `xpath` is an XPath expression
+for expressions matching the discouraged usage. `bad_expr` is the result of executing this logic
+on the expression's XML representation.
+
+Let's examine the XPath a bit more closely:
+
+```xpath
+//expr                  # global search for 'expr' nodes (R expressions), at any nesting level
```
AshesITR, line 140: "global (//)" for even more clarity?

vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
+  *                     # match any node
+  [1]                   # match the first such node
+  [self::SYMBOL]        # match if the current node is a 'SYMBOL' (i.e., a 'name' in R)
+]                       #
+```
+
+Taken together, that means we want to match `expr` nodes preceded by the `%>%` infix operator whose first child node is a `name`.
+That maps pretty closely to the description of what the `pipe_call_linter` is looking for, but there is subtlety in mapping
+between the R code you're used to and how they show up in the XML representation. `expr` nodes in particular take some practice
+to get accustomed to -- use the plentiful XPath-based linters in `lintr` as a guide to get extra practice.
```
AshesITR, line 156: Maybe also reference the excellent w3schools tutorials here again? Also worth noting that `xml2` implements XPath 1.0.

vignettes/creating_linters.Rmd
```diff
@@ -11,113 +11,186 @@ vignette: >
...
+  type = "warning"
 )
 ```
-Lastly, build a `lint` object which describes the issue.  See `?Lint` for a
-description of the arguments. You do not have to return a lint for every
-iteration of your loop.  Feel free to return `NULL` or `list()` for tokens which
-do not need to be linted. You can even return a `list` of `lint` objects if
-more than one lint was found.
+
+Finally, we pass the matching XML node to `xml_nodes_to_lint()`, which returns `Lint` objects corresponding to any "bad" usages
```
AshesITR, line 167: "node(s)" or "node set"?

vignettes/creating_linters.Rmd
```diff
@@ -157,29 +226,34 @@ such as the type, line number, column number, e.g.
...
 
 ## Adding your linter to the default_linters ##
-If your linter is non-project specific you can add it to `default_linters`.
-This object is created in the file `zzz.R`.  The name ensures that it will always run after all
-the linters are defined.  Simply add your linter name to the `default_linters`
+
+If your linter is not project-specific you can add it to `default_linters`.
+This object is created in the file `zzz.R` (this name ensures that it will always run after all
+the linters are defined). Simply add your linter name to the `default_linters`
 list before the `NULL` at the end.
```
AshesITR, line 253: Probably should say that the default linters are meant for compliance to the tidyverse style, and that adding a default linter requires introducing a lint to default_linter_testcode.R.

Alternatively, we could remove this section since that's very {lintr} specific?

AshesITR (commented):
Great work so far!

vignettes/creating_linters.Rmd
MichaelChirico: pushed into parenthetical with doc reference

MichaelChirico:
Will return to incorporate more feedback later.

vignettes/creating_linters.Rmd
MichaelChirico: actually, let's just get rid of the need for this altogether...

AshesITR: What do you mean? It was a reaction to `xmlparsedata` bugs in the past, where `parsed_content` was still fine but the XML was invalid.
Or do you think we should no longer run parsed_content based linters if the XML is invalid?

MichaelChirico: Let's discuss in #1147

vignettes/creating_linters.Rmd
MichaelChirico: I also considered whether we might want to drop it... I think on balance it's better to keep, for now. maybe eventually it belongs in a "CONTRIBUTING" document, though.

AshesITR: Sounds good.

vignettes/creating_linters.Rmd
MichaelChirico: OK it looks like #1147 will land. I'll adjust the wording here quickly and then we can proceed.

vignettes/creating_linters.Rmd
```diff
@@ -157,29 +229,37 @@ such as the type, line number, column number, e.g.
...
   list(
     list(line_number = 1, column_number = 5),
     list(line_number = 1, column_number = 13),
   )
-  assignment_linter())
+  assignment_linter()
+)
 ```
 
 It is always better to write too many tests rather than too few.
```
AshesITR, line 250: Maybe insert a heading "Contributing to {lintr} directly" here?

AshesITR:
Very nice. I think we're basically done.

vignettes/creating_linters.Rmd
MichaelChirico: I kept most of the testing section as is, and pulled out the lintr-specific part into its own new section at the end.

vignettes/creating_linters.Rmd
```diff
@@ -157,29 +221,51 @@ such as the type, line number, column number, e.g.
...
+`filter` argument to `devtools::test()`.
+
+Linter tests should be put in the
+[tests/testthat/](https://github.com/r-lib/lintr/tree/master/tests/testthat)
+folder.  The test filename should be the linter name prefixed by `test-`, e.g.
+`test-pipe_continuation_linter.R`.
+
+### Adding your linter to the default_linters ##
+
+_NB: this section and the next are specific to adding new linters to {linter}_
```
AshesITR, line 259: This remark is no longer necessary.

vignettes/creating_linters.Rmd
```diff
@@ -157,29 +221,51 @@ such as the type, line number, column number, e.g.
...
-## Adding your linter to the default_linters ##
-If your linter is non-project specific you can add it to `default_linters`.
-This object is created in the file `zzz.R`.  The name ensures that it will always run after all
-the linters are defined.  Simply add your linter name to the `default_linters`
-list before the `NULL` at the end.
+## Contributing to `{lintr}`
+
+### More details about writing tests for new `{lintr}` linters
+
+The `{lintr}` package uses [testthat](https://github.com/hadley/testthat) for
```
AshesITR, line 247: Isn't it `r-lib/testthat`?

MichaelChirico: good catch!

PR #1148, opened 2022-05-19
Fixed regex refactor

MichaelChirico:


MichaelChirico:
ugh. bad rebasing. abort.

PR #1153, opened 2022-05-19
Add xml_missing() when parsing fails

### LINKED ISSUE #1147: Always populate xml_parsed_content and full_xml_parsed_content (perhaps with xml_missing)
Just confirmed that

```
xml2::xml_find_all(xml2::xml_missing, "//a")
# {xml_nodeset (0)}
```

so most of our linters will just work with the replacement.

Unfortunately

```
xml2::xml_find_first(xml2::xml_missing, "//a")
```

is an error, so we may have to guard the code against attempting `xml_find_first()` on missing documents.

MichaelChirico:
Closes #1147

Passing all tests as of now. What other tests would be helpful?

### DIFF
--- a/R/commas_linter.R
+++ b/R/commas_linter.R
@@ -15,6 +15,7 @@ commas_linter <- function() {
 
     res <- re_matches(source_expression$lines, re, global = TRUE, locations = TRUE)
 
+    # no match --> return list(data.frame(NA, NA))
     if (length(res) == 1L && nrow(res[[1L]]) == 1L && all(is.na(res[[1L]]))) {
       return(list())
     }
--- a/R/commas_linter.R
+++ b/R/commas_linter.R
@@ -15,6 +15,10 @@ commas_linter <- function() {
 
     res <- re_matches(source_expression$lines, re, global = TRUE, locations = TRUE)
 
+    if (length(res) == 1L && nrow(res[[1L]]) == 1L && all(is.na(res[[1L]]))) {
+      return(list())
+    }
+
     lapply(seq_along(res), function(id) {
       line_number <- names(source_expression$lines)[id]
 
--- a/man/default_undesirable_functions.Rd
+++ b/man/default_undesirable_functions.Rd
@@ -53,7 +53,6 @@ instead.
 instead.
 \item \code{\link[=require]{require()}} modifies the global search path. Use roxygen2's @importFrom statement in packages, and \code{\link[=library]{library()}}
 or \code{::} in scripts.
-\item \code{\link[=return]{return()}} is not necessary as the last value of a function is automatically returned. # TODO remove this
 \item \code{\link[=sapply]{sapply()}} isn't type stable. Use \code{\link[=vapply]{vapply()}} with an appropriate \verb{FUN.VALUE=} argument to obtain type stable
 simplification.
 \item \code{\link[=setwd]{setwd()}} modifies the global working directory. Use \code{\link[withr:with_dir]{withr::with_dir()}} for a temporary change instead.
--- a/tests/testthat/test-get_source_expressions.R
+++ b/tests/testthat/test-get_source_expressions.R
@@ -200,8 +200,23 @@ test_that("linters pass with xml_missing() content", {
   #   evidence that things are working as intended.
   bad_source <- withr::local_tempfile()
   writeLines("a = 1\nb = 2", bad_source)
-  source_expression <- get_source_expressions(bad_source)
+  expressions <- get_source_expressions(bad_source)$expressions
+
+  # "zap" the xml_parsed_content to be xml_missing -- this gets
+  #   around the issue of creating a file that fails to parse now,
+  #   but later fails in a different way -> xml not missing.
+  for (ii in seq_along(expressions)) {
+    if ("xml_parsed_content" %in% names(expressions[[ii]])) {
+      expressions[[ii]]$xml_parsed_content <- xml2::xml_missing()
+    } else {
+      expressions[[ii]]$full_xml_parsed_content <- xml2::xml_missing()
+    }
+  }
+
+  # test all linters at least pass (no error) when finding xml_missing
   for (linter in linters_with_tags(tag = NULL)) {
-    expect_error(linter(source_expression), NA)
+    for (expr in expressions) {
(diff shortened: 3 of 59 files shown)


MichaelChirico:
FYI the meat of the change is in R/is_lint_level.R and R/utils.R

MichaelChirico:
Ran this locally:

```
writeLines("\\", tmp<-tempfile())
lintr::lint(tmp, linters_with_tags(tags=NULL))
l=lintr::lint(tmp, linters_with_tags(tags=NULL))
dput(l)
# structure(list(structure(list(filename = "/tmp/RtmpxgyDZs/file385fe3207fa68a", 
#     line_number = 1L, column_number = 1L, type = "error", message = "unexpected input", 
#     line = "\\", ranges = NULL, linter = NA_character_), class = "lint")), class = "lints")
```

(learned that not _all_ syntax errors lead to `xml_missing()`, but a file with just `\` does)

man/default_undesirable_functions.Rd
```diff
@@ -53,7 +53,6 @@ instead.
 instead.
 \item \code{\link[=require]{require()}} modifies the global search path. Use roxygen2's @importFrom statement in packages, and \code{\link[=library]{library()}}
 or \code{::} in scripts.
-\item \code{\link[=return]{return()}} is not necessary as the last value of a function is automatically returned. # TODO remove this
```
MichaelChirico, line 56: missed roxygenize in #1149

AshesITR:
It think that test is all we need.
NB that the current behaviour might be fixed so we should manually zap the XML data from an otherwise good file and expect lint to be silent.

MichaelChirico:
PTAL. It's a bit wonky but I think it'll suffice. I wasn't sure which file to add this test... LMK if you've a better idea.

tests/testthat/test-get_source_expressions.R
```diff
@@ -192,3 +192,31 @@ test_that("1- or 2-width octal expressions give the right STR_CONST values", {
...
+      expressions[[ii]]$xml_parsed_content <- xml2::xml_missing()
+    } else {
+      expressions[[ii]]$full_xml_parsed_content <- xml2::xml_missing()
+    }
+  }
+
+  # test all linters at least pass (no error) when finding xml_missing
+  for (linter in linters_with_tags(tag = NULL)) {
+    for (expr in expressions) {
+      expect_error(linter(expr), NA)
```
AshesITR, line 219: Wouldn't we want to tighten this to `expect_silent()` at least for our own linters?
Also, I can imagine the failure message would be awful to debug. Is this a usecase for {patrick}?

MichaelChirico: `expect_silent()` doesn't quite capture it (since the `a = 1` didn't throw a failure for `assignment_linter()`).

`expect_length(., 0L)` fits the bill better, but because we're using a bit of an awkward interface to lint, we have to use the workaround to #1160 -- `nrow(as.data.frame())`.

R/commas_linter.R
```diff
@@ -15,6 +15,10 @@ commas_linter <- function() {
 
     res <- re_matches(source_expression$lines, re, global = TRUE, locations = TRUE)
 
+    if (length(res) == 1L && nrow(res[[1L]]) == 1L && all(is.na(res[[1L]]))) {
```
MichaelChirico, line 18: `commas_linter()` was also affected by a similar issue to #1160; the fix here is easy

tests/testthat/test-get_source_expressions.R
AshesITR: Oh I think we need to check both, i.e. no output and no lint, then? I'd say we should make our linters consistently return `list()` and the use the simpler test.

MichaelChirico: from typical user perspective it's the same because `flatten_lints()` gives the same output... so i think we're better off leaving that part as a separate fix in #1160.

MichaelChirico: I think we want `lint()` to be silent, i don't think that works with the current test setup.

AshesITR: > from typical user perspective it's the same because `flatten_lints()` gives the same output... so i think we're better off leaving that part as a separate fix in #1160.

I agree it should be a separate PR, just think we should fix it before merging this one so we get a nicer test from the start.

AshesITR: > I think we want `lint()` to be silent, i don't think that works with the current test setup.

I'm just hoping to catch any stray warnings that may or may not be produced by linters when they encounter `xml_missing()` for now.

MichaelChirico: I think we want expect_warning(., NA) then

MichaelChirico: > just think we should fix it before merging this one so we get a nicer test from the start.

I think it's fine, the test is only a little big uglier for now. would agree if it was on the R side.

the main thing is I think the fix will take some effort. don't see a need to hold up this PR (the last code change of the release for now) while we fix internals

AshesITR: Ah thanks for pointing this out. I didn't realize it was that much work.

Is `expect_length(flatten_lints(...), 0L)` an option?

MichaelChirico: flatten_lints not exported right? it'll work but I'm usually hesitant to use unexported stuff. here's probably an ok case as an expedient... is it that much better than nrow(as.df())?

MichaelChirico: > I didn't realize it was that much work.

yea i think it needs a refactor.

current structure is `lapply(foo, function(x) if (bar) then_lint)`. we have to break the inner `if` outside the `lapply` IINM.

AshesITR: It does seem more concise to me, and even more so with a wrapping `expect_warning(..., NA)`...

Fine with either approach as long as we test for warning-free and lint-free results though.

MichaelChirico: OK, I think it should be ready to go

AshesITR (approved):

MichaelChirico:
Should this get a NEWS item? can add to #833 if so...

PR #1159, opened 2022-05-20
explain undesirable operators in Rd and lint message

AshesITR:
Sister to #1133 for `all_undesirable_operators`.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -137,35 +137,34 @@ all_undesirable_functions <- modify_defaults(
 )
 
 #' @rdname default_undesirable_functions
+#' @format
 #' @export
-default_undesirable_functions <- do.call(modify_defaults, c(
-  list(default = list()),
-  all_undesirable_functions[c(
-    "attach",
-    "browser",
-    "debug",
-    "debugcall",
-    "debugonce",
-    "detach",
-    ".libPaths",
-    "library",
-    "mapply",
-    "options",
-    "par",
-    "require",
-    "sapply",
-    "setwd",
-    "sink",
-    "source",
-    "Sys.setenv",
-    "Sys.setlocale",
-    "trace",
-    "undebug",
-    "untrace"
-  )]
-))
+default_undesirable_functions <- all_undesirable_functions[names(all_undesirable_functions) %in% c(
+  "attach",
+  "browser",
+  "debug",
+  "debugcall",
+  "debugonce",
+  "detach",
+  ".libPaths",
+  "library",
+  "mapply",
+  "options",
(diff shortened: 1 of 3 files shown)


AshesITR:
While I was at it, I cleaned up the Rd by adding empty `@format` tags to all child objects and removed the unnecessary `do.call()` constructs.

R/zzz.R
```diff
@@ -82,6 +82,16 @@ default_linters <- modify_defaults(
...
 #'  * [untrace()] is only useful for interactive debugging with [trace()]. It should be removed.
 #'
+#' The following operators are sometimes regarded as undesirable:
+#'
+#'  * [`:::`][base::ns-dblcolon] accesses non-exported functions inside packages. Code relying on these is likely to
+#'    break in future versions of the package because the functions are not part of the public interface and may be
+#'    changed or removed by the maintainers without notice.
+#'    Use public functions via `::` instead.
+#'  * [`<<-`][base::assignOps] and `->>` assign variables outside of the current environment, breaking functional
+#'    encapsulation by causing side-effects. This makes code harder to debug and reason about.
```
MichaelChirico, line 92: to me it's not the side effects _per se_, but the sometimes hard-to-predict behavior of cascading assign. e.g. a basically OK usage of `<<-` is

```
warning <- NULL
something <- tryCatch(
  do_something(),
  warning = function(cond) {
    warning <<- cond$message
    return(2)
  }
)
```

where the separation in the code between `warning` and the `<<-` assignment to it is not much, it's more readable.

So our recommendation is focused on assigning to a specific environments (don't need `new.env()` if such an environment already exists). Also the part about `new.env()` needs more nuance -- `new.env()` has to be defined in calling scope, i.e., replacing `x <<- 2` with `e <- new.env(); e$x <- 2` is not likely to work.

AshesITR: Regarding the nuance I hope the "mutable state" implies that the env needs to be defined in whatever scope needs mutable state.
We could remove the new.env() but I've experienced that new R developers rarely have a firm understanding of environments / frames and wouldn't know what to do to get an "environment".

I like the advice to explicitly specify the target environment. Any suggestions for a concise wording?

MichaelChirico: How about

```
* [`<<-`][base::assignOps] and `->>` assign outside the current environment in a way that can be hard to reason through. Prefer fully-encapsulated functions wherever possible, or, if necessary, assign to a specific environment with [assign()]. Recall that you can create an environment at the desired scope with [new.env()].

MichaelChirico (approved):

Issue #1160, opened 2022-05-20
implicit_integer_linter return object is inconsistent with other linters for NULL lint


MichaelChirico:
```
writeLines("1L", tmp <- tempfile())
expr <- lintr::get_source_expressions(tmp)
lintr::implicit_integer_linter()(expr$expressions[[1L]])
# list(NULL)
```

whereas most linters would just return `lint()`.

`flatten_lints()` handles this so it's not really distinguishable in regular usage, but it's making the test in #1153 a bit more difficult because `length(list(NULL))` is 1 whereas `length(list())` is 0.

MichaelChirico:
Other affected linters:

 - `make_linter_from_regex()` --> `no_tab_linter()`
 - `spaces_inside_linter()`
 - `undesirable_operator_linter()`

AshesITR:
The `lapply(ids_with_token...)` idiom could be replaced by `compact(lapply(ids_with_token...))` to fix this, right?

MichaelChirico:
yes, but otoh I thought it might make sense to take the chance to refactor these linters while we're at it

PR #1169, opened 2022-05-22
refactor spaces_inside_linter to use xpath

MichaelChirico:
Part of #1160 

@AshesITR PTAL -- our column numbers are off again here.

`match_after_end` will work for the LHS case (`[` or `(`) but not the RHS case.

Two options:

 1. Refactor `xml_nodes_to_lints()` again -- maybe `match_after_end` ➡️ `match_column = c("center", "after", "before")`, where `"before"` works for the `]`/`)` cases here?
 2. Just don't use `xml_nodes_to_lints()` here
 3. other ideas?


AshesITR:
Is it possible to solve with the current implementation of `xml_nodes_to_lints()` if the two XPaths are split into separate nodesets?
Also, wondering what column results in optimal usability. If the new columns are better, I'm also open to changing the expected results.

R/spaces_inside_linter.R
```diff
@@ -1,7 +1,8 @@
 #' Spaces inside linter
 #'
-#' Check that parentheses and square brackets do not have spaces directly inside them, i.e., directly following an
-#' opening delimiter or directly preceding a closing delimiter.
+#' Check that parentheses and square brackets do not have spaces directly
+#'   inside them, i.e., directly following an opening delimiter or directly
+#'   preceding a closing delimiter.
```
AshesITR, line 5: Is there a reason for adhering to line length < 120?

MichaelChirico: my emacs not wide enough and I'm too noob to figure out how to adjust 😢

but more generally I think of our 120 as an "if needed for readability feel free to spill over 80" guideline

AshesITR:
Okay, I pondered this for a while, with two observations:

1. the old locations are good, they shouldn't change
2. `xml_nodes_to_lints()` is probably not useful here.

By the way, I have issues with the `print.lints()` method on this branch and maybe on others:
```
> lint(text = "a( 1 )")
<text>:1:2: style: Do not place spaces around code in parentheses or square brackets.
NA
 ^
<text>:1:6: style: Do not place spaces around code in parentheses or square brackets.
NA
     ^

```

MichaelChirico:
> xml_nodes_to_lints() is probably not useful here.

that means to abandon the xpath approach, or to roll our own conversion from xml match --> Lint?

AshesITR:
I'd use a custom `lapply()`.
How to the two versions compare performance-wise?
If the old implementation is faster, no need to switch IMO. It would then be better to just `compact()` the linters output so we get the desired consistency.

MichaelChirico:
I'll test performance out, but i also think the xpath version is a fair amount more readable.

AshesITR:
I agree to that.

MichaelChirico:
(hopefully the adjustments needed to avoid `xml_nodes_to_lints()` don't tank the readability 😄)

R/spaces_inside_linter.R
```diff
@@ -10,59 +11,31 @@
...
+        and not(preceding-sibling::*[1][self::OP-COMMA])
+        and @start != preceding-sibling::*[1]/@end + 1
+        and @line1 = preceding-sibling::*[1]/@line2
+      ) or (
+        (self::OP-LEFT-BRACKET or self::OP-LEFT-PAREN)
+        and not(following-sibling::*[1][self::COMMENT])
+        and @end != following-sibling::*[1]/@start - 1
+        and @line1 = following-sibling::*[1]/@line2
+      )
+    ]"
```
AshesITR, line 32: What about a segmented XPath (in the factory code to avoid performance hit)

```r
xp_closing_cond <- "
  not(preceding-sibling::*[1][self::OP-COMMA])
  and @start != preceding-sibling::*[1]/@end + 1
  and @line1 = preceding-sibling::*[1]/@line2
"

xp_opening_cond <- "
  not(following-sibling::*[1][self::COMMENT])
  and @end != following-sibling::*[1]/@start - 1
  and @line1 = following-sibling::*[1]/@line2
"

xpath <- glue::glue("
  //OP-RIGHT-BRACKET[{xp_closing_cond}] | 
  //OP-RIGHT-PAREN[{xp_closing_cond}] |
  //OP-LEFT-BRACKET[{xp_opening_cond}] |
  //OP-LEFT-PAREN[{xp_opening_cond}]
")
```

R/spaces_inside_linter.R
AshesITR: Ah I always use the full 120 columns because my editor windows are pretty wide 😄  [SO Seems to have something for you](https://stackoverflow.com/questions/92971/how-do-i-set-the-size-of-emacs-window)

MichaelChirico: thanks for googling for me :laughing: might as well be Lao to me though

MichaelChirico:
PTAL. I still used `xml_nodes_to_lint()`... the positions look fine to me:

Selection_183
Selection_182

both land you where you can delete the whitespace  which is what's important.

On `main`, the difference is both markers are to the _left_ of the whitespace, so user consistently deletes forward, whereas here, we land to the left of _opening_ whitespace and to the right of _closing_ whitespace.

I think that's fine -- in both cases there's a single key to press on many keyboards (or a key combo to press for forward-delete on laptops). If anything, if we insisted on consistency, I'd say it's better to be to the _right_ in both cases since backwards-delete is usually easier.

AshesITR:
Being to the left also causes the `print()` to look nicer.

MichaelChirico:
Hmm I see what you mean:

```r
l=lint("test.R", spaces_inside_linter())
print(l)
# /home/michael/github/lintr/test.R:1:3: style: Do not place spaces after square brackets.
# a[   1   ]
#  ~^
# /home/michael/github/lintr/test.R:1:10: style: Do not place spaces before square brackets.
# a[   1   ]
#          ^
```

OTOH, the source marker is maybe not very nice for a wider expression:

Selection_184

This branch is also _way_ faster -- 20x vs. `main` (see details)

On balance, between (1) the improved message (now says "before" when `print()` lands on the bracket); (2) consistently press one button (<kbd>Backspace</kbd> or <kbd>Delete</kbd>); (4) improved efficiency; and (3) the easier maintenance & readability, I still think this approach is preferable.

<details>
benchmarked on tools/R/QC.R, isolating the `get_source_expressions` overhead:

```r
library(lintr)
l = spaces_inside_linter()
f = "~/svn/R-devel/src/library/tools/R/QC.R"
e = get_source_expressions(f)
system.time(lapply(e$expressions, l))
# to check overlap
write.csv(as.data.frame(lint(f, l)), "branch.csv")
```
</details>

AshesITR:
It would be awesome if we modified the ranges to include the entire whitespace and we can add a `preceding-siblling::*[1]` to the OP-RIGHT-PAREN path to arrive at the `1` in this case.
Ideal output imo would be

```
a[   1   ]
  ^~~

a[   1   ]
      ^~~
```

MichaelChirico:
I think that's a separate PR, WDYT?

Now the refactor handles XML logic and fixes the no-match output, with tiny nudge to source markers (in the right direction I think).

Follow-up can further improve lint metadata.

AshesITR:
Filed #1205

AshesITR (approved):

PR #1172, opened 2022-05-22
make_linter_from_regex returns list() if no matches

MichaelChirico:
Part of #1160 

Don't think there's any way to push this one into xpath. In particular our internal linter using this (`no_tab_linter()`) is only linting on whitespace, which the XML tree is basically unaware of.


AshesITR:
This will make the class of the returned list `lints`.
Do we want to make `xml_nodes_to_lints()` also return a `lints` list for consistency? (Not part of this PR of course)

R/make_linter_from_regex.R
```diff
@@ -46,6 +50,8 @@ make_linter_from_regex <- function(regex,
         all_matches,
         line_numbers
       )
+
+      lapply(lints, function(x) x[vapply(x, length, integer(1L)) > 0L])
```
AshesITR, line 54: Isn't `lengths()` available for a while now?

MichaelChirico: thanks for making me check -- I know it's not in 3.1.

It's 3.2, which we use:

https://github.com/r-lib/lintr/blob/77558530416c4fb9017313abf14aae4c3b23d92a/R/backport_linter.R#L111-L115

AshesITR:
We should either conditionally import `backports::lengths()` or Depend on `R >= 3.2` so our dependencies are explicitly stated and correct. Slightly in favor of backports since we already use it for `R_user_dir()`.

MichaelChirico:
sorry for the confusion -- we're already on 3 2:

https://github.com/r-lib/lintr/blob/89c37d25d079adaafcb22e2e474d4c4c5b4a41a7/DESCRIPTION#L18

PR #1173, opened 2022-05-22
convert undesirable_operator_linter to xpath

MichaelChirico:
Part of #1160

### DIFF
--- a/R/infix_spaces_linter.R
+++ b/R/infix_spaces_linter.R
@@ -38,6 +38,7 @@ infix_metadata <- data.frame(stringsAsFactors = FALSE, matrix(byrow = TRUE, ncol
   "LBB", "[[",
   "OP-LEFT-PAREN", "(",
   "OP-QUESTION", "?",
+  "OP-DOLLAR", "$",
   NULL
 )))
 names(infix_metadata) <- c("xml_tag", "string_value")
@@ -56,6 +57,13 @@ infix_metadata$low_precedence <- infix_metadata$string_value %in% c(
 )
 # comparators come up in several lints
 infix_metadata$comparator <- infix_metadata$string_value %in% c("<", "<=", ">", ">=", "==", "!=")
+
+# undesirable_operator_linter needs to distinguish
+infix_overload <- data.frame(
+  exact_string_value = c("<-", "<<-", "=", "->", "->>", "^", "**"),
+  xml_tag = rep(c("LEFT_ASSIGN", "RIGHT_ASSIGN", "OP-CARET"), c(3L, 2L, 2L)),
+  stringsAsFactors = FALSE
+)
 
 #' Infix spaces linter
 #'
--- a/R/undesirable_operator_linter.R
+++ b/R/undesirable_operator_linter.R
@@ -19,33 +19,42 @@ op_types <- c(
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 undesirable_operator_linter <- function(op = default_undesirable_operators) {
-  Linter(function(source_expression) {
-    lapply(
-      ids_with_token(source_expression, op_types, fun = `%in%`),
-      function(id) {
-        token <- with_id(source_expression, id)
-        op_name <- token[["text"]]
-        if (op_name %in% names(op)) {
-          line_num <- token[["line1"]]
-          start_col_num <- token[["col1"]]
-          end_col_num <- token[["col2"]]
-          msg <- sprintf("Operator `%s` is undesirable.", op_name)
-          alt_op <- op[[op_name]]
-          if (!is.na(alt_op)) {
-            msg <- paste(msg, alt_op)
-          }
+  undesirable_operator_metadata <-
+    merge(infix_metadata, infix_overload, by = "xml_tag", all.x = TRUE)
 
-          Lint(
-            filename = source_expression[["filename"]],
-            line_number = line_num,
-            column_number = start_col_num,
-            type = "warning",
-            message = paste0(msg, collapse = " "),
-            line = source_expression[["lines"]][[as.character(line_num)]],
-            ranges = list(c(start_col_num, end_col_num))
-          )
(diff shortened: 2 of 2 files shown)


R/infix_spaces_linter.R
```diff
@@ -57,6 +58,13 @@ infix_metadata$low_precedence <- infix_metadata$string_value %in% c(
 # comparators come up in several lints
 infix_metadata$comparator <- infix_metadata$string_value %in% c("<", "<=", ">", ">=", "==", "!=")
 
+# undesirable_operator_linter needs to distinguish
+infix_overload <- data.frame(
+  exact_string_value = c("<-", "<<-", "=", "->", "->>", "^", "**"),
+  xml_tag = rep(c("LEFT_ASSIGN", "RIGHT_ASSIGN", "OP-CARET"), c(3L, 2L, 2L)),
```
AshesITR, line 64: I remember `EQ_ASSIGN` to be a thing. Is that a thing for older R versions?

R/undesirable_operator_linter.R
```diff
@@ -19,33 +19,42 @@ op_types <- c(
...
-          if (!is.na(alt_op)) {
-            msg <- paste(msg, alt_op)
-          }
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    xpath <- glue::glue("//*[ { paste0('self::', operator_nodes, collapse = ' or ') } ]")
```
AshesITR, line 42: `xp_or()`?

AshesITR:
Will this work with custom undesired operators in the same way it used to?
E.g. if linting `%something%` was possible before, is it still possible now?

AshesITR (changes requested):
Please fix & add a test for a custom SPECIAL operator, i.e.
```
expect_lint(
  "a %oo% b",
  rex::rex("Operator `%oo%` is undesirable"), 
  undesirable_operator_linter(list("%oo%" = NA))
)
```

MichaelChirico:
Nice! Also caught that the HEAD version doesn't really handle `=` correctly... I am assuming if user wants `undesirable_operator_linter(list("=" = NA))`, they're targeting equals-assignment. But HEAD also considers `lm(data = mtcars)` a violation.

R/infix_spaces_linter.R
MichaelChirico: already handled actually IINM, but added tests anyway. Non-assignment `=` were indeed wrongly handled.

R/undesirable_operator_linter.R
```diff
@@ -19,33 +9,59 @@ op_types <- c(
...
-            line = source_expression[["lines"]][[as.character(line_num)]],
-            ranges = list(c(start_col_num, end_col_num))
-          )
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    xpath <- glue::glue("//*[ { xp_or(operator_nodes) } ]")
```
AshesITR, line 49: Could be moved into the factory as well, since that's where most of the XPath is constructed.

AshesITR:
By the way: `//*[...]` is much much slower than `//... | //...`:

```
> xml <- xml2::read_xml(xmlparsedata::xml_parse_data(parse(text = "a <- 2\nb <- 3 + 4\nc <- 5 %% 6")))
> bench::mark(
+   xml2::xml_find_all(xml, "//*[self::OP-PLUS or self::SPECIAL[text() = '%%']]"),
+   xml2::xml_find_all(xml, "//OP-PLUS | //SPECIAL[text() = '%%']")
+ )
# A tibble: 2 x 13
  expression                                                                         min
  <bch:expr>                                                                    <bch:tm>
1 xml2::xml_find_all(xml, "//*[self::OP-PLUS or self::SPECIAL[text() = '%%']]")   57.3us
2 xml2::xml_find_all(xml, "//OP-PLUS | //SPECIAL[text() = '%%']")                 38.7us
```

R/undesirable_operator_linter.R
MichaelChirico: I would keep it here... I think it's clearer/ more consistent w other linters

MichaelChirico:
> By the way: `//*[...]` is much much slower than `//... | //...`:
> 
> ```
> > xml <- xml2::read_xml(xmlparsedata::xml_parse_data(parse(text = "a <- 2\nb <- 3 + 4\nc <- 5 %% 6")))
> > bench::mark(
> +   xml2::xml_find_all(xml, "//*[self::OP-PLUS or self::SPECIAL[text() = '%%']]"),
> +   xml2::xml_find_all(xml, "//OP-PLUS | //SPECIAL[text() = '%%']")
> + )
> # A tibble: 2 x 13
>   expression                                                                         min
>   <bch:expr>                                                                    <bch:tm>
> 1 xml2::xml_find_all(xml, "//*[self::OP-PLUS or self::SPECIAL[text() = '%%']]")   57.3us
> 2 xml2::xml_find_all(xml, "//OP-PLUS | //SPECIAL[text() = '%%']")                 38.7us
> ```

that's unfortunate 😂

will change it, but my guess is we never notice because the overhead of iterating over expressions is much higher than these 20us

MichaelChirico:
I think we should always benchmark on tools/QC.R 😂

R/undesirable_operator_linter.R
MichaelChirico: it might be a separate issue to tackle for performance more globally.

will definitely change to use |

AshesITR:
good idea. Is there a download.file URL for it? 😄 
I think we should add a `benchmark.R` script to `.dev` that runs on some files. WDYT?

MichaelChirico:
> I think we should add a benchmark.R script to .dev that runs on some files. WDYT?

I am planning to add a flag to .dev/compare_branches to help with that. Not 100% sure the API... working on that soon.

AshesITR (approved):

PR #1175, opened 2022-05-22
update fixed_regex_linter lint message

### LINKED ISSUE #1158: Improve warning message for `fixed_regex_linter`
> For static regular expression patterns, set `fixed = TRUE`. Note that this includes regular expressions that can be expressed as fixed patterns, e.g. [.] is really just . and \\$ is really just $ if there are no other regular expression specials. For functions from the 'stringr' package, the way to declare a static string is to wrap the pattern in stringr::fixed(). If this is being used in a dbplyr context (i.e., translated to sql), replace the regular expression with the `LIKE` operator using the `%LIKE%` infix function. Lastly, take care to remember that the `replacement` argument of `gsub()` is affected by the `fixed` argument as well.

I have a few questions/comments around  this warning, especially the second sentence.

1. "Note that this includes" - what is "this"? Why does "this" "include" something? Is this supposed to say that "the previous statement" "holds true"/"applies"/... to "regular expressions that can be expressed as fixed patterns"?
2. What does "[.]" apply to? Is that to mean that folks shouldn't use "[H]ello world", but "Hello world"? Or is this to discourage people from actually writing "[.]" and writing "." instead? I suppose it's the former, but then the use of regular expressions in statements about regular expressions is really making it unclear.
3. Why is "\\$ ... really just $ if there are no other regular expression specials"? To my understanding, `$` is an end-of-expression placeholder when `fixed = FALSE` (or omitted), and it's a literal dollar sign when `fixed = TRUE`. What do "other regular expression specials" have to do with that?
4. The part of the message in 3 feels to me like it has been added to explain *why* lintr warns about something. However, it seems to warn about everything, irrespective of any regular expression specials - e.g., expressions such as `grepl("b", "abc")`. [Is that for performance? Because I find this much more readable than `grepl("b", "abc", fixed = TRUE)`, and I am not aware of any good alternative in R.] Anyway, I wonder why lintr explains why it warns about something specific if it warns about pretty much everything.

AshesITR:
fixes #1158


AshesITR:
@MichaelChirico This PR fixes the last lint in our codebase IIRC (commented_code).

AshesITR:
What would be the (conditional) hints, that help you find the proper fixed expression more quickly?
I think it should technically be possible to directly specify the fixed equivalent to all of the tokens?

MichaelChirico:
the first thing that came to mind was basically to subset the existing message based on which specials are detected.

obviously a static version user can copy/paste is pretty ideal but maybe a lot of work

AshesITR:
I have something in mind that should yield the fixed string. When I find time, I'll rebase and update this PR.

The gist is to capture each token and then replace it by the fixed equivalent last-to-first.

AshesITR:
@MichaelChirico PTAL I think it works, but definitely worth running on a few packages to make sure it always returns a correct string.

R/fixed_regex_linter.R
```diff
@@ -61,46 +61,28 @@ fixed_regex_linter <- function() {
...
     patterns <- xml2::xml_find_all(xml, xpath)
+    pattern_strings <- get_r_string(patterns)
+    is_static <- is_not_regex(pattern_strings)
+
+    fixed_equivalent <- get_fixed_string(pattern_strings[is_static])
+    msg <- paste(
+      "This regular expression is static, i.e., its matches can be expressed as a fixed substring expression, which",
+      "is faster to compute.",
+      "Here, you can use", encodeString(fixed_equivalent, quote = '"', justify = "none"), "as a fixed substring.",
+      "For static regular expression patterns, set fixed = TRUE or use stringr::fixed()."
```
MichaelChirico, line 72: let's customize this too while we're at it -- either recommend fixed=TRUE or stringr::fixed() depending on the regex flavor

MichaelChirico: With this in hand I think we can dramatically simplify the lint message:

This regex pattern pattern is equivalent to X with fixed=TRUE.

maybe adding "it is more efficient to use fixed=TRUE when possible and typically easier to read."

wdyt @bersbersbers -- would the message be clear enough if the fixed equivalent is provided?

I guess the purely no special character case would also warrant special handling then...

R/fixed_regex_linter.R
```diff
@@ -119,13 +101,82 @@ is_not_regex <- function(str) {
...
+is_not_regex <- function(str) {
   # need to add single-line option to allow literal newlines
-  grepl(paste0("(?s)", rx_static_regex), str, perl = TRUE) | grepl(rx_static_regex, str)
+  grepl(paste0("(?s)", rx_static_regex), str, perl = TRUE)
+}
+
+#' Compute a fixed string equivalent to a static regular expression
+#'
+#' @param static_regex A regex for which `is_not_regex()` returns `TRUE`
+#' @return A string such that `grepl(static_regex, x)` is equivalent to `grepl(get_fixed_string(static_regex), x, fixed = TRUE)`
```
MichaelChirico, line 141: delint

R/fixed_regex_linter.R
```diff
@@ -119,13 +101,82 @@ is_not_regex <- function(str) {
...
+#'
+#' @noRd
+get_fixed_string <- function(static_regex) {
+  if (length(static_regex) == 0L) {
+    return(character())
+  } else if (length(static_regex) > 1L) {
+    return(vapply(static_regex, get_fixed_string, character(1L)))
+  }
+  fixed_string <- ""
```
MichaelChirico, line 150: maybe more efficient to use list(nchar(static_regex)) (fixed version has to be weakly shorter right?) and do.call(paste0) at the end

R/fixed_regex_linter.R
AshesITR: Honestly I like the general advice first, special information second.
I'm fine with dynamic choice of `fixed = TRUE` vs. `stringr::fixed()` but wouldn't go further.

tests/testthat/test-fixed_regex_linter.R
```diff
@@ -32,7 +32,7 @@ test_that("fixed_regex_linter skips allowed usages", {
 
 test_that("fixed_regex_linter blocks simple disallowed usages", {
   linter <- fixed_regex_linter()
-  msg <- rex::rex("For static regular expression patterns, set `fixed = TRUE`.")
+  msg <- rex::rex("This regular expression is static")
```
MichaelChirico, line 35: let's include the static equivalent part of the regex in the test message, that's the most important part IMO

R/fixed_regex_linter.R
AshesITR: Not sure if that's efficient since we grossly over-allocate on most regexes.

MichaelChirico: but it's just slot pointers overallocated. my instinct is overwriting is less efficient.

OTOH when I wrote this I hadn't read carefully enough -- thought we were building the fixed version char-by-char but it's actually chunk-wise, so I think it's fine

R/fixed_regex_linter.R
bersbersbers: I'm fine with either - I think outputting `fixed_equivalent` helps a lot already.

tests/testthat/test-fixed_regex_linter.R
AshesITR: I've added a patrick test for all patterns used throughout the rest of the test suite, to keep diff sizes low in case we change lint messages down the line.

MichaelChirico (approved):

PR #1176, opened 2022-05-22
improve linter factory detection

### LINKED ISSUE #1174: Possible improvement for 3.0.0 breaking change
Reported by downstream here:

https://github.com/microsoft/LightGBM/issues/5228

Slightly simpler reprex:

```
lintr::lint(text = "print(1)", linters = lintr::assignment_linter)
# Error: `fun` must be a function taking exactly one argument.
# In addition: Warning message:
#  The use of linters of class 'function' was deprecated in lintr version 2.0.1.9001. Use linters classed as 'linter' (see ?Linter) instead.
```

Can we not do `fun()` for the user in this release after issuing "please use a linter" warning?

Also, I'm wondering if the warning message can't be improved:

> The use of linters of class 'function' was deprecated in lintr version 2.0.1.9001. Use linters classed as 'linter' (see ?Linter) instead.

I think the most likely scenario where a user encounters this warning is just like the above -- they've got a script using the un-called form, or they've grown used to supplying it themselves interactively. So we should add this to the end, right?

> For example, if you supplied `lintr::assignment_linter`, please use `lintr::assignment_linter()` instead.

AshesITR:
fixes #1174


AshesITR:
`covr::package_coverage()` rewrites each `call` into `if (TRUE) { covr::count(...); call }`.
The way I fixed is by also diving into `if (TRUE) {` expressions as well as `{` expressions.

@MichaelChirico this probably affects `is_s3_generic()` as well. The only way to trigger the bug is to call `is_s3_generic()` within covr on a function from the coverage-tested package, i.e. to call `lintr` during package tests.
That means if I am correct, `lintr::expect_lint_free()` can trigger it if it's placed in a test file. Not sure if this happens in the wild, though.
See a similar workaround for `trace()`d functions in the current `isS3stdGeneric` implementation:

```
bdexpr <- body(if (methods::is(f, "traceable")) f@original else f)
```

R/lint.R
```diff
@@ -370,7 +370,7 @@ define_linters <- function(linters = NULL) {
 validate_linter_object <- function(linter, name) {
   if (inherits(linter, "linter")) {
```
MichaelChirico, line 371: easier to understand as

```
if (!inherits(linter, "linter") && is.function(linter)) { ... }
```

no?

R/lint.R
```diff
@@ -390,6 +390,16 @@ validate_linter_object <- function(linter, name) {
   linter
 }
 
+is_linter_factory <- function(fun) {
+  # A linter factory is a function whose last call is to Linter()
+  bdexpr <- body(fun)
+  # covr internally transforms each call into if (TRUE) { covr::count(...); call }
```
MichaelChirico, line 396: nit: maybe better to move that out of the `while()` condition since it only need to be tested once at most?

OTOH, evaluating this while loop more than once is already in funky territory -- means authors intentionally writing adversarial code AFAICT...

MichaelChirico:
> See a similar workaround for trace()d functions in the current isS3stdGeneric implementation:

Does `covr` offer anything similar? maybe cc @jimhester here...

anyway, as long as tests pass on the `covr` branch, I wouldn't worry about it too much... kinda painful to maintain this overhead for `covr` so i'd avoid it where possible...

R/lint.R
AshesITR: I wanted to stick close to the S3 generic detection here, since that has been well tested.

MichaelChirico: works for me :+1:

MichaelChirico (approved):

PR #1183, opened 2022-05-22
delint tests/testthat/test-spaces_inside_linter.R

AshesITR:


tests/testthat/test-spaces_inside_linter.R
```diff
@@ -37,20 +37,20 @@ test_that("returns the correct linting", {
...
-    list(message = msg, line_number = 1, column_number = 3, type = "style"),
+    list(message = msg, line_number = 1L, column_number = 3L, type = "style"),
     linter
   )
 
   expect_lint(
     "a( 1 )",
-    list(list(message = msg, line_number = 1, column_number = 3, type = "style"),
-         list(message = msg, line_number = 1, column_number = 5, type = "style")),
+    list(list(message = msg, line_number = 1L, column_number = 3L, type = "style"),
```
MichaelChirico, line 52: should we edit expect_lint to use identical?

MichaelChirico (approved):

tests/testthat/test-spaces_inside_linter.R
AshesITR: I think we can be more lenient.

MichaelChirico: I think that's a good break TBH -- type consistency is a good think.

(to clarify I don't care that much about using doubles in the test files, but rather about custom linters that produce non-integer values for column_number, etc. so we could also check is.integer() instead)

PR #1195, opened 2022-05-23
allow detecting any function ending in UseMethod as an S3 generic

MichaelChirico:
This is just a copy of the branch in #857. I am suspecting the original fork needs to update its master branch for CI... testing if running here works.


MichaelChirico:
Not sure what GitHub's doing but it's duplicated the CI runs for this and the other PR, even though the branches are coming from different forks. Anyway, the CI is passing now, so closing this one :crossed_fingers:

PR #1197, opened 2022-05-23
improve performance of XPath linters using //* predicates

AshesITR:
Rewrote all `//*[(self::xxx or self::yyy) and zzz]` constructs to the faster `//xxx[zzz] | //xxx[zzz]` XPath representation.

### DIFF
--- a/R/infix_spaces_linter.R
+++ b/R/infix_spaces_linter.R
@@ -84,43 +84,44 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces 
     op <- "!="
     lint_message <- "Put exactly one space on each side of infix operators."
   }
+
+  infix_tokens <- infix_metadata[
+    infix_metadata$low_precedence & !infix_metadata$string_value %in% exclude_operators,
+    "xml_tag"
+  ]
+
+  # NB: preceding-sibling::* and not preceding-sibling::expr because
+  #   of the foo(a=1) case, where the tree is <SYMBOL_SUB><EQ_SUB><expr>
+  # NB: preceding-sibling::* for the unary case, e.g. x[-1]
+  # NB: the last not() disables lints inside box::use() declarations
+  xp_condition <- glue::glue("
+    preceding-sibling::*
+    and (
+      (
+        @line1 = preceding-sibling::*[1]/@line1
+        and @col1 {op} preceding-sibling::*[1]/@col2 + 2
+      ) or (
+        @line1 = following-sibling::*[1]/@line1
+        and following-sibling::*[1]/@col1 {op} @col2 + 2
+      )
+    )
+    and not(
+      self::OP-SLASH[
+        ancestor::expr/preceding-sibling::OP-LEFT-PAREN/preceding-sibling::expr[
+          ./SYMBOL_PACKAGE[text() = 'box'] and
+          ./SYMBOL_FUNCTION_CALL[text() = 'use']
+        ]
+      ]
+    )
+  ")
+  xpath <- paste(glue::glue("//{infix_tokens}[{xp_condition}]"), collapse = " | ")
+
   Linter(function(source_expression) {
     if (!is_lint_level(source_expression, "expression")) {
       return(list())
     }
 
     xml <- source_expression$xml_parsed_content
-    infix_tokens <- infix_metadata[
-      infix_metadata$low_precedence & !infix_metadata$string_value %in% exclude_operators,
-      "xml_tag"
-    ]
-
-    # NB: preceding-sibling::* and not preceding-sibling::expr because
-    #   of the foo(a=1) case, where the tree is <SYMBOL_SUB><EQ_SUB><expr>
-    # NB: preceding-sibling::* for the unary case, e.g. x[-1]
-    # NB: the last not() disables lints inside box::use() declarations
-    xp_condition <- glue::glue("
-      preceding-sibling::*
-      and (
-        (
-          @line1 = preceding-sibling::*[1]/@line1
-          and @col1 {op} preceding-sibling::*[1]/@col2 + 2
-        ) or (
-          @line1 = following-sibling::*[1]/@line1
-          and following-sibling::*[1]/@col1 {op} @col2 + 2
-        )
-      )
-      and not(
-        self::OP-SLASH[
-          ancestor::expr/preceding-sibling::OP-LEFT-PAREN/preceding-sibling::expr[
-            ./SYMBOL_PACKAGE[text() = 'box'] and
-            ./SYMBOL_FUNCTION_CALL[text() = 'use']
-          ]
-        ]
-      )
-    ")
-    xpath <- paste(glue::glue("//{infix_tokens}[{xp_condition}]"), collapse = " | ")
-
     bad_expr <- xml2::xml_find_all(xml, xpath)
 
     xml_nodes_to_lints(bad_expr, source_expression = source_expression, lint_message = lint_message, type = "style")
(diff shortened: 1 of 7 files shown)


R/infix_spaces_linter.R
```diff
@@ -119,7 +118,8 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces
           ]
         ]
       )
-    ]")
+    ")
+    xpath <- paste(glue::glue("//{infix_tokens}[{xp_condition}]"), collapse = " | ")
```
MichaelChirico, line 122: this needs a different benchmark IMO, we didn't measure with vs. without conditions on each node

R/object_usage_linter.R
```diff
@@ -195,20 +195,14 @@ get_assignment_symbols <- function(xml) {
...
-      "*[
-        (
-          (self::expr and (LEFT_ASSIGN or EQ_ASSIGN))
-          or ((self::expr_or_assign_or_help or self::equal_assign) and EQ_ASSIGN)
-        )
-        and expr[2][FUNCTION]
-      ]
-      /expr[2]
-      ",
+      "expr[LEFT_ASSIGN or EQ_ASSIGN]/expr[2][FUNCTION]",
```
MichaelChirico, line 203: ditto... separate benchmark

R/package_hooks_linter.R
```diff
@@ -91,10 +91,14 @@ package_hooks_linter <- function() {
     # NB: base only checks the SYMBOL_FUNCTION_CALL version, not SYMBOL.
     library_require_xpath <- "
     //expr[SYMBOL[text() = '.onAttach' or text() = '.onLoad']]
-    /following-sibling::expr//*[
-      (self::SYMBOL or self::SYMBOL_FUNCTION_CALL)
-      and (text() = 'require' or text() = 'library' or text() = 'installed.packages')
-    ]
+    /following-sibling::expr
```
MichaelChirico, line 94: again different... benchmark is for //* which is "start from root and search", which is way more overhead than "search all siblings" which is almost always one or two nodes

R/infix_spaces_linter.R
AshesITR: good thinking. I'll cook up a benchmark based on default-linter-testcode.R

AshesITR:
Ran a more thorough benchmark on `tests/testthat/default_linter_testcode.R` with interesting results:

```
modified_linters <- c(
  "assignment_linter",
  "backport_linter",
  "infix_spaces_linter",
  "object_usage_linter",
  "package_hooks_linter",
  "undesirable_function_linter",
  "vector_logic_linter"
)

gert::git_branch_checkout("main")
devtools::load_all()
old_linters <- lapply(modified_linters, function(linter_name) {
  eval(call(linter_name))
})
gert::git_branch_checkout("fix/1196-performance")
devtools::load_all()
new_linters <- lapply(modified_linters, function(linter_name) {
  eval(call(linter_name))
})

exprs <- get_source_expressions("tests/testthat/default_linter_testcode.R")
benchmarks <- vector(mode = "list", length = length(modified_linters))

for (i in seq_along(modified_linters)) {
  cat("## ", modified_linters[i], ":\n", sep = "")
  benchmark <- bench::mark(
    lapply(exprs$expressions, old_linters[[i]]),
    lapply(exprs$expressions, new_linters[[i]]),
    min_time = 1.0
  )
  benchmarks[[i]] <- benchmark
  print(benchmark)
  cat("\n\n")
}

vapply(setNames(benchmarks, modified_linters), function(bm) c("old", "new")[which.max(bm[["itr/sec"]])], character(1L))
```

Produces the following verdict:
```
          assignment_linter             backport_linter         infix_spaces_linter 
                      "new"                       "new"                       "old" 
        object_usage_linter        package_hooks_linter undesirable_function_linter 
                      "new"                       "old"                       "old" 
        vector_logic_linter 
                      "old" 
```

I'm reverting the implementations where "old" is faster, although not 100% sure for some implementations if that is just because of the relatively simple structure of the testcode.

AshesITR:
@MichaelChirico I've managed to halve the runtimes of `undesirable_function_linter` and `infix_spaces_linter` by moving the XPath construction out of the linting closure.
Now all linters are either unchanged or faster than main, as tested by the benchmark above.

AshesITR:
On tools/QC.R:
```
## infix_spaces_linter:
Warning: Some expressions had a GC in every iteration; so filtering is disabled.
# A tibble: 2 x 13
  expression                                      min median `itr/sec` mem_alloc `gc/sec`
  <bch:expr>                                  <bch:t> <bch:>     <dbl> <bch:byt>    <dbl>
1 lapply(exprs$expressions, old_linters[[i]])   798ms  798ms      1.25   644.6KB    1.25 
2 lapply(exprs$expressions, new_linters[[i]])   215ms  219ms      4.56    11.9KB    0.913
# ... with 7 more variables: n_itr <int>, n_gc <dbl>, total_time <bch:tm>,
#   result <list>, memory <list>, time <list>, gc <list>
```

AshesITR:
The `get_source_expressions()` improvement will definitely help.
Performance improvements were also visible for just XPath changes (that was the comparison from before), but I agree the refactoring of computing XPaths to outside the linter closure provides most of the speedup.

I suggest we try to move all XPaths to the factories if possible and try to make them as readable as possible.

AshesITR:
WDYT about the new `get_function_assignments()` XPath? I actually find it more readable because the old XPath has very contrived logical conjunctions.

Ready to merge IMO.

R/undesirable_function_linter.R
```diff
@@ -24,17 +24,16 @@ undesirable_function_linter <- function(fun = default_undesirable_functions,
...
       paste0(
         "not(parent::expr/preceding-sibling::expr[SYMBOL_FUNCTION_CALL[",
         xp_text_in_table(c("library", "require")),
         "]])"
       ),
       "not(preceding-sibling::OP-DOLLAR)"
-    ), "]")
-
+    )
+    xpath <- paste(glue::glue("//{tokens}[{xp_condition}]"), collapse = " | ")
```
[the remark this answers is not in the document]
AshesITR: Not really necessary after the rollbacks. But worth keeping an eye out for, if the pattern emerges again.

R/undesirable_function_linter.R
```diff
@@ -14,27 +14,27 @@ undesirable_function_linter <- function(fun = default_undesirable_functions,
...
+  xp_condition <- xp_and(
+    xp_text_in_table(names(fun)),
+    paste0(
+      "not(parent::expr/preceding-sibling::expr[SYMBOL_FUNCTION_CALL[",
+      xp_text_in_table(c("library", "require")),
+      "]])"
+    ),
+    "not(preceding-sibling::OP-DOLLAR)"
+  )
+  xpath <- paste(glue::glue("//{tokens}[{xp_condition}]"), collapse = " | ")
```
MichaelChirico, line 32: WDYT about being more explicit instead of using collapse?

```
if (symbol_is_undesirable) {
  xpath <- glue::glue("//SYMBOL_FUNCTION_CALL[{xp_condition}] | //SYMBOL[{xp_condition}]")
} else {
  xpath <- glue::glue("//SYMBOL_FUNCTION_CALL[{xp_condition}]
}
```

or

```
xpath <- glue::glue("//SYMBOL_FUNCTION_CALL[{xp_condition}]"
if (symbol_is_undesirable) {
  xpath <- paste(xpath, glue::glue("| //SYMBOL[{xp_condition}]"))
}
```

?

AshesITR: I like the first suggestion

MichaelChirico (approved):

Issue #1104, opened 2022-04-26
Add argument `scope = c("function", "variable", "constant", "class", ?)` to `object_name_linter()`


AshesITR:
This would allow a user to define custom naming conventions per scope if desired.
e.g. we use `snake_case` for functions and variables, `SNAKE_CASE` for constants and `CamelCase` for classes.
IIUC, google style mandates `CamelCase` for functions.

The way this would work if a user wanted to would be
```
linters_with_tags(
  tags = "default",
  object_name_linter(scope = c("function", "variable")),
  constant_name_linter = object_name_linter("SNAKE_CASE", scope = "constant"),
  class_name_linter = object_name_linter("CamelCase", scope = "class")
)
```

Definition of the scopes would be:
 * function = assignment targets of `//FUNCTION` exprs
 * constant = assignment targets of exactly one known-constant expr
 * class = assignment targets of `R6::R6Class()` exprs
 * variable = everything else

MichaelChirico:
excellent proposal, thanks!

AshesITR:
Any other interesting and identifiable scopes?

I had shiny modules in mind, but these, especially the UI part, are hardly discernable from functions.

MichaelChirico:
everything I could think of was covered in your initial post (well, maybe add S4 classes?)

AshesITR:
I've never used S4 myself. What would be a good way to detect S4 class assignments?

MichaelChirico:
actually I think this linter would to adapt pretty substantially for S4... all the names are in string literals I think (also not an S4 user myself)?

see e.g. Matrix

https://github.com/cran/Matrix/blob/master/R/AllClass.R

AshesITR:
Seems like a separate feature, though some linters provide some (limited) support for S4, e.g., object_usage_linter scans S4 setMethod functions (assuming they are provided positionally)
I think we can safely add S4 class assigments to the class scope and S4 method definitions to the function scope at a later time.

PR #1198, opened 2022-05-23
remove package prefix from auto-generated names

### LINKED ISSUE #1193: Noisy warnings from new abbreviated linter exclusions
I've been seeing this a lot lately, finally got a reprex:

```
lintr::lint(text = "a = 1 # nolint: implicit_integer.", linters = lintr::assignment_linter())
# <text>:1:3: style: Use <-, not =, for assignment.
# a = 1 # nolint: implicit_integer.
#   ^
# Warning message:
# In add_exclusions(exclusions, nolints[i], linters_string, exclude_linter_sep,  :
#   Could not find linter named ‘implicit_integer’ in the list of active linters. Make sure the linter is uniquely identified by the given name or prefix.
```

That is, the code (1) throws a lint and (2) has a `#nolint` directive specifying a linter that's not in the current set of `linters`.

This makes running one or a few linters on code that's generally covered by more linters quite noisy. Is there anything we can do here? Should "list of active linters" be expanded to "list of known linters"?

AshesITR:
Partially fixes #1193

```
lintr::lint(text = "a = 1 # nolint: implicit_integer.", linters = lintr::implicit_integer_linter())
```

is silent with this fix because `lintr::implicit_integer_linter()` is correctly named `"implicit_integer_linter"`.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -63,10 +63,11 @@ names2 <- function(x) {
 linter_auto_name <- function(which = -3L) {
   call <- sys.call(which = which)
   nm <- paste(deparse(call, 500L), collapse = " ")
-  regex <- rex(start, one_or_more(alnum %or% "." %or% "_"))
+  regex <- rex(start, one_or_more(alnum %or% "." %or% "_" %or% ":"))
   if (re_matches(nm, regex)) {
     match <- re_matches(nm, regex, locations = TRUE)
     nm <- substr(nm, start = 1L, stop = match[1L, "end"])
+    nm <- re_substitutes(nm, rex::rex(start, alnums, "::"), "")
   }
   nm
 }
--- a/R/with.R
+++ b/R/with.R
@@ -34,7 +34,7 @@ modify_defaults <- function(defaults, ...) {
           # Very long input might have newlines which are not caught
           #  by . in a perl regex; see #774
         re_substitutes(args, rex("(", anything), "", options = "s"),
-        rex(start, anything, '["'),
+        rex(start, anything, '["' %or% "::"),
         ""
       ),
       rex('"]', anything, end),
--- a/tests/testthat/test-settings.R
+++ b/tests/testthat/test-settings.R
@@ -114,7 +114,7 @@ test_that("linters_with_defaults doesn't break on very long input", {
         )
       ))
     ),
-    "lintr::undesirable_function_linter"
+    "undesirable_function_linter"
   )
 })
 


R/utils.R
```diff
@@ -63,10 +63,11 @@ names2 <- function(x) {
 linter_auto_name <- function(which = -3L) {
   call <- sys.call(which = which)
   nm <- paste(deparse(call, 500L), collapse = " ")
-  regex <- rex(start, one_or_more(alnum %or% "." %or% "_"))
+  regex <- rex(start, one_or_more(alnum %or% "." %or% "_" %or% ":"))
   if (re_matches(nm, regex)) {
     match <- re_matches(nm, regex, locations = TRUE)
     nm <- substr(nm, start = 1L, stop = match[1L, "end"])
+    nm <- re_substitutes(nm, rex::rex(start, alnums, "::"), "")
```
MichaelChirico, line 70: what happens if user (probably accidentally) runs:

```
lintr::lint(file, linters = list(lintr::assignment_linter, my_custom_package::assignment_linter))
```

where they've written their own `assignment_linter` and perhaps inadvertently included `lintr`' implementation (e.g. picked up from settings?).

i guess that should error up-front, but what's the current behavior?

AshesITR: That will work just fine, producing two linters with the same name:
```
> lintr::lint(text = "a = 1", linters = list(assignment_linter(), lintr::assignment_linter()))
<text>:1:3: style: Use <-, not =, for assignment.
a = 1
  ^
<text>:1:3: style: Use <-, not =, for assignment.
a = 1
  ^
```

Having two linters with the same name currently disables partial matching for those, however:
```
> lintr::lint(text = "a = 1 # nolint: a.", linters = list(assignment_linter(), assignment_linter = lintr::assignment_linter()))
Warning in add_exclusions(exclusions, nolints[i], linters_string, exclude_linter_sep,  :
  Could not find linter named ‘a’ in the list of active linters. Make sure the linter is uniquely identified by the given name or prefix.
<text>:1:3: style: Use <-, not =, for assignment.
a = 1 # nolint: a.
  ^
<text>:1:3: style: Use <-, not =, for assignment.
a = 1 # nolint: a.
  ^
> lintr::lint(text = "a = 1 # nolint: assignment_linter.", linters = list(assignment_linter(), assignment_linter = lintr::assignment_linter()))
```

I think there might be a valid use case once #1104 is implemented, namely to have `# nolint: object_name_linter.` suppress multiple versions of the linter, i.e. for different scopes.
WDYT about using `unique(names(linters))` for the partial matching to not disable that for this scenario, but let it happen without an error?
I don't expect end users to be wary of this when they want to configure multiple `object_name_linter()`s

MichaelChirico: > WDYT about using unique(names(linters)) for the partial matching to not disable that for this scenario, but let it happen without an error?

I think the downside is silently ignoring ambiguity in other context. e.g. `# nolint: spaces.` right now matches `spaces_inside_linter` and `spaces_left_parentheses_linter` -- do we want that to work without warning? If that's intentional it seems too clever to me. (same happens with `# nolint: paren`).

So for now, I'd rather keep warning. Maybe once we have an actual use case in hand it'll be clearer what to do.

MichaelChirico (approved):

PR #1200, opened 2022-05-23
generate top level parsed content map instead of tree in `get_source_expressions()`

AshesITR:
I noticed `get_source_expressions()` was very slow on `tools/QC.R` and set out to optimize it.

Instead of computing the parse tree as a graph ahead of time and then recursing this data structure for each and every top level expression, we now generate a mapping from parsed content to top level expression ahead of time and then just index into that mapping.

Results speak for themselves:

```
> system.time(get_source_expressions(test_file))
   user  system elapsed 
  4.420   0.002   4.422 
> devtools::load_all()
ℹ Loading lintr
> system.time(get_source_expressions(test_file))
   user  system elapsed 
  1.651   0.000   1.651 
```

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -78,7 +78,7 @@ get_source_expressions <- function(filename, lines = NULL) {
   )
   source_expression$content <- get_content(source_expression$lines)
   parsed_content <- get_source_expression(source_expression, error = function(e) lint_parse_error(e, source_expression))
-  tree <- generate_tree(parsed_content)
+  top_level_map <- generate_top_level_map(parsed_content)
 
   if (inherits(e, "lint") && !nzchar(e$line)) {
     # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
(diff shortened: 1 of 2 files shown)


AshesITR:
The test failures are because of a change in row order (row indices were hard-coded) whereas now they are in the same order as in the original `getParseData()` call.
Fixing up the tests now.

tests/testthat/test-get_source_expressions.R
```diff
@@ -170,20 +170,20 @@ test_that("Can extract line number from parser errors", {
 })
 
 test_that("1- or 2-width octal expressions give the right STR_CONST values", {
-  with_content_to_parse("'\\1'", expect_identical(pc[[1L]][2L, "text"], "'\\1'"))
-  with_content_to_parse('"\\1"', expect_identical(pc[[1L]][2L, "text"], '"\\1"'))
+  with_content_to_parse("'\\1'", expect_identical(pc[[1L]][1L, "text"], "'\\1'"))
```
MichaelChirico, line 173: probably it would be better to test this directly against `getParseData()` somehow... someday...

MichaelChirico (approved):

AshesITR:
I found one more optimization, let me test that

MichaelChirico:
Awesome stuff!!! 🚀

I suspect there are more improvements available for `get_source_expressions()`...

PR #1201, opened 2022-05-23
generate expression xml parse data from full xml parse data

AshesITR:
Another 20% performance improvement by parsing the XML only once and creating the expression documents from the full document.

```
> system.time(get_source_expressions(test_file))
   user  system elapsed 
  2.099   0.001   2.099 
> devtools::load_all()
ℹ Loading lintr
> system.time(get_source_expressions(test_file))
   user  system elapsed 
  1.635   0.001   1.636 
```

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -84,7 +84,7 @@ get_source_expressions <- function(filename, lines = NULL) {
     # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
     expressions <- list()
   } else {
-     xml_parsed_content <- safe_parse_to_xml(parsed_content)
+    xml_parsed_content <- safe_parse_to_xml(parsed_content)
 
     expressions <- lapply(
       X = top_level_expressions(parsed_content),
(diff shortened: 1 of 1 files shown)


R/get_source_expressions.R
```diff
@@ -93,14 +95,23 @@ get_source_expressions <- function(filename, lines = NULL) {
       top_level_map
     )
 
+    expression_xmls <- lapply(
+      xml2::xml_find_all(xml_parsed_content, "/exprlist/*"),
+      function(top_level_expr) xml2::xml_add_parent(xml2::xml_new_root(top_level_expr), "exprlist")
```
AshesITR, line 100: This is the magic 😄  (faster than parsing a new XML is copying the node from the existing parsed XML and adding the `exprlist` parent node.

MichaelChirico: does this actually make a copy / are our XML trees always read-only?

i have this in mind:

https://github.com/r-lib/xml2/issues/341

R/get_source_expressions.R
```diff
@@ -93,14 +95,25 @@ get_source_expressions <- function(filename, lines = NULL) {
       top_level_map
     )
 
+    if (!is.null(xml_parsed_content)) {
+      expression_xmls <- lapply(
+        xml2::xml_find_all(xml_parsed_content, "/exprlist/*"),
+        function(top_level_expr) xml2::xml_add_parent(xml2::xml_new_root(top_level_expr), "exprlist")
+      )
+      expressions[] <- Map(function(expr, xml) {
```
MichaelChirico, line 103: this overwrites the whole list just to add sub-elements -- could you benchmark that vs. (maybe not exactly right) `for (ii in seq_along(expressions)) expressions[[ii]]$xml_parsed_content <- expression_xmls[[ii]]`?

R/get_source_expressions.R
MichaelChirico: (awesome idea btw & makes perfect sense to me, just trying to stress test it)

AshesITR: `xml_new_root()` has a `.copy` argument, which defaults to `TRUE` in our usecase. This ensures a new document is created with a copy of that particular selected node as the root element.

R/get_source_expressions.R
AshesITR: Yes, will test that.

AshesITR: It is indeed 200ms faster 👍🏻

MichaelChirico: nice! that makes this branch another 33% vs main?

MichaelChirico (approved):

Issue #1186, opened 2022-05-22
Cyclomatic complexity lints in lintr


AshesITR:
After fixing all implicit integer lints, we can almost use `.lintr_new`, except for cyclomatic complexity:

 - [ ] `lint`: 40 -> 29
 - [x] `lint_parse_error`: 24
 - [x] `commas_linter`: 23
 - [x] `object_usage_linter`: 21
 - [x] `sprintf_linter`: 19
 - [x] `normalize_exclusions`: 18
 - [x] `fix_eq_assigns`: 18
 - [x] `available_linters`: 18
 - [x] `namespace_linter`: 18
 - [x] `closed_curly_linter`: 17
 - [x] `split_path`: 17
 - [x] `fix_tab_indentations`: 16
 - [x] `lint_dir`: 16
 - [x] `open_curly_linter`: 16

PR #1206, opened 2022-05-24
add helpers, simplify logic in path_linters

MichaelChirico:
Part of #1186


R/path_linters.R
```diff
@@ -101,41 +100,37 @@ is_valid_long_path <- function(path, lax = FALSE) {
...
+split_path <- function(dirs, prefix) {
+  # add root dir if needed
+  i <- 1L
+  for (dir in seq_along(dirs)) {
+    if (nzchar(dirs[[i]])) {
+      break
+    }
+    i <- i + 1L
+  }
+  i <- i - 1L
```
AshesITR, line 122: So `i` is the last index such that `!nzchar(dirs[[i]])`.
But `nzchar` is vectorized, so

```r
i <- match(TRUE, nzchar(dirs)) - 1L
```

right?

MichaelChirico: basically -- need to account for `dirs = ""` too. done.

R/path_linters.R
```diff
@@ -101,41 +100,32 @@ is_valid_long_path <- function(path, lax = FALSE) {
...
-    substr(path, 1L, 1L)
-  )
+  Map(split_path, strsplit(path, sep), substr(path, 1L, 1L))
+}
+
+split_path <- function(dirs, prefix) {
+  # add root dir if needed
+  nonempty_dirs <- nzchar(dirs)
+  i <- match(TRUE, nonempty_dirs) - 1L
+  if (is.na(i)) i <- length(dirs)
```
AshesITR, line 117: Use `default` arg of `match()` instead.

AshesITR (approved):

PR #1209, opened 2022-05-24
Clarify NEWS item about new function factory interface

MichaelChirico:


AshesITR:
Is it customary to include code chunks in NEWS files?

MichaelChirico:
I think it's fine. e.g. `ggplot2` has some:

https://github.com/tidyverse/ggplot2/blob/main/NEWS.md

`data.table` uses it a fair amount:

https://github.com/Rdatatable/data.table/blob/master/NEWS.md

PR #1211, opened 2022-05-24
improve location information for spaces_inside_linter

### LINKED ISSUE #1205: Improve spaces_inside_linter location info
It would be awesome if we modified the ranges to include the entire whitespace and we can add a `preceding-siblling::*[1]` to the OP-RIGHT-PAREN path to arrive at the `1` in this case.
Ideal output imo would be

```
a[   1   ]
  ^~~

a[   1   ]
      ^~~
```

__Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/1169#issuecomment-1135436480__

AshesITR:
Fixes #1205


MichaelChirico (approved):
Thanks!

might be nice to start writing tests where we can directly read the output of the test instead of doing mental math on the column numbers. something like expect_lint_output() that shows how the lint will print directly in the test file.

but that's an issue for another time.

AshesITR:
To be honest, I looked at the console output of `print(lint(...))` until satisfied and updated the expected numbers 😆

MichaelChirico:
same as I'd do now! would make our lives easier if we built that into the testing directly...

PR #1214, opened 2022-05-24
refactor commas_linter to XPath

AshesITR:
Part of #1186

### DIFF
--- a/R/commas_linter.R
+++ b/R/commas_linter.R
@@ -9,107 +9,29 @@
 #' @importFrom utils head
 #' @export
 commas_linter <- function() {
+  xpath_before <- "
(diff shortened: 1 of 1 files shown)


AshesITR:
Crazy how much code can be saved by XPath 😮

MichaelChirico:
glorious xpath 🙌🙌🙌

MichaelChirico (approved):
any need to mess with the source markers here?

AshesITR:
Yes, could be done as a follow-up, similar to `spaces_inside_linter`.
Currently, all lints mark the comma, which is fine for spaces before, but useless for spaces after.

PR #1216, opened 2022-05-24
reduce cyclocomp of sprintf_linter() to 6

AshesITR:
I extended `xml_nodes_to_lints()` to also vectorize on `lint_message` so I could use it.

### DIFF
--- a/R/sprintf_linter.R
+++ b/R/sprintf_linter.R
@@ -21,51 +21,51 @@ sprintf_linter <- function() {
 
     sprintf_calls <- xml2::xml_find_all(xml, xpath)
 
-    get_range_text <- function(content, line1, col1, line2, col2) {
-      lines <- content[line1:line2]
-      lines[length(lines)] <- substr(lines[length(lines)], 1L, col2)
-      lines[1L] <- substr(lines[1L], col1, nchar(lines[1L]))
-      lines
-    }
-
-    line1 <- as.integer(xml2::xml_attr(sprintf_calls, "line1"))
-    col1 <- as.integer(xml2::xml_attr(sprintf_calls, "col1"))
-    line2 <- as.integer(xml2::xml_attr(sprintf_calls, "line2"))
-    col2 <- as.integer(xml2::xml_attr(sprintf_calls, "col2"))
-
-    is_missing <- function(x) {
-      is.symbol(x) && !nzchar(x)
-    }
-
-    result <- .mapply(function(line1, col1, line2, col2) {
-      text <- get_range_text(source_expression$file_lines, line1, col1, line2, col2)
-      expr <- tryCatch(parse(text = text, keep.source = FALSE)[[1L]], error = function(e) NULL)
-      if (is.call(expr) &&
-        is.language(expr[[1L]]) &&
-        is.character(expr[[2L]])) {
-        if (length(expr) >= 3L) {
-          for (i in 3L:length(expr)) {
-            if (!is_missing(expr[[i]]) && !is.atomic(expr[[i]])) {
-              expr[[i]] <- 0L
-            }
-          }
-        }
-
-        res <- tryCatch(eval(expr, envir = baseenv()), warning = identity, error = identity)
-        if (inherits(res, "condition")) {
-          Lint(
-            filename = source_expression$filename,
-            line_number = line1,
-            column_number = col1,
-            type = "warning",
-            message = conditionMessage(res),
-            line = source_expression$file_lines[line1],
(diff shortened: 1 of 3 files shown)


R/xp_utils.R
```diff
@@ -43,7 +44,13 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
...
   }
   if (inherits(xml, "xml_nodeset")) {
-    lints <- lapply(xml, xml_nodes_to_lints, source_expression, lint_message, type, match_after_end)
+    if (is.character(message) && length(message) > 1L) {
+      lints <- .mapply(xml_nodes_to_lints, list(xml = xml, lint_message = lint_message), list(
+        source_expression = source_expression, type = type, match_after_end = match_after_end
+      ))
+    } else {
+      lints <- lapply(xml, xml_nodes_to_lints, source_expression, lint_message, type, match_after_end)
+    }
```
AshesITR, line 53: Not sure about the most elegant way. A simple `mapply()` will likely choke on `source_expressions` being a list.

R/sprintf_linter.R
```diff
@@ -21,51 +21,52 @@ sprintf_linter <- function() {
...
-      lines[length(lines)] <- substr(lines[length(lines)], 1L, col2)
-      lines[1L] <- substr(lines[1L], col1, nchar(lines[1L]))
-      lines
-    }
-
-    line1 <- as.integer(xml2::xml_attr(sprintf_calls, "line1"))
-    col1 <- as.integer(xml2::xml_attr(sprintf_calls, "col1"))
-    line2 <- as.integer(xml2::xml_attr(sprintf_calls, "line2"))
-    col2 <- as.integer(xml2::xml_attr(sprintf_calls, "col2"))
+    message <- vapply(sprintf_calls, function(xml) {
```
MichaelChirico, line 24: this is quite a long inline function, at a glance I don't see it inheriting from the function environment -- let's make it its own helper?

R/xp_utils.R
```diff
@@ -27,7 +27,8 @@ xp_text_in_table <- function(table) {
 #'   to the `Lint` object. If `lint_message` is a `function`,
 #'   this function is first applied to `xml` (so it should be a
 #'   function taking an `xml_node` as input and must produce a
-#'   length-1 character as output).
+#'   length-1 character as output). If `lint_message` is a character vector the same length as `xml`,
+#'   the `i`-th lint will be given the `i`-th message
```
MichaelChirico, line 31: final `.`

R/xp_utils.R
MichaelChirico: Seemed to do OK:

```
mapply(c, as.list(1:10), as.list(10:1), SIMPLIFY=FALSE)
mapply(c, replicate(10, sample(5, sample(5, 1)), simplify = FALSE), as.list(10:1), SIMPLIFY=FALSE)
mapply(c, replicate(10, as.list(sample(5, sample(5, 1))), simplify = FALSE), as.list(10:1), SIMPLIFY=FALSE)
```

AshesITR: What does
```
mapply(list, xml = 1:10, source_expression = list(1:6), message = paste("m", 1:10))
```
do?

MichaelChirico: throws a warning because of 6 < 10, but it gets recycled and then elt `i` of output is `list(xml[[ii]], source_expression[[ii %% 10]], message[[ii]])` (when `SIMPLIFY=FALSE`, then `simplify2array` of this otherwise)

AshesITR: Yeah, but what we want is `source_expressions` to stay the same for every call, i.e. be recycled as length-1.
I'll test out if the `.mapply()` branch also works for the `length(lint_message) == 1L` case.

AshesITR: No, it doesn't because if `lint_message` is a function, `.mapply()` stops with "object of type closure is not subsettable".

MichaelChirico: the issue is `source_expression` has 6 elements and `mapply` will try and iterate over them right? won't `source_expression = list(source_expression)` solve that?

MichaelChirico: c.f.

```
l = as.list(1:6)
x = 1:10
y = 10:1
mapply(c, x, list(l), y, SIMPLIFY=FALSE)
```

AshesITR: Yeah, but it doesn't really help in any way compared to just passing it via `MoreArgs`, which is essentially what happens with the `.mapply()` call I wrote.

MichaelChirico (approved):

PR #1217, opened 2022-05-24
reduce cyclocomp of normalize_exclusions() to 15

AshesITR:
Part of #1186

 - use names2() instead of branching conditional on is.null(names(...)) and treating names() == "" separately
 - move full_line_exclusions out of the any(unnamed) branch

### DIFF
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -213,7 +213,7 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
   }
 
   x <- as.list(x)
-  unnamed <- names2(x) == ""
+  unnamed <- !nzchar(names2(x))
   if (any(unnamed)) {
 
     # must be character vectors of length 1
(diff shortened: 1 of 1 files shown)


R/exclude.R
```diff
@@ -212,12 +212,15 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
     return(list())
   }
 
-  # no named parameters at all
-  if (is.null(names(x))) {
+  x <- as.list(x)
+  unnamed <- names2(x) == ""
+  if (any(unnamed)) {
```
MichaelChirico, line 217: I'd do

```r
named <- nzchar(names2(x))
if (!all(named))
```

(mainly to use `nzchar()`)

MichaelChirico: (or use `!nzchar()` to keep using `unnamed[i]` below)

R/exclude.R
```diff
@@ -228,56 +231,34 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
...
+  if (any(full_line_exclusions)) {
 
-      if (any(bad)) {
-        stop("Full line exclusions must be numeric or integer vectors. items: ",
-             toString(which(bad)),
-             " are not!",
-             call. = FALSE)
-      }
+    # must be integer or numeric vectors
+    bad <- vapply(
```
MichaelChirico, line 244: maybe better to just get the `vapply(x, is.numeric)` vector & compare with `full_line_exclusions` as a vector?

MichaelChirico (approved):

PR #1221, opened 2022-05-25
simplify open_curly_linter using xpath

### LINKED ISSUE #1220: Drop lint for trailing whitespace after { ?
As seen here:

https://github.com/r-lib/lintr/blob/c68c11e269bf2b5c15335cbfde163f05734417a9/tests/testthat/test-open_curly_linter.R#L35-L37

But not matched by the new `brace_linter`:

```
lint("a <- function() {  \n}", brace_linter())
```

I think the new behavior is better -- `trailing_whitespace_linter()` is the one designed to lint in this case.

OTOH, `open_curly_linter()` is being marked as deprecated -- is it worth making such a change?

This came up working on #1186. Linting on whitespace also makes it difficult to convert to XPath since the XML tree is unaware of the whitespace.

MichaelChirico:
Closes #1220; part of #1186

### DIFF
--- a/R/open_curly_linter.R
+++ b/R/open_curly_linter.R
@@ -10,64 +10,55 @@
 #' @export
 open_curly_linter <- function(allow_single_line = FALSE) {
   lintr_deprecated("open_curly_linter", new = "brace_linter", version = "2.0.1.9001", type = "Linter")
-  Linter(function(source_expression) {
-    lapply(
-      ids_with_token(source_expression, "'{'"),
-      function(id) {
 
-        parsed <- with_id(source_expression, id)
-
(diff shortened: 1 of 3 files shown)


AshesITR (approved):
LGTM, but I can't find the test case `"a <- function() { \n}"` (only `"a <- function() {\n}"`).
Was that intentional?

MichaelChirico:
> LGTM, but I can't find the test case `"a <- function() { \n}"` (only `"a <- function() {\n}"`). Was that intentional?

I guess we should keep that in, just changing it to a true negative

AshesITR (approved):

PR #1222, opened 2022-05-25
use helpers in namespace_linter

MichaelChirico:
Towards #1186 

TIL I don't understand cyclomatic complexity at all :joy:

The flow looks much improved to me, but alas, the complexity of the new helper is HIGHER than at head :facepalm: 

Filing PR anyway because I still think it's an improvement and a good starting point for more simplification.


R/namespace_linter.R
```diff
@@ -33,97 +33,124 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
-          line = source_expression$file_lines[[line1]],
-          ranges = list(c(col1, col2))
-        ))
+    results <- lapply(
+      seq_along(ns_nodes),
+      function(i) {
+        build_namespace_lint(pkgs[[i]], syms[[i]], pkg_nodes[[i]], sym_nodes[[i]], check_exports, check_nonexports)
       }
-    })
+    )
```
AshesITR, line 41: Use `mapply()`?

R/namespace_linter.R
```diff
@@ -33,97 +33,124 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
 
     results[!vapply(results, is.null, logical(1L))]
   })
 }
+
+build_namespace_lint <- function(package, symbol, package_node, symbol_node, check_exports, check_nonexports) {
+  if (!package %in% installed_packages) {
+    return(build_uninstalled_lint(package, package_node, source_expression))
+  }
+  if (!check_exports && !check_nonexports) {
```
AshesITR, line 51: Not sure if `cyclocomp()` knows that `return()` is terminal. Maybe it'll report a better value if these are `else if`s?

R/namespace_linter.R
```diff
@@ -33,97 +33,124 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
+  Lint(
+    filename = source_expression$filename,
+    line_number = line1,
+    column_number = col1,
+    type = "warning",
+    message = sprintf("Package '%s' is not installed.", package),
+    line = source_expression$file_lines[[line1]],
+    ranges = list(c(col1, col2))
+  )
+}
```
AshesITR, line 156: Aren't all these special cases of `xml_nodes_to_lint()`?

AshesITR:
Feels like we can refactor this even more using the technique from the `sprintf_linter()` PR.

MichaelChirico:
working on a different approach now... IIUC the main issue here is trying to be conservative about calling `getNamespace()`, `getNamespaceExports()`, and `ls(<namespace>)` unless necessary, which makes straight vectorization impossible. So I'm basically vectorizing in chunks.

AshesITR:
I was thinking something like
```
messages <- vapply(ns_nodes, function(node) {
   # do the sequential checks
  NA_character_ # if all good
}, character(1L))

xml_nodes_to_lints(sym_nodes[!is.na(messages)], lint_message = messages[!is.na(messages)], ...)
```

MichaelChirico:
I think the nuance here is some lints land on `pkg::` and others land on `::foo`. we could go to `parent::expr` for consistency but that'll throw off the markers.

Two more things I want to improve here:

(1) lots of shared code in the `build*` helpers to go from lint component vectors --> `Lint()` calls
(2) put things into a data.frame to subset instead of repetitive vector subsetting

but checking in now to get feedback on the basic idea of the refactor before putting in more time

R/namespace_linter.R
MichaelChirico: (completely refactored again)

R/namespace_linter.R
MichaelChirico: (completely refactored again)

R/namespace_linter.R
MichaelChirico: i think so... will need the vectorized messages though in the current approach

AshesITR:
I'd try to go down the route with vectorized `lint_message`.
Not sure the data frame is worth it, I think the repeated indexing is easier to understand.

Regarding lint locations: The unneeded `:::` seems like the only lint where placing the column actually matters.
WDYT about always linting
```
foo(linted:::part())
    ~~~~~~^~~~~~~
```

MichaelChirico:
> WDYT about always linting

any way to do that with current `xml_nodes_to_lints()` setup?

AshesITR:
I'm afraid no. The `column_marker` is the problem.
Would need to allow a custom XPath expression for the marker (`number(./NS-GET/@col1)` maybe?)

MichaelChirico:
awesome! great to get mileage out of vectorized `lint_message` like this already

tests/testthat/test-namespace_linter.R
```diff
@@ -85,7 +85,7 @@ test_that("returns the correct linting", {
 
   expect_lint(
     "stats::sd(c(1,2,3))\nstats::sdd(c(1,2,3))",
-    list(line = "stats::sdd(c(1,2,3))"),
+    list(line = c("2" = "stats::sdd(c(1,2,3))")),
```
MichaelChirico, line 88: PTAL -- consequence of shifting to `xml_nodes_to_lints()` -- unsure if this is fine or maybe reflects something to edit in `xml_nodes_to_lints()` -- `lines[[line1]]`? any reason not to strip attributes?

MichaelChirico: ok. I'll do that in separate PR in case it breaks other stuff. as well as add some more direct tests of xml_nodes_,to_lints() which are clearly needed!

AshesITR (approved):

PR #1229, opened 2022-05-25
simplifications to fix_eq_assigns

MichaelChirico:
Part of #1186


R/get_source_expressions.R
```diff
@@ -592,88 +592,78 @@ tab_offsets <- function(tab_columns) {
...
+  )
 
-  true_locs <- which(expr_locs == TRUE)
-  for (i in seq_along(true_locs)) {
+  for (i in seq_len(n_expr)) {
     start <- true_locs[i]
 
     end <- true_locs[i]
     j <- end + 1L
-    while (j <= length(expr_locs) && expr_locs[j] == FALSE) {
```
MichaelChirico, line 639: this can likely be vectorized (looks like a `findInterval()` thing?) but it's a bit hard to test as my 3.6.3 doesn't run this region except for a trivial error test case (`{ x = }` in test-error.R)

MichaelChirico: are we confident our tests are sufficient? don't want to touch it unless we know it's covered well (i.e., in a case where `n_expr > 1`)

AshesITR: The line isn't hit on R 4.2.0 either...

MichaelChirico: covr should tell us that, and i think it's likely hit (it's hit in the weird edge case of syntax error `{ x = }` in test-error.R), but the hit is in a trivial case when `n_expr=1`, so we don't really learn anything.

what'll be interesting is to stop() if `n_expr > 1`... basically wrote as such in the TODO.

AshesITR (approved):

PR #1234, opened 2022-05-25
refactor xml_nodes_to_lints() lint location to XPath

### LINKED ISSUE #1215: improve source markers for commas_linter
Yes, could be done as a follow-up, similar to `spaces_inside_linter`.
Currently, all lints mark the comma, which is fine for spaces before, but useless for spaces after.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/issues/1214#issuecomment-1136419741_

AshesITR:
Fixes #1215 

- [x] move `xml_nodes_to_lints()` to its own file
- [x] always return unnamed line info in `Lint()`s
- [x] allow passing a `list()` containing `xml_node`s, even if it's not classed `xml_nodeset`
- [x] refactor `match_after_end` usages to XPath based locations
- [x] use new XPath based locations to improve location information for `commas_linter()`
- [x] update tests for `commas_linter()`
- [x] add tests for `xml_nodes_to_lints()`
- [x] refactor `spaces_inside_linter()` to use `xml_nodes_to_lints()`


AshesITR:
`spaces_inside_linter()` can also probably be refactored to use `xml_nodes_to_lints()` now.
Waiting for a round of feedback before I go for that, but I consider it within the scope of this PR.

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,86 @@
...
+#' @param range_start_xpath XPath expression to return the range start location of the lint.
+#'   Defaults to the start of the expression matched by `xml`.
+#' @param range_end_xpath XPath expression to return the range end location of the lint.
+#'   Defaults to the end of the expression matched by `xml`.
+#' @return For `xml_node`s, a `lint`. For `xml_nodeset`s, `lints` (a list of `lint`s).
+#' @export
+xml_nodes_to_lints <- function(xml, source_expression, lint_message,
+                               type = c("style", "warning", "error"),
+                               column_number_xpath = "number(./@col1)",
+                               range_start_xpath = "number(./@col1)",
```
MichaelChirico, line 30: let's set range_start_xpath=column_number_xpath by default, that way they change in concert instead of user needing to write the path twice

AshesITR: Sounds good, but I would swap the role (`column_number_xpath = range_start_xpath`)

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,86 @@
...
+          type = type,
+          column_number_xpath = column_number_xpath,
+          range_start_xpath = range_start_xpath,
+          range_end_xpath = range_end_xpath
+        )
+      )
+    } else {
+      lints <- lapply(
+        xml, xml_nodes_to_lints,
+        source_expression = source_expression, lint_message = lint_message, type = type,
```
MichaelChirico, line 51: each argument on its own line please

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,86 @@
...
+      )
+    }
+    class(lints) <- "lints"
+    return(lints)
+  } else if (!inherits(xml, "xml_node")) {
+    stop("Expected an xml_nodeset or xml_node, got an object of class(es): ", toString(class(xml)))
+  }
+  type <- match.arg(type, c("style", "warning", "error"))
+  line1 <- xml2::xml_attr(xml, "line1")
+  col1 <- as.integer(xml2::xml_find_num(xml, range_start_xpath))
```
MichaelChirico, line 63: does xml_find_num not return integer 😢

MichaelChirico: filed https://github.com/r-lib/xml2/issues/365

AshesITR: Unfortunately, no. There also isn't a `xml_find_int()`.
```
> class(xml2::xml_find_num(xml2::xml_new_root("a"), "number(1)"))
[1] "numeric"
```

MichaelChirico: let's define our own wrapper for now? just wrapping as.integer(xml_find_num())

AshesITR: These three lines are the only uses currently. Are you sure that's not overkill?

R/commas_linter.R
```diff
@@ -24,12 +24,16 @@ commas_linter <- function() {
     before_lints <- xml_nodes_to_lints(
       xml2::xml_find_all(xml, xpath_before),
       source_expression = source_expression,
-      lint_message = "Commas should never have a space before."
+      lint_message = "Commas should never have a space before.",
+      range_start_xpath = "number(./preceding-sibling::*[1]/@col2 + 1)",
```
MichaelChirico, line 28: let's have a practice of commenting what the xpath is accomplishing in (non-trivial) non-standard cases?

AshesITR: SGTM

R/spaces_inside_linter.R
```diff
@@ -35,42 +35,35 @@ spaces_inside_linter <- function() {
...
-      Lint(
-        filename = source_expression$filename,
-        line_number = line1,
-        column_number = col1,
-        message = switch(
+
+    left_lints <- xml_nodes_to_lints(
+      left_expr,
+      source_expression = source_expression,
+      lint_message = function(expr) {
```
MichaelChirico, line 42: we could also vectorize this pretty easily

R/xml_nodes_to_lints.R
MichaelChirico: slightly... what I have in mind is if that xml2 FR does land, well likely backport the same wrapper for a while (?)

up to you, not a blocker

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,89 @@
...
+#'   `xml_nodeset` object (to generate several `Lint`s), e.g. as returned by
+#'   [xml2::xml_find_all()] or [xml2::xml_find_first()].
+#' @param source_expression A source expression object, e.g. as
+#'   returned typically by [lint()], or more generally
+#'   by [get_source_expressions()].
+#' @param lint_message The message to be included as the `message`
+#'   to the `Lint` object. If `lint_message` is a `function`,
+#'   this function is first applied to `xml` (so it should be a
+#'   function taking an `xml_node` as input and must produce a
+#'   length-1 character as output). If `lint_message` is a character vector the same length as `xml`,
```
MichaelChirico, line 17: do we want to add a test to enforce the equal-lengths part more strictly?

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,89 @@
...
+#'   the `i`-th lint will be given the `i`-th message.
+#' @param column_number_xpath XPath expression to return the column number location of the lint.
```
MichaelChirico, line 19: let's add a reference link to xml_find_num

R/xml_nodes_to_lints.R
AshesITR: Nonequal lengths will generate a recycling warning by `.mapply()`, right?

R/xml_nodes_to_lints.R
```diff
@@ -0,0 +1,89 @@
...
+#'   Defaults to the start of the expression matched by `xml`.
+#' @param range_start_xpath XPath expression to return the range start location of the lint.
+#'   Defaults to the start of the expression matched by `xml`.
```
MichaelChirico, line 22: not quite accurate re:default

R/xml_nodes_to_lints.R
MichaelChirico: right... tidyverse usually hates recycling though and tends to error instead

MichaelChirico:
last few comments here. good overall! I think this API maximizes flexibility & has good defaults.

R/T_and_F_symbol_linter.R
```diff
@@ -50,8 +50,8 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
           sprintf(fmt, replacement_map[[symbol]], symbol)
         },
         type = "style",
-        column_number_xpath = "number(./@col2 + 1)",
-        range_end_xpath = "number(./@col2 + 1)"
+        column_number_xpath = "number(./@col2 + 1)", # mark at end
+        range_end_xpath = "number(./@col2 + 1)" # end after T/F for easy fixing
```
MichaelChirico, line 54: case where it's trivial (bump by 1) doesn't need a note I'd say. but I'm fine either way

R/xml_nodes_to_lints.R
```diff
@@ -3,6 +3,13 @@
 #' Convenience function for converting nodes matched by XPath-based
 #'   linter logic into a [Lint()] object to return.
 #'
+#' @details
+#' The location XPaths, `column_number_xpath`, `range_start_xpath` and `range_end_xpath` are evaluated using
+#' [xml2::xml_find_num()] and will usually be of the form
+#'   "number(./relative/xpath)"
+#' Note that the location line number cannot be changed and lints spanning multiple lines will ignore `range_end_xpath`.
+#' `column_number_xpath` and `range_start_xpath` must always refer to locations on the starting line of the `xml` node.
```
MichaelChirico, line 11: should right? we don't enforce that?

R/T_and_F_symbol_linter.R
AshesITR: I find the reason why we do it interesting enough to document.

R/xml_nodes_to_lints.R
AshesITR: Seems impossible, as the `number()` XPath can do anything to arrive.
What we could check is if `start <= column_number <= nchar(line)`, but I'd leave this sanity checking out of this already large PR.

R/spaces_inside_linter.R
AshesITR: With a minor tweak we can even use just one `xml_nodes_to_lints()` call 😄

R/xml_nodes_to_lints.R
MichaelChirico: oh, not suggesting we enforce it! just changing the doc wording :)

R/xml_nodes_to_lints.R
```diff
@@ -67,7 +68,7 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
     class(lints) <- "lints"
     return(lints)
   } else if (!inherits(xml, "xml_node")) {
-    stop("Expected an xml_nodeset or xml_node, got an object of class(es): ", toString(class(xml)))
+    stop("Expected an xml_nodeset, a list of xml_nodes or an xml_node, got an object of class(es): ", toString(class(xml)))
```
MichaelChirico, line 71: too wide now

R/xml_nodes_to_lints.R
```diff
@@ -95,3 +96,8 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
     ranges = list(c(col1, col2))
   )
 }
+
+is_nodeset_like <- function(xml) {
+  inherits(xml, "xml_nodeset") ||
+    (is.list(xml) && all(vapply(xml, inherits, logical(1L), what = "xml_node")))
+}
```
MichaelChirico, line 103: terminal newline

R/xml_nodes_to_lints.R
AshesITR: Let's stick with this for now, we also have plenty of `as.integer(xml2::xml_attr(...))` calls.

R/xml_nodes_to_lints.R
AshesITR: Sounds like a good candidate for a sanity-check follow up (in addition to the `start <= column <= nchar(line)` check discussed below.

R/xml_nodes_to_lints.R
AshesITR: ah 😆

AshesITR:
The build failure looks sporadic. Restarting should fix it.

MichaelChirico (approved):

PR #1235, opened 2022-05-25
refactor closed_curly_linter() to XPath (reduces cyclocomp to 4)

AshesITR:
Part of #1186 

I just copied the refactored XPath logic from `brace_linter()`.


MichaelChirico:
Great! glad I didn't work more on this :)

One thing I did catch is that the docs don't capture all the exceptions (e.g. `})` or `}]` or `},`), shall we add notes to the docs here?

https://github.com/r-lib/lintr/compare/main...cyclo-closed-curly

MichaelChirico (approved):

Issue #1199, opened 2022-05-23
Performance: Generate XPaths ahead of time


AshesITR:
Some of the XPath generation code uses `glue::glue()` with configuration-dependent (but not file-dependent) input.
Currently, on every `lint()` these XPaths are recomputed, amounting to quite a few ms spent on `glue::glue()` e.g. in the test suite.

It is much more efficient to compute the XPath in the linter factory instead, whenever that is possible.

MichaelChirico:
#1222 had me thinking:

`namespace_linter` is a case where there's a lot of logic we may be stuck keeping inside the linter, because it runs `.packages()`, `getNamespace()`, etc. which in principle could differ between (lintr config) compile time & run time.

Where that's not the case, I think it's fine to move logic into the factory by default (for consistency), even when glue() is not involved.

PR #1246, opened 2022-05-26
finish expect_*_linter migration to XPath logic in factory

MichaelChirico:
Part of #1199 

Giving up on only putting one vectorized lint per PR -- if you are reviewing sequentially, it should be pretty standard by now. I'll split off separate PRs again later if there's something unique or unusual.

### DIFF
--- a/R/expect_named_linter.R
+++ b/R/expect_named_linter.R
@@ -8,6 +8,14 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 expect_named_linter <- function() {
+   xpath <- "//expr[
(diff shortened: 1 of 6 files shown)


AshesITR (approved):
Does it even make sense to support the function interface to `lint_message` anymore? It can always be replaced by `vapply(xml, lint_message_fun, character(1L))`.

MichaelChirico:
> Does it even make sense to support the function interface to `lint_message` anymore? It can always be replaced by `vapply(xml, lint_message_fun, character(1L))`.

Definitely have the same thought -- let me get through the sweep of linters first to be sure there's no surprises. But it sounds right.

PR #1251, opened 2022-05-26
support multi-line calls in sprintf() again

### LINKED ISSUE #1249: `sprintf_linter()` errors for multi-line `sprintf()` statements with comments
`Error in parsed_expr[[2L]] (sprintf_linter.R#24): subscript out of bounds`

This is because `xml2::xml_text()` doesn't respect newlines and thus
```r
sprintf(
  "asd", # comment
  b
)
```

when run through `xml_text()` turns into `sprintf("asd", # commentb)`, which is not parseable.

AshesITR:
fixes #1249
added a regression test

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -238,3 +238,36 @@ get_r_string <- function(s, xpath = NULL) {
   is.na(out) <- is.na(s)
   out
 }
+
+#' Convert XML node to R code within
+#'
+#' NB this is not equivalent to `xml2::xml_text(xml)` in the presence of line breaks
+#'
+#' @param xml An `xml_node`.
+#'
+#' @return A source-code equivalent of `xml` with unnecessary whitespace removed.
+#'
+#' @noRd
+get_r_code <- function(xml) {
+  # shortcut if xml has line1 and line2 attrs and they are equal
+  # if they are missing, xml_attr() returns NA, so we continue
+  if (isTRUE(xml2::xml_attr(xml, "line1") == xml2::xml_attr(xml, "line2"))) {
+    return(xml2::xml_text(xml))
+  }
+  # find all unique line numbers of atoms (i.e. expressions
+  line_numbers <- sort(unique(xml2::xml_find_num(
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -238,3 +238,36 @@ get_r_string <- function(s, xpath = NULL) {
...
+#'
+#' @param xml An `xml_node`.
+#'
+#' @return A source-code equivalent of `xml` with unnecessary whitespace removed.
+#'
+#' @noRd
+get_r_code <- function(xml) {
+  # shortcut if xml has line1 and line2 attrs and they are equal
+  # if they are missing, xml_attr() returns NA, so we continue
+  if (isTRUE(xml2::xml_attr(xml, "line1") == xml2::xml_attr(xml, "line2"))) {
```
MichaelChirico, line 254: this makes it non-vectorized, where `xml_text()` was earlier -- is that intentional?

R/utils.R
```diff
@@ -238,3 +238,36 @@ get_r_string <- function(s, xpath = NULL) {
...
+    return(xml2::xml_text(xml))
+  }
+  # find all unique line numbers of atoms (i.e. expressions
```
MichaelChirico, line 257: just missing `)` or anything else?

R/utils.R
AshesITR: `capture_sprintf_warning()` is already non-vectorized, so it's fine.

R/utils.R
AshesITR: actually a relic of an older approach which wouldn't have worked with multi-line strings.

MichaelChirico:
would it be simpler to back up to using the `source_expression$lines` in this case once we know the line/column where the `sprintf()` call starts/stops

AshesITR:
We have a `get_content()`. That would work for all nodes with full location info (line1, line2, col1, col2).
I'll see if there is an easy way to use that.

MichaelChirico:
Worth adding a TODO instead?

PR #1252, opened 2022-05-26
support named fmt argument and skip linting calls with dots

### LINKED ISSUE #626: sprintf_linter not working on edge cases
In the following cases, `sprintf_linter` might not work properly:

1. `...` is used in the argument (https://github.com/jimhester/lintr/pull/625#issuecomment-735397084):

```r
f <- function(id, ...) {
  sprintf("%d %d, %d", id, ...) ### too few arguments
}
```

2. Supplying arguments in non-typical order (https://github.com/jimhester/lintr/pull/625#issuecomment-735398166):

```r
f(42, fmt = "bla")
```

These cases might be not worth fixing as they are quite far from standard usage.

AshesITR:
fixes #626

### DIFF
--- a/R/sprintf_linter.R
+++ b/R/sprintf_linter.R
@@ -7,7 +7,7 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 sprintf_linter <- function() {
-  xpath_positional <- "//expr[
+  xpath <- "//expr[
     expr/SYMBOL_FUNCTION_CALL[text() = 'sprintf'] and
     (
       OP-LEFT-PAREN/following-sibling::expr[1]/STR_CONST or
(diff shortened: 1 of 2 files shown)


R/sprintf_linter.R
```diff
@@ -72,10 +82,7 @@ zap_extra_args <- function(parsed_expr) {
 capture_sprintf_warning <- function(xml) {
   text <- xml2::xml_text(xml)
   parsed_expr <- try_silently(parse(text = text, keep.source = FALSE)[[1L]])
-  if (!is.character(parsed_expr[[2L]])) { # sprintf(fmt = "") must be a constant format
-    return(NA_character_)
-  }
```
AshesITR, line 77: This couldn't occur because the XPath guarantees as STR_CONST to be present.

AshesITR:
WDYT about partially supporting `...`?
What comes to mind is parse the `fmt` to detect the necessary number of arguments, by working with the regex `(?<!%)%(\d+\$)?[^%]` we should find all format specifiers and capture their index parameter. From there it should be straight-forward to find out how many arguments should be placed in `...` to satisfy the need.

MichaelChirico:
Sounds reasonable enough... I would file as a separate issue though, doesn't sound like a big priority :) 

See also here for a pretty thoroughly-tested regex for printf templates:

https://github.com/MichaelChirico/potools/blob/b273ea86210398215dc3105c5ecaab007bab6190/R/utils.R#L126-L136

MichaelChirico (approved):

PR #1281, opened 2022-05-26
 refactor commented_code_linter() to use xml_nodes_to_lints()

### LINKED ISSUE #1274: refactor `comment_linters()` to use xml_nodes_to_lints()

AshesITR:
fixes #1274


R/comment_linters.R
```diff
@@ -54,32 +54,34 @@ commented_code_linter <- function() {
...
+    is_parsable <- which(vapply(code_candidates[, "code"], parsable, logical(1L)))
 
-    # TODO(michaelchirico): convert this to xml_nodes_to_lint()?
-    lapply(rownames(na.omit(code_candidates)), function(code_candidate) {
-      is_parsable <- parsable(code_candidates[code_candidate, "code"])
-      if (is_parsable) {
-        comment_node <- all_comment_nodes[[as.integer(code_candidate)]]
-        line_number <- as.integer(xml2::xml_attr(comment_node, "line1"))
-        column_offset <- as.integer(xml2::xml_attr(comment_node, "col1")) - 1L
+    lint_list <- xml_nodes_to_lints(
```
MichaelChirico, line 59: hmm I'd say no need to force xml_nodes_to_lint if it's not up to the task -- here may be such a case

AshesITR: I like that it abstracts away the boilerplate. The implementation looks much cleaner imo.

AshesITR: on tools/QC.R we lose 270ms (34 lints produced) but this is surely recovered by the optimizations to `xml_nodes_to_lints()` in #1284.

If you want, we merge that first and I do another performance comparison?

MichaelChirico: I'm not going to lose sleep over 270ms in one of the most unwieldy R scripts around :)

MichaelChirico (approved):

R/comment_linters.R
AshesITR: True :) will hold off just out of scientific curiosity.

PR #1286, opened 2022-05-26
don't lint `a[NULL]` in `extraction_operator_linter()`

### LINKED ISSUE #1273: `extraction_operator_linter()` shouldn't lint `code[NULL]`
spin-off from #1261:

>Why do we lint `NULL_CONST` at all?

```r
letters[NULL]
#> character(0)
letters[[NULL]]
#> Error in letters[[NULL]]: attempt to select less than one element in get1index
```

AshesITR:
fixes #1273

needs a news bullet in #834 once merged

### DIFF
--- a/R/extraction_operator_linter.R
+++ b/R/extraction_operator_linter.R
@@ -7,7 +7,7 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 extraction_operator_linter <- function() {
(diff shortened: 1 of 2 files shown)


MichaelChirico:
> needs a news bullet in #834 once merged

can we start adding these as to-do bullets to #833 ? I'm afraid they'll get lost

MichaelChirico (approved):

AshesITR:
The reason I mention the PR is so they show up in the PR timeline so they don't get lost 😁

MichaelChirico:
that's true -- still maybe a bit noisy. bullets seems more direct

Issue #1257, opened 2022-05-26
Deprecate is.function(lint_message) in xml_nodes_to_lints()


MichaelChirico:
Not really necessary now that `lint_message` accepts vectors -- h/t @AshesITR for the improvement.

PRs for #1199 handled most of the work here; what's left mostly needs some refactoring to get it going (I'll tally a list here once all the #1199 PRs have landed on `main`)

`grep -Fr "lint_message = function" R` gives us these to start:

 - [x] #1289 `yoda_test_linter()`
 - [x] #1291 `seq_linter()`
 - [x] #1294 `unused_import_linter()`
 - [x] #1295 `undesirable_operator_linter()`
 - [x] #1296 `package_hooks_linter()`

AshesITR:
Oh, that's a nicer way to track  the checklists, without creating mini-issues.

PR #1291, opened 2022-05-26
vectorize lint_message in seq_linter; add a test

MichaelChirico:
Part of #1257

Definitely still room for improvement here... another day :)


R/seq_linter.R
```diff
@@ -16,6 +16,16 @@ seq_linter <- function() {
     and expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
   ]")
 
+  ## The actual order of the nodes is document order
+  ## In practice we need to handle length(x):1
+  get_fun <- function(expr, n) {
+    funcall <- xml2::xml_find_chr(expr, sprintf("string(./expr[%d])", n))
+    fun <- gsub("\\(.*\\)", "(...)", trimws(funcall))
```
AshesITR, line 23: Is this `xp_call_name(expr, depth  = n)` with a `(...)` suffix?

MichaelChirico: not quite, no... the call we're at is `:`(x, y) in infix form, so call logic doesn't quite apply. and it only applies to one half of the expr (`1:length(x)` or `length(x):1`)...

i recall trying to simplify this previously and it being a bit of a pain. willing to revisit again later but for now focused on vectorizing as straightforwardly as possible.

AshesITR (approved):

Issue #1292, opened 2022-05-26
`release_bullets()` for future releases?


AshesITR:
What `release_bullets()` should we define for future (> 3.0.0) release issues of lintr?

CC @MichaelChirico @hadley for input.

hadley:
What release tasks do you currently have to manually remember to do?

MichaelChirico:
@hadley this is both of our first release of `lintr`, so I think we're both figuring it out as we go along :)

I think we should always do a .dev/compare_branches of old CRAN vs latest dev on a sizeable chunk of packages -- that's pretty much our revdeps since we don't expect many downstream packages to be on CRAN (maybe as Suggests?)

It would be good to have a nice source of scripts to get a sense of how things change outside of packages too. I will try and run internally at Google.

hadley:
Hmmm, and you can't even really use revdepcheck, since `.lintr` won't be included in CRAN packages.

AshesITR:
True (revdepcheck returns no problems), but do you know if the CRAN sources include it?
Then we could at least run on all packages that Suggest lintr.

MichaelChirico:
CRAN sources won't have anything .Rbuildignore'd:

https://github.com/r-lib/lintr/blob/19308c1d61dae2a9cc921a9cbd921c1df31c867e/.Rbuildignore#L13

I suggest we look for GitHub URLs in our reverse suggests and run compare_branches for all of them too to try and give a heads up about (1) upcoming breaking changes and (2) upcoming delta of lints for their repo

MichaelChirico:
oh that's right, `available.packages()` doesn't include the `URL` key of the package's DESCRIPTIONs -- @hadley is there a tool to retrieve those that you're aware of? if not we can scrape the canonical CRAN pages...

AshesITR:
@MichaelChirico You can use `read.dcf()` with an open URL file:

```r
f <- file("https://cran.r-project.org/web/packages/ggplot2/DESCRIPTION", open = "r")
read.dcf(f, "URL")[[1L]]

#> [1] "https://ggplot2.tidyverse.org,\nhttps://github.com/tidyverse/ggplot2"
```

MichaelChirico:
What's discussed here is seen in .dev/revdep*. We can add more incrementally later, not sure I see a reason to keep this open now.

PR #1304, opened 2022-05-27
Add a list of downstream repos, and script to update

MichaelChirico:
Related: #1292


revdep/get_urls.R
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/Rscript
+# Script to get URL keys from CRAN
+#   of packags that Suggest or Import lintr
+library(xml2)
+
+cran_db = as.data.frame(available.packages())
+lintr_re = "(^|\\s)lintr(,|\\s+\\(|$)"
```
AshesITR, line 7: `\blintr\b` is a more compact regex for this.

revdep/get_urls.R
```diff
@@ -0,0 +1,38 @@
...
+}
+
+urls = vector("list", length(lintr_pkg))
+url_xpath = "string(//tr[td[starts-with(text(), 'URL')]]/td/a)"
+# for loop makes debugging easier
+for (ii in seq_along(urls)) {
+  urls[[ii]] <- lintr_pkg[[ii]] %>%
+    retry_read() %>%
+    xml2::xml_find_chr(cran_pg, url_xpath)
+}
```
AshesITR, line 33: `read.dcf(..., fields = "URL")` from the raw description (`/DESCRIPTION` instead of `/index.html`) seems more elegant.
NB requires passing `file("https://...", open = "r")` to `read.dcf` because it doesn't automatically support URLs.

revdep/reverse-imports-urls
```diff
@@ -0,0 +1,48 @@
+http://github.com/soodoku/abbyyR
```
AshesITR, line 1: This file should be `.gitignore`'d imo, compare `revdepcheck` which only versions `failures.md`, `problems.md` and `README.md`.

revdep/get_urls.R
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/Rscript
```
AshesITR, line 1: I'd put this in `.dev` - `revdep` is used by `revdepcheck`. We shouldn't interfere with its folder structure IMO.

MichaelChirico: oh, that was my first thought, but then I saw this folder. didn't realize it's related to `revdepcheck`

revdep/reverse-imports-urls
MichaelChirico: that makes sense, I did have in mind a way for users to report their repos to us if they're not linked on CRAN. but that should be in a separate file since this one's automatic and the other manual. maybe reverse-import-urls-(public vs private) ?

revdep/get_urls.R
MichaelChirico: great! much better

revdep/reverse-imports-urls
AshesITR: Maybe something like `extra-check-urls` could be developed if and when the need arises.

AshesITR:
I noticed `.dev` isn't covered by our `lintr` suite. Do we want to do something about it?

.dev/get_urls.R
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/Rscript
+# Script to get URL keys from CRAN
+#   of packags that Suggest or Import lintr
+
+cran_db = as.data.frame(available.packages())
```
AshesITR, line 5: Use `<-' for assignments

.dev/get_urls.R
```diff
@@ -0,0 +1,47 @@
...
+  unname(substr(urls, starts, starts + sapply(matches, attr, "match.length") - 1L))
+}
+
+urls = character(length(lintr_pkg))
+# for loop makes debugging easier
+for (ii in seq_along(urls)) {
+  urls[ii] <- extract_url(lintr_pkg[ii])
+}
+
+urls = sort(extract_github_repo(urls[!is.na(urls)]))
```
AshesITR, line 45: Where we could add manual URLs.
Would be nice to automatically get a list of packages with missing git repos.
I noticed `fst` and `fstcore` for example.
Those are at `https://github.com/fstpackage/fst/` and `https://github.com/fstpackage/fstcore` respectively.

AshesITR: Could crawl `BugReports` as well, that would have found fst: `"https://github.com/fstpackage/fst/issues"`

.dev/get_urls.R
```diff
@@ -0,0 +1,47 @@
...
+    url(desc_url, open = "r")
+  }
+  Sys.sleep(initial_sleep)
+  url_conn = tryCatch(url(desc_url, open = "r"), error = retry, warning = retry)
+  pkg_data = read.dcf(url_conn, "URL")
+  drop(pkg_data)
+}
+
+extract_github_repo = function(urls) {
+  matches = gregexpr("https?://git(hub|lab).com/[a-zA-Z0-9_-]+/[a-zA-Z0-9._-]+", urls)
```
AshesITR, line 28: No need for global matching. You can use `regexpr()`.

.dev/get_urls.R
MichaelChirico: we could also back up to the cran org mirrors on GitHub, emailing the maintainers if they don't have a repo listed

MichaelChirico: problem is we won't see their .lintr... we could default to running _all_ our linters for those repos

MichaelChirico:
I guess we should also include a recommendation to add us as Suggests in the README? there's typically CRAN requirement all but it'll make things easier

.dev/get_urls.R
AshesITR: Let's see how many packages need to be manually found after crawling BugReports. Appsilon packages, e.g. shiny.semantic also provide a github issue link.

I wouldn't loose sleep if a few downstream packages can only be tested on the default lintr config.

revdep/reverse-imports-urls
MichaelChirico: i'll leave it in for now for easier reviewing but will remove+ignore before merging this

MichaelChirico:
> I noticed .dev isn't covered by our lintr suite. Do we want to do something about it?

i think it's fine... not really production scripts, fine if they're a little rough around the edges

.dev/get_urls.R
MichaelChirico: for the bugreports one... probably easier to just put those in the extra- manual file

.dev/get_urls.R
```diff
@@ -0,0 +1,50 @@
...
+  desc_url <- sprintf(desc_url_fmt, pkg)
+  retry <- function(cond) {
+    message("Sleepy... 😴")
+    Sys.sleep(retry_sleep)
+    url(desc_url, open = "r")
+  }
+  Sys.sleep(initial_sleep)
+  url_conn <- tryCatch(url(desc_url, open = "r"), error = retry, warning = retry)
+  pkg_data <- read.dcf(url_conn, "URL")
+  drop(pkg_data)
```
AshesITR, line 24: If you do `c("URL", "BugReports")` and then `paste(drop(pkg_data), collapse = "\n")`, you'll find a bunch more packages.

MichaelChirico: right but for the next step well also have to post-process the URLs 😩

but ok you convinced me. it's only a few extra lines.

AshesITR: Postprocessing works without modifications because the /issues path isn't part of the match :)

MichaelChirico: true... i guess there's an assumption that if both contain a github URL, they're the same, which is acceptable

AshesITR: Yep (or, at least, the one listed under URL is the correct one)

.dev/reverse-imports-no-repos
```diff
@@ -0,0 +1,15 @@
+package,maintainer
+adaptalint,conway.max1@gmail.com
```
MichaelChirico, line 2: definitely also want to .gitignore this one to avoid posting emails (even thought they're easily available from CRAN) -- again keeping it in the PR for now for viz

.dev/reverse-imports-no-repos
```diff
@@ -1,14 +1,15 @@
...
-DataFakeR
-datarobot
-DominoDataCapture
-NHSRplotthedots
-Plasmidprofiler
-SamplerCompare
-TDA
-wavefunction
+package,maintainer
+adaptalint,conway.max1@gmail.com
```
AshesITR, line 2: Found it:
https://github.com/maxconway/adaptalint

MichaelChirico: nice!

suggest this as the approach for cases like this:

https://github.com/maxconway/adaptalint/pull/2

MichaelChirico: definitely highlights some manual friction that's bound to happen as the boundary between the packages we can check automatically and those we track manually changes over time

MichaelChirico: Added more PRs:

 - https://github.com/cloudyr/aws.alexa/pull/18
 - https://github.com/compTAG/r-tda/pull/41
 - https://github.com/openpharma/DataFakeR/pull/3
 - https://github.com/phac-nml/plasmidprofiler/pull/1
 
Two more are already added in the dev versions:

 - https://github.com/nhs-r-community/NHSRplotthedots/blob/e54f3d6ab9ce497d2a625e34ab88dae3481ee5c5/DESCRIPTION#L20
 - https://github.com/pharmaverse/admiral/blob/07189fcbbd13a585456fdd826c5d3a7bfe92fe68/DESCRIPTION#L48

Since there's so few left I'll just e-mail the maintainers of the rest:

 - autoharp
 - biolink
 - ConNEcT
 - datarobot
 - DominoDataCapture
 - SamplerCompare
 - wavefunction

MichaelChirico: `SamplerCompare`/`wavefunction`, `autoharp`, and `datarobot` maintainers all wrote back confirming no public repo, so I think it's safe to just use the CRAN mirror as a back-up for those and just report deltas.

MichaelChirico:
Added current output to the PR description. Feel free to chime in with more things you'd like to see/collect.

Still haven't had a chance to read the deltas in detail. But did notice that we are doing ~30% better for overall lint performance (even though edgarWebR is still way slower, I think we can ignore that, esp. if this PR is merged: https://github.com/mwaldstein/edgarWebR/pull/24) :tada:

.dev/revdep-repos
```diff
@@ -0,0 +1,78 @@
+http://github.com/soodoku/abbyyR
+http://github.com/soodoku/tuber
```
AshesITR, line 2: nit: https:// everywhere?

.dev/revdep_compare_releases.R
```diff
@@ -0,0 +1,246 @@
...
+  old_only <- old[!new, on = c("package", "filename", "line_number")]
+  old_only[, version := NULL]
+
+  withr::local_options(datatable.print.nrows = Inf)
+
+  message("Count of lints found on ", new_version, " but not on ", old_version, ": ", nrow(new_only))
+  message("Count of these by linter:")
+  print(new_only[, .N, by = linter][order(-N)])
+  message("Sample of <=10 hits from each linter:")
+  print(new_only[sample(.N), head(.SD, 10L), by = linter][, .(package, filename, linter, line)])
```
AshesITR, line 167: this is a bit verbose.
Maybe also reorder to `linter, line, package, filename`?
The actual line of code seems more important to assess accuracy.

.dev/revdep-repos
MichaelChirico: this is taken directly from CRAN, but I have filed PRs to do this, so it'll be picked up eventually:

https://github.com/soodoku/abbyyR/pull/13
https://github.com/soodoku/tuber/pull/119

.dev/revdep_compare_releases.R
MichaelChirico: this is mainly for our own inspection, locally, so I'm not so worried about verbosity, though the column ordering can be improved.

the next step is to produce e-mails with the delta for each individual package -- there we'll be more careful about presentation.

MichaelChirico: column order didn't matter too much -- the issue is there are v. wide lines that always bump the other columns. so I combined `line [linter]` as one column to force them always to appear together (ditto for `package:filename`)

MichaelChirico:
@AshesITR PTAL. Added the e-mail templates -- these again will not be stored on GH (though the `revdep-email-template` file will be).

I'll put out a call for new repos now (I prefer if downstreams declare themselves to us, btw, rather than getting in downstreams' hair uninvited and/or dealing with low-quality/unmaintained repos [it's also a bit more of a security risk]).

We can fine-tune the output a bit (or handle any bigger feedback), then I'll remove the temp artefacts & we can merge

AshesITR:
I took a glimpse now.

* Some maintainers might get multiple mails. Do we want to bundle all info in one mail?
* The lint comparisons for `brace_linter()` and its predecessors always show up as a change, but I don't think this is material. Can we somehow exclude these sporadic hits?
* Would it be worthwhile to dynamically list what is attached? e.g. if a failure is detected.
* For the failures pertaining to a removed deprecated function, we could repeat the new linter to ease transition after an update to 3.0.0 has already been made. I see `absolute_paths_linter()` causing a few failures.

Also, maybe a good idea to run the checks with a soft-deprecated shim for the removed linters so we can check the remainder of the packages using now removed linters.
IINM we can achieve this by supplying the removed linter functions in globalenv() before calling lint_package().

MichaelChirico:
> Some maintainers might get multiple mails. Do we want to bundle all info in one mail?

Do you know what `revdepcheck` does here? i can see benefits of both ways. currently there are at most a few e-mails to be received so it may not be worth over-engineering...

> The lint comparisons for brace_linter() and its predecessors always show up as a change, but I don't think this is material. Can we somehow exclude these sporadic hits?

Note that the differences are not matching on linter name, just on package/filename/line_number. So if `brace_linter()` is showing up, either the source marker has moved, or it's a new/removed lint. In any case, it should be surfaced to make sure it's intentional.

> Would it be worthwhile to dynamically list what is attached? e.g. if a failure is detected.

Not following what this would look like...

> For the failures pertaining to a removed deprecated function, we could repeat the new linter to ease transition after an update to 3.0.0 has already been made. I see absolute_paths_linter() causing a few failures.

I think there's been transition enough -- as noted in the NEWS, they've been marked as deprecated already since 2017. Time to pull the plug.

> Also, maybe a good idea to run the checks with a soft-deprecated shim for the removed linters so we can check the remainder of the packages using now removed linters. IINM we can achieve this by supplying the removed linter functions in globalenv() before calling lint_package().

great idea :+1:

MichaelChirico:
OK @AshesITR cleaned up the PR here, should be ready for merge if there are no blockers

.dev/revdep-email-template
```diff
@@ -0,0 +1,29 @@
+Hello {maintainer}! Thank you for using {{lintr}} in your package {{{package}}}!
+
+{{lintr}} will soon be releasing an update to CRAN -- be sure to check the latest NEWS to get an overview
+of the upcoming improvements and bug fixes.
+
+We checked the results of running lint_package() on your package as found at {repo_url} (hash: {git_hash}) using
+the CRAN version {old_release} and the current version in development (commit hash {main_hash}).
+We do this ourselves because of {{lintr}}'s position as a developer tool --
```
AshesITR, line 8: "position" feels awkward to me. Is "role" appropriate here?

.dev/revdep-email-template
```diff
@@ -0,0 +1,29 @@
+Hello {maintainer}! Thank you for using {{lintr}} in your package {{{package}}}!
+
+{{lintr}} will soon be releasing an update to CRAN -- be sure to check the latest NEWS to get an overview
+of the upcoming improvements and bug fixes.
```
AshesITR, line 4: Maybe add a link to `https://lintr.r-lib.org/news/index.html` here?

.dev/revdep-email-template
```diff
@@ -0,0 +1,29 @@
...
+
+We checked the results of running lint_package() on your package as found at {repo_url} (hash: {git_hash}) using
+the CRAN version {old_release} and the current version in development (commit hash {main_hash}).
+We do this ourselves because of {{lintr}}'s position as a developer tool --
+normal CRAN revdep checking is not as helpful for detecting breakages.
+
+On your package, lint_package() took {old_duration}s on CRAN vs. {main_duration}s in our latest version. This
+includes applying your .lintr config if one is available.
+
+I am also attaching the following, if they exist:
```
AshesITR, line 14: "We are also attaching..."

AshesITR (commented):
Looks good so far, thanks!

.dev/revdep_compare_releases.R
```diff
@@ -0,0 +1,333 @@
+#!/usr/bin/env Rscript
+library(withr)
+library(pkgload)
+library(data.table)
+library(glue)
+
+repo_data <- rbind(
+  data.table::fread("revdep-repos"),
+  data.table::fread("revdep-extra-repos", skip = 2L),
```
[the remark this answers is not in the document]
MichaelChirico: not so magical in this case :)

didn't work on my version of data.table at least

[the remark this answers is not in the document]
AshesITR: Then `skip = "package,repo"` would be more expressive I think.

AshesITR (approved):

PR #1309, opened 2022-05-28
Some clean-up of tests to be more canonical

MichaelChirico:
Mostly being more consistent about using `testthat::test_path()`, and some style touch-ups


tests/testthat/test-lint_package.R
```diff
@@ -16,17 +16,19 @@ test_that(
...
 
   expected_lines <- c(
     # from abc.R
     "abc = 123",
     # from jkl.R
-    "jkl = 456", "mno = 789"
+    "jkl = 456",
+    "mno = 789"
   )
   read_settings(NULL)
```
AshesITR, line 28: This actually tests the default settings. (Related to the bug in the `.lintr` PR?)
Better to `skip_if(!is.null(find_config(...)))` and let `lint_package()` notice that the default config is needed?

Requires copying the package to a tempdir so our `.lintr` isn't found.

MichaelChirico: sounds less trivial than this PR... follow-up? 😬

AshesITR: Should be straight-forward now with the new solution of using `lintr_test_config`. That file is unlikely to exist in parent directories by accident.

AshesITR (approved):

tests/testthat/test-lint_package.R
```diff
@@ -13,20 +13,24 @@
 
 test_that(
   "`lint_package` does not depend on path to pkg - no excluded files", {
+
```
MichaelChirico, line 16: please check new lints

MichaelChirico:
thanks for the clean-up... i was still trying to digest your suggestion :)

tests/testthat/test-lint_package.R
AshesITR: sorry, GitHub web UI isn't exactly helpful there ^^

AshesITR (approved):

PR #1321, opened 2022-05-29
updates to backport_linter()

### LINKED ISSUE #1316: Add `except =` to `backport_linter()`
If a `backports_linter()` lint was fixed by importing the function from `{backports}`, the lint won't go away.

### LINKED ISSUE #1318: Update `r-devel` version in `backport_linter`
Currently sees `4.2.0` as r-devel, which is no longer true.

### LINKED ISSUE #1319: Add proper oldrel versions to `backport_linter()`
oldrel is customarily defined as the last patch in a `x.y` series, even if no new functions were introduced in that release, the specification should use that version number.

### LINKED ISSUE #1320: Add `backport_linter("oldrel-4")` to our linters
`oldrel-4` makes the most sense to me because that's what our GHA pipeline runs on.
Currently amounts to `3.4.4`

AshesITR:
Various updates to `backport_linter()`

- [x] add argument `except`. Fixes #1316
- [x] add empty R 4.3.0 to update `r-devel`. Fixes #1318
- [x] add old patched release version numbers to yield correct `oldrel-x` specs. Fixes #1319
- [x] add `backports_linter("oldrel-4", except = "R_user_dir")` to our `.lintr` configs (currently no lints). Fixes #1320

### DIFF
--- a/R/backport_linter.R
+++ b/R/backport_linter.R
@@ -3,12 +3,17 @@
 #' Check for usage of unavailable functions. Not reliable for testing r-devel dependencies.
 #'
 #' @param r_version Minimum R version to test for compatibility
+#' @param except Character vector of functions to be excluded from linting.
+#'  Use this to list explicitly defined backports, e.g. those imported from the {backports} package or manually
+#'  defined in your package.
+#'
 #' @evalRd rd_tags("backport_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-backport_linter <- function(r_version = getRversion()) {
+backport_linter <- function(r_version = getRversion(), except = character()) {
   r_version <- normalize_r_version(r_version)
   backport_blacklist <- backports[r_version < R_system_version(names(backports))]
+  backport_blacklist <- lapply(backport_blacklist, function(nm) setdiff(nm, except))
 
   names_xpath <- "//SYMBOL | //SYMBOL_FUNCTION_CALL"
 
@@ -35,7 +40,8 @@ backport_linter <- function(r_version = getRversion()) {
     # which(arr.ind) returns things in the same order as which()
     needs_backport_version_idx <- ((which(needs_backport) - 1L) %% length(backport_blacklist)) + 1L
     lint_message <- sprintf(
-      "%s (R %s) is not available for dependency R >= %s.",
+      paste("%s (R %s) is not available for dependency R >= %s.",
+            "Use the `except` argument of `backport_linter()` to configure available backports."),
       all_names[bad_idx],
       names(backport_blacklist)[needs_backport_version_idx],
       r_version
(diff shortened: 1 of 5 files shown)


AshesITR:
Looks like network issues, this time with CRAN, again...

AshesITR:
TODO for #834: Add issue numbers to the `backport_linter()` bullet.

R/backport_linter.R
```diff
@@ -3,12 +3,17 @@
 #' Check for usage of unavailable functions. Not reliable for testing r-devel dependencies.
 #'
 #' @param r_version Minimum R version to test for compatibility
+#' @param except Character vector of functions to be excluded from linting.
+#'  Use this to list explicitly defined backports, e.g. those imported from the {backports} package or manually
```
MichaelChirico, line 7: shouldn't we be able to reuse object_usage_linter logic to detect if either of these is true instead?

AshesITR: Nope, because we can't see all files.

AshesITR: Also, we'd have to parse the NAMESPACE of the linted package to check if an alternative implementation is imported, e.g. from backports.

AshesITR: Are you sure you ran this in a session where the linted package wasn't loaded, backports wasn't loaded and R didn't define the function?
I expect a lint there.

AshesITR: Also note that backport linter looks for SYMBOL and SYMBOL_FUNCTION_CALL with an explicit blacklist so this linter definitely has the problem.

MichaelChirico: > and R didn't define the function?

oh whoops :facepalm: that's why it was different for work (3.6.3) vs. home (4.2.0) :)

my point is still the same -- why are we only "solving" this for `backport_linter()`, when it seems a broader issue? Isn't `codetools::checkUsage()` our friend here?

AshesITR: Backport Linter doesn't look for names. Otherwise, it wouldn't properly lint e.g. when running `backport_linter("3.0.0")` under R 4.2.0.

MichaelChirico: it's possible to work though right? since we maintain the dictionary of backports.

sorry for all the pushback, I just find this approach rather awkward for the user when it seems possible to handle this all on their behalf.

AshesITR: It's impossible to work around this issue.
Imagine this setup, linting under R 4.2.0 with `backport_linter("3.0.0")`:

file1.R
```r
my_fun <- function() {
  R_user_dir("mypkg")
}
```

`object_usage_linter()` won't complain because `tools` is loaded and R 4.2.0 provides `R_user_dir`.
`backport_linter()` finds `R_user_dir` in its blacklist. How should it check if the lint should be supressed?
The function is defined (available via `tools`).
It could also be imported from `backports` via `NAMESPACE`, in which case it would be visible if the linted package was loaded first.
Or it could be defined in another file, `file2.R`, within the same package.
Or it could be conditionally imported (like we do), in which case it would come from `tools` in R 4.2.0 despite not needing a lint.

Not too keen on splitting this PR because we need the `excecpt = ` argument for our own usage.

MichaelChirico: All of those possibilities seem knowable to me -- just a matter of effort required. I think we should set a goal to eventually eliminate this argument / populate it automatically, but am fine to leave it for another day.

MichaelChirico (approved):

PR #1323, opened 2022-05-29
fix object_usage_linter() XPath for `assign()` and `setMethod()`

### LINKED ISSUE #1322: `object_usage_linter()` tries too many assignments
Seen in our own package when running `lint_package()`:

```
..........Warning in formals(fun) : argument is not a function
Warning in body(fun) : argument is not a function
Warning in formals(fun) : argument is not a function
Warning in body(fun) : argument is not a function
...................................................Warning in formals(fun) : argument is not a function
Warning in body(fun) : argument is not a function
................................................................................................................................................
```

The reason is the XPath for `object_usage_linter()` finds e.g.

```r
assign(
    envir = cache,
    x = digest_content(linters, filename),
    value = lints,
    inherits = FALSE
  )
```

which is not a function assignment.

AshesITR:
fixes #1322

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -21,7 +21,7 @@ object_usage_linter <- function(interpret_glue = TRUE) {
     "expr_or_assign_or_help[EQ_ASSIGN]/expr[2][FUNCTION]",
     "equal_assign[EQ_ASSIGN]/expr[2][FUNCTION]",
     # assign() and setMethod() assignments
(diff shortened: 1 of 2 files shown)


MichaelChirico:
tests?

AshesITR:
will add.

AshesITR:
@MichaelChirico good call. Turns out neither `assign()` nor `setMethod()` were tested before and `setMethod()` was dysfunctional because it takes the function as the third positional argument and not the second.

AshesITR:
NB: add info to #834 about the fix to `setMethods()` - that was not working in 2.0.1 either.

MichaelChirico (approved):

Issue #1324, opened 2022-05-29
`editors.html` has 9 MB


AshesITR:
Investigating the `R CMD check` note I found out it's because of `editors.html`, which includes the images as inline data.

The largest images are:

 - sublime: 4.65 MB
 - flycheck: 1.877 MB
 - vim - syntastic: 1.67 MB
 - vscode: 0.40 MB
 - RStudio: 0.27 MB
 - Atom: 0.15 MB

AshesITR:
I think we have four options:

 1. make `editors.Rmd` an article (`.Rbuildignore`'d, cf. `usethis::use_article()`) 
 This would make setup guidance unavailable offline.
 2. (is this even possible)? only include the images in `pkgdown`, not in the CRAN vignette.  
 This would include the essential bits offline while keeping the website version good looking.
 3. remove the images entirely. This would make the website version less pretty but solve the size problem for the vignette.
 4. compress the images to try and reduce their required space (e.g. by reducing resolution and framerate)  
 This seems a lot of work and no guaranteed success.

I would prefer 2. if that is possible, and 1. if it is not.
@MichaelChirico WDYT?

AshesITR:
Here's what I could do w.r.t. file size reduction so far:

```
-rw-r--r-- 1 alex alex 109K 29. Mai 13:04 atom.png
-rw-r--r-- 1 alex alex 830K 29. Mai 13:20 emacs.gif
-rw-r--r-- 1 alex alex  36K 29. Mai 13:31 emacs-still.gif
-rw-r--r-- 1 alex alex 195K 29. Mai 13:03 rstudio.png
-rw-r--r-- 1 alex alex 987K 29. Mai 13:23 sublime.gif
-rw-r--r-- 1 alex alex  94K 29. Mai 13:34 sublime-still.gif
-rw-r--r-- 1 alex alex 726K 29. Mai 13:22 vim-syntastic.gif
-rw-r--r-- 1 alex alex  65K 29. Mai 13:34 vim-syntastic-still.gif
-rw-r--r-- 1 alex alex 185K 29. Mai 13:11 vscode.png
```

The animated gifs will be too large for the vignette, even `gifsicle --colors 16 --lossy` couldn't manage to get `sublime.gif` small enough (recall we have a budget of around 1 MB for the total file size)

PR #1325, opened 2022-05-29
WIP reduce editor vignette size

AshesITR:
WIP towards #1324 

 - add compressed versions of external files


AshesITR:
@hadley Is there a way to preview the pkgdown site generated by a PR?
Also, `pkgdown` warns me locally about
```
Warning messages:
1: Failed to parse example for topic 'lint' 
2: Missing images in 'vignettes/editors.Rmd': 'reference/figures/rstudio.png', 'reference/figures/emacs.gif', 'reference/figures/vim-syntastic.gif', 'reference/figures/sublime.gif', 'reference/figures/atom.png', 'reference/figures/vscode.png'
ℹ pkgdown can only use images in 'man/figures' and 'vignettes' 
```

but the files actually exist under `doc/`. Do you know what that is about?

The vignette tries to embed the images via, e.g.:
```
![RStudio Example](man/figures/rstudio.png "Rstudio Example")
```

hadley:
Why are you embedding the images in vignette in the first place? Doesn't that make the vignette larger?

hadley:
@AshesITR unfortunately we haven't yet figured out a good way to preview pkgdown with GitHub pages. It is possible, however, to do so with netlify, so for pkgdown itself we use netlify instead (https://github.com/r-lib/pkgdown/blob/main/.github/workflows/pkgdown.yaml). Setting up netlify is a more of hassle so we don't generally recommend it.

AshesITR:
Embedding seems to be the default behaviour.
The advantage is the PDF vignette contains the images as well, right?
I managed to get it down to an acceptable size by conditionally using a static version for R CMD build and the dynamic version (compressed by `gifslice`) for pkgdown.
Quite happy with it for now. WDYT?

AshesITR:
@MichaelChirico `doc/`is back down to 595.9 kB with this PR, by embedding all images in still-form (I picked a frame from each of the gifs) and running image compression programs on all image files.

.Rbuildignore
```diff
@@ -27,3 +25,5 @@
 ^_pkgdown\.yml$
 ^docs$
 ^pkgdown$
+^vignettes/[^-]+.gif$
+^man/figures$
```
hadley, line 29: This is surprising — and I'm pretty sure it will break the rendering of the README on CRAN.

vignettes/editors.Rmd
```diff
@@ -12,13 +12,21 @@ knitr::opts_chunk$set(
...
+maybe_still <- function(url) {
+  if (pkgdown::in_pkgdown()) {
+    url
+  } else {
+    gsub("\\.gif$", "-still.gif", url)
+  }
+}
 ```
 
 ## RStudio
```
MichaelChirico, line 25: WDYT about a note at the top here that directs users to the pkgdown version to see full animation?

AshesITR: Not sure how useful this is to a user browsing `vignette("editors")`. Maybe instead add a note to `?lintr-package`?

MichaelChirico: I do think it's quite useful.

> Note: this vignette is best viewed in the browser, where we can render full animations of editor flows. See <link>.

?

MichaelChirico (approved):

PR #1307, opened 2022-05-27
add references in undesirable_* docs

MichaelChirico:
Closes #1306

PR #1326, opened 2022-05-29
Don't embed gifs in vignette

hadley:
@AshesITR maybe this is what you're looking for?

### DIFF
--- a/vignettes/editors.Rmd
+++ b/vignettes/editors.Rmd
@@ -1,6 +1,8 @@
 ---
 title: "Editor setup"
-output: rmarkdown::html_vignette
+output: 
+  rmarkdown::html_vignette:
+    self_contained: false
 vignette: >
   %\VignetteIndexEntry{Editor setup}
   %\VignetteEngine{knitr::rmarkdown}
@@ -118,7 +120,7 @@ You can also configure what linters are used.
 e.g. disabling the assignment linter and using a different line length cutoff.
 In the SublimeLinter User Settings
 
-```json
+``` json
     {
       "linters": {
         "lintr": {


AshesITR:
Also a viable alternative, thanks for pointing this out.
Are there any downsides such as problems with CRAN if any of the images become unreachable?

MichaelChirico:
@hadley for context, we saw this GHA failure and got a little spooked -- if regular CRAN checks fail to render because of intermittent misses on external links, will we be at risk of archiving?

https://github.com/r-lib/lintr/pull/1307#issuecomment-1140398704

hadley:
Ah, yes that's definitely a risk. I asked on the RStudio internal rmarkdown channel and there doesn't seem to be any way to suppress this behaviour with the version of pandoc that CRAN has (it's only a warning in pandoc >2.18).

Personally, I'd still lean towards taking on this risk over including a lot of extra gifs in the package that few people are likely to ever see.

AshesITR:
I'm slightly in favor of including the supporting files in lintr. The overhead is quite small after compression (which I already did), so we have only ~ 500KB of auxiliary files, the result is guaranteed to never fail to build because of network issues, and the vignette source code is easier to read, because the image URLs contain the name of the application depicted instead of a random ID.

OTOH the risk of imgur taking down pictures isn't very high as long as there is some traffic on them and githubusercontent (hosts the vs-code image) will probably be accessible as long as the owner account isn't deleted.

Regarding sizes, standalone = false is obviously the winner:
`R CMD build` size for #1325 is 3.9M, for #1326 it's 524K and for current main (c842e4e04daa30a3dcea857730f385009dcce302) it is 7.0M

AshesITR:
It's worth checking I think.
We should also take a look manually how the vignettes render with that.

AshesITR:
I just tested this locally. Unfortunately, self-contained FALSE removes the images from the built `doc/editors.html`, so this would be equivalent to removing the images altogether.

Thus, closing for the moment.

PR #1339, opened 2022-06-01
Accept default= in modify_defaults

### LINKED ISSUE #1336: Rename of default= to defaults= is hard to debug (linters_with_defaults)
Just discovered this hard-to-debug issue:

```
# pre-3.0.0 usage
my_linters <- with_defaults(default = list(), no_tab_linter())
writeLines(names(my_linters))
# no_tab_linter
```

OK, we get a warning:

> Warning message: Function with_defaults was deprecated in lintr version 2.0.9001. Use linters_with_defaults instead.

So we switch:

```
my_linters <- linters_with_defaults(default = list(), no_tab_linter())
writeLines(names(my_linters))
# assignment_linter
# ...
# default
# ...
# vector_logic_linter
```

i.e., `default=` is taken as part of the `...` arguments, and we don't start from a blank (`list()`) override.

Probably we should allow `default=` to work in `linters_with_defaults()`, with warning.

(Or perhaps this pluralization was an oversight?)

### LINKED ISSUE #1338: modify_defaults() is missing from NEWS
IINM this is a new export since 2.0.1 but I don't see it in the NEWS.

MichaelChirico:
Closes #1336 and closes #1338

Handling it in `linters_with_defaults()` alone is pretty painful because of the `substitute()`-based logic for auto-naming, so I implemented the patch inside `modify_defaults()` instead.

Originally wanted to just throw a warning with `modify_defaults()` too, but it seems unlikely to use `default=` by accident there because `defaults=` is a required argument. So added some uglier logic to allow `modify_defaults(default = , ...)` to work, assuming that was intentional.


R/with.R
```diff
@@ -24,9 +24,34 @@ modify_defaults <- function(defaults, ...) {
...
+      )
+    ) {
+      warning(
+        "'default' is not an argument to linters_with_defaults(). Did you mean 'defaults'? ",
+        "This warning will be removed when with_defaults() is fully deprecated."
+      )
+      defaults <- vals$default
+      vals$default <- NULL
+      valid_idx <- nms != "default"
+      nms <- nms[valid_idx]
```
AshesITR, line 45: Is there a way to use `Recall()` after the repair to keep the change localized?

Also shouldn't do this if there is also a `defaults` argument.

MichaelChirico: Didn't we eliminate `Recall()` as viable for the other similar checks in `lint()`/`lint_package()` because of headaches with `...`?

> Also shouldn't do this if there is also a defaults argument.

OK, actually that will make things a lot easier

AshesITR: It would definitely be cleaner to put the warning in `linters_with_defaults()` though...
Shouldn't something like this work?

```r
dots <- list(...)
if (missing(defaults) && "default" %in% names2(dots)) {
  warning("...")
  names(dots)[names(dots) == "default"] <- "defaults"
  return(do.call(modify_defaults, dots))
}
```

AshesITR: In what way? Maybe we can also use `substitute(alist(...))` here to generate the correct call?
Possibly even modifying `match.call()` instead.

MichaelChirico: OK, seems working now after pulling the naming logic into a helper.

AshesITR (approved):
Nice, looks much cleaner now 👍🏻

PR #1348, opened 2022-06-02
fix xpath for commas_linter and paren_body_linter not to hang on certain files

### LINKED ISSUE #1340: lint_package() hangs on edgarWebR
Found during work on #1304, haven't had a chance to investigate. Also haven't confirmed if the issue is present in 2.0.1.

```
git clone git@github.com:mwaldstein/edgarWebR.git
Rscript -e "lintr::lint_package("edgarWebR")
```

hangs for at least several minutes.

MichaelChirico:
Closes #1340

It looks like `'and'` conditions in XPath benefit from early exit which does us a huge favor here.

Also might want to look into where else `//expr` is inefficient & we'd be better off doing `//$SPECIFIC_NODE[...]/parent::expr/parent::expr`. For readability for now, let's stick with the current approach except where it proves demonstrably over-slow.


MichaelChirico:
Not sure what's going on with CI here... but I identified `paren_body_linter` as also hanging on the repo. Investigating and I'll add any fix to this PR.

AshesITR:
Looks like a case for requesting a re-run of failed jobs in the GHA interface. I'll refrain from that since you're going to push new commits anyway, right?

MichaelChirico:
Still just instantaneously failing :cry: 

Good news is `lintr` now runs on `edgarWebR` again. I think `cyclocomp_linter()` still struggles with those massive files, but I'm not too worried about it. Really those files should be excluded in the first place.

MichaelChirico:
Required tests passed at least. Should be good to go.

MichaelChirico:
@AshesITR any pending issue here?

AshesITR:
I didn't find time to review yet.

AshesITR (approved):

Issue #778, opened 2021-03-11
Add proper benchmarking functionality to .dev/compare_branches.R


AshesITR:
I think properly benchmarking the two branches is in scope for the script but would require a much bigger PR... filed this quickly as a first pass to give a rough idea (I don't expect it to work for, say, <30% performance difference, but something like 2x/3x should show up)

_Originally posted by @MichaelChirico in https://github.com/jimhester/lintr/issues/777#issuecomment-796612676_

kpagacz:
What would you like to see as a result of benchmarking?

Print microbenchmark results for some examples of each linter?
Allow custom examples - run lintr on a source file?

MichaelChirico:
Nice! Forgot about this issue. Indeed I have in mind to record timings for each linter. It should be fine to track total time spent on each linter over a large enough sample of files.

PR #1353, opened 2022-06-03
use expr[1][SYMBOL_FUNCTION_CALL] where appropriate+possible

### LINKED ISSUE #1349: Use expr[1][SYMBOL_FUNCTION_CALL] consistently
#1340 and its solution highlights the importance of the point raised by @AshesITR in #1310 -- we should use `expr[1]` where possible when specifying XPaths on `<expr>` nodes. This will almost surely have a notable performance implication for edge cases like `c(...)` where `...length()` is huge, but will likely have a good benefit even in the more common case where `...length()` is, say, dozens rather than 100s or 1000s.

MichaelChirico:
Closes #1349

There's plenty more to do on XPath optimization, but this is a nice start/low-hanging fruit.

#778 should also help clarify where optimization is likely to have the most payoff too.

### DIFF
--- a/R/expect_not_linter.R
+++ b/R/expect_not_linter.R
@@ -11,8 +11,8 @@
 #' @export
 expect_not_linter <- function() {
   xpath <- "//expr[
-    SYMBOL_FUNCTION_CALL[text() = 'expect_true' or text() = 'expect_false']
-    and following-sibling::expr[1][OP-EXCLAMATION]
+    expr[1][SYMBOL_FUNCTION_CALL[text() = 'expect_true' or text() = 'expect_false']]
+    and expr[2][OP-EXCLAMATION]
   ]"
 
   Linter(function(source_expression) {
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -13,8 +13,8 @@
 #' @export
 any_duplicated_linter <- function() {
   any_duplicated_xpath <- "//expr[
-    expr[SYMBOL_FUNCTION_CALL[text() = 'any']]
-    and expr[expr[SYMBOL_FUNCTION_CALL[text() = 'duplicated']]]
+    expr[1][SYMBOL_FUNCTION_CALL[text() = 'any']]
+    and expr[expr[1][SYMBOL_FUNCTION_CALL[text() = 'duplicated']]]
     and (
       not(OP-COMMA)
       or OP-COMMA[
@@ -37,8 +37,8 @@ any_duplicated_linter <- function() {
   length_unique_xpath <- "
   //expr[EQ or NE or GT or LT]
   /expr[
-    expr[SYMBOL_FUNCTION_CALL[text() = 'length']]
-    and expr[expr[
+    expr[1][SYMBOL_FUNCTION_CALL[text() = 'length']]
+    and expr[expr[1][
       SYMBOL_FUNCTION_CALL[text() = 'unique']
       and (
         following-sibling::expr =
@@ -46,7 +46,7 @@ any_duplicated_linter <- function() {
           /parent::expr
           /parent::expr
           /expr
-          /expr[SYMBOL_FUNCTION_CALL[text()= 'length']]
+          /expr[1][SYMBOL_FUNCTION_CALL[text()= 'length']]
           /following-sibling::expr
         or
         following-sibling::expr[OP-DOLLAR or LBB]/expr[1] =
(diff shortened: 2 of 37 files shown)


MichaelChirico:
Oh, just found this nugget in the XPath documentation:

https://www.w3.org/TR/1999/REC-xpath-19991116/#node-sets

> `preceding::foo[1]` returns the first foo element in _reverse document_ order, because the axis that applies to the `[1]` predicate is the preceding axis; by contrast, `(preceding::foo)[1]` returns the first foo element in _document order_

i.e., `preceding-sibling::expr[SYMBOL_FUNCTION_CALL]` should be replaceable by `(preceding-sibling::expr)[1][SYMBOL_FUNCTION_CALL]`.

Tucking that away as a mental note for now -- that's a pretty obscure usage that I don't want to get too greedy about using unless there's a clear performance benefit.

MichaelChirico:
Added `preceding-sibling::expr[last()]` now, that looks pretty readable. Not benchmarked.

MichaelChirico:
Quick benchmarks on our favorite huge dump file from #1340:

```
library(xml2)
library(xmlparsedata)

p = parse("https://raw.githubusercontent.com/mwaldstein/edgarWebR/master/tests/cache/browse-edgar-11457c.R")
xml = read_xml(xml_parse_data(p))

system.time(xml_find_all(xml, "//expr"))
#    user  system elapsed 
#   0.087   0.000   0.089 
system.time(xml_find_all(xml, "//expr[1]"))
#    user  system elapsed 
#   0.055   0.000   0.056 

system.time(xml_find_all(xml, "//expr[SYMBOL_FUNCTION_CALL]"))
#    user  system elapsed 
#   0.111   0.000   0.112 
system.time(xml_find_all(xml, "//expr[1][SYMBOL_FUNCTION_CALL]"))
#    user  system elapsed 
#   0.073   0.000   0.073
```

So a 30-40% improvement in this limiting case from using `[1]`. Still not sure about `last()`.

MichaelChirico:
I wasn't really satisfied with that benchmark -- here's probably a better one:

```
writeLines(replicate(1e4, sprintf("c(%s)", toString(sample(100, sample(100, 1))))), tmp <- tempfile())

xml = xml2::read_xml(xmlparsedata::xml_parse_data(parse(tmp)))

system.time(xml2::xml_find_all(xml, "//expr[expr[SYMBOL_FUNCTION_CALL[text() = 'c']]]"))
#    user  system elapsed 
#   0.847   0.180   1.027 
system.time(xml2::xml_find_all(xml, "//expr[expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]]"))
#    user  system elapsed 
#   0.672   0.024   0.697 
system.time(xml2::xml_find_all(xml, "//expr[SYMBOL_FUNCTION_CALL[text() = 'c']]"))
#    user  system elapsed 
#   0.681   0.000   0.680 
system.time(xml2::xml_find_all(xml, "//expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]"))
#    user  system elapsed 
#   0.493   0.000   0.492 
system.time(xml2::xml_find_all(xml, "//SYMBOL_FUNCTION_CALL"))
#    user  system elapsed 
#   0.286   0.000   0.285 

system.time(xml2::xml_find_all(xml, "//expr/preceding-sibling::expr[SYMBOL_FUNCTION_CALL]"))
#    user  system elapsed 
# 526.236   0.168 526.188 
system.time(xml2::xml_find_all(xml, "//expr/preceding-sibling::expr[last()][SYMBOL_FUNCTION_CALL]"))
#    user  system elapsed 
#  40.659   0.000  40.645
```

So clearly `[last()]` is better, but still painful! So we should be wary of expressions like this where we think they are vulnerable to edge cases.

R/expect_not_linter.R
```diff
@@ -11,8 +11,8 @@
 #' @export
 expect_not_linter <- function() {
   xpath <- "//expr[
-    SYMBOL_FUNCTION_CALL[text() = 'expect_true' or text() = 'expect_false']
-    and following-sibling::expr[1][OP-EXCLAMATION]
+    expr[1][SYMBOL_FUNCTION_CALL[text() = 'expect_true' or text() = 'expect_false']]
+    and expr[2][OP-EXCLAMATION]
```
AshesITR, line 15: This actually changes the matched expr node from the function call node (i.e. only the function symbol) to its parent node (i.e. the entire call).

Is this intentional?

AshesITR (commented):
LGTM just want to confirm the location change is intentional. Feel free to merge if it is.
Will affect the ranges only, since the column number is the same.

AshesITR (approved):

R/expect_not_linter.R
MichaelChirico: yep -- did it for consistency with the other linters.

Also just double checked, and the marker is definitely superior:

```
# on main
lint("\nexpect_true(!A)", expect_not_linter())
<text>:2:1: warning: expect_false(x) is better than expect_true(!x), and vice versa.
expect_true(!A)
^~~~~~~~~~~

# this branch
<text>:2:1: warning: expect_false(x) is better than expect_true(!x), and vice versa.
expect_true(!A)
^~~~~~~~~~~~~~~
```

If anything, I'd move it to put the marker on the `!A` part... for another day/PR

PR #1365, opened 2022-06-06
Skip unloadable namespaces when populating imports

### LINKED ISSUE #1360: Breaking change (ish): lint_package() fails if loadNamespace() doesn't work for some deps
Finished a run of #1304 thanks to #1348 and am peeking at results now.

The first thing I'm noticing is that `lint_package()` fails on a lot of repos where it still succeeds on v2.0.1, in particular with error messages like:

```
[babette]: there is no package called ‘beastier’
[BiasCorrector] there is no package called ‘shiny’
[BTYDplus]: there is no package called ‘BTYD’
[caretEnsemble]: there is no package called ‘caret’
[cloudos]: there is no package called ‘ggplot2’
```

Haven't investigated why, but if this breaking change is intentional, we should mention it (I don't recall anything in the NEWS)

MichaelChirico:
Closes #1360

The tests are not very satisfying -- they basically just guarantee running `lint()` doesn't fail in this case.

Considered making a little "cache" to throw a warning, but opted for a long explanation in the documentation instead. I think that's a good compromise. See #1366 for a way forward on this.

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -10,14 +10,27 @@ namespace_imports <- function(path = find_package()) {
 
   full_imports <- lengths(imports) == 1L
 
-  # this loads the namespaces, but is the easiest way to do it
-  imports[full_imports] <- lapply(imports[full_imports], function(x) list(x, getNamespaceExports(asNamespace(x))))
+  imports[full_imports] <- lapply(imports[full_imports], safe_get_exports)
 
+  funs <- lapply(imports, `[[`, 2L)
   data.frame(
-    pkg = unlist(lapply(imports, function(x) rep(x[[1L]], length(x[[2L]])))),
-    fun = unlist(lapply(imports, `[[`, 2L)),
+    pkg = rep(unlist(lapply(imports, `[[`, 1L)), lengths(funs)),
+    fun = unlist(funs),
     stringsAsFactors = FALSE
   )
+}
+
+# this loads the namespaces, but is the easiest way to do it
+# test package available packages to avoid failing out as in #1360
+#   typically, users are running this on their own package directories and thus
+#   will have the namespace dependencies installed, but we can't guarantee this.
+safe_get_exports <- function(ns) {
+  exports <- tryCatch(
+    getNamespaceExports(asNamespace(ns)),
+    error = function(cond) character()
+  )
+
+  list(ns, exports)
 }
 
 # filter namespace_imports() for S3 generics
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -25,10 +25,10 @@ namespace_imports <- function(path = find_package()) {
 #   typically, users are running this on their own package directories and thus
 #   will have the namespace dependencies installed, but we can't guarantee this.
 safe_get_exports <- function(ns) {
-  exports <- tryCatch(
-    getNamespaceExports(asNamespace(ns)),
-    error = function(cond) character()
-  )
+  exports <- character()
+  if (requireNamespace(ns, quietly = TRUE)) {
+    exports <- getNamespaceExports(asNamespace(ns))
+  }
 
   list(ns, exports)
 }
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -1,23 +1,15 @@
 # Parse namespace files and return imports exports, methods
 namespace_imports <- function(path = find_package()) {
-  imports <- tryCatch({
-    pkg_name <- suppressWarnings(pkg_name(path))
-    data <- parseNamespaceFile(package = basename(path), package.lib = file.path(path, ".."))
-    data$imports
-  }, error = function(e) {
-    list()
-  })
-
-  full_imports <- lengths(imports) == 1L
-
-  imports[full_imports] <- lapply(imports[full_imports], safe_get_exports)
-
-  funs <- lapply(imports, `[[`, 2L)
-  data.frame(
-    pkg = rep(unlist(lapply(imports, `[[`, 1L)), lengths(funs)),
-    fun = unlist(funs),
-    stringsAsFactors = FALSE
+  namespace_data <- tryCatch(
+    parseNamespaceFile(basename(path), package.lib = file.path(path, "..")),
+    error = function(e) NULL
   )
+
+  if (length(namespace_data$imports) == 0L) {
+    return(empty_namespace_data())
+  }
+
+  do.call(rbind, lapply(namespace_data$imports, safe_get_exports))
 }
 
 # this loads the namespaces, but is the easiest way to do it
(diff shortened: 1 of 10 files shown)


R/namespace.R
```diff
@@ -10,16 +10,29 @@ namespace_imports <- function(path = find_package()) {
...
 
-  # this loads the namespaces, but is the easiest way to do it
-  imports[full_imports] <- lapply(imports[full_imports], function(x) list(x, getNamespaceExports(asNamespace(x))))
+  imports[full_imports] <- lapply(imports[full_imports], safe_get_exports)
 
+  funs <- lapply(imports, `[[`, 2L)
   data.frame(
-    pkg = unlist(lapply(imports, function(x) rep(x[[1L]], length(x[[2L]])))),
-    fun = unlist(lapply(imports, `[[`, 2L)),
+    pkg = rep(unlist(lapply(imports, `[[`, 1L)), lengths(funs)),
```
AshesITR, line 17: Use `rep(full_imports, lengths(funs))` here after refactoring `safe_get_exports()` to return only the functions?

R/namespace.R
```diff
@@ -10,16 +10,29 @@ namespace_imports <- function(path = find_package()) {
...
 
+# this loads the namespaces, but is the easiest way to do it
+# test package availablity to avoid failing out as in #1360
+#   typically, users are running this on their own package directories and thus
+#   will have the namespace dependencies installed, but we can't guarantee this.
+safe_get_exports <- function(ns) {
+  exports <- tryCatch(
+    getNamespaceExports(asNamespace(ns)),
+    error = function(cond) character()
+  )
```
AshesITR, line 31: Maybe better `if (requireNamespace(ns, quietly = TRUE))`?
Catchall tryCatch might be difficult to debug down the line.

R/object_name_linters.R
```diff
@@ -24,6 +24,25 @@ object_name_xpath <- local({
...
+#'
+#' Suppose, for example, you have `import(upstream)` in your NAMESPACE,
+#'   which makes available its exported function
+#'   `a_really_quite_long_function_name` that you then use in your package.
+#'   Then, if `upstream` is not installed when this linter runs, a lint
+#'   will be thrown on this object (even though you don't "own" its name).
+#'
+#' There are three workarounds: (1) always namespace-qualify usages,
+#'   because this linter ignores names in `pkg::foo()` form; (2) use
+#'   `@importFrom(pkg, foo)` instead of a blanket `import(pkg)` in your
```
AshesITR, line 41: Mixed roxygen / NAMESPACE syntax.

R/object_name_linters.R
```diff
@@ -24,6 +24,25 @@ object_name_xpath <- local({
...
+#'   If this fails (e.g., the package is not yet installed), the linter
+#'   won't be able to ignore such usages.
+#'
+#' Suppose, for example, you have `import(upstream)` in your NAMESPACE,
+#'   which makes available its exported function
+#'   `a_really_quite_long_function_name` that you then use in your package.
+#'   Then, if `upstream` is not installed when this linter runs, a lint
+#'   will be thrown on this object (even though you don't "own" its name).
+#'
+#' There are three workarounds: (1) always namespace-qualify usages,
```
AshesITR, line 39: Options?

R/object_name_linters.R
```diff
@@ -24,6 +24,25 @@ object_name_xpath <- local({
...
+#'   because this linter ignores names in `pkg::foo()` form; (2) use
+#'   `@importFrom(pkg, foo)` instead of a blanket `import(pkg)` in your
+#'   NAMESPACE, because this linter takes these imported names as given;
+#'   and of course (3) install the package so that it's available in
```
AshesITR, line 43: Make this the first option?

tests/testthat/dummy_packages/missing_dep/R/foo.R
```diff
@@ -0,0 +1,7 @@
+downstream <- function() {
+  a_really_really_long_local_object_name <- 1
+  aSilLLyObjEctNaME <- 2
+
+  myFancyUpstream::upstreams_really_long_imported_object_name()
+  myFancyUpstream::uPStreams_SilLY.objecT.Name()
```
AshesITR, line 6: Also add a case of `importFrom()`?

R/namespace.R
MichaelChirico: Sorry, not seeing it...

(1) we get a list with elements like (a) `list(ns, imported_funs)` or (b) `list(ns)`
(2) `safe_get_exports()` fills out the (b)-type elements to match the style of (a)-type
(3) combine to get the output data.frame

the presence of the (a)-type elements makes it harder to refactor, but I may be missing something.

R/namespace.R
MichaelChirico: also reads better, now that I've resigned myself to this being the right approach :)

tests/testthat/dummy_packages/missing_dep/R/foo.R
MichaelChirico: oh, good call. turns out we need to skip more... need to investigate further:

https://github.com/r-lib/lintr/blob/cc9a221763ffe01492eb3b4672915a78ea8c67f9/R/namespace.R#L29

R/namespace.R
AshesITR: ah nevermind, I didn't see the named import part because it wasn't in the diff.

R/object_name_linters.R
```diff
@@ -24,6 +24,27 @@ object_name_xpath <- local({
...
+#'   will be thrown on this object (even though you don't "own" its name).
+#'
+#' There are three options to get lintr to work correctly:
+#'   (1) install the package so that it's available in
+#'   the session where this linter is running; (2) always namespace-qualify
+#'   usages, because this linter ignores names in `pkg::foo()` form; and
+#'   (3) use `importFrom(pkg, foo)` instead of a blanket `import(pkg)` in your
+#'   NAMESPACE, because this linter takes these imported names as given;
+#'   and of course (3) install the package so that it's available in
+#'   the session where this linter is running.
```
AshesITR, line 46: (3) appears twice now.
Also, what about using a md enumeration?

1. install the package so that it's available to this linter while it's running.
2. only use namespace-qualified functions, e.g. `pkg::foo()`, because they are ignored by this linter.
3. only use selective imports (`importFrom(pkg, foo)`) instead of wildcard imports (`import(pkg)`) in your `NAMESPACE`, because these are always assumed to be available by this linter.

MichaelChirico: NVM, I think this list is wrong anyway. Resolving.

R/namespace.R
MichaelChirico: Reworked now, PTAL.

AshesITR (approved):

PR #1369, opened 2022-06-06
mention old source_file argument still works

MichaelChirico:


AshesITR:
huh? Isn't the argument name of the linter closure simply ignored because we provide it positionally?

MichaelChirico:
Yes; this is referring to the arguments to `with_id()` and `ids_with_token()`

MichaelChirico:
Just merging this into #1370 instead

PR #1375, opened 2022-06-08
benchmarking in .dev/compare_branches

### LINKED ISSUE #778: Add proper benchmarking functionality to .dev/compare_branches.R

MichaelChirico:
Closes #778

I'm having some hardware (software?) issue where the drive with my CRAN mirror keeps disconnecting itself, which makes a long-running comparison pretty tough. Hopefully figure that out soon (and anyway it's more motivation to support just _ad hoc_ downloading packages from a mirror), but anyway we can merge this now.


.dev/compare_branches.R
```diff
@@ -318,17 +319,44 @@ run_workflow <- function(what, packages, linter_names, branch, number) {
...
+      gert::git_branch_checkout(branch, force = TRUE),
+      error = function(cond) {
+        tag_hash <- gert::git_tag_list(branch)$commit
+        if (length(tag_hash) == 0L) {
+          stop("Unable to find branch or tag '", branch, "'")
+        }
+        # if an actual tag is provided, we'll get 0 or 1 results. but
+        #   git_tag_list() supports wildcards too, e.g. git_tag_list("v2*"), so safeguard:
+        if (length(tag_hash) > 1L) {
+          warning("Matched more than one tag! Selecting the first of: ", toString(tag_hash))
```
AshesITR, line 337: Might be more helpful to specify the selected tag name rather than the associated commit hash.

.dev/compare_branches.R
```diff
@@ -318,17 +319,44 @@ run_workflow <- function(what, packages, linter_names, branch, number) {
...
+          tag_hash <- tag_hash[1L]
+        }
+        # no way to checkout a commit directly, so create a branch based to it instead -- gert#147
+        # also don't have checkout --force, so just reset to prevent that from blocking -- gert#177
+        gert::git_reset_hard()
+        gert::git_branch_create(paste(sample(letters), collapse = ""), ref = tag_hash)
```
AshesITR, line 343: directly follow up by branch_delete to avoid polluting the local git repo and avoid accidental pushes of these temp branches?

MichaelChirico: could, but remember this is in a copy of the repo in `tempdir()`... maybe we should `git remote rm` on the copy to prevent any pushes?

AshesITR: SGTM

MichaelChirico:
OK @AshesITR this is ready for review... pending TODO can be handled in follow-up, it's mostly cosmetic. This now works:

```bash
.dev/compare_branches.R --branch=v2.0.1 --sample_size=20 --benchmark=true
```

(not sure why `optparse` doesn't allow `--benchmark` to work since we declare the parameter as `logical` :thinking:)

.dev/compare_branches.R
```diff
@@ -419,6 +551,39 @@ if (has_target) {
...
+              linter,
+              function(package) data.table::data.table(filename = names(package), duration = unlist(package))
+            )
+          )
+        )
+      )
+    )
+  )
+  # delete noisy/redundant information from filename
+  timings_data[, filename := sub(file.path(tempdir(), .BY$package, ""), "", filename), by = package]
```
AshesITR, line 578: Nice gem. I didn't know about `.BY` 😄

AshesITR (approved):
LGTM

.dev/compare_branches.R
MichaelChirico: needed because sub doesn't vectorize on pattern

Issue #1358, opened 2022-06-04
Avoid //expr XPaths


MichaelChirico:
The following currently linters use `//expr`:

- [x] any_duplicated_linter.R
- [x] any_is_na_linter.R
- [x] brace_linter.R
- [ ] class_equals_linter.R
- [x] condition_message_linter.R
- [x] conjunct_test_linter.R
- [x] consecutive_stopifnot_linter.R
- [x] duplicate_argument_linter.R
- [x] equals_na_linter.R
- [x] expect_comparison_linter.R
- [ ] expect_identical_linter.R
- [x] expect_length_linter.R
- [ ] expect_named_linter.R
- [ ] expect_not_linter.R
- [ ] expect_null_linter.R
- [x] expect_s3_class_linter.R
- [ ] expect_true_false_linter.R
- [x] expect_type_linter.R
- [x] fixed_regex_linter.R
- [x] function_left_parentheses_linter.R
- [x] ifelse_censor_linter.R
- [x] inner_combine_linter.R
- [x] literal_coercion_linter.R
- [x] missing_package_linter.R
- [x] nested_ifelse_linter.R
- [x] object_usage_linter.R
- [x] outer_negation_linter.R
- [x] package_hooks_linter.R
- [x] paste_linter.R
- [x] pipe_call_linter.R
- [ ] redundant_ifelse_linter.R
- [ ] regex_subset_linter.R
- [x] seq_linter.R
- [x] sprintf_linter.R
- [x] string_boundary_linter.R
- [x] strings_as_factors_linter.R
- [x] system_file_linter.R
- [x] unneeded_concatenation_linter.R
- [ ] unused_import_linter.R
- [x] yoda_test_linter.R

---

_Original issue raised_

As seen in #1353, #1340, #1310, there are subtle performance implications to the way we write our XPaths.

One thing that became clear is that writing `//NODE1[NODE2]` is slower than `//NODE2[parent::NODE1]` if `NODE2` is far less frequent than `NODE1`.

And `<expr>` is by far the most common node; here's a guess at the general frequency by tabulating across r-devel and my local R packages/scripts:

```r
library(data.table)
library(knitr)

get_r_files = function(dir) list.files(dir, pattern = "\\.R$", recursive = TRUE, full.names = TRUE)
get_freq = function(f) {
  pc = tryCatch(parse(f), error = identity, warning = identity)
  if (inherits(pc, "condition")) return(NULL)
  setDT(getParseData(pc))[, .N, by = token]
[...]

PR #1377, opened 2022-06-08
Adds `expect_that()` to undesirable functions linter

IndrajeetPatil:
~~If code is expected to produce an error, using `expect_that(x, throws_error())` works but is unadvised. It is better to use the function dedicated specifically for checking errors: `expect_error(x)`. The additional benefit is that you can also check for the expected error message using regular expressions.~~

First time making a PR to _lintr_, so any help in getting the PR aligned with coding standards (if I'm violating any) adopted in this repo is greatly appreciated.


R/expect_error_linter.R
```diff
@@ -0,0 +1,37 @@
...
+#' dedicated specifically for checking errors: `expect_error(x)`. The additional
+#' benefit is that you can also check for the expected error message using
+#' regular expressions.
+#'
+#' @evalRd rd_tags("expect_error_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+expect_error_linter <- function() {
+  # one case to match: expect_that(x, throws_error())
+  xpath <- glue::glue("//expr[
```
AshesITR, line 14: Try to avoid `//expr` for performance reasons (cf. #1358)

AshesITR: Also, e.g. in https://github.com/r-lib/lintr/blob/5cbc007df1032ec6d7b0d11c3abe7f692019166c/R/expect_comparison_linter.R#L15-L19 we exclude expressions from linting if the `info` argument is present. Is this necessary here, too?

AshesITR (commented):
Thanks for taking the time to create a PR - it looks quite good already.
From a meta point of view: Is there any circumstance where `expect_that()` is preferrable to other expectations?

It might also be worthwhile to call this linter `expect_that_linter()` and have it lint all flavors of `expect_that(x, some_condition)`where there is an appropriate equivalent `expect_some_condition(x)`. WDYT?

MichaelChirico:
> It might also be worthwhile to call this linter `expect_that_linter()` and have it lint all flavors of `expect_that(x, some_condition)`where there is an appropriate equivalent `expect_some_condition(x)`. WDYT?

Came here to say just that -- the [documentation](https://testthat.r-lib.org/reference/expect_that.html) explicitly says:

> An old style of testing that's no longer encouraged.

So linting this to encourage folks to move off of it (maybe with `# nolint` if there's something that can only be done with `test_that`? I am unfamiliar with this approach so I don't know) would be appropriate IMO

MichaelChirico:
But then, isn't this just a use case for `undesirable_function_linter`? :thinking:

IndrajeetPatil:
Thanks for the feedback, both. 

Yeah, I initially thought of writing `expect_that_linter`, but wasn't sure if it was too aggressive since `expect_that()` was superseded only [recently](https://testthat.r-lib.org/news/index.html#testthat-300).

But, given that we are just issuing a warning, I think it might be worth it. 

> But then, isn't this just a use case for `undesirable_function_linter`?

Wasn't aware of this linter. Yeah, I would agree then that this should just be part of this linter.

I will update the PR accordingly.

IndrajeetPatil:
@MichaelChirico, @AshesITR WDYT? Is this better?

R/zzz.R
```diff
@@ -55,6 +56,7 @@ default_linters <- modify_defaults(
 #'  * [debugonce()] is only useful for interactive debugging. It should be removed.
 #'  * [detach()] modifies the global search path. Detaching environments from the search path is rarely necessary in
 #'    production code.
+#'  * [testthat::expect_that()] This style of testing is no longer encouraged by `{testthat}` package. Using a more specific expect_ function (e.g. `expect_equal()`, `expect_s3_class()`, etc.) is recommended.
```
MichaelChirico, line 59: line width

IndrajeetPatil: Yeah, I wasn't sure about that because the other items in the list don't seem to follow 80-width rule.

At any rate, I've changed the new entry to follow this rule.

MichaelChirico: we do 80 kinda organically, our max is 120 though

MichaelChirico:
Again per the docs:

https://github.com/r-lib/testthat/blob/426eb523f6772c76175d6e8d7ca04355b1ee7ad5/R/expect-that.R#L8-L9

So actually this function is deprecated in 3e... that eliminates any hesitation I had about adding it as a default undesirable. But let's also make note of that in the lint message/documentation.

IndrajeetPatil:
> But let's also make note of that in the lint message/documentation

You mean we should mention that this function is deprecated in 3e in the message/documentation?

MichaelChirico:
> You mean we should mention that this function is deprecated in 3e in the message/documentation?

yep, exactly

IndrajeetPatil:
@MichaelChirico I've updated the message. Lemme know if you want me to change the wording.

MichaelChirico:
be sure to add a NEWS bullet & cite yourself!

IndrajeetPatil:
Done!

MichaelChirico:
LGTM, but now I'm wondering -- why stop here? shouldn't we add _all_ deprecated tidyverse functions here? 🤔

WDYT @AshesITR should we just merge this now and consider that in follow-up?

IndrajeetPatil:
Hmm, I would say that that would be expanding the scope of this PR beyond its initial purview. 

It'd be better to create a new issue solely for _tidyverse_'s deprecated functions, and then I can make a separate PR targeting that particular issue.

Sounds reasonable?

AshesITR:
I agree. Let's create a follow-up issue (maybe one per package?) to keep the PR focused.

Are there other deprecated functions in testthat?

IndrajeetPatil:
[expect_known_output()](https://testthat.r-lib.org/reference/expect_known_output.html), 
[expect_known_value()](https://testthat.r-lib.org/reference/expect_known_output.html), 
[expect_known_hash()](https://testthat.r-lib.org/reference/expect_known_output.html), 
[expect_equal_to_reference()](https://testthat.r-lib.org/reference/expect_known_output.html)
[expect_equivalent()](https://testthat.r-lib.org/reference/expect_equivalent.html)
[expect_reference()](https://testthat.r-lib.org/reference/expect_reference.html)
[expect_is()](https://testthat.r-lib.org/reference/expect_is.html)
[setup()](https://testthat.r-lib.org/reference/teardown.html)
[teardown()](https://testthat.r-lib.org/reference/teardown.html)

I am not sure about linting these, though. 

If you are using 3rd edition of _testthat_, _testthat_ itself will warn you about these functions being deprecated in the 3rd edition.

MichaelChirico:
there are a bunch:

expect_more_than(), expect_less_than()
context()
expect_equivalent(), expect_reference()
with_mock(), local_mock()
expect_is()

everything in these files:

https://github.com/r-lib/testthat/blob/f8ccf033a74d89e2d1d3cb2ec3c6c4f63ca69055/R/old-school.R

https://github.com/r-lib/testthat/blob/f8ccf033a74d89e2d1d3cb2ec3c6c4f63ca69055/R/deprec-condition.R

https://github.com/r-lib/testthat/blob/6666662844274e8fa1988c8e0cfecf0b13399ee1/R/expect-known.R

----

an added complication is some are plain deprecated, others are 3e-deprecated only. maybe what we want to do is export an object to encapsulate all this.

because that would entail a user-facing change vs this PR as of now, I lean towards tabling this until after 3.0.0 release so we can handle the design more carefully, WDYT?

IndrajeetPatil:
> I lean towards tabling this until after 3.0.0 release so we can handle the design more carefully, WDYT?

Sounds good. I am marking this as a draft for now. We can come back to it later.

Similar to `undesirable_function_linter()`, I am wondering if we should introduce a new `deprecated_function_linter()` in the next release. Functions like `expect_that()` are more suitable for this linter than undesirable linter.

IndrajeetPatil:
Closing in favour of #1386.

MichaelChirico:
Thanks again for the PR! Sorry it didn't land, & thank you for your understanding. Looking forward to the bigger & better version!

PR #1378, opened 2022-06-09
de-lint

AshesITR:
GH mobile doesn't show lints :(


MichaelChirico:
I should also follow our own Emacs instructions to prevent this on my side :sweat_smile:

MichaelChirico (approved):

PR #1383, opened 2022-06-12
add tag "executing" to linters that call eval() or loadNamespace() on arbitrary input

### LINKED ISSUE #1380: Add a tag to linters that loadNamespace()
I think these linters have different enough requirements to warrant their own tag, it would make excluding them simpler for users that can't run `loadNamespace()` reliably in their environment.

AshesITR:
Fixes #1380 

I searched the code for `eval()`, `loadNamespace()` and `requireNamespace()`.
The following linters turned up and were *not* tagged, since the possible input is quite restricted.
LMK if they should also be tagged.

 * `sprintf_linter()` via `capture_sprintf_warning()`: Input is confined to known-constants. Security of this linter depends on security of `sprintf()` on arbitrary _atomic constant_ input.
 * `fixed_regex_linter()` via `get_token_replacement()`: Input is confined to matches of a regular language. We know it's a string literal that matches our fixed regex. Security depends on correctness of this regex.

I've tagged

 * `namespace_linter()`: calls `getNamespace()` and `.getNamespaceInfo()` directly.
 * `object_length_linter()` via `namespace_imports()`: calls `requireNamespace()` and `getNamespaceExports()`.
 * `object_name_linter()`: ditto
 * `object_usage_linter()`: calls `eval()` on function declarations directly. Security depends on correctness of the XPath, `get_content()` and that `eval()` of a top-level `FUNCTION` expression is safe for any code.
 * `unused_import_linter()`: calls `getNamespace()` and `.getNamespaceInfo()` directly.


MichaelChirico (approved):
LGTM. I have close to 0 security expertise so this is mostly suppositional.

I guess loadNamespace() is the most "dangerous" since e.g. a GtiHub package could execute arbitrary code in .onLoad(). For CRAN packages we rely on CRAN security checks.

evalq(foo <- function(...) { ... }) doesn't seem risky to me unless we run foo() (or at least, security is owned by the R parser)

PR #1391, opened 2022-06-14
Fix typos

salim-b:


R/lint.R
```diff
@@ -12,7 +12,7 @@
 #'
 #' * `lint()` lints a single file.
 #' * `lint_dir()` lints all files in a directory.
-#' * `line_pakage()` lints all likely locations for R files in a package, i.e.
+#' * `lint_package()` lints all likely locations for R files in a package, i.e.
```
MichaelChirico, line 15: could you run roxygenize as well?

salim-b: I didn't actually clone the repo locally first, just did the fix via GH's web interface. But now I did also rebuild the doc :)

MichaelChirico: thanks! I know there's a bot to run document so you wouldn't need a local patch.... but haven't figured that out myself.

thanks for the extra effort

MichaelChirico (approved):

PR #1395, opened 2022-06-15
wide line lint

MichaelChirico:


AshesITR (commented):
document()?

MichaelChirico:
doesn't roxygen do its own string wrapping? I thought it wouldn't matter

MichaelChirico:
I stand corrected :)

done

PR #1398, opened 2022-06-16
`seq_linter` lints on dplyr and data.table flavors

### LINKED ISSUE #1396: seq_linter should lint on 1:n() [dplyr] and 1:.N [data.table]
Both of these are flavor-specific versions of `1:nrow(.)` that we should recognize and lint

IndrajeetPatil:
Closes #1396

### DIFF
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -1,7 +1,7 @@
 #' Sequence linter
 #'
 #' Check for `1:length(...)`, `1:nrow(...)`, `1:ncol(...)`, `1:NROW(...)` and `1:NCOL(...)` expressions in base-R.
-#' Additionally, `1:n()` (from dplyr package) and `1:.N` (from data.table package) are also checked.
+#' Additionally check `1:n()` (from dplyr) and `1:.N` (from data.table).
 #' These often cause bugs when the right-hand side is zero.
 #' It is safer to use [base::seq_len()] or [base::seq_along()] instead.
 #'
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -18,7 +18,7 @@ seq_linter <- function() {
     and (
           expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
           or
-          expr[SYMBOL='.N']
+          expr[SYMBOL = '.N']
         )
   ]")
 
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -1,6 +1,7 @@
 #' Sequence linter
 #'
-#' Check for `1:length(...)`, `1:nrow(...)`, `1:ncol(...)`, `1:NROW(...)` and `1:NCOL(...)` expressions.
+#' Check for `1:length(...)`, `1:nrow(...)`, `1:ncol(...)`, `1:NROW(...)` and `1:NCOL(...)` expressions in base-R.
+#' Additionally, `1:n()` (from dplyr package) and `1:.N` (from data.table package) are also checked.
 #' These often cause bugs when the right-hand side is zero.
 #' It is safer to use [base::seq_len()] or [base::seq_along()] instead.
 #'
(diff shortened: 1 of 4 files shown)


R/seq_linter.R
```diff
@@ -1,19 +1,25 @@
...
 
+  # `.N` from {data.table} is special since it's not a function but a symbol
   xpath <- glue::glue("//expr[
     expr[NUM_CONST[text() =  '1' or text() =  '1L']]
     and OP-COLON
-    and expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
+    and (
+          expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
+          or
+          expr[SYMBOL='.N']
```
MichaelChirico, line 21: oh, I didn't know this works!

usually we'd do

SYMBOL[text() = '.N']

let me check the docs to ensure there's nothing subtle about this XPath usage

tests/testthat/test-seq_linter.R
```diff
@@ -49,6 +49,19 @@ test_that("finds 1:length(...) expressions", {
     rex("dim(...)", anything, "Use seq_len"),
     linter
   )
+
+  expect_lint(
+    "function(x) { mutate(x, .id = 1:n()) }",
+    rex("n(...)", anything, "Use seq_len"),
```
MichaelChirico, line 55: not ideal that the message is producing n(...) since n() doesn't take any arguments:

https://dplyr.tidyverse.org/reference/context.html

would you mind having a go at special-casing n() in the message generation so that the message is n() not n(...)?

if you're not having any luck, that's OK -- we can fix it in follow-up. the message for seq_linter needs improvement more generally anyway.

IndrajeetPatil: Good catch!

Yeah, I can give it a try.

R/seq_linter.R
MichaelChirico: nice! looks like this is legit:

> If one object to be compared is a node-set and the other is a string, then the comparison will be true if and only if there is a node in the node-set such that the result of performing the comparison on the string-value of the node and the other string is true.

https://www.w3.org/TR/1999/REC-xpath-19991116/#section-Expressions

Its for != where things are a bit trickier:

https://stackoverflow.com/a/4629660/3576984

tests/testthat/test-seq_linter.R
IndrajeetPatil: @MichaelChirico Just pushed something to deal with `n()` in special way.

Let me know if you think it's not the best way to proceed on this.

tests/testthat/test-seq_linter.R
```diff
@@ -49,6 +49,33 @@ test_that("finds 1:length(...) expressions", {
...
+  ))
+
+  expect_lint(
+    "function(x) { x[, .id := 1:.N] }",
+    rex(".N", anything, "Use seq_len"),
+    linter
+  )
+
+  # `.N` is a symbol, so the message should not print it as a function
+  expect_error(expect_lint(
```
MichaelChirico, line 73: I would just improve the regex I'm the first test rather than add a second test

tests/testthat/test-seq_linter.R
```diff
@@ -49,6 +49,33 @@ test_that("finds 1:length(...) expressions", {
...
   )
+
+  expect_lint(
+    "function(x) { mutate(x, .id = 1:n()) }",
+    rex("n()", anything, "Use seq_len"),
+    linter
+  )
+
+  # `dplyr::n()` takes no arguments, so the message shouldn't have `n(...)`
+  expect_error(expect_lint(
```
MichaelChirico, line 60: not necessary, already implied by the first test

MichaelChirico: (btw I think testthat offers expect_failure() for expecting failed expectations, and expect_success for the reverse)

IndrajeetPatil:
@MichaelChirico Thanks! I've improved the regex and removed the redundant tests.

MichaelChirico (approved):

MichaelChirico:
Thanks for your work! 🙌

PR #1407, opened 2022-06-17
prioritize plain R code over knitr

### LINKED ISSUE #743: R files generating knitr templates are parsed as Rmd
If I have a file that generates `knitr` templates, e.g.

    writeLines('
    CreateTemplate <- function(x) {
    return(sprintf("
    ```{r code}
    foo <- function(x) x+%d
    foo(5)
    ```", x))
    }', tmp <- tempfile())

`parse(tmp)` works just fine, however, `lintr` assumes this is an Rmd based on matching some chunk regex:

```
lintr::lint(tmp)
# /tmp/RtmpN8X7Dj/file3d0536f05e5d:4:1: error: Missing chunk end for chunk (maybe starting at line 4).
# ```{r code}
# ^
```

Should we try parsing the file as R code first, then checking Rmd chunks afterwards?

### LINKED ISSUE #879: lintr:::get_source_expressions fails when a code comment contains an r script command surrounded by backticks
We were surprised our files linted correctly despite having obvious linting issues, so I have investigated and narrowed it down to this minimal reprex:
```r
# `r print("7")`
function() 2<=3
```
The above doesn't lint.

I further narrowed the issue down to backticks in combination with r and something because the following do lint correctly:
```r
# `print("7")`
function() 2<=3

# `r`
function() 2<=3

# r print("7")
function() 2<=3
```
In short - all three are needed: backticks, r and something in addition to r.

Then, I dug inside inflix_spaces_linter and tracked the issue to the parsing function, Given a file "test.R", `lintr:::get_source_expressions("test.R")` returns:
image
Without the comment, it parses correctly:
image

Environment:
lintr version 2.0.1
R: 4.1.0

Is this a known issue?

### LINKED ISSUE #1406: `extract_r_source()` is overly eager with detecting knitr patterns
It seems the test code is linted as Rmd for some reason?!
The NA should be investigated in a separate issue. Maybe we stumbled on an actual bug in the extractor code.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/1403#discussion_r900382755_

AshesITR:
fixes #1406 
fixes #743 
fixes #879

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 
 ## Bug fixes
 
-* `get_source_expressions()` no longer fails on R files that match a knitr pattern (#1406, @AshesITR).
+* `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
 * Parse error lints now appear with the linter name `"error"` instead of `NA` (#1405, @AshesITR).  
   Also, linting no longer runs if the `source_expressions` contain invalid string data that would cause error messages
   in other linters. 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,10 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table)  (#1396, @IndrajeetPatil).
 
+## Bug fixes
+
+* `get_source_expressions()` no longer fails on R files that match a knitr pattern (#1406, @AshesITR).
+
 # lintr 3.0.0
 
 ## Breaking changes
--- a/R/comment_linters.R
+++ b/R/comment_linters.R
@@ -79,7 +79,7 @@ commented_code_linter <- function() {
 
 # is given text parsable
 parsable <- function(x) {
-  if (is.na(x)) {
+  if (anyNA(x)) {
     return(FALSE)
   }
   res <- try_silently(parse(text = x))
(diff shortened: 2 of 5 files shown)


R/comment_linters.R
```diff
@@ -79,7 +79,7 @@ commented_code_linter <- function() {
 
 # is given text parsable
 parsable <- function(x) {
-  if (is.na(x)) {
+  if (anyNA(x)) {
```
AshesITR, line 82: This is required so that `parsable()` is vectorized over `x`.

MichaelChirico:
does this fix #743

AshesITR:
yes.

MichaelChirico:
#879 too?

(sorry I knew there were two but search was eluding me)

NEWS.md
```diff
@@ -7,6 +7,7 @@
 
 ## Bug fixes
 
+* `get_source_expressions()` no longer fails on R files that match a knitr pattern (#1406, @AshesITR).
```
MichaelChirico, line 10: please add the other issue #s here too

tests/testthat/test-get_source_expressions.R
```diff
@@ -249,6 +249,13 @@ test_that("#1262: xml_parsed_content gets returned as missing even if there's no
   expect_identical(source_expressions$expressions[[1L]]$full_xml_parsed_content, xml2::xml_missing())
 })
```
MichaelChirico, line 251: nit: i usually like to add tests from duplicate issues just to ensure future updates don't regress one issue but not the others, i.e. just copy-paste the reported reprex with expected output. i think we're pretty confident that the issues are the exact same root cause, but better safe than sorry.

MichaelChirico (approved):

PR #1408, opened 2022-06-17
set error linter name to "error"

### LINKED ISSUE #1405: github comment bot failed to find the linter name, prints NA
Note the NA -- seems the linter name has not been picked up correctly:

![Check warning on line 44 in tests/testthat/test-trailing_blank_lines_linter.R\nGitHub Actions\n/ lint tests/testthat/test-trailing_blank_lines_linter.R#L44\nfile=tests/testthat/test-trailing_blank_lines_linter.R,line=44,col=1,[NA] Missing chunk end for chunk (maybe starting at line 44).](https://user-images.githubusercontent.com/7606389/174351263-89786089-086f-4602-8683-88855040671e.png)

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/1403#discussion_r900376909_

AshesITR:
fixes #1405


AshesITR:
As it turns out, the unparseable files actually caused `lint()` to error due to `line_length_linter()` running `nchar()` on `file_lines`.
I opted to simply refuse to lint files with completely invalid content (it's not parseable anyway).

R/lint.R
```diff
@@ -88,9 +88,11 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
   }
 
   lints <- list()
-  for (expr in source_expressions$expressions) {
-    for (linter in names(linters)) {
-      lints[[length(lints) + 1L]] <- get_lints(expr, linter, linters[[linter]], lint_cache, source_expressions$lines)
+  if (!is_tainted(source_expressions$lines)) {
```
MichaelChirico, line 91: does that mean such files go undetected?

what if I have an encoding issue without realizing it, and am trying to figure out why `a = 1` isn't triggering assignment linter, how do I diagnose that?

MichaelChirico: partially orthogonal: should we make a note of this behavior in the ?lint documentation?

AshesITR: It will produce an `error` lint. Only other lints are disabled and only if the parse error leaves butchered `file_lines`.

AshesITR: > partially orthogonal: should we make a note of this behavior in the ?lint documentation?

Good idea. "If files contain unparseable encoding problems, only the encoding problem will be linted to avoid unintelligible error messages from other linters"?

MichaelChirico: SGTM. LGTM otherwise

MichaelChirico (approved):

PR #1410, opened 2022-06-17
fix documentation reference to exclusions

### LINKED ISSUE #1401: Using `exclude:` in `.lintr` results in error
When using the 3.0.0 CRAN release of lintr, listing with the following `.lintr` file results in an unexpected error:

```r
linters: linters_with_defaults(
    line_length_linter(100) 
  )
exclude: list(
    "bad.R" = list(
      1, # global exclusions are unnamed
      line_length_linter = 4:6
    )
  )
```

(adapted from the [Get Started](https://lintr.r-lib.org/articles/lintr.html) page)

Error message when running `lint_dir()`:

```
$ Rscript -e 'lintr::lint_dir(".")'
Error in `[.data.frame`(rex::re_matches(lines, exclude, locations = TRUE),  : 
  undefined columns selected
Calls: <Anonymous> ... do.call -> <Anonymous> -> lapply -> FUN -> [ -> [.data.frame
Execution halted
```

To reproduce: run `Rscript -e 'lintr::lint_dir(".")'` from a directory containing the `.lintr` file above, and the following `bad.R`:

```r
tmp = "value"

# regular comment
# really long comment ............................................................................................................
# really long comment ............................................................................................................
# really long comment ............................................................................................................
# regular comment
```

AshesITR:
Exclude is the regex for `# nolint`, exclusions is for specifying exclusions.

fixes #1401

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 * `get_source_expressions()` no longer omits trailing non-code lines from knitr files (#1400, @AshesITR).  
   This fixes the location information for `trailing_blank_lines_linter()` in RMarkdown documents without terminal
   newlines.
+* The `vignette("lintr")` incorrectly cited `exclude` as the key for setting file exclusions in `.lintr` when it is 
+  actually `exclusions`. (#1401, @AshesITR)
 
 # lintr 3.0.0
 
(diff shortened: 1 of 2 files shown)


MichaelChirico:
LGTM, but the names do seem needlessly overlapped. how about exclusion_regex and file_exclusions instead? (as follow-up)

MichaelChirico (approved):

AshesITR:
That will be a breaking change, though.
I'd actually like the simple `exclude: ...` for exclusions - it's very intuitive.
`exclusion_regex` seems good to me as well, maybe rename the start and end regexes as well, then.

Issue #1043, opened 2022-04-02
Rtex code extraction problematic


AshesITR:
When bringing #165 up to speed, I noticed a test failure for Rtex, caused by the behavior of `get_source_expressions()`:

```r
get_source_expressions("tests/testthat/knitr_formats/test.Rtex")$lines[[25]]
#> [1] " "
```

The original file line in question is `"%"`, where the `%` is masked by whitespace as part of the Rtex markup.
This results in a false positive `trailing_whitespace_linter()` hit for that line once it's extended to cover all lines and not just lines with expressions.

MichaelChirico:
Copying over some more context from a TODO comment:

> `file_lines` contains a whitespace on the final line for Rtex, because that is used to mark the Rtex escape char "%" as well. c.f. `get_source_expressions("tests/testthat/knitr_formats/test.Rtex")$lines[[25]]`

PR #1411, opened 2022-06-18
implement indentation_linter

### LINKED ISSUE #63: indentation consistency

AshesITR:
supersedes #497
fixes #63


AshesITR:
Putting this out here looking for feedback.
CC @dgkf

AshesITR:
So far, every mis-indented line causes a lint.
That means if, e.g., en entire block is mis-indented, there will be tons of lints with identical messages.

```
  top_level_misindent <- function() {
    foo
    bar
  }
```

Causes four lints, one per line.

AshesITR:
So, there were a ton of indentation lints in lintr itself.
One particular problem is the following with hanging indents:

```r
lapply(1:10, function(x) {
  x^2
})

structure(1:10 %>%
            cumsum() %>%
            mean(), class = "meh")
```

Currently, the first code doesn't lint, because hanging indents are suppressed if there is a multi-line expression on the same line as the `(`.
This however causes the second piece of code to lint, because no hanging indent situation is detected.

Any thoughts on how hanging indents should be defined and detected and what exception there should be?

MichaelChirico:
This one definitely seems susceptible to edge cases -- have you tried any .dev/compare_branches here to see what sort of patterns might be missed "in the wild"? (can only turn up false positives of course).

also, at a glance there's a lot going on & several loops -- I'm curious the performance relative to other default linters.

AshesITR:
I've hand sifted all lints on lintr (which brought up quite a few edge cases that made it to the tests), but haven't run compare_branches yet.
That's definitely a good idea to do, but I'm also interested in preliminary feedback e.g. regarding the hanging indents.

Will try to get compare_branches to run.

Regarding performance, I'm not too worried because the XPath should produce relatively few hits on each expression and is written to avoid common nodes (such as expr), but will produce a benchmark when satisfied with the behavior.

AshesITR:
I ran on the top 20 CRAN packages, yielding quite a few lints.

CLI has some odd formatted `tryCatch()` code, which doesn't expect a paren-indent because it's similar to the `lapply` example.
https://github.com/r-lib/cli/blob/76a2733b2a15893537c1ac95dee7dbd6bb885513/inst/examples/apps/news.R#L11-L26

Hanging indents of `if` conditions with logical conjunctions are expected to be indented by one more level, unlike here.
I can't find guidance in the tidyverse style guide on multi-line if conditions, just a hunch that they are discouraged alltogether.
https://github.com/r-lib/cli/blob/76a2733b2a15893537c1ac95dee7dbd6bb885513/R/ansi-hyperlink.R#L83-L84
Styler does something very ugly:
```r
styler::style_text("
if (a &&
    b) {
  boo
}
")

# and

styler::style_text("
if (a &&
      b) {
  boo
}
")

#> if (a &&
#>   b) {
#>   boo
#> }
```

This one seems like it definitely shouldn't lint, but currently lints ll. 28-37:
https://github.com/r-lib/cli/blob/76a2733b2a15893537c1ac95dee7dbd6bb885513/inst/scripts/up.R#L22-L43

Ditto here, I think a hanging indent should trigger, although L232 should lint because the argument binds to `||` in the preceding line.
https://github.com/r-lib/cli/blob/76a2733b2a15893537c1ac95dee7dbd6bb885513/R/ansi.R#L231-L234

The huge number of lints produced will require careful examination.
For now, I'll try to get the lints on `cli` to be satisfying.

AshesITR:
Satisfied with cli now.
ggplot2 features a style that triggers the hanging indent rule, causing lots of lints.
Maybe we need an option to allow unnamed arguments on the same line as the function call, followed by only named arguments on subsequent lines, to be block-indented instead of hanging-indented. WDYT @MichaelChirico?

R/scale-.R:801

```r
#' @rdname ggplot2-ggproto
#' @format NULL
#' @usage NULL
#' @export
ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
  drop = TRUE,
  na.value = NA,
  n.breaks.cache = NULL,
  palette.cache = NULL,

  is_discrete = function() TRUE,

  train = function(self, x) {
    if (length(x) == 0) {
      return()
    }
    self$range$train(x, drop = self$drop, na.rm = !self$na.translate)
  }

# ...
)
```

AshesITR:
Here's a current tally of lints:

```
     package    N
 1:        sf 8155 -- uses indent = 1
 2:       rgl 4015 -- uses indent = 4
 3:     rgeos 2295 -- uses indent = 4
 4:   ggplot2 1278
 5:   stringi 1228 --  uses indent = 4
 6:     vctrs  463
 7:     dplyr  297
 8:     rlang  263
 9:       cli  215
10:  magrittr  209
11:   stringr  152
12:    tibble  151
13:  devtools  113
14:  processx   89
15:      glue   81
16:      ragg   28
17:    pillar   15
18: lifecycle    5
```

Deduplicating consecutive lines:
```
      package   N
 1:   stringi 738
 2:   ggplot2 468
 3:     vctrs 137
 4:     dplyr 116
 5:       cli 108
 6:     rlang  72
 7:  processx  69
 8:  magrittr  65
 9:   stringr  63
10:      glue  34
11:    tibble  33
12:  devtools  26
13:      ragg  19
14:    pillar   4
15: lifecycle   3
```

Both stringi and ggplot2 feature function declarations of this form:
```r
fun <- function(a, b)
{
  code
}
```

Since that style is caught by `brace_linter()`, indentation linter should let it slide.
Working in an exception for the hanging indent of `function()` s if they start with `{`.

AshesITR:
Here are the total timing stats of a820b6e by linter on the 20 packages.
`indentation_linter` took 1.3% of the total time spent, putting it at rank 59/73 of all linters by computational cost.

<details>
<pre>
                              linter total_duration relative_duration
 1: function_left_parentheses_linter     1291882077        0.01594003
 2:                     brace_linter     1245462984        0.01536728
 3:              class_equals_linter     1238749606        0.01528445
 4:           missing_package_linter     1234203467        0.01522836
 5:              infix_spaces_linter     1220307232        0.01505689
 6:        strings_as_factors_linter     1214248867        0.01498214
 7:               line_length_linter     1202948999        0.01484272
 8:        duplicate_argument_linter     1202036031        0.01483145
 9:                paren_body_linter     1193722645        0.01472888
10:                 pipe_call_linter     1181258225        0.01457508
11:              regex_subset_linter     1174803442        0.01449544
12:          nonportable_path_linter     1174491722        0.01449160
13:      undesirable_operator_linter     1174082240        0.01448654
14:       trailing_whitespace_linter     1170982557        0.01444830
15:                 semicolon_linter     1155365517        0.01425560
16:              todo_comment_linter     1150029714        0.01418977
17:          implicit_integer_linter     1146832117        0.01415031
18:            commented_code_linter     1145512625        0.01413403
19:    unneeded_concatenation_linter     1144953767        0.01412714
20:             inner_combine_linter     1141912308        0.01408961
21:       extraction_operator_linter     1138723378        0.01405026
22:                    commas_linter     1137843155        0.01403940
23:             single_quotes_linter     1132992952        0.01397956
24:             ifelse_censor_linter     1132216423        0.01396998
25:               fixed_regex_linter     1131842636        0.01396536
26:          unreachable_code_linter     1131047764        0.01395556
27:             absolute_path_linter     1130687039        0.01395111
28:                 yoda_test_linter     1130656935        0.01395073
29:                     paste_linter     1129108002        0.01393162
30:             expect_length_linter     1125918201        0.01389227
31:      trailing_blank_lines_linter     1125338691        0.01388512
32:             unused_import_linter     1124164869        0.01387063
33:                 any_is_na_linter     1122566754        0.01385091
34:             conjunct_test_linter     1119299362        0.01381060
35:               system_file_linter     1118847791        0.01380503
36:            T_and_F_symbol_linter     1116341918        0.01377411
37:                       seq_linter     1113902970        0.01374401
38:         pipe_continuation_linter     1111198979        0.01371065
39:                    no_tab_linter     1110540563        0.01370253
40:         expect_true_false_linter     1105829983        0.01364440
41:      numeric_leading_zero_linter     1105823903        0.01364433
42:              expect_named_linter     1105256359        0.01363733
43:             package_hooks_linter     1104493164        0.01362791
44:             object_length_linter     1103934584        0.01362102
45:         expect_comparison_linter     1103228729        0.01361231
46:               expect_type_linter     1102793176        0.01360693
47:          expect_identical_linter     1099999290        0.01357246
48:           expect_s3_class_linter     1093578888        0.01349324
49:                expect_not_linter     1092318514        0.01347769
50:           expect_s4_class_linter     1092093070        0.01347491
51:          missing_argument_linter     1090397590        0.01345399
52:                   sprintf_linter     1088241068        0.01342738
53:                  backport_linter     1083290939        0.01336630
54:                assignment_linter     1081623392        0.01334573
55:      undesirable_function_linter     1076559637        0.01328325
56:                 equals_na_linter     1069672637        0.01319827
57:          literal_coercion_linter     1068198277        0.01318008
58:               expect_null_linter     1063382822        0.01312067
59:               indentation_linter     1057241804        0.01304489
60:            any_duplicated_linter     1051494281        0.01297398
61:             spaces_inside_linter     1050104439        0.01295683
62:          redundant_ifelse_linter     1029116511        0.01269787
63:         condition_message_linter     1025801896        0.01265697
64:              vector_logic_linter     1019118627        0.01257451
65:     consecutive_stopifnot_linter     1005379514        0.01240499
66:               object_name_linter     1003824867        0.01238580
67:   spaces_left_parentheses_linter     1003486443        0.01238163
68:                 namespace_linter      998576641        0.01232105
69:             nested_ifelse_linter      992548021        0.01224666
70:           string_boundary_linter      985214122        0.01215617
71:            outer_negation_linter      980182593        0.01209409
72:              object_usage_linter      969743071        0.01196528
73:                 cyclocomp_linter      956833921        0.01180600
                              linter total_duration relative_duration
</pre>
</details>

MichaelChirico:
Thanks for the work so far! Heads up it may take me longer to review this one as I'm on vacation.

MichaelChirico:
> Maybe we need an option to allow unnamed arguments on the same line as the function call, followed by only named arguments on subsequent lines, to be block-indented instead of hanging-indented. WDYT @MichaelChirico?

I think this is OK by the [style guide](https://style.tidyverse.org/syntax.html#long-lines), and `styler` also allows this:

> Short unnamed arguments can also go on the same line as the function name, even if the whole function call spans multiple lines.

```r
styler::style_text('
ScaleDiscrete <- ggproto("ScaleDiscrete", Scale,
  drop = TRUE,
  na.value = NA,
  n.breaks.cache = NULL,
  palette.cache = NULL,
  is_discrete = function() TRUE,
  train = function(self, x) {
    if (length(x) == 0) {
      return()
    }
    self$range$train(x, drop = self$drop, na.rm = !self$na.translate)
  }

  # ...
)')
# [unchanged]
```

MichaelChirico:
> Since that style is caught by brace_linter(), indentation linter should let it slide. Working in an exception for the hanging indent of function() s if they start with {.

Agree here

AshesITR:
> So yes, we should not lint by default. I wouldn't even make it an option for now.

The difficulty with that is, that hanging indents are _also_ okay and I (currently) only provision for one unique expected indentation level.

We don't want to lint any of the following code, right?

```r
lapply(very_long_x_arg, fun_arg,
       extra_arg = 42)

lapply(very_long_x_arg, fun_arg,
  extra_arg = 42
)
```

MichaelChirico:
I would lint the first one -- that style is only allowed for function definitions IINM. observe what styler does to it. See also https://github.com/tidyverse/style/pull/174

AshesITR:
IINM auto-indent of RStudio and PyCharm does make these hanging though...
I also tend to use that style if overall readability benefits from the fewer lines used up by that vs. a fully block-indented call.
I personally don't like the trailing arguments on the first call if there is no hanging indent because they are very easy to overlook.

MichaelChirico:
I think the style guide is pretty clear here, which is what we enforce:

> If a function call is too long to fit on a single line, use one line each for the function name, each argument, and the closing `)`

Anything else has to be done through arguments.

IndrajeetPatil:
Should #497 be closed in favour of this one? Both are about implementing indentation linters.

dgkf:
@IndrajeetPatil - I'm fine with that (I'm the author of #497). 

I'm quite fond of the approach that I settled on, but other priorities came up before I could flesh it out. Interest of seeing _an_ approach finally get accepted, I'm happy to see all the progress being made here.

IndrajeetPatil:
@dgkf Thanks a lot for the quick response and for being generous about this! :)

Please let us know if we miss anything here that you had considered while working on #497.

dgkf:
Thanks @IndrajeetPatil - the approaches are very different. That's not a bad thing! This approach is much more focused, and I think that's great. It fits the ethos of `lintr` better.

> I was writing up a reply and went to link to some examples of the implementation and realized I never pushed my most recent work to GitHub :sob:. I pushed up the code [to a new branch](https://github.com/dgkf/lintr/tree/dgkf/indentation_linter_refactor) if you're curious to see some of the ideas I was exploring, but this code is literally whatever stale work-in-progress happened to be on my machine at the moment. 

Just to highlight a few interesting concepts that I explored:

- An indentation style has [its own constructor](https://github.com/dgkf/lintr/blob/dgkf/indentation_linter_refactor/R/indent_styles.R), and the low-level [indentation_linter accepts these styles as parameters](https://github.com/dgkf/lintr/blob/d8e81895f4db2fe14acb793586c6f1e46b7ecccf/R/indentation_linter.R#L11-L16) for different parts of the code (function headers, function calls, control flow blocks, etc)
- Indentation styles can be unions - for example the [default function call style](https://github.com/dgkf/lintr/blob/d8e81895f4db2fe14acb793586c6f1e46b7ecccf/R/indentation_linter.R#L14) permits _either_ indenting newlines by one indentation, _or_ indenting to the call's opening parenthesis.
- Then the identification of tokens that fail these criteria [ends up being relatively standardized](https://github.com/dgkf/lintr/blob/d8e81895f4db2fe14acb793586c6f1e46b7ecccf/R/indentation_linter.R#L30-L34), with some minor differences to identify the parts of code that the indentation applies to.
- Indentation ends up being a set of 5 linters (the 5 style arguments - one for function header indentation, one for function call indentation, etc..). Because they all depend on [the same data on the parse data.frame](https://github.com/dgkf/lintr/blob/d8e81895f4db2fe14acb793586c6f1e46b7ecccf/R/indentation_linter.R#L27), it reduces redundant calculation to tuck them all in the same place. 

A few of these [sub-linters are fully implemented](https://github.com/dgkf/lintr/blob/d8e81895f4db2fe14acb793586c6f1e46b7ecccf/R/indentation_linter.R#L208-L226) if you're curious to see what that ends up looking like. 

In short, I need to offer a very deep apology for not sharing this work earlier. I had assumed it was up already. I think there are some interesting ideas in here, but the scope is very different. I think you have a more tangible pass at indentation here. If you're interested in just taking a peek at some examples, I wrote [pretty comprehensive tests](https://github.com/dgkf/lintr/blob/dgkf/indentation_linter_refactor/tests/testthat/test-indentation_linter.R) for the ones that were fully implemented or works-in-progress.

Just as a parallel, because I thought this idea of defining linter rules as functions was so cool, I wrote up an entirely [separate experimental linter, {reflow}](https://github.com/dgkf/reflow) that is effectively a syntax-aware regex-like DSL, using R's conditions to bubble up failed lint messages when the parsing fails. It ended up being horrendously slow, so I didn't pursue it further.

IndrajeetPatil:
@AshesITR This is increasingly diverging from the `main` branch with quite a few merge conflicts already. 
Maybe time to catch up with the `main`?

AshesITR:
Feel free to chip in as I don't have much time to contribute to lintr at the moment.

MichaelChirico:
Bumping to https://github.com/r-lib/lintr/milestone/3 following the most recent comment

AshesITR:
I've resolved all merge conflicts so work on this branch can continue.
Note to self: Don't de-lint lintr for new linters until the PR is merged.

AshesITR:
@MichaelChirico PTAL. The CI failure is because `indentation_linter()` applies to us as well.
If the review is positive, I will see to fixing the lints.
In an effort to prevent future merge conflicts if the review takes longer, I won't cause any trivial changes by de-linting at this time.

NEWS.md
```diff
@@ -88,6 +90,8 @@
 
 * `routine_registration_linter()` for identifying native routines that don't use registration (`useDynLib` in the `NAMESPACE`; @MichaelChirico)
 
+*  `indentation_linter()`: see above in 'Changes to defaults'
```
MichaelChirico, line 93: I figure it's good to add this under "new linters" in case there are any readers just focus on that when scanning the NEWS, WDYT?

R/expect_lint.R
```diff
@@ -94,9 +94,8 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
           }
           check <- check[[field]]
           value <- lint[[field]]
-          msg <- sprintf(
-            "check #%d: %s %s did not match %s",
-            itr, field, deparse(value), deparse(check)
+          msg <- sprintf("check #%d: %s %s did not match %s",
```
MichaelChirico, line 97: revert this?

R/extract.R
```diff
@@ -132,15 +132,8 @@ replace_prefix <- function(lines, prefix_pattern) {
...
-  blanks <- function(n) {
-    vapply(Map(rep.int, rep.int(" ", length(n)), n, USE.NAMES = FALSE),
-      paste, "",
-      collapse = ""
-    )
-  }
-
-  regmatches(lines[non_na], m[non_na]) <-
-    Map(blanks, lapply(regmatches(lines[non_na], m[non_na]), nchar))
+  prefix_lengths <- lapply(regmatches(lines[non_na], m[non_na]), nchar)
```
MichaelChirico, line 135: looks OK, is it relevant to this PR?

MichaelChirico: oh, is this fixing an existing indentation_linter violation?

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
...
+#'       additional_arg = 42
+#'   )
+#'
+#'   # always complies
+#'   map(x, f,
+#'       additional_arg = 42)
+#'   ```
+#'
+#' @evalRd rd_tags("indentation_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
```
MichaelChirico, line 24: link to the tidyverse guide section?

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
...
+#'   map(x, f,
+#'       additional_arg = 42
+#'   )
+#'
+#'   # always complies
+#'   map(x, f,
+#'       additional_arg = 42)
+#'   ```
+#'
+#' @evalRd rd_tags("indentation_linter")
```
MichaelChirico, line 23: add some `@examples`, esp demonstrating the parameters

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
...
+#'   map(x, f,
+#'     additional_arg = 42
+#'   )
+#'
+#'   # complies only with use_hybrid_indent = FALSE:
+#'   map(x, f,
+#'       additional_arg = 42
+#'   )
+#'
+#'   # always complies
```
MichaelChirico, line 18: causes me some "alarm" that {styler} will edit this style:

```r
styler::style_text("
map(x, f,
    additional_arg = 42)
")

# map(x, f,
#   additional_arg = 42
# )
```

Is totally addressing this just outside the scope of this linter? or should the defaults for this linter move the user towards the {styler} output?

The style guide section is not _quite_ conclusive here, but collating a few BAD examples I think we can conclude {styler} is doing the "right" thing per style guide:

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
...
+#'   map(x, f,
+#'       additional_arg = 42)
+#'   ```
+#'
+#' @evalRd rd_tags("indentation_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+indentation_linter <- function(indent = 2L, use_hybrid_indent = TRUE) {
+  paren_tokens <- c("OP-LEFT-BRACE", "OP-LEFT-PAREN", "OP-LEFT-BRACKET", "LBB")
```
MichaelChirico, line 27: nit: call it `paren_tokens_left` for balance?

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
...
+                    ]/following-sibling::*[1]/@line2")
+      ),
+      collapse = " | "
+    ),
+    ")"
+  )
+
+  xp_indent_changes <- paste(
+    c(
+      glue::glue("//{paren_tokens}[not(@line1 = following-sibling::expr[
```
MichaelChirico, line 68: with all of these XPaths, I do wonder about performance, could you do a quick check on some large files vs. some other default linters to see if that concern is valid?

My first thought is: would it be possible to do a quick sanity check on the input file, like `gregexpr("^\\s+", lines)` and if all of the matches are multiples of 2, we're done already? I guess not, since something like this would pass

```r
x =
    too_far_out()
```

Maybe if everything's a multple of 2 and `diff(indent)` is never more than 2?

R/indentation_linter.R
```diff
@@ -0,0 +1,163 @@
+#' Check that indentation is consistent
+#'
+#' @param indent Block indentation level, used for multi-line code blocks (`{ ... }`), function calls (`( ... )`) and
```
MichaelChirico, line 3: I think it's not clear what this means from this description, maybe at least mention the default is `2` following the tidyverse recommendation?

R/indentation_linter.R
MichaelChirico: Maybe we need `function_right_parentheses_linter()`, or just combine into a single `function_parentheses_linter()`?

MichaelChirico (commented):
I notice there's a lot of "false positives" where indentation is used to align across lines (e.g. tests/testthat/test-implicit_integer_linter.R)... it would be great if this linter could recognize that and not throw a lint. Probably best to leave it as a follow-up...

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,423 @@
...
+    trim_some("
+      list(
+           1,
+           2)
+    "),
+    NULL,
+    linter
+  )
+
+  # comments do not trigger hanging indent rule
```
MichaelChirico, line 55: i don't see any cases when the comment is mis-indented, please add tests for this e.g.

```r
list(
 # comment at 1 space
  key = "correct spaced code"
)
```

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,423 @@
...
+      function(a = 1L,
+               b = 2L) {
+        a + b
+      }
+    "),
+    NULL,
+    linter
+  )
+
+  expect_lint(
```
MichaelChirico, line 140: this is just the previous example, plus assignment, right? could you add a quick comment to the test?

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,423 @@
...
+      a %>%
+        b()
+    "),
+    NULL,
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      (a + b + c) /
```
MichaelChirico, line 189: i think there's some ambiguity on whether something like

```r
y ~
  x1 + 
    x2
```

should lint... on the one hand, it's like `~` and `+` independently induce an indent. on the other, all operators introduce one max indent.

FWIW, {styler} aligns with the choice here:

```r
styler::style_text("
y ~
  x1 +
    x2
")

# y ~
#   x1 +
#   x2
```

but maybe that's an option to explore? (probably best for follow-up)

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,423 @@
...
+    "Indentation",
+    linter
+  )
+})
+
+test_that("indentation within string constants is ignored", {
+  expect_lint(
+    trim_some("
+      x <- '
+        an indented string
```
MichaelChirico, line 300: better to also include some lines that are indented the "wrong" amount (e.g. 1 or 3 spaces)

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,423 @@
...
+  expect_lint(code_hybrid, "Hanging indent", indentation_linter(use_hybrid_indent = FALSE))
+
+  expect_lint(code_non_hybrid, "Indent", indentation_linter())
+  expect_lint(code_non_hybrid, NULL, indentation_linter(use_hybrid_indent = FALSE))
+
+  expect_lint(code_always_ok, NULL, indentation_linter())
+  expect_lint(code_always_ok, NULL, indentation_linter(use_hybrid_indent = FALSE))
+})
+
+test_that("consecutive same-level lints are suppressed", {
```
MichaelChirico, line 391: I think it's a good feature, but also not something we do for other linters (e.g. `no_tab_linter()` is the first that comes to mind as a linter that tends to hit on many lines in a row).

Do you think it's worth maintaining the code to remove consecutive lints? Should we consider adding that logic to the other linters too? Maybe it's an argument we could add to `Lint()` or `flatten_lints()` or something like that?

MichaelChirico:
Impressive work here! Thanks so much.

A lot of my comments can probably be punted to future CLs, but OTOH, this will be a new default, so we should balance more in favor of fixing things up front. Just something I had in mind when making more nit-picky reviews.

NEWS.md
AshesITR: Makes more sense - good idea. (I.e. " New `indentation_linter()`" is part of the default linters under "Changes to defaults" and the rest under "New linters").

R/indentation_linter.R
AshesITR: Hmm I think we should at least add a parameter to allow suppressing this lint, because it will be produced by RStudios and PyCharms auto-indent.
Alternatively, we control the expected indentation of function calls based on whether the closing parenthesis has arguments on its line? That would let both styles, hanging indent and ggplot-like indent, pass.
I think the part of the style guide that says function arguments on multiline function calls need to be unnamed if on the first line and named if on subsequent lines should be put into its own linter, though.

R/extract.R
AshesITR: IIRC I didn't understand the original code very well and it's pretty terrible in terms of performance.
Also, not sure if indenting after `<-` should be encouraged. The old code would be more readable if `Map(` were on the preceding line.
It's such a small change that I wouldn't make a separate PR for this. LMK if I should revert.

``` r
blanks <- function(n) {
  vapply(Map(rep.int, rep.int(" ", length(n)), n, USE.NAMES = FALSE),
    paste, "",
    collapse = ""
  )
}
n <- rpois(1000, 10)
microbenchmark::microbenchmark(lapply(n, blanks), lapply(n, strrep, x = " "))
#> Unit: microseconds
#>                        expr      min        lq     mean   median        uq
#>           lapply(n, blanks) 7607.310 8014.3745 8610.423 8216.366 9332.1900
#>  lapply(n, strrep, x = " ")  578.054  607.4585  664.728  628.980  644.0755
#>        max neval
#>  11738.313   100
#>   1922.868   100
```

<sup>Created on 2022-10-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/indentation_linter.R
AshesITR: Let's worry about performance when the functionality is final.
I'm not too worried because most of them don't use `*` or `expr` for entry.
Also, I didn't experience noticable performance problems during my testing of CRAN packages.

R/indentation_linter.R
AshesITR: Oh wait, I think that's what I did with `use_hybrid_indent = TRUE` 😆

R/indentation_linter.R
AshesITR: Note for future: If we run into trouble with performance (looking at tools?), the `self::*` could be transformed as to never select `expr` nodes, at the cost of exploding the XPath with all other allowed XML tags.

AshesITR:
> I notice there's a lot of "false positives" where indentation is used to align across lines (e.g. tests/testthat/test-implicit_integer_linter.R)... it would be great if this linter could recognize that and not throw a lint. Probably best to leave it as a follow-up...

The way to satisfy indentation linter and still keep everything aligned is to left-align the columns.
Maybe suitable for follow-up but hard to discern from actual wrong indentation.
These situations seem suited for `# nolint start: indentation`.

tests/testthat/test-indentation_linter.R
AshesITR: This seems like a peculiarity in the way `~` is parsed. Good catch, but should be a follow up I think.
See
``` r
lintr::lint(text = "y ~\n  x1 +\n  x2", linter = lintr::indentation_linter())
#> <text>:3:2: style: [indentation_linter] Indentation should be 4 spaces but is 2 spaces.
#>   x2
#>  ^~~
lintr::lint(text = "y +\n  x1 +\n  x2", linter = lintr::indentation_linter())
```

<sup>Created on 2022-10-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/test-indentation_linter.R
AshesITR: I think this is highly context dependent whether consecutive lints need to be "collapsed" into one. This linter tends to produce massive amounts of lints if e.g. a top-level indentation is wrong, see the tables with / without duplicates on the tested packages. But for `no_tab_linter()` I agree it could also be implemented such that it only lints the first line per file that it finds. OTOH if we ever add auto-fixing, this might be detrimental.

AshesITR: Filed #1717

tests/testthat/test-indentation_linter.R
AshesITR: Filed #1718

R/indentation_linter.R
AshesITR: WDYT about linting

```
map(x, f,
    additional_arg = 42)
```

With a multiline function call linter that also checks that each argument on the subsequent lines is named and the arguments on the same line as the function call are unnamed?
Could be flagged as "Closing parenteses of multiline function calls should always be on their own line".

AshesITR: #1719

AshesITR:
Here is a table of the most recent lint counts

```
      package    N
 1:        sf 4705 -- indent = 1
 2:       rgl 1635 -- indent = 4
 3:     rgeos 1146 -- indent = 4
 4:   stringi 1023 -- indent = 4
 5:   ggplot2  536 -- see below
 6:       cli  272 -- see below
 7:     vctrs  257 -- tab indents, see below
 8:     dplyr  152 -- looks okay
 9:   stringr  132 -- see below
10:      glue  101 -- see below
11:  processx   93 -- see below
12:     rlang   90 -- see below
13:  magrittr   75 -- see below
14:    tibble   35 -- see below
15:  devtools   21 -- see below
16:      ragg    8 -- see below
17:    pillar    2 -- see below
18: lifecycle    1 -- looks okay
```

ggplot2, processx, cli, stringr and glue feature (some) multi line function calls of the form
```r
fun(a, b,
  c, d)
```
Those should, IMO, be linted and converted to
```r
fun(a, b,
  c, d
)
```
However, seeing the sheer number, maybe an option is in order to enable that.

cli, rlang and devtools have a few interesting spots where a large `local({})` block isn't indented.
Those lints seem fine to me

tibble has a few lints in tests where a multi line string describes the test:
```r
test_that("multi line
           test name", {
  test_code
})
```
These seem like a false positive to me, i.e. shouldn't lint.

magrittr has a bunch of lints due to indented R code blocks (where the entire block is indented and so is the block start pattern), such as [here](https://github.com/tidyverse/magrittr/blob/main/vignettes/tradeoffs.Rmd).
These are false positives but IMO should be treated by the code extractor in a similar fashion to #1043 

Regarding vctrs use of tab indents I think we should implicitly assume that `"\t"` is `indent` wide instead of 1.

ragg features conjunct `if` clauses
```r
if (a &&
    b) {
  code
}
```

pillar has two lints related to assignments:
```r
basic <-
  ggplot2::ggplot(data, ggplot2::aes(x, y)) +
  ggplot2::geom_point() +
  ggplot2::theme_minimal(36)
```
Which, similar to the `y ~ x1 + x2` example from #1718, cause the current logic to expect a second indent after the first `+`, i.e.
```r
basic <-
  ggplot2::ggplot(data, ggplot2::aes(x, y)) +
    ggplot2::geom_point() +
    ggplot2::theme_minimal(36)
```

AshesITR:
After giving this some more thought:
Handling mixtures of tabs and spaces for indenting is very ugly when trying to obtain correct lint locations, so I'd punt on the vctrs specialty.
What I'd want to solve in this PR is give an option that allows for the style seen in ggplot2.

I think the most elegant way is to refactor `use_hybrid_indent=` to a character parameter that allows for three styles regarding "hanging" function calls:

```r
# 1. always use hanging indent -- as produced by RStudio and PyCharm auto-indent
fun(a,
    b
)

fun(a,
    b)

# 2. (default) tidy style -- hanging if closing parenthesis is not on its own line
fun(a,
  b
)

fun(a,
    b)

# 3. never use hanging indent -- seen in ggplot2 IINM
fun(a,
  b
)

fun(a,
  b)
```

@MichaelChirico, @IndrajeetPatil  do you have naming suggestions? I'm having trouble finding a concise but understandable argument name. Best I can come up with is `use_hanging_indent = c("tidy", "always", "never")`, but I'm not particularly sold.

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,445 @@
+test_that("indentation linter flags unindented expressions", {
```
MichaelChirico, line 1: another test I'd like to see, make sure that multiple consecutive "indenting" operators only outdent by one level, e.g.

```r
local({
  some_code
})
```

is OK, and the logic isn't thinking the indentation should be

```r
local({
    some_code
})
```

(I think this is already handled / would have been picked up from your spot checks on other packages, but it'll be good to keep as a regression test if we ever refactor)

R/indentation_linter.R
MichaelChirico: SG. Also for future reference, per the frequency table here, I guess the biggest concern would be `//OP-LEFT-PAREN`:

https://github.com/r-lib/lintr/issues/1358#issue-1260855389

MichaelChirico:
> Maybe suitable for follow-up but hard to discern from actual wrong indentation. These situations seem suited for # nolint start: indentation.

Agree it's tricky to always get right, but {styler} does have some logic that helps. Maybe the way to approach it is piecemeal -- if we can think of a good rule for detecting "false positives" in a certain case, add it, gradually improving.

MichaelChirico:
> These [testthat with multiline test description] seem like a false positive to me, i.e. shouldn't lint.

Agreed

MichaelChirico:
> These [space-designated code blocks in magrittr] are false positives but IMO should be treated by the code extractor in a similar fashion to https://github.com/r-lib/lintr/issues/1043

Agreed, seems like a more general issue, no need to fix here I'd say

MichaelChirico:
> Regarding vctrs use of tab indents I think we should implicitly assume that "\t" is indent wide instead of 1.

Agreed, the linter should be useful even for packages using tabs-not-spaces, and `\t`==`indent` seems the easiest way to reconcile this without needing options.

But as it's not part of the style guide, happy to defer on that to a future PR.

MichaelChirico:
> {styler} butchers this (IMO):

Also agree. Are you using CRAN or dev {styler}? I know there's been a lot of improvements in dev. If dev, maybe worth an issue? IMO that should be edited to:

```r
if (
  a &&
    b
) {
}
```

OTOH, I think {styler}'s current choice does stick to the letter of the guide (except maybe the trailing `)`?)... not sure if it's easy to fix automatically

MichaelChirico:
> do you have naming suggestions? 

I like the idea of offering a few options. `use_hanging_indent` implies a boolean to me though, so how about `hanging_indent_style = c(...)`?

The 3 option names you gave sound good to me. Unless there's some universal reference to apply here, that's about as good as we'll be able to muster. Definitely a place for "See examples" in the docs & we use the examples you gave.

R/indentation_linter.R
```diff
@@ -50,11 +81,11 @@ indentation_linter <- function(indent = 2L, use_hybrid_indent = TRUE) {
     "number(",
     paste(
       c(
-        glue::glue("self::{paren_tokens}/following-sibling::{paren_tokens_right}/preceding-sibling::*[1]/@line2"),
-        glue::glue("self::*[{xp_and(paste0('not(self::', paren_tokens, ')'))}]/following-sibling::SYMBOL_FUNCTION_CALL/
+        glue::glue("self::{paren_tokens_left}/following-sibling::{paren_tokens_right}/preceding-sibling::*[1]/@line2"),
+        glue::glue("self::*[{xp_and(paste0('not(self::', paren_tokens_left, ')'))}]/following-sibling::SYMBOL_FUNCTION_CALL/
```
MichaelChirico, line 85: too wide now :)

maybe an opportunity to switch to the style I sort of settled on elsewhere, where `/` paths become hanging indents if they're too wide:

```
self::*[...]
  /following-sibling::SYMBOL_FUNCTION_CALL
  /parent::expr
  /following-sibling::expr[1]
  /@line2
```

tests/testthat/test-indentation_linter.R
```diff
@@ -0,0 +1,445 @@
...
+  linter <- indentation_linter(indent = 2L)
```
MichaelChirico, line 2: One more request for extra tests -- currently all the tests just try to match `"Indentation"` in the message, but we should have some tests that ensure the customization is correct

.dev/compare_branches.R
```diff
@@ -344,6 +344,9 @@ get_linter_from_name <- function(linter_name) {
     #   did not make it into the release?
     if (linter_name == "line_length_linter" && !is.integer(formals(linter_name)$length)) {
       eval(call(linter_name, 80L))
+    } else if (endsWith(linter_name, ")")) {
```
MichaelChirico, line 347: could you file this as its own PR & I can approve right away?

AshesITR:
Okay, I've spotted a problem with the only remaining false-positive:

```r
test_that("multi-line
           test describtion", {
})

hanging_call(looking,
             basically,
             identical = function() {
                but_with_a_need_for_hanging_indent
             })
```

Therefor, I think this is ready to merge now.
I've run multiple configurations of the linter against the testing packages to see if the adaptations turn out okay.
The result is the packages are surprisingly inconsistent in their indentation style 😆 

Here is a tally of lints using default settings:
```
      package    N
 1:        sf 4710 -- 1416 with indent = 1
 2:       rgl 1646 -- 6584 with indent = 4
 3:     rgeos 1227 -- 428 with indent = 4
 4:   stringi 1023 -- 112 with indent = 4
 5:   ggplot2  514 -- 543 with hanging_indent_style = "never"
 6:       cli  271
 7:     vctrs  208
 8:     dplyr  126
 9:      glue  101
10:  processx   92
11:     rlang   89
12:   stringr   74
13:  magrittr   59
14:    tibble   33
15:  devtools   21
16:      ragg    8
17:    pillar    2
18: lifecycle    1
```

To get a gist of the consistency within "problematic" packages, I intersected the default lints with the customized parametrization to find only completely broken indentations.
Here are these counts where neither the default settings nor the style resulting in most of the default lints are valid indentations:
```
   package    N
1:     rgl 1318 -- some valid lints, some should disappear with hanging_indent_style = "always"
2:      sf 1005 -- extremely inconsistent indents (sometimes different levels in the same file, tabs and spaces mixed)
3:   rgeos  292 -- I see mostly commented code where the comments aren't indented properly
4: ggplot2  212 -- inconsistent indents; especially R/theme-defaults.r has no consistent indentation rules.
5: stringi  112 -- most should disappear with hanging_indent_style = "always"
```

@MichaelChirico are you okay with the propsal? Then I'll see to fixing our lints and get this merged.

.dev/compare_branches.R
MichaelChirico: No :)

MichaelChirico:
LGTM. I think you are the master of indentation now. You'll need a cross & garlic to read r-devel source code from now on 😄

MichaelChirico:
> Okay, I've spotted a problem with the only remaining false-positive:

BTW, QQ here to check my understanding, the difference between the testthat example:

```r
test_that("multi-line
           test describtion", {
})
```

and the comparison example:

```r
hanging_call(looking,
             basically,
             identical = function() {
                but_with_a_need_for_hanging_indent
             })
```

is that `identical=` is a new argument on its own line, whereas `{` is coming following the previous argument. So I guess it's actually closer to the ggplot2 examples of indent styles, so would this be a closer comparison?

```r
hanging_call(looking,
             basically, function() {
                but_with_a_need_for_hanging_indent
             })
```

MichaelChirico:
Great work here @AshesITR !! thanks for the final push to get it over the line.

Original issue = 6.5 years ago
Original implementation = 2.5 years ago
Started this PR = 4 months ago & 36 commits

🚀 🚀 🚀

PR #1414, opened 2022-06-19
adds workaround to using .lintr files with vim ALE

andyquinterom:
This allows me to use .lintr files for configuration using vim ALE


AshesITR (approved):
Thanks for helping to improve the documentation!

PR #1419, opened 2022-06-19
lint commented argument code as well

### LINKED ISSUE #386: flagging commented arguments of a function being tested
If a line of code is commented, then lintr produces a lint of type 
> **Commented code should be removed.**

Example of this would be the following line of code-
```
# testthat::expect_equal(broom_df1$p.value, tidy_df1$p.value, tolerance = 0.001)
```

Does it also make sense to produce a lint of type
> **Commented arguments should be removed.**

For example, here the second argument is commented out-
```
 d <- t.test(
        x = dplyr::starwars$height,
       # var.equal = TRUE,
        conf.level = .99,
        mu = 175
      )

testthat::expect_equal(d$statistic[[1]], -0.1661693, tolerance = 0.001)
```
This can typically happen when interactively writing tests and it will be nice for `lintr` to highlight such vestiges.

AshesITR:
fixes #386


NEWS.md
```diff
@@ -5,6 +5,10 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table)  (#1396, @IndrajeetPatil).
 
+## New features
+
+* `commented_code_linter()` now lints commented argument code as well (#386, @AshesITR).
```
MichaelChirico, line 10: I don't find the description here very clear -- could you revisit? maybe a quick example is warranted?

MichaelChirico:
PTAL at the new lints surfaced

R/comment_linters.R
```diff
@@ -54,7 +53,10 @@ commented_code_linter <- function() {
     all_comment_nodes <- xml2::xml_find_all(source_expression$full_xml_parsed_content, "//COMMENT")
     all_comments <- xml2::xml_text(all_comment_nodes)
     code_candidates <- re_matches(all_comments, code_candidate_regex, global = FALSE, locations = TRUE)
-    is_parsable <- which(vapply(code_candidates[, "code"], parsable, logical(1L)))
+    extracted_code <- code_candidates[, "code"]
+    # ignore trailing ',' when testing for parsability
+    extracted_code <- rex::re_substitutes(extracted_code, rex::rex(",", any_spaces, end), "")
```
MichaelChirico, line 58: include trailing comments too?

```r
# na.rm = TRUE, # ignore missing entries
```

R/comment_linters.R
```diff
@@ -54,7 +53,10 @@ commented_code_linter <- function() {
     all_comment_nodes <- xml2::xml_find_all(source_expression$full_xml_parsed_content, "//COMMENT")
     all_comments <- xml2::xml_text(all_comment_nodes)
     code_candidates <- re_matches(all_comments, code_candidate_regex, global = FALSE, locations = TRUE)
-    is_parsable <- which(vapply(code_candidates[, "code"], parsable, logical(1L)))
+    extracted_code <- code_candidates[, "code"]
+    # ignore trailing ',' when testing for parsability
```
MichaelChirico, line 57: what about leading comment style?

```r
foo(x
  , key = val
  # , bar = baz
)
```

NEWS.md
```diff
@@ -5,6 +5,11 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
+## New features
+
+* `commented_code_linter()` now lints commented argument code, containing a trailing comma, as well (#386, @AshesITR).
```
MichaelChirico, line 10: Understand better now. I guess we could also include a test that this extension applies regardless of whether the comment occurs inside a function call/definition, but it also smells like encoding false-positive-ish behavior, WDYT?

R/comment_linters.R
AshesITR: That's not so easy to properly detect without introducing false negatives I think: `x <- "this#should lint"`.

Or am I missing something?

NEWS.md
AshesITR: What do you mean? The parse tree is oblivious to where the comment occurs.

R/comment_linters.R
MichaelChirico: Hmm I had in mind `,\s*(#.*)?$`, but you're right it technically wouldn't lint on `# "na.rm = TRUE, # comment in string"`. Seems a bit obscure, but we can consider that later if a user requests.

NEWS.md
MichaelChirico: Right -- but I believe the intent of #386 was to only cover such comments inside a function call/definition. e.g. now we'll get false positives like:

```
# check the following for further details:
#   short-link/helpful-doc-1,
#   short-link/helpful-doc-2.
```

(IINM line 2 will newly lint, line 3 linted before)

AshesITR: OTOH if the entire function call was block-commented, we'd have a bunch of false negatives.

I think we'd be better off by tightening the logic to require the commented code to potentially cause side effects, i.e. contain a function call or an assignment.

MichaelChirico: > I think we'd be better off by tightening the logic to require the commented code to potentially cause side effects, i.e. contain a function call or an assignment.

I think I wasn't sure if you planned to attempt that in this PR. Maybe leave it as a TODO otherwise?

AshesITR: I'd leave this for a later PR, there are already issues so not sure we need to add a TODO to the code.

MichaelChirico: SGTM; resolved

MichaelChirico (approved):

PR #1420, opened 2022-06-19
add function_argument_linter()

### LINKED ISSUE #450: Add a linter to check that arguments after an argument with default value have a default value as well
It seems good practice to avoid arguments without a default value if a preceding argument already has a default value:

    # Bad
    function(x, y = 1, z) {
      # some code
    }

    # Good
    function(x, y = 1, z = 0) {
      # some code
    }

    # Good as well
    function(x, z, y = 1) {
      # some code
    }

AshesITR:
Fixes #450.

Name chosen so that we can extend this linter to perform other checks on function argument lists in the future.
e.g. `tidyverse` usually has defaults for arguments that appear after `...`


NEWS.md
```diff
@@ -20,6 +20,9 @@
 
 * `unreachable_code_linter()` ignores trailing comments if they match a closing nolint block (#1347, @AshesITR).
 
+* New `function_argument_linter()` to enforce that arguments with defaults appear last in function declarations
```
MichaelChirico, line 23: link the design guide here?

R/function_argument_linter.R
```diff
@@ -0,0 +1,29 @@
+#' Function argument linter
+#'
+#' Check that arguments with defaults come last in all function declarations.
```
MichaelChirico, line 3: Ditto -- external link to the design guide

R/function_argument_linter.R
```diff
@@ -0,0 +1,29 @@
...
+#'
+#' @evalRd rd_tags("function_argument_linter")
+#' @seealso
+#'   [linters] for a complete list of linters available in lintr.
+#' @export
+function_argument_linter <- function() {
+  xpath <- "//FUNCTION/following-sibling::EQ_FORMALS[1]/
+              following-sibling::SYMBOL_FORMALS[not(following-sibling::*[1][self::EQ_FORMALS])]"
```
MichaelChirico, line 11: would a comment interfere with this `following-sibling::*[1]`?

tests/testthat/test-function_argument_linter.R
```diff
@@ -0,0 +1,9 @@
+test_that("function_argument_linter returns the correct linting", {
+  linter <- function_argument_linter()
+  msg <- rex::rex("Arguments without defaults should come before arguments with defaults.")
+
+  expect_lint("function(x, y = 1, z) {}", list(message = msg, column_number = 20L), linter)
```
MichaelChirico, line 5: what happens with `...`? Add some tests & ensure `...` is handled correctly

tests/testthat/test-function_argument_linter.R
```diff
@@ -0,0 +1,9 @@
+test_that("function_argument_linter returns the correct linting", {
```
MichaelChirico, line 1: add some tests of multi-line signatures. XPath means we _should_ be fine, but no harm to include.

the same logic applies to lambda functions too, right? I think typically people are not defining lambdas with defaults, but I think syntactically it's OK (don't have R 4.1 in front of me ATM) to write `\(x, y = 2, z)`. Are those in scope?

MichaelChirico (commented):
Would you mind running this on all (a bunch of?) the tidyverse/r-lib packages? I have a feeling we may be missing some common case, or that it's violated a lot in practice. Any special care needed for S3 methods?

R/function_argument_linter.R
AshesITR: That would have to be a comment between the argument name and its equals sign. Is that even allowed? Never saw it in practice.

MichaelChirico: I had this in mind:

```
function(
  really_long_argument_name = # TODO: deprecate this
    1
)
```

but = dropping to the next line should still parse

AshesITR: You example would be safe. Bit it doesn't hurt to add the `not(self::COMMENT)` condition I think.

Note I don't have a lot of free time to work on lintr at the moment so it may take a while till I get to this.

MichaelChirico: > Note I don't have a lot of free time to work on lintr at the moment so it may take a while till I get to this.

No problem at all. I also need to pull back from `lintr` relatively -- definitely over-committed in the push to get 3.0.0 out :)

tests/testthat/test-function_argument_linter.R
AshesITR: Added them. Just requires `//OP-LAMBDA` in addition to `//FUNCTION`.

tests/testthat/test-function_argument_linter.R
AshesITR: Added code and tests to ignore `...`, i.e. allow `function(a, b = 2, ...) {}`, but not `function(a, b = 2, ..., c)`.

R/function_argument_linter.R
AshesITR: done.

R/function_argument_linter.R
AshesITR: done.

NEWS.md
AshesITR: done.

R/function_argument_linter.R
```diff
@@ -1,14 +1,17 @@
...
-#'   [linters] for a complete list of linters available in lintr.
+#'   [linters] for a complete list of linters available in lintr. \cr
+#'   <https://design.tidyverse.org/args-data-details.html>
 #' @export
 function_argument_linter <- function() {
-  xpath <- "//FUNCTION/following-sibling::EQ_FORMALS[1]/
-              following-sibling::SYMBOL_FORMALS[not(following-sibling::*[1][self::EQ_FORMALS])]"
+  xpath <- paste(glue::glue(
+    "//{c('FUNCTION', 'OP-LAMBDA')}/following-sibling::EQ_FORMALS[1]/
+              following-sibling::SYMBOL_FORMALS[not(following-sibling::*[not(self::COMMENT)][1][self::EQ_FORMALS])]"
```
MichaelChirico: nit: feels like a lot of whitespace here, no?

MichaelChirico (approved):

MichaelChirico (approved):

PR #1421, opened 2022-06-19
object_name_linter: add `regexes=` argument for custom style regexes

### LINKED ISSUE #822: Adding shiny module names to custom whitelist
I am often writing Shiny modules and following the suggested naming convention for a module called e. g. _launcher_ I would name the modules `launcherUI()` and `launcherServer()` according to [Mastering Shiny](https://mastering-shiny.org/scaling-modules.html#naming-conventions).

This of course triggers the `object_name_linter` and suggests "snake_case". This is totally fine for the rest of my package as I want to stick to snake case convention for all other functions.

Is there a way to whitelist specific function names during linting based on their name? Currently I use `# nolint` directly in the code, but I would prefer to have a configuration option where I could specify function names that are allowed. (like in the example above)

AshesITR:
Fixes #822


R/object_name_linters.R
```diff
@@ -44,15 +44,36 @@ object_name_xpath <- local({
...
+#'   If named, the names will be used in the lint message. Otherwise, "custom" will be used as a name for the style.
+#'   Quotes (`` `"' ``) and specials (`%` and trailing `<-`) are not considered part of the name to be matched.
 #' @evalRd rd_tags("object_name_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-object_name_linter <- function(styles = c("snake_case", "symbols")) {
-  styles <- match.arg(styles, names(style_regexes), several.ok = TRUE)
+object_name_linter <- function(styles = c("snake_case", "symbols"), regexes = character()) {
+  if (length(styles) > 0L) {
+    # Allow `object_name_linter(NULL, "my_regex")`
```
MichaelChirico, line 55: not a big fan of `(NULL, "my_regex")` as a signature... can't we allow `object_name_linter(regex = "my_regex")` and use `missing()` on the arguments instead?

R/object_name_linters.R
```diff
@@ -44,15 +44,36 @@ object_name_xpath <- local({
...
+    style_list <- list()
+  }
+  if (length(regexes) > 0L) {
+    if (!is.character(regexes)) {
+      stop("`regexes` must be a character vector.")
+    }
+    if (is.null(names(regexes))) {
+      names(regexes) <- "custom"
+    }
+    style_list <- c(style_list, as.list(regexes))
```
MichaelChirico, line 68: what happens if user provides an exact replica of one of the existing `style_regexes`? do we need `unique()`?

tests/testthat/test-object_name_linter.R
```diff
@@ -151,3 +150,67 @@ test_that("object_name_linter won't fail if an imported namespace is unavailable
     3L
   )
 })
+
+test_that("object_name_linter supports custom regexes", {
```
MichaelChirico, line 154: don't have any tests of the default-named `"custom"` case. I'm particularly interested how it looks when there's >1 default-named regex to check

R/object_name_linters.R
AshesITR: I thought about it but it seemed confusing to me because we still want `object_name_linter()` to use the default `styles`, so behavior of `missing(styles)` depends on `missing(regexes)`.

R/object_name_linters.R
AshesITR: Not really. There is no harm with a name matching multiple styles.

MichaelChirico: What does the message look like for matching multiple "custom" styles?

R/object_name_linters.R
MichaelChirico: >  behavior of missing(styles) depends on missing(regexes).

I don't see a big problem with that...

Both look pretty clear to me in usage:

```r
object_name_linter() # use default 'styles'
object_name_linter(regex = "regex") # use regex, not 'styles'
```

R/object_name_linters.R
AshesITR: See L76, duplicates are removed in the message, so the message will read "Variable and function name style should be custom."

MichaelChirico: Got it. Would still like to see some tests around this to ensure that's what's happening / continues to happen.

R/object_name_linters.R
AshesITR: Okay.

tests/testthat/test-object_name_linter.R
```diff
@@ -203,14 +210,28 @@ test_that("object_name_linter supports custom regexes", {
...
   )
+
+  # Default regex naming works
+  expect_lint(
+    trim_some("
+      a <- 42L
+      b <- 1L
+      c <- 2L
+    "),
+    list(line_number = 3L, message = rex::rex("Variable and function name style should be custom.")),
```
MichaelChirico, line 224: WDYT about something like:

> Variable and function name style should match custom pattern(s) %s

I think it's likely many people who read the lint message aren't authors -- the current message will lead to a confusing experience.

e.g. contributors to a package on GH shouldn't need to know any details about lintr in order to get actionable feedback if their PR is blocked by lintr.

in our case, linting is happening in an even more decentralized fashion -- lints show up in the editor and you would have to do some digging to find out what 'custom' means.

AshesITR: Excellent idea. We can auto-name the regexes by themselves to achieve a more actionable message. "... should match snake_case or ^bla$" also works for me in the context of predefined styles.

MichaelChirico:
Hey @IndrajeetPatil minor feedback -- I would avoid doing "merge to main" updates for PRs so much (exception: handling merge conflicts), a few reasons:

(1) it makes it harder to tell which PRs actually have substantive updates. For example, this PR was highlighted in my `/pulls` page because of "new changes since I last viewed":

Selection_187

(2) it creates a lot of inbox noise (I get GitHub updates in e-mail), and because Gmail generally threads the updates for a single PR, it makes it harder when coming back after some absence to scroll through the thread and pick up the "merge main" mails from the mails with comments -- I'm worried I'll miss something important if it sneaks in amid several "merge main" messages

(3) it does muddy up the commit history -- even though we squash+merge to main, the PR history will still reflect many more commits than necessary, it can make future reference a bit harder

The PR reviewer can handle the merge just before submitting since it's just a simple button click. Merge conflicts are a bit tougher, since sometimes a merge resolution might affect how the PR is reviewed. But conflicts on the NEWS are probably best left until just before submission as well.

Thanks for your commitment to keeping everything clean & tidy!

IndrajeetPatil:
Thanks for the feedback, @MichaelChirico. 

Sure, I can stop updating branches. The main reason I was doing this was to avoid merge conflicts, which are already present in some PRs which have been pending for a while.

But, all the issues you have mentioned stemming from this practice are also valid, so, henceforth, I will let the contributor update their own PRs whenever they come back to working on them again.

MichaelChirico:
> Sure, I can stop updating branches. The main reason I was doing this was to avoid merge conflicts, which are already present in some PRs which have been pending for a while.

Yep, it makes sense. Indeed I used to do the same myself on `data.table`. But eventually, I picked up that if the PR will only be merge after needing to do 3 or 4 rounds of conflict resolution, it would be better to do all the resolution at once, when `main` is in as near a state as possible to how the code will actually be merged. Better to save your efforts :)

IndrajeetPatil:
@AshesITR Resolved a merge conflict to prevent this one from drifting too much away from the main.

AshesITR:
Sorry for the delayed reaction.
I'll see if I can invest some time today.
No worries if that's too late to the party for 3.0.2 😊

AshesITR:
@MichaelChirico PTAL what you think. Still needs a NEWS entry, but the rest is complete.
I also took the freedom to split object_name_linters.R into separate files for the two linters to improve on the consistency `R/*.R` -> `tests/testhat/test-*.R`

tests/testthat/test-object_name_linter.R
```diff
@@ -155,3 +155,88 @@ test_that("object_name_linter won't fail if an imported namespace is unavailable
...
+  )
+
+  # Default regex naming works
+  expect_lint(
+    trim_some("
+      a <- 42L
+      b <- 1L
+      c <- 2L
+    "),
+    list(line_number = 3L, message = rex::rex("Variable and function name style should match /^a$/ or /^b$/.")),
```
MichaelChirico, line 229: looks great! I do wonder if `/regex/` is widely understood, but that can easily be fixed later if needed.

R/object_name_linter.R
```diff
@@ -44,6 +44,11 @@ object_name_xpath <- local({
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
 #'   match at least one of these styles.
+#' @param regexes A (possibly named) character vector specifying a custom naming convention.
+#'   If named, the names will be used in the lint message. Otherwise, "custom" will be used as a name for the style.
+#'   Quotes (`` `"' ``) and specials (`%` and trailing `<-`) are not considered part of the name to be matched.
```
MichaelChirico, line 49: should this (Quotes...matched) be mentioned in the above details? it may fit more with the discussion of S3 patterns, WDYT?

R/object_name_linter.R
```diff
@@ -44,6 +44,11 @@ object_name_xpath <- local({
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
```
MichaelChirico, line 45: aside: could this be substituted with `@evalRd`? do I understand what this code is doing correctly? (neat trick btw, never seen it before...)

tests/testthat/test-object_name_linter.R
AshesITR: Yeah, I had the same question, but thought it was clearer than using any other form of quoting or not quoting it at all.

R/object_name_linter.R
AshesITR: Hmm sounds good. Also, the docs regading the auto-naming are out-of date 🤦🏻‍♂️

R/object_name_linter.R
AshesITR: IIRC `@evalRd` doesn't work "within" other roxygen tags. The trick goes back to Jim, though. Maybe he has more insight.

MichaelChirico: OK, let's punt for now

R/object_name_linter.R
```diff
@@ -143,17 +180,17 @@ check_style <- function(nms, style, generics = character()) {
       # If they are not conforming, but are S3 methods then ignore them
       conforming[!conforming][has_generic] <- TRUE
     }
-    # exclude namespace hooks like .onLoad, .Last.lib, etc (#500)
-    is_special <- is_special_function(nms[!conforming])
+    # exclude namespace hooks like .onLoad, .Last.lib, etc (#500) and ...
+    is_special <- is_special_function(nms[!conforming]) | nms[!conforming] == "..."
```
MichaelChirico, line 184: nice catch! is this covered by the `function(...) ..1 + ..2` test below?

PS I'm surprised `..1`/`..2` are excepted then?

AshesITR: `..1` and `..2` aren't checked, since they are only used, but neither assigned nor explicitly declared as formals.

AshesITR: And yes, I discovered the "bug" when writing the test you mention.

MichaelChirico: oh right, but `...` are "assigned" as `SYMBOL_SUB`. thanks!

MichaelChirico (approved):
Looks great, thanks!

AshesITR:
This still needs a NEWS. Do you want to write a bullet?

MichaelChirico:
whoops, got too excited :)

feel free to write one if you're still at it, I am logging off for the day

Issue #991, opened 2022-03-23
Produce SARIF reports ?


nvuillam:
V6 version of [MegaLinter](https://megalinter.github.io/) (currently in development) will support [SARIF](https://sarifweb.azurewebsites.net/) format, which is standard for SAST tools (cf https://megalinter.github.io/v6-alpha/supported-linters/ )

It would help a lot with the integration if your great R linter was able to provide SARIF output, is it something you would like to consider ?

Thanks :)

MichaelChirico:
The standard is ~300 pages long :)

https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.pdf

Happy to take PRs providing a conversion function. For now I'm not particularly keen to write anything myself.

nvuillam:
Nobody ever read it, just looking at the examples is enough to implement
the output JSON format:)

About the PR, unfortunately i don't have any experience in R :/

Le jeu. 24 mars 2022 à 00:08, Michael Chirico ***@***.***> a
écrit :

AshesITR:
This is an example JSON for a single (!) lint, for anyone interested in writing a PR for this. It's from Page 211 of the standard.

```json
  {
  "version": "2.1.0",
  "runs": [
{
"tool": {
        "driver": {
          "name": "CodeScanner",
          "rules": [
            {
              "id": "C2001",
              "fullDescription": {
"text": "A variable was used without being initialized. This can result in runtime errors such as null reference exceptions."
              },
              "messageStrings": {
"default": {
"text": "Variable \"{0}\" was used without being initialized."
 } }
} ]
} },
"artifacts": [
  {
"location": {
"uri": "src/collections/list.cpp", "uriBaseId": "SRCROOT"
},
    "sourceLanguage": "c"
  }
],
"results": [
  {
    "ruleId": "C2001",
    "ruleIndex": 0,
          "message": {
            "id": "default",
            "arguments": [
"count" ]
          },
          "locations": [
            {
              "physicalLocation": {
"artifactLocation": {
"uri": "src/collections/list.cpp", "uriBaseId": "SRCROOT",
"index": 0
                },
                "region": {
                  "startLine": 15
                }
              },
              "logicalLocations": [
{
"fullyQualifiedName": "collections::list::add"
} ]
} ]
} ]
} ]
}
```

nvuillam:
[...]

Issue #1188, opened 2022-05-22
Package logo


hadley:
Do y'all have any ideas for a package logo? I can help get it designed if you have some idea what it might look like

AshesITR:
Thanks for the offer! (see also #981)

I personally like the [SonarLint Logo](https://www.sonarlint.org/logos). Maybe that can serve as an inspiration?
The squiggly line is a good metaphor as it relates to spellchecking.

MichaelChirico:
copying notes from the dup'd issue:

@IndrajeetPatil:
> I love talking about `{lintr}`, but I always find it a tad bit dissatisfying to have slides with just the name of the package without a cool hex sticker to go along with it! 🙃
> 
> I wish I were creative enough to design and contribute one, but maybe this can be crowdsourced (on Twitter, e.g.)?
> 
> Thanks for the awesome work, either way! Looking forward to the `3.0.0` release.

@AshesITR:
> If created, this could also be featured here: https://github.com/GuangchuangYu/hexSticker

AshesITR:
@hadley Do you think this will make it in time for the 3.0.0 release?

hadley:
@AshesITR sorry for dropping the ball on this; probably not.

shaopeng-gh:
@hadley fyi I am working on starter workflow for lintr and it requires a icon/logo, looks like we can put it to use right away when it is ready :)
The required format I need to use is SVG, don't worry at all if your output is PNG I can use free converter to convert it.
(my input is if the logo is text type, the text/font should be clear/large proportion enough when the display size is small as icon, so that user can tell what text it is.)

IndrajeetPatil:
How about this one? Just a suggestion. 

![image](https://user-images.githubusercontent.com/11330453/183625044-a24a9a27-e6e5-4fc1-a166-d3f39bd14dfd.png)

hadley:
@IndrajeetPatil I like the basic idea, but I think it needs to be a dryer (not a washing machine) to convey "lint".

hadley:
Another approach would be do something like a wool jumper (e.g. https://www.shutterstock.com/image-vector/green-sweater-lint-balls-1550178998) with little red wigglies as lint?

AshesITR:
Another idea:

https://www.shutterstock.com/de/image-photo/freisteller-fliege-binder-78825286

Maybe a bow tie could be combined somehow with a red squiggle to indicate good style?

AshesITR:
[...]

Issue #981, opened 2022-03-22
Add `{lintr}` hex sticker?


IndrajeetPatil:
I love talking about `{lintr}`, but I always find it a tad bit dissatisfying to have slides with just the name of the package without a cool hex sticker to go along with it! 🙃

I wish I were creative enough to design and contribute one, but maybe this can be crowdsourced (on Twitter, e.g.)?

Thanks for the awesome work, either way! Looking forward to the `3.0.0` release.

AshesITR:
If created, this could also be featured here:
https://github.com/GuangchuangYu/hexSticker

MichaelChirico:
marking this one as duplicate

IndrajeetPatil:
Closed in favour of https://github.com/r-lib/lintr/issues/1188

PR #1425, opened 2022-06-21
Support SARIF format result

shaopeng-gh:
Related issue: https://github.com/r-lib/lintr/issues/991
Usage: lintr::sarif_output(lintr::lint_dir("C:/repos/project1_root"), "sarif-results.sarif")


AshesITR:
Hi @shaopeng-gh, thanks for the PR.
The pkgdown build failure can be fixed by adding `sarif_output` to `_pkgdown.yml`, e.g. on line 41.

R/lint.R
```diff
@@ -561,6 +561,161 @@ checkstyle_output <- function(lints, filename = "lintr_results.xml") {
...
+      if (length(rule_index_exists) == 0L ||
+          is.na(rule_index_exists[1L])) {
+        rule_index_exists <- 0L
+      }
+    }
+
+    if (rule_index_exists == 0L) {
+      new_rule <- list(
+        id = lint$linter,
+        fullDescription = list(text = lint$message),
```
AshesITR, line 681: Note that  `lint$message` is usually a customized message that incorporates context information for the specific piece of code that was linted, e.g. `undesirable_function_linter()` generates a message naming the concrete function that was detected. Is that okay for SARIF?

shaopeng-gh: Yes, currently this field lint$message is used for both the each result's message and also rule description. Rule description is required field can not be empty. In each result's message user can see the correct message for that instance of the violation which is correct and what we want; the rule description should be static and not have any variable in the string. In the future if we have a separated property like lint$rule_description we should use it. Many the rules lint$message I see is static so this is a good starting version.

AshesITR: Thanks for the explanation!

AshesITR (approved):

shaopeng-gh:
@AshesITR thanks!
1. not familiar with the package upgrade process, can a new version to be published?
I will work on the starter workflow after a new version is published and available to use, I see sample in the internet:
`
- uses: r-lib/actions/setup-r-dependencies@v2
        with:
          extra-packages: lintr
`
This should be correct right? And should be good to get the latest version, after a new version is published.

2. the starter workflow will need a icon for the linter, I only find one at the top left of page https://github.com/r-lib, will use it, let me know if there is a better one. (And it can also be changed later)

AshesITR:
Hi @shaopeng-gh, we just recently released version 3.0.0. Maybe we can get a patch release coming some time soon.
Regarding an icon, this is tracked by #981.

shaopeng-gh:
@AshesITR 
sounds good, 
I have created the [lintr starter workflow pr](https://github.com/actions/starter-workflows/pull/1621) and set in draft status, 
we just need to replace the logo image when it is done in https://github.com/r-lib/lintr/issues/1188 (fyi @hadley), 
and set the pr to ready to review status after a new version released.

PR #1431, opened 2022-06-26
Update lintr.Rproj

IndrajeetPatil:
AFAIK, this has been dropped by RStudio.

### DIFF
--- a/lintr.Rproj
+++ b/lintr.Rproj
@@ -17,6 +17,5 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
-PackageCleanBeforeInstall: Yes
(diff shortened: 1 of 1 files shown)


AshesITR:
How would a freshly generated `.Rproj` file look like?
Also, is there any problem with leaving old settings in the file? All I can imagine is that old RStudio versions might complain if the Rproj file is missing a key...

IndrajeetPatil:
This was never present in the old RStudio versions. It was introduced temporarily in daily builds and then dropped.

IndrajeetPatil:
E.g. https://github.com/tidyverse/dplyr/commit/c5977661a039fa69ef82b2dcafe2f7166506e48c

AshesITR (approved):

PR #1432, opened 2022-06-26
Update message for redundant ifelse linter

### LINKED ISSUE #1430: Message for ifelse(x, 0, 1) should be as.numeric(!x)
Currently we get `as.numeric(x)` in the replacement message for the following:

```r
ifelse(y > 0, 0, 1)
```

But the replacement should really be `!x`, i.e., `y <= 0`

IndrajeetPatil:
Closes #1430

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,9 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
-* `redundant_ifelse_linter()` produces correct message (#1432, @IndrajeetPatil).
+* `redundant_ifelse_linter()` produces correct message in the cases 
+  where the `yes` condition for `ifelse()` is `0`/`0L`, and `no` condition 
+  is `1`/`1L` (#1432, @IndrajeetPatil).
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,9 +8,8 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
-* `redundant_ifelse_linter()` produces correct message in the cases 
-  where the `yes` condition for `ifelse()` is `0`/`0L`, and `no` condition 
-  is `1`/`1L` (#1432, @IndrajeetPatil).
+* `redundant_ifelse_linter()` correctly suggests negation when 
+  the `yes` condition is `0` (#1432, @IndrajeetPatil).
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
+* `redundant_ifelse_linter()` produces correct message (#1432, @IndrajeetPatil).
+
 ## Bug fixes
 
 * `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
-* `redundant_ifelse_linter()` correctly suggests negation when 
+* `redundant_ifelse_linter()`'s lint message correctly suggests negation when 
   the `yes` condition is `0` (#1432, @IndrajeetPatil).
 
 ## Bug fixes
--- a/R/redundant_ifelse_linter.R
+++ b/R/redundant_ifelse_linter.R
@@ -49,12 +49,10 @@ redundant_ifelse_linter <- function(allow10 = FALSE) {
       # [1] call; [2] logical condition
       first_arg <- xml2::xml_find_chr(num_expr, "string(expr[3]/NUM_CONST)")
       second_arg <- xml2::xml_find_chr(num_expr, "string(expr[4]/NUM_CONST)")
+      is_numeric_01 <- first_arg %in% c("0", "1") | second_arg %in% c("0", "1")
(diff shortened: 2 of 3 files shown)


tests/testthat/test-redundant_ifelse_linter.R
```diff
@@ -47,12 +47,12 @@ test_that("redundant_ifelse_linter blocks usages equivalent to as.numeric, optio
 
   expect_lint(
     "ifelse(x > 5, 1, 0)",
-    rex::rex("Prefer as.numeric(x) to ifelse(x, 1, 0)"),
+    rex::rex("Prefer as.numeric(!x) to ifelse(x, 1, 0)"),
```
MichaelChirico, line 50: this is incorrect

IndrajeetPatil: Oops. Will fix that.

R/redundant_ifelse_linter.R
```diff
@@ -46,17 +46,17 @@ redundant_ifelse_linter <- function(allow10 = FALSE) {
...
+      # [1] call; [2] logical condition
       first_arg <- xml2::xml_find_chr(num_expr, "string(expr[3]/NUM_CONST)")
       second_arg <- xml2::xml_find_chr(num_expr, "string(expr[4]/NUM_CONST)")
+      replacement_numeric <- ifelse(first_arg == "0" & second_arg == "1", "as.numeric(!x)", "as.numeric(x)")
       replacement <- ifelse(
         first_arg %in% c("0", "1") | second_arg %in% c("0", "1"),
-        "as.numeric",
-        "as.integer"
+        replacement_numeric,
+        "as.integer(x)"
```
MichaelChirico, line 56: shouldn't ifelse(x, 0L, 1L) also recommend negation?

sorry for all the back & forth here...

IndrajeetPatil: Ah, sorry, missed that case here. Yes, we should also recommend negation here.

It should be `as.integer(!x)`, right? Since the original conditions are `0L` and `1L`?

``` r
x <- 8

ifelse(x > 5, 0L, 1L)
#> [1] 0

as.numeric(! (x > 5))
#> [1] 0
as.integer(! (x > 5))
#> [1] 0
```

<sup>Created on 2022-06-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MichaelChirico: right, exactly

MichaelChirico:
Getting much closer!

I may have missed this in my quick glance, but I don't think we have any tests yet for the following case:

```
ifelse(A, 0, 1L)
```

Where the alternative is `as.numeric(!A)`. Could you add some tests of the mixed integer+numeric case and ensure the update is working correctly for those cases as well? Thanks!

R/redundant_ifelse_linter.R
```diff
@@ -46,17 +46,17 @@ redundant_ifelse_linter <- function(allow10 = FALSE) {
...
-      # [1] call; [2] logical condiditon
+      # [1] call; [2] logical condition
       first_arg <- xml2::xml_find_chr(num_expr, "string(expr[3]/NUM_CONST)")
       second_arg <- xml2::xml_find_chr(num_expr, "string(expr[4]/NUM_CONST)")
-      replacement <- ifelse(
-        first_arg %in% c("0", "1") | second_arg %in% c("0", "1"),
-        "as.numeric",
-        "as.integer"
-      )
+      is_numeric_01 <- first_arg %in% c("0", "1") | second_arg %in% c("0", "1")
```
MichaelChirico, line 52: I would think a flow like

ifelse(is_numeric_01, "as.numeric", "as.integer")
is_negated <- first_arg %in% c("0", "0L")
ifelse(is_negated, "!x", "x")

then build the message with those two bits

would be simpler/easier to follow, is there a reason that won't work?

IndrajeetPatil: Thanks for the suggestion! 

That totally works; I was just being stingy in creating objects. But I guess this is a bit more readable.

MichaelChirico: if we really want to skimp we don't need to allocate either is_* variable since they're not used only once. either ways basically don't though. these objects are almost always tiny (maybe a couple dozen elements in the most extreme realistic case)

R/redundant_ifelse_linter.R
```diff
@@ -46,17 +46,17 @@ redundant_ifelse_linter <- function(allow10 = FALSE) {
...
+      replacement_is_numeric <- ifelse(is_numeric_01, "as.numeric", "as.integer")
```
MichaelChirico, line 53: not: better naming. current reads like it's a boolean variable. maybe "coercion_function"?

R/redundant_ifelse_linter.R
```diff
@@ -46,17 +46,17 @@ redundant_ifelse_linter <- function(allow10 = FALSE) {
...
+      is_negated <- first_arg %in% c("0", "0L")
+      replacement_is_negated <- ifelse(is_negated, "!x", "x")
```
MichaelChirico, line 55: not: ditto. maybe "replacement_argument"?

tests/testthat/test-redundant_ifelse_linter.R
```diff
@@ -52,7 +52,19 @@ test_that("redundant_ifelse_linter blocks usages equivalent to as.numeric, optio
   )
   expect_lint(
     "ifelse(x > 5, 0, 1)",
-    rex::rex("Prefer as.numeric(x) to ifelse(x, 0, 1)"),
+    rex::rex("Prefer as.numeric(!x) to ifelse(x, 0, 1)"),
+    redundant_ifelse_linter()
+  )
+
+  # mixing int and num
```
MichaelChirico, line 59: nit: add tests for mixed 1,0 too

NEWS.md
```diff
@@ -8,6 +8,10 @@
 * `seq_linter()` additionally lints on `1:n()` (from dplyr) 
   and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
 
+* `redundant_ifelse_linter()` produces correct message in the cases
```
MichaelChirico, line 11: much better! suggested tweak:

> correctly suggests negation when the `yes` condition is 0

(0 not in code so that it implies both the numeric and integer cases; and mentioning 1 is redundant since the lint only throws if 0/1 are paired)

MichaelChirico:
Very close now thanks! small stuff left

PR #1436, opened 2022-06-28
add version requirement for OP-LAMBDA

MichaelChirico:
Alternatively, we could just bump the version requirement to 1.0.5


IndrajeetPatil:
It has been over a year since `xmlparsedata 1.0.5` release, so I'd personally prefer if we bumped the required version instead.

IndrajeetPatil (approved):
Thanks!

PR #1437, opened 2022-06-28
include rlang functions in literal coercion

### LINKED ISSUE #1382: include rlang::int(1) and similar in literal_coercion_linter
`rlang::int()` is really meant to be a typed builder like `c(1, 2, 3)` with specified output type like

```
rlang::int(1, 3, 5) == c(1L, 3L, 5L)
rlang::chr("a", "b", "c") = c("a", "b", "c")
```

but if `...lengths() == 1`, it seems reasonable to just use a literal instead.

IndrajeetPatil:
Closes #1382

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,10 +3,14 @@
 * `modify_defaults()` no longer uses the mistaken `"lintr_function"` S3 class, instead applying the
   `"linter"` class also common to `Linter()`. `Linter()` also includes `"function"` in the S3
   class of its output to facilitate S3 dispatch to `function` methods where appropriate (#1392, @MichaelChirico).
+
 ## Changes to defaults
 
-* `seq_linter()` additionally lints on `1:n()` (from dplyr) 
-  and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
+* `seq_linter()` additionally lints on `1:n()` (from {dplyr}) 
+  and `1:.N` (from {data.table}) (#1396, @IndrajeetPatil).
+
+* `literal_coercion_linter()` lints {rlang}'s atomic vector constructors if 
+  the argument is a scalar (#1437, @IndrajeetPatil).
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,9 @@
 * `literal_coercion_linter()` lints {rlang}'s atomic vector constructors if 
   the argument is a scalar (#1437, @IndrajeetPatil).
 
+* `redundant_ifelse_linter()`'s lint message correctly suggests negation when 
+  the `yes` condition is `0` (#1432, @IndrajeetPatil).
+
 ## Bug fixes
 
 * `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,10 @@
 * `redundant_ifelse_linter()`'s lint message correctly suggests negation when 
   the `yes` condition is `0` (#1432, @IndrajeetPatil).
 
+## New and improved features
+
+* `unreachable_code_linter()` ignores trailing comments if they match a closing nolint block (#1347, @AshesITR).
+
 ## Bug fixes
 
 * `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,8 +9,9 @@
 * `seq_linter()` additionally lints on `1:n()` (from {dplyr}) 
   and `1:.N` (from {data.table}) (#1396, @IndrajeetPatil).
 
-* `literal_coercion_linter()` lints {rlang}'s atomic vector constructors if 
-  the argument is a scalar (#1437, @IndrajeetPatil).
+* `literal_coercion_linter()` lints {rlang}'s atomic vector constructors 
+  (`int()`, `chr()`, `lgl()`, `dbl()`) if the argument is a scalar 
+  (#1437, @IndrajeetPatil).
 
 * `redundant_ifelse_linter()`'s lint message correctly suggests negation when 
   the `yes` condition is `0` (#1432, @IndrajeetPatil).
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
   and `1:.N` (from {data.table}) (#1396, @IndrajeetPatil).
 
 * `literal_coercion_linter()` lints {rlang}'s atomic vector constructors 
-  (`int()`, `chr()`, `lgl()`, `dbl()`) if the argument is a scalar 
+  (i.e., `int()`, `chr()`, `lgl()`, and `dbl()`) if the argument is a scalar 
   (#1437, @IndrajeetPatil).
 
 * `redundant_ifelse_linter()`'s lint message correctly suggests negation when 
--- a/R/literal_coercion_linter.R
+++ b/R/literal_coercion_linter.R
@@ -34,7 +34,9 @@ literal_coercion_linter <- function() {
     )
   ]"
   xpath <- glue::glue("//expr[
-    expr[1][SYMBOL_FUNCTION_CALL[ {coercers} ]] and {not_extraction_or_scientific} and count(expr) = 2
+    expr[1][SYMBOL_FUNCTION_CALL[ {coercers} ]]
+    and {not_extraction_or_scientific}
+    and count(expr) = 2
   ]")
 
   Linter(function(source_expression) {
--- a/R/literal_coercion_linter.R
+++ b/R/literal_coercion_linter.R
@@ -15,7 +15,7 @@ literal_coercion_linter <- function() {
   rlang_coercers <- xp_text_in_table(
     c("lgl", "int", "dbl", "chr")
   )
-  coercers <- xp_text_in_table(
+  base_coercers <- xp_text_in_table(
       paste0("as.", c("logical", "integer", "numeric", "double", "character"))
   )
   # notes for clarification:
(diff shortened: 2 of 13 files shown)


R/literal_coercion_linter.R
```diff
@@ -12,23 +12,31 @@
...
+        not(OP-DOLLAR)
+        and (
+          NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
+          or STR_CONST[not(following-sibling::*[1][self::EQ_SUB])]
+        )
+      ]
+    ) or
+    (
+      expr[1][SYMBOL_FUNCTION_CALL[ {rlang_coercers} ]]
+      and not(OP-COMMA)
```
MichaelChirico, line 38: does rlant support list2 construction, i.e., does `lgl(1,)` work? if so, may need count(expr) here instead

R/literal_coercion_linter.R
```diff
@@ -45,7 +53,7 @@ literal_coercion_linter <- function() {
       source_expression = source_expression,
       lint_message = paste(
         "Use literals directly where possible, instead of coercion.",
-        "c.f. 1L instead of as.integer(1), or NA_real_ instead of as.numeric(NA)."
+        "c.f. 1L instead of as.integer(1) or rlang::int(1), or NA_real_ instead of as.numeric(NA)."
```
MichaelChirico, line 56: we should customize this to match what's actually linted. but it predates this PR, so we can put that as a follow-up issue

R/literal_coercion_linter.R
```diff
@@ -12,23 +12,31 @@
...
-        NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
-        or STR_CONST[not(following-sibling::*[1][self::EQ_SUB])]
-      )
-    ]
+    (
+      expr[1][SYMBOL_FUNCTION_CALL[ {coercers} ]]
+      and expr[2][
+        not(OP-DOLLAR)
+        and (
+          NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
```
MichaelChirico, line 31: note this condition, which let's a carve out for `as.integer(1e6)`. shouldnt we also allow `int(1e6)` then?

R/literal_coercion_linter.R
```diff
@@ -12,23 +12,31 @@
...
 #' @export
 literal_coercion_linter <- function() {
-  coercers <- xp_text_in_table(paste0(
-    "as.",
-    c("logical", "integer", "numeric", "double", "character")
-  ))
+  rlang_coercers <- xp_text_in_table(
+    c("lgl", "int", "dbl", "chr")
+  )
+  coercers <- xp_text_in_table(
```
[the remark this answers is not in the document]
IndrajeetPatil: Done.

R/literal_coercion_linter.R
IndrajeetPatil: Let's also try to fix this issue in the current PR.

How should we be customizing this? Any good example from existing linters for me draw inspiration?

R/literal_coercion_linter.R
IndrajeetPatil: We should definitely allow this. I will fix it.

R/literal_coercion_linter.R
IndrajeetPatil: Good catch! I have changed xpath to catch/lint this possibility as well.

R/literal_coercion_linter.R
IndrajeetPatil: Although unlikely to occur in the wild, for the sake of completeness, should `as.character('e')` be considered a lint?

R/literal_coercion_linter.R
MichaelChirico: is it always possible to suggest an exact alternative? I think it should be. so

as.integer(1) -> suggest 1L
as.numeric(NA) -> suggest NA_real_

etc. looking again I think it's definitely a big enough undertaking to warrant its own PR

IndrajeetPatil: Okay, sounds good. I can work on it in a different PR.

R/literal_coercion_linter.R
IndrajeetPatil: Okay, I have added tests for this context for both base and _rlang_ coercers, and all tests are passing. 

So, if necessary, this should be good for another round of review.

NEWS.md
```diff
@@ -3,10 +3,14 @@
...
   class of its output to facilitate S3 dispatch to `function` methods where appropriate (#1392, @MichaelChirico).
+
 ## Changes to defaults
 
-* `seq_linter()` additionally lints on `1:n()` (from dplyr) 
-  and `1:.N` (from data.table) (#1396, @IndrajeetPatil).
+* `seq_linter()` additionally lints on `1:n()` (from {dplyr}) 
+  and `1:.N` (from {data.table}) (#1396, @IndrajeetPatil).
+
+* `literal_coercion_linter()` lints {rlang}'s atomic vector constructors if
```
MichaelChirico, line 12: nit: give an example in case the generic description doesn't click for the reader

> (e.g. int(), chr(), lgl(), etc)

(or maybe even just list them all)

IndrajeetPatil: Done.

R/literal_coercion_linter.R
```diff
@@ -12,23 +12,41 @@
...
+          NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
+          or STR_CONST[not(following-sibling::*[1][self::EQ_SUB])]
+        )
+      ]
+    ) or
+    (
+        (
+          expr[1][SYMBOL_FUNCTION_CALL[ {rlang_coercers} ]]
+          and expr[2][
+          not(OP-DOLLAR)
```
MichaelChirico, line 41: probably makes sense to save this condition as a variable and `glue()` it in since it's reused

R/literal_coercion_linter.R
MichaelChirico: #1441

tests/testthat/test-literal_coercion_linter.R
```diff
@@ -29,22 +41,50 @@ patrick::with_parameters_test_that(
...
+patrick::with_parameters_test_that(
+  "literal_coercion_linter blocks rlang disallowed usages",
+  expect_lint(
+    sprintf("%s(%s)", out_type, input),
+    rex::rex("Use literals directly where possible, instead of coercion."),
+    literal_coercion_linter()
+  ),
+  # even if `as.character(1)` works, `chr(1)` doesn't, so no corresponding test case
+  .cases = tibble::tribble(
+    ~.test_name,  ~out_type, ~input,
```
MichaelChirico, line 71: can this be combined with the previous suite of parametrized tests?

MichaelChirico:
just about ready to go -- looking good!

tests/testthat/test-literal_coercion_linter.R
IndrajeetPatil: The key difference between these two tests is the `sprintf` function here:

- For base coercers

```r
  expect_lint(
    sprintf("as.%s(%s)", out_type, input),
    rex::rex("Use literals directly where possible, instead of coercion."),
    literal_coercion_linter()
  )
```

- For rlang coercers

```r
  expect_lint(
    sprintf("%s(%s)", out_type, input),
    rex::rex("Use literals directly where possible, instead of coercion."),
    literal_coercion_linter()
  )
```

So maybe a new column called `prefix` can be added with a value of `"as."` for base coercers, while `""` for rlang coercers, and then it will be `sprintf("%s%s(%s)", prefix, out_type, input),`.

WDYT?

MichaelChirico: no that's overcomplicated, thanks, this works fine!

R/literal_coercion_linter.R
```diff
@@ -12,23 +12,41 @@
...
+    (
+      expr[1][SYMBOL_FUNCTION_CALL[ {base_coercers} ]]
+      and expr[2][
+        not(OP-DOLLAR)
+        and (
+          NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
+          or STR_CONST[not(following-sibling::*[1][self::EQ_SUB])]
+        )
+      ]
+    ) or
```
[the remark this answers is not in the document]
IndrajeetPatil: Done; I have simplified this xpath a bit, which makes it a bit more readable.

R/literal_coercion_linter.R
IndrajeetPatil: Thanks! Simplifies the path significantly. Done!

R/literal_coercion_linter.R
```diff
@@ -18,32 +18,24 @@ literal_coercion_linter <- function() {
   base_coercers <- xp_text_in_table(
       paste0("as.", c("logical", "integer", "numeric", "double", "character"))
   )
+
   # notes for clarification:
   #  - as.integer(1e6) is arguably easier to read than 1000000L
   #  - in x$"abc", the "abc" STR_CONST is at the top level, so exclude OP-DOLLAR
   #  - need condition against STR_CONST w/ EQ_SUB to skip quoted keyword arguments (see tests)
   #  - for {rlang} coercers, both `int(1)` and `int(1, )` need to be linted
+  xpath_e <- "expr[2][
```
MichaelChirico, line 27: better name needed... not_extraction_or_scientific?

IndrajeetPatil: Good point. Using the suggested name.

R/literal_coercion_linter.R
```diff
@@ -12,23 +12,31 @@
...
-      not(OP-DOLLAR)
-      and (
-        NUM_CONST[not(contains(translate(text(), 'E', 'e'), 'e'))]
-        or STR_CONST[not(following-sibling::*[1][self::EQ_SUB])]
-      )
-    ]
+    (
+      expr[1][SYMBOL_FUNCTION_CALL[ {base_coercers} ]] and {not_extraction_or_scientific}
+    ) or (
+      expr[1][SYMBOL_FUNCTION_CALL[ {rlang_coercers} ]] and {not_extraction_or_scientific} and count(expr) = 2
```
MichaelChirico, line 38: IINM count(expr)=2 also applies to the base case, right? it's just kind of redundant since trailing ',' is not allowed. but we could combine the conditions here for conciseness, WDYT?

IndrajeetPatil: That worked! Great idea. 

Done.

MichaelChirico (approved):

PR #1446, opened 2022-07-10
skip evaluation that hangs on bad input

### LINKED ISSUE #1443: lintr hangs on openxlsx
Hello,
in one of the many packages for which I use `lintr`, `lintr` started hanging after the last version 3.0. This can be reproduced both locally (R 4.2.1 on Mac and `lintr`s HEAD) and in the Github action.

`lintr::lint_package()` hangs on [`openxlsx`](https://github.com/ycphs/openxlsx). This can be seen with the latest CRAN release of `openxlsx` and also in the `development` branch. While the `openxlsx` code is known to be a mess, `lintr` did not hang before its release 3.0. Unfortunately there is not much else I can report, there is no verbose option that I am aware of. I tried to disable the linter mentioned in #1427 but that did not help. Let me know if you need anything else.

Thank you for the time you have invested in the development and maintenance of `lintr`!

MichaelChirico:
Closes #1443

Not totally happy we've solved the problem with a high level of generality, but it does close the issue at hand.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,7 +31,7 @@
 * Parse error lints now appear with the linter name `"error"` instead of `NA` (#1405, @AshesITR).  
   Also, linting no longer runs if the `source_expressions` contain invalid string data that would cause error messages
   in other linters.
-   + Also precludes hanging on Rmd files with some syntax errors (#1443, @MichaelChirico).
+* Prevent `lint()` from hanging on Rmd files with some syntax errors (#1443, @MichaelChirico).
 * `get_source_expressions()` no longer omits trailing non-code lines from knitr files (#1400, #1415, @AshesITR).  
   This fixes the location information for `trailing_blank_lines_linter()` in RMarkdown documents without terminal
   newlines.
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,7 +30,8 @@
 * `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
 * Parse error lints now appear with the linter name `"error"` instead of `NA` (#1405, @AshesITR).  
   Also, linting no longer runs if the `source_expressions` contain invalid string data that would cause error messages
-  in other linters. 
+  in other linters.
+   + Also precludes hanging on Rmd files with some syntax errors (#1443, @MichaelChirico).
 * `get_source_expressions()` no longer omits trailing non-code lines from knitr files (#1400, #1415, @AshesITR).  
   This fixes the location information for `trailing_blank_lines_linter()` in RMarkdown documents without terminal
   newlines.
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -79,12 +79,12 @@ get_source_expressions <- function(filename, lines = NULL) {
   names(source_expression$lines) <- seq_along(source_expression$lines)
   source_expression$content <- get_content(source_expression$lines)
   parsed_content <- get_source_expression(source_expression, error = function(e) lint_parse_error(e, source_expression))
-  top_level_map <- generate_top_level_map(parsed_content)
 
-  if (inherits(e, "lint") && !nzchar(e$line)) {
+  if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line))) {
     # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
     expressions <- list()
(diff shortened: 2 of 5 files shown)


R/get_source_expressions.R
```diff
@@ -79,12 +79,12 @@ get_source_expressions <- function(filename, lines = NULL) {
   names(source_expression$lines) <- seq_along(source_expression$lines)
   source_expression$content <- get_content(source_expression$lines)
   parsed_content <- get_source_expression(source_expression, error = function(e) lint_parse_error(e, source_expression))
-  top_level_map <- generate_top_level_map(parsed_content)
 
-  if (inherits(e, "lint") && !nzchar(e$line)) {
+  if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line))) {
```
MichaelChirico, line 83: As the parser fails out on the last line of the `NA`-blanked version of `lines`, `lint_parse_error_r42()` returns one of these `NA` lines. `nzchar(NA)` is `TRUE`, and `nzchar(NA, keepNA=TRUE)` would fail for `if (NA)`.

R/get_source_expressions.R
```diff
@@ -79,12 +79,12 @@ get_source_expressions <- function(filename, lines = NULL) {
...
     # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
     expressions <- list()
   } else {
+    top_level_map <- generate_top_level_map(parsed_content)
```
MichaelChirico, line 87: not needed in the `inherits("lint")` branch, so no need to run this code there. I _think_ it's safe to assume the `while()` loop in `generate_top_level_map()` will work as long as `e` is `NULL`, so didn't touch that. But for example, we could add a conservative test that the `while()` loop must terminate, say, at at most the # rows of the parse content.

R/get_source_expressions.R
AshesITR: Since this is the second bug relating to NA lines, can we add a check to `Lint()` to assert that `line` is not `NA`?
This should expedite bug hunts in these cases. WDYT?

NEWS.md
```diff
@@ -30,7 +30,8 @@
 * `get_source_expressions()` no longer fails on R files that match a knitr pattern (#743, #879, #1406, @AshesITR).
 * Parse error lints now appear with the linter name `"error"` instead of `NA` (#1405, @AshesITR).  
   Also, linting no longer runs if the `source_expressions` contain invalid string data that would cause error messages
-  in other linters. 
+  in other linters.
+   + Also precludes hanging on Rmd files with some syntax errors (#1443, @MichaelChirico).
```
AshesITR, line 34: IMO this should be a separate bullet.

MichaelChirico: Thanks! I originally thought this was a regression introduced in dev. Confirmed the same issue is on 3.0.0, so split to its own item.

R/get_source_expressions.R
MichaelChirico: Hmm, this caused some other tests to fail. Agree it's a good idea, but let's handle it in a follow-up instead.

MichaelChirico: #1450

R/tree-utils.R
```diff
@@ -6,9 +6,12 @@ generate_top_level_map <- function(pc) {
   tl_parent <- pc$parent
   tl_parent[pc$parent <= 0L] <- tl_ids
   i_not_assigned <- which(!tl_parent %in% tl_ids)
-  while (length(i_not_assigned)) {
+  while ((prev_length <- length(i_not_assigned)) > 0L) {
     tl_parent[i_not_assigned] <- pc$parent[match(tl_parent[i_not_assigned], pc$id)]
     i_not_assigned <- which(!tl_parent %in% tl_ids)
+    if (length(i_not_assigned) >= prev_length) {
+      stop("Logical error: unassigned set did not shrink. Check file syntax and please report as a bug.")
```
AshesITR, line 13: Add GitHub issue link? As-is the error message might not be clearly linked to lintr in production.

AshesITR (commented):
Basically good now.

R/tree-utils.R
MichaelChirico: Just added 'lintr' -- should be slightly more stable than referencing the GH link (GH site structure changes, or "turns evil" & we change hosts, etc)

AshesITR (approved):

PR #1451, opened 2022-07-16
Comment on preceding line fine with `brace_linter()`

### LINKED ISSUE #1433: brace_linter exception opening exception should apply for preceding comma-then-comment, too
Compare:

```r
lintr::lint('
test_that(
  "Some longer test description",
  {
    # Do test stuff
  }
)',
lintr::brace_linter())
```

vs

```r
lintr::lint('
test_that(
  "Some longer test description", # comment
  {
    # Do test stuff
  }
)',
lintr::brace_linter())
# <text>:4:3: style: [brace_linter] Opening curly braces should never go on their own line and should always be followed by a new line.
#   {
#   ^
```

The exception should apply to the latter case as well

### LINKED ISSUE #1434: brace_linter exception opening exception should apply when a comment precedes the lone brace, too
(nearly the same as #1433, and maybe both fixed at the same time, but filed separately since the cause is just different enough)

Compare:

```r
lintr::lint('
switch(stat,
  o = {
    FUN <- FO
    bw <- 0.01
  },
  s = {
    FUN <- FS
    bw <- 0.005
  },
  {
    FUN <- FE
    bw <- 2
  }
)',
lintr::brace_linter())
```

vs

```r
lintr::lint('
switch(stat,
  o = {
    FUN <- FO
    bw <- 0.01
  },
  s = {
    FUN <- FS
    bw <- 0.005
  },
  # else
  {
    FUN <- FE
    bw <- 2
  }
)',
lintr::brace_linter())
# <text>:12:3: style: [brace_linter] Opening curly braces should never go on their own line and should always be followed by a new line.
#   {
#   ^
```

The latter also shouldn't lint -- the `# else` comment is quite common/helpful, especially for long `switch()` statements.

IndrajeetPatil:
Closes #1433
Closes #1434


tests/testthat/test-brace_linter.R
```diff
@@ -100,6 +100,43 @@ test_that("brace_linter lints braces correctly", {
     linter
   )
 
+  # , #comment<\n>{ is allowed
```
MichaelChirico, line 103: I guess reword this to turn off the comment linter hit

tests/testthat/test-brace_linter.R
```diff
@@ -100,6 +100,43 @@ test_that("brace_linter lints braces correctly", {
...
+          do_first(x)
+          do_second(x)
+        }
+      )
+    "),
+    NULL,
+    linter
+  )
+
+  # #comment<\n>{ is allowed
```
MichaelChirico, line 120: same here, though I'm surprised this is turning up a lint

R/brace_linter.R
```diff
@@ -29,9 +29,15 @@ brace_linter <- function(allow_single_line = FALSE) {
...
     )",
     # allow `(`, `,` and `%>%` on preceding line
+    #
+    # note that '{' is not supported in RHS call of base-R's native pipe (`|>`),
+    # so this doesn't need to be supported
     "not(
       @line1 = parent::expr/preceding-sibling::*[1][
-        self::OP-LEFT-PAREN or self::OP-COMMA or (self::SPECIAL and text() = '%>%')
+        self::OP-LEFT-PAREN or
+        self::OP-COMMA or
```
MichaelChirico, line 38: this doesn't quite feel right to me but I cant think of a counterexample at the moment 🤭

would you mind trying to run .dev/compare_branches on a few packages to see what turns up?

my thinking is we _do_ still want to throw a lint if the node before the comment doesn't match these existing criteria. so the XPath (untested) should be something like preceding-sibling::*[not(self::COMMENT)][1][...] where ... is the existing test.

IndrajeetPatil: > I cant think of a counterexample at the moment 

Is this a good counterexample? This should produce a lint, correct?

``` r
lintr::lint(
  "# bla
  local(
    {
      x <- 2
    }
  )",
  lintr::brace_linter()
)
```

<sup>Created on 2022-07-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MichaelChirico: that also doesn't produce a lint on `HEAD`, and is one of the carve-outs we have regarding the preceding line:

https://github.com/r-lib/lintr/blob/75d4d3116107f33f258ff6c0eeb8a65464888fa1/R/brace_linter.R#L31

MichaelChirico: I ran .dev/compare_branches.R on a sample of 100 packages, found an example that lints at HEAD but not in this branch:

https://github.com/cran/gets/blob/beb581d92e3a0d448ef1a69932c98036fcb4baf9/R/gets-base-source.R#L262-L264

The comment does not excuse the incorrect use of `{` where it should be on the previous line with `function ()`

IndrajeetPatil: Awesome! Thanks for finding a counterexample. I have added a (currently failing) test for this.

> preceding-sibling::*[not(self::COMMENT)][1][...] 

I tried this, but the test for #1434 fails. I will see how it can be modified to make sure all tests pass.

IndrajeetPatil: No matter the xpath I try, I seem to get only 2 out 3 tests to pass. So I'm stuck here.

MichaelChirico: Subtle one! we were using a fixed test of `@line1 = @prev::line2 + 1`, but where there's an intervening comment (or comments), `+1` won't be right.

This condition is to lint on cases like `foo(x, { y })` where `,` and `{` are on the same line, so we don't actually care about `+1` -- we only care that `@line1 != @prev::line2`.

MichaelChirico (approved):

NEWS.md
```diff
@@ -7,7 +7,7 @@
 ## Changes to defaults
 
 * `brace_linter()` allows opening curly braces on a new line when there is 
-  a comment on the preceding line (#1451, @IndrajeetPatil).
+  a comment ending the preceding line (#1433 and #1434, @IndrajeetPatil).
```
IndrajeetPatil, line 10: I think all of these are wrong for PRs I've contributed because I thought we were writing the PR numbers that fixed the issue, and not the issue numbers. 🤦‍♂️

I will make a separate PR to fix all of them. Sorry about that.

MichaelChirico: I don't think it matters enough to make a cleanup PR -- I just think it's more direct for NEWS reader to land on the bug report page than the PR page (which will link tot he bug page).

the implicit assumption is the reader cares about the nature of the bug (to see if they're affected) more than the code to fix it.

Issue #1453, opened 2022-07-17
false positive in nested vector_logic_linter usage when '&' is a method


MichaelChirico:
`&` and `|` are generics, so methods can be defined for which the output is not `logical`.

This is the case for base `octmode` class which has an `&` method to do bitwise-and (much like C/Python's `&`) and returns `octmode` (see `?"&.octmode"`). As seen in base R:

https://github.com/r-devel/r-svn/blob/9bcb4bef2979e608804fb37ab6471b4894bfa2b2/src/library/utils/R/tar.R#L497

```r
if (is.null(extra_flags) && grepl("/(configure|cleanup)$", f) &&
    (mode & "111") != as.octmode("111")) {
    #     ^ marked as a lint
    warning(gettextf("file '%s' did not have execute permissions: corrected", f), domain = NA, call. = FALSE)
    mode <- mode | "111"
}
```

The giveaway here is that a comparison operator is used in the `<expr>` above `&` -- IINM `vector_logic_linter()` assumes parent infixes will be `&`/`&&`/`|`/`||`.

My first instinct was just to call this a known edge case & move on, because in general we can't swap `&&` for all methods of `&`, but for this particular method we can detect it's fine statically.

MichaelChirico:
Very similar case seen in sf:

https://github.com/r-spatial/sf/blob/3362108a005ab77ba83953430dbaa6edf4b26075/R/wkb.R#L128-L132

```r
zm = if ((info & as.raw(12)) == as.raw(12))
  "XYZM"
else if (info & as.raw(8))
  "XYZ"
else if (info & as.raw(4))
```

PR #1457, opened 2022-07-17
another note re: edge cases in | vs ||

MichaelChirico:
Related to #1453 again.

### DIFF
--- a/R/vector_logic_linter.R
+++ b/R/vector_logic_linter.R
@@ -17,7 +17,7 @@
 #'   early, if `a` is `TRUE`,  the second condition will never be evaluated
 #'   and `b` will not be assigned. Such usage is not allowed by the Tidyverse
 #'   style guide, and the code can easily be refactored by pulling the
-#'   assignment outside of the condition, so using `||` is still preferable.
+#'   assignment outside the condition, so using `||` is still preferable.
 #'
 #' @evalRd rd_tags("vector_logic_linter")
 #' @seealso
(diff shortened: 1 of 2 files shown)


R/vector_logic_linter.R
```diff
@@ -11,6 +11,14 @@
 #'   `&&` / `||` are not perfect substitutes because `&` is doing
 #'   method dispatch in an incompatible way.
 #'
+#' Moreover, be wary of code that may have side effects, most commonly
+#'   assignments. Consider `if ((a <- foo(x)) | (b <- bar(y))) { ... }`
+#'   vs. `if ((a <- foo(x)) || (b <- bar(y))) { ... }`. Because `||` exits
+#'   early, if `a` is `TRUE`,  the second condition will never be evaluated
+#'   and `b` will not be assigned. Such usage is not allowed by the Tidyverse
+#'   style guide, and the code can easily be refactored by pulling the
+#'   assignment outside of the condition, so using `||` is still preferable.
```
IndrajeetPatil, line 20:
```suggestion
#'   assignment outside the condition, so using `||` is still preferable.
```

IndrajeetPatil (approved):
Very useful note! Thanks.

R/vector_logic_linter.R
MichaelChirico: indeed looks like the more common choice 

https://books.google.com/ngrams/graph?content=outside+the%2Coutside+of+the&year_start=1800&year_end=2019&corpus=26&smoothing=3&case_insensitive=true

AshesITR: Probably pull sth out of sth would be more correct.

IndrajeetPatil (approved):

PR #1463, opened 2022-07-20
fix grammar on undesirable_function_linter default messages

### LINKED ISSUE #1462: undesirable_function_linter mangles suggested action and reason
The output message states "... As an alternative, It isn't type stable. Use vapply() with an appropriate `FUN.VALUE=` argument to obtain type stable simplification.."

It seems that "As an alternative, " is out of place and it should be rather something as "... It isn't type stable. As an alternative, use vapply() with an appropriate `FUN.VALUE=` argument to obtain type stable simplification."

It is the same situation with other undesirable functions including `par`, `library`, `require` etc.

``` r
print(packageVersion("lintr"))
#> [1] '3.0.0.9000'

lintr::lint(
  text = "sapply(1:2, seq_len)",
  linters = lintr::undesirable_function_linter()
)
#> <text>:1:1: style: [undesirable_function_linter] Function "sapply" is undesirable. As an alternative, It isn't type stable. Use vapply() with an appropriate `FUN.VALUE=` argument to obtain type stable simplification..
#> sapply(1:2, seq_len)
#> ^~~~~~
```

<sup>Created on 2022-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MichaelChirico:
Closes #1462


R/zzz.R
```diff
@@ -98,43 +98,79 @@ default_linters <- modify_defaults(
...
-  "untrace" = "It is only useful for interactive debugging with trace(). It should be removed."
+  "attach" = 
+    "use roxygen2's @importFrom statement in packages, or `::` in scripts. attach() modifies the global search path",
+  "browser" =
+    "remove this likely leftover from debugging. It pauses execution when run",
+  "debug" = paste(
+    "remove this likely leftover from debugging.",
+    "It traps a function and causes execution to pause when that function is run"
+  ),
+  "debugcall" =paste(
```
IndrajeetPatil, line 109:
```suggestion
  "debugcall" = paste(
```

IndrajeetPatil:
For a few example messages, see: https://github.com/r-lib/lintr/issues/1462#issuecomment-1197006144

IndrajeetPatil (approved):
Thanks! The messages read quite nice now.

PR #1468, opened 2022-07-24
Remove false negative in `seq_linter()`

### LINKED ISSUE #1467: seq_linter should catch seq(length()), redirecting to seq_along()
```r
seq(0)
# [1] 1 0
```

i.e., `seq(length(x))` is subject to the same unexpected/error-prone edge-case behavior as other usages caught by `seq_linter()`

IndrajeetPatil:
Closes #1467

TODO

- [x] Get rid of false negative
- [x] Tests should pass
- [x] Correct lint message should be produced
- [x] Update documentation (needs bumping roxygen2 version)
- [x] Address code review comments

# Example

``` r
library(lintr)

lint(
  text = "seq(length(mtcars))",
  linters = seq_linter()
)
#> <text>:1:1: warning: [seq_linter] seq(length(...)) is likely to be wrong in the empty edge case. Use seq_along(...) instead.
#> seq(length(mtcars))
#> ^~~~~~~~~~~~~~~~~~~

lint(
  text = "seq(nrow(mtcars))",
  linters = seq_linter()
)
#> <text>:1:1: warning: [seq_linter] seq(nrow(...)) is likely to be wrong in the empty edge case. Use seq_len(...) instead.
#> seq(nrow(mtcars))
#> ^~~~~~~~~~~~~~~~~
```

<sup>Created on 2022-07-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -63,10 +63,18 @@ seq_linter <- function() {
       "seq_along",
       "seq_len"
     )
-    lint_message <- sprintf(
-      "%s:%s is likely to be wrong in the empty edge case. Use %s() instead.",
-      dot_expr1, dot_expr2, replacement
-    )
+
+    if (any(grepl("seq", dot_expr1, fixed = TRUE))) {
+      lint_message <- sprintf(
+        "%s(%s) is likely to be wrong in the empty edge case. Use %s(%s(...)) instead.",
+        dot_expr1, dot_expr2, dot_expr1, replacement
+      )
+    } else {
+      lint_message <- sprintf(
+        "%s:%s is likely to be wrong in the empty edge case. Use %s() instead.",
+        dot_expr1, dot_expr2, replacement
+      )
+    }
 
     xml_nodes_to_lints(badx, source_expression, lint_message, type = "warning")
   })
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -13,13 +13,19 @@ seq_linter <- function() {
 
   # `.N` from {data.table} is special since it's not a function but a symbol
   xpath <- glue::glue("//expr[
-    expr[NUM_CONST[text() =  '1' or text() =  '1L']]
-    and OP-COLON
-    and (
-          expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
-          or
-          expr[SYMBOL = '.N']
-        )
+    (
+      expr[SYMBOL_FUNCTION_CALL[text() = 'seq']] and
+      expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
+    )
+    or (
+      expr[NUM_CONST[text() =  '1' or text() =  '1L']]
+      and OP-COLON
+      and (
+            expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
+            or
+            expr[SYMBOL = '.N']
+          )
+    )
   ]")
 
   ## The actual order of the nodes is document order
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -14,21 +14,19 @@
 #' @export
 seq_linter <- function() {
   bad_funcs <- c("length", "n", "nrow", "ncol", "NROW", "NCOL", "dim")
+  bad_func_xpath <- glue::glue(
+    "expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]"
+  )
 
   # `.N` from {data.table} is special since it's not a function but a symbol
   xpath <- glue::glue("//expr[
     (
-      expr[SYMBOL_FUNCTION_CALL[text() = 'seq']] and
-      expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
-    )
-    or (
+      expr[SYMBOL_FUNCTION_CALL[text() = 'seq']]
+      and {bad_func_xpath}
+    ) or (
       expr[NUM_CONST[text() =  '1' or text() =  '1L']]
       and OP-COLON
-      and (
-            expr[expr[(expr|self::*)[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]]]
-            or
-            expr[SYMBOL = '.N']
-          )
+      and ( {bad_func_xpath} or expr[SYMBOL = '.N'] )
     )
   ]")
 
--- a/R/seq_linter.R
+++ b/R/seq_linter.R
@@ -17,7 +17,9 @@ seq_linter <- function() {
 
   # Exact `xpath` depends on whether bad function was used in conjunction with `seq()`
   bad_func_xpath_with_seq <- glue::glue(
-    "expr[1][SYMBOL_FUNCTION_CALL[text() = 'seq']]/following::expr[1]/expr[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]"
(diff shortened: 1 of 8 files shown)


R/seq_linter.R
```diff
@@ -57,10 +67,18 @@ seq_linter <- function() {
       "seq_along",
       "seq_len"
     )
-    lint_message <- sprintf(
-      "%s:%s is likely to be wrong in the empty edge case. Use %s() instead.",
-      dot_expr1, dot_expr2, replacement
-    )
+
+    if (any(grepl("seq", dot_expr1, fixed = TRUE))) {
```
MichaelChirico, line 71: should this be an `ifelse`? what happens for `c(seq(length(x)), 1:nrow(x))`?

IndrajeetPatil: Good catch! Fixed it, and added a test as well:

``` r
library(lintr)

lint(
  text = "c(seq(length(x)), 1:nrow(y))",
  linters = seq_linter()
)
#> <text>:1:3: warning: [seq_linter] seq(length(...)) is likely to be wrong in the empty edge case. Use seq_along(...) instead.
#> c(seq(length(x)), 1:nrow(y))
#>   ^~~~~~~~~~~~~~
#> <text>:1:19: warning: [seq_linter] 1:nrow(...) is likely to be wrong in the empty edge case. Use seq_len() instead.
#> c(seq(length(x)), 1:nrow(y))
#>                   ^~~~~~~~~
```

<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

IndrajeetPatil:
Btw, shouldn't the message here be different:

``` r
library(lintr)

lint(
  text = "1:nrow(data.frame())",
  linters = seq_linter()
)
#> <text>:1:1: warning: [seq_linter] 1:nrow(...) is likely to be wrong in the empty edge case. Use seq_len() instead.
#> 1:nrow(data.frame())
#> ^~~~~~~~~~~~~~~~~~~~
```

<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Current:

> 1:nrow(...) is likely to be wrong in the empty edge case. Use seq_len() instead.

Expected:

> 1:nrow(...) is likely to be wrong in the empty edge case. Use seq_len(nrow(...)) instead.

Since that's the correct replacement:

```r
1:nrow(data.frame())
#> [1] 1 0

seq_len(nrow(data.frame()))
#> integer(0)
```

MichaelChirico:
> shouldn't the message here be different

it could certainly be improved... I think of it as "good enough" for now since the main distinction is seq_len v seq_along. would be a good follow-up issue/PR though.

tests/testthat/test-seq_linter.R
```diff
@@ -5,6 +5,28 @@ test_that("other : expressions are fine", {
   expect_lint("function(x) { 1:(length(x) || 1) }", NULL, linter)
 })
 
+test_that("seq_len(...) or seq_along(...) expressions are fine", {
+  linter <- seq_linter()
+  expect_lint("function(x) { seq_len(x) }", NULL, linter)
```
MichaelChirico, line 10: could you add tests for other usages of seq() that shouldn't lint? e.g. seq(2, length(x)) and seq(length(x), 2) (since 2:length(x) doesn't lint)

IndrajeetPatil: Damn, that lints:

``` r
library(lintr)

lint(
  text = "function(x) { seq(2, length(x)) }",
  linters = seq_linter()
)
#> <text>:1:15: warning: [seq_linter] seq(2) is likely to be wrong in the empty edge case. Use seq_len(...) instead.
#> function(x) { seq(2, length(x)) }
#>               ^~~~~~~~~~~~~~~~~
```

<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I think the xpath I am using is incorrect. 

It shouldn't be 

`seq()` + anything in between + bad_funcs

But rather

`seq()` + nothing in between + bad_funcs

IndrajeetPatil: Bingo!

``` r
library(lintr)

lint(
  text = "function(x) { seq(2, length(x)) }",
  linters = seq_linter()
)

lint(
  text = "function(x) { seq(length(x), 2) }",
  linters = seq_linter()
)
```

<sup>Created on 2022-07-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/seq_linter.R
```diff
@@ -11,15 +15,25 @@
 seq_linter <- function() {
   bad_funcs <- c("length", "n", "nrow", "ncol", "NROW", "NCOL", "dim")
 
+  # Exact `xpath` depends on whether bad function was used in conjunction with `seq()`
+  bad_func_xpath_with_seq <- glue::glue(
+    "expr[1][SYMBOL_FUNCTION_CALL[text() = 'seq']]/following::expr[1]/expr[SYMBOL_FUNCTION_CALL[ {xp_text_in_table(bad_funcs)} ]]"
```
IndrajeetPatil, line 20: @MichaelChirico I am open to suggestion on how this xpath can be changed to avoid the `line_length_linter` lint.

R/seq_linter.R
```diff
@@ -14,19 +14,22 @@
 #' @export
 seq_linter <- function() {
   bad_funcs <- c("length", "n", "nrow", "ncol", "NROW", "NCOL", "dim")
-  bad_func_xpath <- glue::glue(
+  bad_func_xpath_with_seq <- glue::glue(
+    "expr[1][SYMBOL_FUNCTION_CALL[text() = 'seq']]/following::expr[1]/expr[SYMBOL_FUNCTION_CALL[{xp_text_in_table(bad_funcs)}]]"
```
MichaelChirico: I am usually doing this:

//node
/successor
/successor

IndrajeetPatil: Thanks, done!

IndrajeetPatil:
@MichaelChirico I think I've addressed all of your comments. So this should be all set.

tests/testthat/test-seq_linter.R
```diff
@@ -5,6 +5,32 @@ test_that("other : expressions are fine", {
...
+
+  expect_lint("function(x) { seq_len(x) }", NULL, linter)
+  expect_lint("function(x) { seq_along(x) }", NULL, linter)
+
+  expect_lint("function(x) { seq(2, length(x)) }", NULL, linter)
+  expect_lint("function(x) { seq(length(x), 2) }", NULL, linter)
+})
+
+test_that("finds seq(...) expressions", {
+  linter <- seq_linter()
```
MichaelChirico, line 19: could you add a test for seq(dim(x)[1]) which is how dim would usually be used? it's pretty rare (never seen it before myself), so if the current approach doesn't work, just file a follow-up issue.

but if the current xpath is good enough just add tests to prevent regression

IndrajeetPatil: Oh, wait. This should actually produce a lint:

``` r
seq(dim(data.frame())[1])
#> [1] 1 0
```

<sup>Created on 2022-07-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MichaelChirico: right... don't worry about it though. just please file a follow-up issue

MichaelChirico: and plz revert the incorrect test :)

IndrajeetPatil: Is it okay if I add the correct test but comment it out for now?

IndrajeetPatil: Created issue to track this: https://github.com/r-lib/lintr/issues/1474

MichaelChirico: yep that's fine, though we'll have to be wary of comment_linter

MichaelChirico: see :)

MichaelChirico (approved):

PR #1473, opened 2022-07-25
Build release and development website versions

### LINKED ISSUE #1456: Build release and development versions of package website
We need to start having separate versions of the website for the release and development package versions, so that the users are not confused by why the output they see doesn't match with what's on the website. 

We can only do this after the next release of the package. If we do it now, IINM, the release website will disappear.

We just need to add the following to `_pkgdown.yml`.

```yml
development:
  mode: auto
  version_label: default      # optional
  version_tooltip: "Version"  # optional
```

For more, see:
https://pkgdown.r-lib.org/articles/how-to-update-released-site.html

IndrajeetPatil:
Closes #1456

To be considered for a merge only after the next CRAN release.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -2,9 +2,14 @@ url: https://lintr.r-lib.org
 template:
   bootstrap: 5
 
-  includes:
-    in_header: |
-      <script defer data-domain="lintr.r-lib.org,all.tidyverse.org" src="https://plausible.io/js/plausible.js"></script>
+includes:
+  in_header: |
+    <script defer data-domain="lintr.r-lib.org,all.tidyverse.org" src="https://plausible.io/js/plausible.js"></script>
+
+development:
+  mode: "auto"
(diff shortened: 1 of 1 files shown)


MichaelChirico:
> To be considered for a merge only after the next CRAN release, and before the version is bumped to devel (.9000).

Hey @IndrajeetPatil I was following this note and not reviewing for now, is there a need for review?

IndrajeetPatil:
@MichaelChirico You were absolutely right not to review it, and that't why I also didn't request a review immediately. 

But given the narrow window where this PR would need to be merged, I just wanted to have it pre-reviewed/approved, so that I don't miss that window. :)

MichaelChirico:
we'll just mark it as part of the release process, and Jim will be responsible for merging it in the correct order

jimhester (approved):

jimhester:
Thanks!

PR #1480, opened 2022-07-28
Edits to docs for linters

IndrajeetPatil:
- Where the text description wasn't clear, adds a small example
- Consistent formatting of function calls or names as code in Roxygen docs to distinguish it from text
- Grammar or spelling corrections

### DIFF
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -1,7 +1,7 @@
 #' Require usage of `anyDuplicated(x) > 0` over `any(duplicated(x))`
 #'
 #' [anyDuplicated()] exists as a replacement for `any(duplicated(.))`, which is
-#'   more efficient for simple objects, and is equally efficient in the worst case scenario.
+#'   more efficient for simple objects, and is at worst equally efficient.
 #'   Therefore, it should be used in all situations instead of the latter.
 #'
 #' Also match usage like `length(unique(x$col)) == nrow(x)`, which can
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -1,4 +1,4 @@
-#' Require usage of anyDuplicated() > 0 over any(duplicated(.))
+#' Require usage of `anyDuplicated(x) > 0` over `any(duplicated(x))`
 #'
 #' [anyDuplicated()] exists as a replacement for `any(duplicated(.))`, which is
 #'   more efficient for simple objects, and is equally efficient in the worst case scenario.
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -1,9 +1,8 @@
 #' Require usage of anyDuplicated() > 0 over any(duplicated(.))
 #'
-#' [anyDuplicated()] exists as a replacement for `any(duplicated(.))` which is
-#'   more efficient for simple objects, and in the worst case is the same
-#'   efficiency. Therefore it should be used in all situations instead of the
-#'   latter.
+#' [anyDuplicated()] exists as a replacement for `any(duplicated(.))`, which is
+#'   more efficient for simple objects, and is equally efficient in the worst case scenario.
+#'   Therefore, it should be used in all situations instead of the latter.
 #'
 #' Also match usage like `length(unique(x$col)) == nrow(x)`, which can
 #'   be replaced by `anyDuplicated(x$col) == 0L`.
--- a/R/any_is_na_linter.R
+++ b/R/any_is_na_linter.R
@@ -1,7 +1,7 @@
 #' Require usage of `anyNA(x)` over `any(is.na(x))`
 #'
 #' [anyNA()] exists as a replacement for `any(is.na(x))` which is more efficient
-#'   for simple objects, and has the same efficiency in the worst case scenario.
+#'   for simple objects, and is at worst equally efficient.
 #'   Therefore, it should be used in all situations instead of the latter.
 #'
 #' @evalRd rd_tags("any_is_na_linter")
--- a/R/any_is_na_linter.R
+++ b/R/any_is_na_linter.R
@@ -1,4 +1,4 @@
-#' Require usage of anyNA over any(is.na(.))
+#' Require usage of `anyNA()` over `any(is.na(.))`
 #'
 #' [anyNA()] exists as a replacement for `any(is.na(.))` which is more efficient
 #'   for simple objects, and has the same efficiency in the worst case scenario.
--- a/R/any_is_na_linter.R
+++ b/R/any_is_na_linter.R
@@ -1,6 +1,6 @@
-#' Require usage of `anyNA()` over `any(is.na(.))`
+#' Require usage of `anyNA(x)` over `any(is.na(x))`
 #'
-#' [anyNA()] exists as a replacement for `any(is.na(.))` which is more efficient
+#' [anyNA(x)] exists as a replacement for `any(is.na(x))` which is more efficient
 #'   for simple objects, and has the same efficiency in the worst case scenario.
 #'   Therefore, it should be used in all situations instead of the latter.
 #'
--- a/R/any_is_na_linter.R
+++ b/R/any_is_na_linter.R
@@ -1,8 +1,8 @@
 #' Require usage of anyNA over any(is.na(.))
 #'
 #' [anyNA()] exists as a replacement for `any(is.na(.))` which is more efficient
-#'   for simple objects, and in the worst case is the same efficiency. Therefore
-#'   it should be used in all situations instead of the latter.
+#'   for simple objects, and has the same efficiency in the worst case scenario.
+#'   Therefore, it should be used in all situations instead of the latter.
 #'
 #' @evalRd rd_tags("any_is_na_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
--- a/R/any_is_na_linter.R
+++ b/R/any_is_na_linter.R
@@ -1,6 +1,6 @@
 #' Require usage of `anyNA(x)` over `any(is.na(x))`
 #'
-#' [anyNA(x)] exists as a replacement for `any(is.na(x))` which is more efficient
+#' [anyNA()] exists as a replacement for `any(is.na(x))` which is more efficient
 #'   for simple objects, and has the same efficiency in the worst case scenario.
 #'   Therefore, it should be used in all situations instead of the latter.
 #'
--- a/R/commas_linter.R
+++ b/R/commas_linter.R
@@ -1,7 +1,6 @@
 #' Commas linter
 #'
-#' Check that all commas are followed by spaces (e.g. `c(a = 4, b = 5)`),
-#' but do not have spaces before them (e.g. `c(a = 4 , b = 5)`).
+#' Check that all commas are followed by spaces, but do not have spaces before them.
 #'
 #' @evalRd rd_tags("commas_linter")
 #' @seealso
--- a/R/commas_linter.R
+++ b/R/commas_linter.R
@@ -1,6 +1,7 @@
 #' Commas linter
 #'
-#' Check that all commas are followed by spaces, but do not have spaces before them.
+#' Check that all commas are followed by spaces (e.g. `c(a = 4, b = 5)`),
+#' but do not have spaces before them (e.g. `c(a = 4 , b = 5)`).
 #'
 #' @evalRd rd_tags("commas_linter")
 #' @seealso
--- a/R/conjunct_test_linter.R
+++ b/R/conjunct_test_linter.R
@@ -1,4 +1,4 @@
-#' Force `&&` conditions in `expect_true()`, `expect_false()` to be written separately
+#' Force `&&` conditions in `expect_true()` and `expect_false()` to be written separately
 #'
 #' For readability of test outputs, testing only one thing per call to
 #'   [testthat::expect_true()] is preferable, i.e.,
--- a/R/conjunct_test_linter.R
+++ b/R/conjunct_test_linter.R
@@ -1,4 +1,4 @@
-#' Force && conditions in expect_true(), expect_false() to be written separately
+#' Force `&&` conditions in `expect_true()`, `expect_false()` to be written separately
 #'
 #' For readability of test outputs, testing only one thing per call to
 #'   [testthat::expect_true()] is preferable, i.e.,
--- a/R/cyclocomp_linter.R
+++ b/R/cyclocomp_linter.R
@@ -2,9 +2,8 @@
 #'
 #' Check for overly complicated expressions. See [cyclocomp::cyclocomp()].
 #'
-#' @param complexity_limit expressions with a cyclomatic complexity higher than
-#' this limit are linted, defaults to 15.
-#' See [cyclocomp::cyclocomp()].
+#' @param complexity_limit Maximum cyclomatic complexity, default 15. Expressions more complex
+#' than this are linted. See [cyclocomp::cyclocomp()].
 #' @evalRd rd_tags("cyclocomp_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @importFrom cyclocomp cyclocomp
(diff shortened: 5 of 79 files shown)


R/unneeded_concatenation_linter.R
```diff
@@ -5,7 +5,9 @@
 #' @param allow_single_expression Logical, default `TRUE`. If `FALSE`, one-expression
 #'   usages of `c()` are always linted, e.g. `c(x)` and `c(matrix(...))`. In some such
 #'   cases, `c()` is being used for its side-effect of stripping non-name attributes;
-#'   it is usually preferable to use [as.vector()] to accomplish the same more readably.
+#'   it is usually preferable to use the more readable [as.vector()] instead. But the
+#'   latter may not always be a suitable replacement (e.g., if `c(r6_object)`
+#'   is used to convert an `R6` class instance to a list).
```
IndrajeetPatil, line 10:
``` r
library(R6)

Person <- R6Class("Person", public = list("name" = NA))
Bob <- Person$new()

c(Bob)
#> [[1]]
#> <Person>
#>   Public:
#>     clone: function (deep = FALSE) 
#>     name: NA

as.vector(Bob)
#> Error in as.vector(x, mode): cannot coerce type 'environment' to vector of type 'any'
```

<sup>Created on 2022-07-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

AshesITR: The note could mention `list(env)` as a suitable replacement for `c(env)` explicitly.

IndrajeetPatil: Thanks. Done!

R/any_duplicated_linter.R
```diff
@@ -1,9 +1,8 @@
-#' Require usage of anyDuplicated() > 0 over any(duplicated(.))
+#' Require usage of `anyDuplicated(x) > 0` over `any(duplicated(x))`
 #'
-#' [anyDuplicated()] exists as a replacement for `any(duplicated(.))` which is
-#'   more efficient for simple objects, and in the worst case is the same
-#'   efficiency. Therefore it should be used in all situations instead of the
-#'   latter.
+#' [anyDuplicated()] exists as a replacement for `any(duplicated(.))`, which is
+#'   more efficient for simple objects, and is equally efficient in the worst case scenario.
```
MichaelChirico, line 4: suggested: for simple objects, and is at worst equally efficient.

R/any_is_na_linter.R
```diff
@@ -1,8 +1,8 @@
-#' Require usage of anyNA over any(is.na(.))
+#' Require usage of `anyNA()` over `any(is.na(.))`
```
MichaelChirico, line 1: try to make the style here consistent with `any_duplicated_linter.Rd`, given how similar the linters are. e.g. there you added `anyDuplicated(x)` but not here. Wording below should also be basically the same.

R/commas_linter.R
```diff
@@ -1,6 +1,7 @@
 #' Commas linter
 #'
-#' Check that all commas are followed by spaces, but do not have spaces before them.
+#' Check that all commas are followed by spaces (e.g. `c(a = 4, b = 5)`),
```
MichaelChirico, line 3: I would leave this as a simple sentence, and refer to the style guide link instead. The style guide makes it clearer that the latter example is bad. We could also add more exposition here saying good and bad examples, but that seems redundant to the style guide. WDYT?

R/conjunct_test_linter.R
```diff
@@ -1,4 +1,4 @@
-#' Force && conditions in expect_true(), expect_false() to be written separately
+#' Force `&&` conditions in `expect_true()`, `expect_false()` to be written separately
```
MichaelChirico, line 1: maybe `expect_true() and expect_false()` (i.e., no `,`)?

R/cyclocomp_linter.R
```diff
@@ -2,7 +2,8 @@
 #'
 #' Check for overly complicated expressions. See [cyclocomp::cyclocomp()].
 #'
-#' @param complexity_limit expressions with a cyclomatic complexity higher than this are linted, defaults to 15.
+#' @param complexity_limit expressions with a cyclomatic complexity higher than
```
MichaelChirico, line 5: suggested:

```
@param complexity_limit Maximum cyclomatic complexity, default 15. Expressions more complex than this are linted.
```

R/duplicate_argument_linter.R
```diff
@@ -1,6 +1,9 @@
 #' Duplicate argument linter
 #'
-#' Check for duplicate arguments in function calls.
+#' Check for duplicate arguments in function calls. Some of these cases will be
+#' blocked by R's syntax itself (e.g. `mean(x = 1:5, x = 2:3)`), but even when
+#' such usage *is* allowed, it is rarely desirable and should be used sparingly
```
MichaelChirico, line 5: Not a fan of this wording, given the utility of usage in `mutate()` to sequentially apply transformations to a single column. How about

Some cases are run-time errors (e.g. `mean(x = 1:5, x = 2:3)`), otherwise this linter is used to discourage explicitly providing duplicate names to objects (e.g. `c(a = 1, a = 2)`). Duplicate-named objects are hard to work with programmatically and should typically be avoided.

R/equals_na_linter.R
```diff
@@ -1,6 +1,7 @@
 #' Equality check with NA linter
 #'
-#' Check for `x == NA` and `x != NA`
+#' Check for `x == NA` and `x != NA`. One should instead use `is.na()` function
```
MichaelChirico, line 3: Not a fan of (IMO) clumsy wording like "one should", how about

Such usage is almost surely incorrect -- checks for missing values should be done with `[is.na()]`.

R/expect_identical_linter.R
```diff
@@ -1,6 +1,6 @@
-#' Require usage of expect_identical(x, y) where appropriate
+#' Require usage of `expect_identical(x, y)` where appropriate
 #'
-#' At Google, [testthat::expect_identical()] should be the default/go-to function for
```
MichaelChirico, line 3: "At Google" was intentional here -- I can't find where it was mentioned, but Hadley mentioned this is not really true for testthat in general. Maybe a more general way to phrase this first sentence is:

This linter enforces the usage of [testthat::expect_identical()] as the default expectation for comparisons in a testthat suite.

R/extraction_operator_linter.R
```diff
@@ -1,7 +1,7 @@
 #' Extraction operator linter
 #'
-#' Check that the `[[` operator is used when extracting a single element from an object, not `[` (subsetting) nor `$`
-#' (interactive use).
```
MichaelChirico, line 4: i prefer the original phrasing here

R/fixed_regex_linter.R
```diff
@@ -3,7 +3,7 @@
 #' Invoking a regular expression engine is overkill for cases when the search
 #'   pattern only involves static patterns.
 #'
-#' NB: for `stringr` functions, that means wrapping the pattern in `stringr::fixed()`.
```
MichaelChirico, line 6: i think lower-case is correct, though apparently it's OK to capitalize sometimes:

https://www.grammarly.com/blog/capitalization-after-colons/#:~:text=Capitalization%3A%20First%20Word%20After%20a,it%20begins%20a%20complete%20sentence.

R/missing_argument_linter.R
```diff
@@ -1,6 +1,7 @@
 #' Missing argument linter
 #'
-#' Check for missing arguments in function calls.
+#' Check for missing arguments in function calls (e.g. `stats::median(1:10, na.rm =)`).
```
MichaelChirico, line 3: I would not use a named argument here -- I don't think I saw a single case of this in our code base, whereas there were dozens of usages of the missing argument just being blank (e.g., `stats::median(1:10,)`). So I would put the more common occurrence in the docs.

R/nested_ifelse_linter.R
```diff
@@ -1,15 +1,15 @@
-#' Block usage of nested ifelse() calls
+#' Block usage of nested `ifelse()` calls
 #'
-#' Calling `ifelse` in nested calls is problematic for two main reasons:
+#' Calling `ifelse()` in nested calls is problematic for two main reasons:
```
MichaelChirico, line 3: Make this a link instead? `[ifelse()]`

R/redundant_ifelse_linter.R
```diff
@@ -1,9 +1,9 @@
-#' Prevent ifelse() from being used to produce TRUE/FALSE or 1/0
+#' Prevent `ifelse()` from being used to produce `TRUE/FALSE` or `1/0`
```
MichaelChirico, line 1: `` `TRUE`/`FALSE` `` is better, right?

R/unneeded_concatenation_linter.R
```diff
@@ -5,7 +5,10 @@
 #' @param allow_single_expression Logical, default `TRUE`. If `FALSE`, one-expression
 #'   usages of `c()` are always linted, e.g. `c(x)` and `c(matrix(...))`. In some such
 #'   cases, `c()` is being used for its side-effect of stripping non-name attributes;
-#'   it is usually preferable to use [as.vector()] to accomplish the same more readably.
+#'   it is usually preferable to use the more readable [as.vector()] instead. But the
```
MichaelChirico, line 8: Suggestion:

`as.vector()` is not always preferable, for example with environments (especially, R6 objects), in which case `list()` is the better alternative.

(note that what you've said is not exactly right -- `c(r6_object)` doesn't do any conversion, it just plops the object into a list, like `list(r6_object)` is doing, i.e., `identical(r6_object, c(r6_object)[[1]])`)

MichaelChirico (commented):
Wow, lots of work here, thanks!

R/commas_linter.R
IndrajeetPatil: I am going to revert this for now, but, TBH, I think each linter should have an examples section with at least one example of code that does produce lint and another that doesn't.

I think the style guide is good to provide the underlying rationale for the lint (provided that the reader actually clicks on that link!), but it doesn't lead to much exploration on the reader's part. Giving concrete examples means that the reader can just copy-paste examples and start toying around with them to see what does and what doesn't lint.

If you agree, I can create a new documentation issue to add examples, and can also probably work on it.

R/any_duplicated_linter.R
IndrajeetPatil: Done.

R/any_is_na_linter.R
IndrajeetPatil: Yeah, good catch. I have changed this to be more consistent.

R/conjunct_test_linter.R
IndrajeetPatil: Definitely. Changed to the suggested phrasing.

R/cyclocomp_linter.R
IndrajeetPatil: Changed to the suggested phrasing.

R/duplicate_argument_linter.R
IndrajeetPatil: Good point. I have reverted this whole change.

IndrajeetPatil: And accepted the suggested text instead.

R/equals_na_linter.R
IndrajeetPatil: Thanks. Changed to the suggested phrasing.

R/expect_identical_linter.R
IndrajeetPatil: Ah, I see. 

Okay, I have changed the first sentence to the one you have suggested.

R/extraction_operator_linter.R
IndrajeetPatil: Reverted.

R/fixed_regex_linter.R
IndrajeetPatil: Thanks for checking!

I was mostly motivated by internal consistency here :)

Screenshot 2022-07-31 at 08 16 13

R/missing_argument_linter.R
IndrajeetPatil: Good point. Using the most common occurrence example now.

R/nested_ifelse_linter.R
IndrajeetPatil: Done.

R/redundant_ifelse_linter.R
IndrajeetPatil: Definitely! Modified it.

R/unneeded_concatenation_linter.R
IndrajeetPatil: Thanks for the clarification.

I am using the suggested phrasing now.

R/commas_linter.R
MichaelChirico: That's definitely worthy of its own issue. My main concern is about duplicate documentation (if the style guide changes, we need to remember update the docs), but not all of our linters are tidyverse guide enforcements, and those would definitely benefit from a good/bad section.

R/any_is_na_linter.R
```diff
@@ -1,8 +1,8 @@
-#' Require usage of anyNA over any(is.na(.))
+#' Require usage of `anyNA(x)` over `any(is.na(x))`
 #'
-#' [anyNA()] exists as a replacement for `any(is.na(.))` which is more efficient
-#'   for simple objects, and in the worst case is the same efficiency. Therefore
-#'   it should be used in all situations instead of the latter.
+#' [anyNA()] exists as a replacement for `any(is.na(x))` which is more efficient
+#'   for simple objects, and has the same efficiency in the worst case scenario.
```
MichaelChirico, line 4: parallel phrasing with `any_duplicated_linter`: "is at worst equally efficient"

R/nested_ifelse_linter.R
```diff
@@ -1,15 +1,15 @@
...
+#' Block usage of nested `ifelse()` calls
 #'
-#' Calling `ifelse` in nested calls is problematic for two main reasons:
+#' Calling [ifelse()] in nested calls is problematic for two main reasons:
 #'   1. It can be hard to read -- mapping the code to the expected output
 #'      for such code can be a messy task/require a lot of mental bandwidth,
 #'      especially for code that nests more than once
-#'   2. It is inefficient -- `ifelse` can evaluate _all_ of its arguments at
-#'      both yes and no (see https://stackoverflow.com/q/16275149); this issue
+#'   2. It is inefficient -- [ifelse()] can evaluate _all_ of its arguments at
```
MichaelChirico, line 7: I think only linking the first usage in a section is the more standard practice... `` `ifelse()` `` instead?

R/fixed_regex_linter.R
MichaelChirico: indeed! let's change that one, too :)

R/nested_ifelse_linter.R
IndrajeetPatil: Done.

R/any_is_na_linter.R
IndrajeetPatil: Done.

R/fixed_regex_linter.R
IndrajeetPatil: Okay, both of them are now lower-case, and, thus, consistent with each other.

R/commas_linter.R
IndrajeetPatil: TBH, I don't see the style guide changing that often, so the examples being out of phase with the guide should happen rarely.

I will create a new issue to track this.

IndrajeetPatil: Done: https://github.com/r-lib/lintr/issues/1492

IndrajeetPatil:
@MichaelChirico All review comments are resolved here now. Let me know if you had any additional comments/concerns.

IndrajeetPatil:
I am waiting for this to be merged before I can work on #1492; I think there aren't any unresolved review comments here.

MichaelChirico (approved):
Thanks!

PR #1483, opened 2022-07-28
Update website reference index

IndrajeetPatil:


_pkgdown.yml
```diff
@@ -21,27 +21,29 @@ reference:
...
 - title: Individual linters
   contents:
-  - all_undesirable_functions
   - ends_with("linter")
 
-- title: Customer linters
+- title: Lists of defaults
+  contents:
+  - all_undesirable_functions
+  - starts_with("default_")
```
IndrajeetPatil, line 29: So that all named lists are together in one section:

Screenshot 2022-07-29 at 00 31 15

_pkgdown.yml
```diff
@@ -21,27 +21,29 @@ reference:
...
+
+- title: Utilities
   contents:
   - Linter
-  - expect_lint
-  - expect_lint_free
+  - starts_with("expect_")
```
MichaelChirico, line 34: this will include a bunch of linters, right? `expect_identical_linter()` and friends...

IndrajeetPatil: Ah, yes. I have changed it to `starts_with("expect_lint")`, which should avoid this and select the only two we have in mind here:

``` r
library(tidyselect)

starts_with(
  "expect_lint",
  vars = c("expect_lint", "expect_lint_free", "expect_identical_linter")
)
#> [1] 1 2
```

<sup>Created on 2022-07-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MichaelChirico: given that there are only two, I'd rather just write them out. it's better to be explicit unless it's really saving us a lot of typing/maintenance/upkeep

IndrajeetPatil: Okay, sounds reasonable. I've reverted this change.

_pkgdown.yml
```diff
@@ -21,10 +21,14 @@ reference:
 
 - title: Individual linters
   contents:
-  - all_undesirable_functions
   - ends_with("linter")
 
-- title: Customer linters
+- title: Lists of defaults
```
MichaelChirico, line 26: I think we can do better on the name here but nothing coming straight to mind... mind having another go at it? I'll also try with a fresh mind tomorrow.

_pkgdown.yml
```diff
@@ -34,14 +38,13 @@ reference:
   - use_lintr
   - xml_nodes_to_lints
 
-- title: internal
+- title: Internal
```
MichaelChirico, line 41: internal also doesn't quite feel right... Meta-tooling?

IndrajeetPatil: Note that this is just to mark functions which are internal and should not to be displayed in [the reference index](https://lintr.r-lib.org/reference/index.html).

And, as expected, it doesn't show up in the TOC or in the index:

Screenshot 2022-07-31 at 09 13 57

IndrajeetPatil: But I have still renamed that section to `Meta-tooling`.

MichaelChirico: oh, I see. indeed I don't see any reason to hide exported functions from the site. thanks!

_pkgdown.yml
IndrajeetPatil: A few options from my side:

- Defaults
- Default options
- Customizable components

MichaelChirico: How about "Common default configurations"?

IndrajeetPatil: Sounds good. Changed to that.

MichaelChirico (approved):

PR #1501, opened 2022-08-21
Fix typo

eitsupi:


R/redundant_ifelse_linter.R
```diff
@@ -2,7 +2,7 @@
 #'
 #' Expressions like `ifelse(x, TRUE, FALSE)` and `ifelse(x, FALSE, TRUE)` are
 #'   redundant; just `x` or `!x` suffice in R code where logical vectors are a
-#'   core data structure. `ifelse(x, 1, 0)` is also `as.numeric(x)`, but even
+#'   core data structure. `ifelse(x, 1, 0)` is also `as.logical(x)`, but even
```
IndrajeetPatil, line 5: I think the docs are correct as they are:

``` r
x <- (2 > 1)
ifelse(x, 1, 0)
#> [1] 1
as.numeric(x)
#> [1] 1
as.logical(x)
#> [1] TRUE
```

<sup>Created on 2022-08-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

eitsupi: Oh, I understand. I definitely didn't understand the context!
This seems to be correct.
Thanks for your quick replay.

PR #1506, opened 2022-08-24
`object_length_linter()` works when no exports

### LINKED ISSUE #1503: lintr::object_length_linter() does not work in the case of dependencies with no exports
Test package:

https://github.com/Bisaloo/testwpp

Error: 

> .Error in data.frame(pkg = ns, fun = getNamespaceExports(ns), stringsAsFactors = FALSE) : 
  arguments imply differing number of rows: 1, 0

This happens because the wpp2017 package does not export any functions. It only provides data. If I understand correctly, this means that the following line

https://github.com/r-lib/lintr/blob/f6cf291a9b28b8c0bf88928aa5167b0396ad687d/R/namespace.R#L29

will try to create a data.frame with one element in the first column and 0 in the second, hence the error.

IndrajeetPatil:
Closes #1503

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -13,7 +13,7 @@ namespace_imports <- function(path = find_package()) {
 }
 
 # this loads the namespaces, but is the easiest way to do it
-# test package availablity to avoid failing out as in #1360
+# test package availability to avoid failing out as in #1360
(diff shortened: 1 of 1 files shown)


IndrajeetPatil:
First need to make sure that this is reproducible. Will open a different PR if it is.

PR #1509, opened 2022-08-25
Handle dependencies with no exports

### LINKED ISSUE #1503: lintr::object_length_linter() does not work in the case of dependencies with no exports

IndrajeetPatil:
Closes #1503

### DIFF
--- a/R/namespace.R
+++ b/R/namespace.R
@@ -13,7 +13,7 @@ namespace_imports <- function(path = find_package()) {
 }
 
 # this loads the namespaces, but is the easiest way to do it
-# test package availablity to avoid failing out as in #1360
+# test package availability to avoid failing out as in #1360
 #   typically, users are running this on their own package directories and thus
 #   will have the namespace dependencies installed, but we can't guarantee this.
 safe_get_exports <- function(ns) {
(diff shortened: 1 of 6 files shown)


R/namespace.R
```diff
@@ -36,14 +41,18 @@ empty_namespace_data <- function() {
...
-    function(i) {
-      fun_obj <- get(ns_imports$fun[i], envir = asNamespace(ns_imports$pkg[i]))
-      is.function(fun_obj) && is_s3_generic(fun_obj)
-    },
-    logical(1L)
-  )
+  is_generic <- logical(0L)
+
+  if (!is.null(ns_imports)) {
+    is_generic <- vapply(
```
MichaelChirico, line 47: how about just using seq_len(NROW(ns_imports))?

MichaelChirico: alternatively, return a 0-row data.frame instead of NULL

IndrajeetPatil: That's cool! Thanks. 

I also added a comment; lemme know if it looks okay to you.

MichaelChirico: I think too verbose tbh

# NROW for the NULL case of 0-export dependencies, #1503

oughta do

MichaelChirico (commented):
Now for the fun part -- could you please add a test?

FWIW it may be possible to use base datasets package

IndrajeetPatil:
> FWIW it may be possible to use base datasets package

Sounds good. Will try to add a test.

IndrajeetPatil:
@MichaelChirico I confirmed that the test fails on `main` branch.

tests/testthat/test-object_length_linter.R
```diff
@@ -43,7 +43,22 @@ test_that("lints S3 generics correctly", {
 
 test_that("object_length_linter won't fail if an imported namespace is unavailable", {
   expect_length(
-    lint_package(test_path("dummy_packages", "missing_dep"), linters = object_length_linter(), parse_settings = FALSE),
+    lint_package(
+      test_path("dummy_packages", "missing_dep"),
+      linters = object_length_linter(),
```
MichaelChirico, line 48: the original issue is about object_usage_linter, right?

IndrajeetPatil: No, it's about `object_length_linter()`: https://github.com/r-lib/lintr/issues/1503

PR #1511, opened 2022-08-25
Tidy up `NEWS` file

IndrajeetPatil:
Via https://github.com/lorenzwalthert/stylermd

New version of the `NEWS` file: https://github.com/r-lib/lintr/blob/tidy_news_file/NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,535 +1,961 @@
 # lintr (development version)
 
-* `modify_defaults()` no longer uses the mistaken `"lintr_function"` S3 class, instead applying the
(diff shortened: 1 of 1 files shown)


MichaelChirico:
could you summarize what changes are being made here?

I don't really see a need to do this (one more purely stylistic thing to maintain), but maybe there's a smaller part that's useful?

cc @AshesITR

jimhester:
I don't think this is a good idea, as it destroys / obscures the history of the news file. In general I prefer the style of maintaining a sentence per line and using soft wrapping if needed in your editor for markdown prose.

PR #1519, opened 2022-08-29
Check on older R versions

IndrajeetPatil:
Since we support `R (≥ 3.2)`


IndrajeetPatil:
Hmm, strange. GHA fails on `3.3` and `3.2` with the error:

> package ‘pak’ is not available

But `{pak}` supports `R (≥ 3.2)`, so not sure what's going on.

PR #1521, opened 2022-08-29
More clean up for tests

IndrajeetPatil:
- DRY in tests for `consecutive_stopifnot_linter()`
- DRY in tests for `test-conjunct_test_linter()`
- Restructuring `closed_curly_linter()` tests to improve readability and consistency with other test files

### DIFF
--- a/tests/testthat/test-conjunct_test_linter.R
+++ b/tests/testthat/test-conjunct_test_linter.R
@@ -1,6 +1,6 @@
-linter <- conjunct_test_linter()
-
 test_that("conjunct_test_linter skips allowed usages of expect_true", {
+  linter <- conjunct_test_linter()
+
   expect_lint("expect_true(x)", NULL, linter)
   expect_lint("testthat::expect_true(x, y, z)", NULL, linter)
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-conjunct_test_linter.R
```diff
@@ -1,107 +1,72 @@
+linter <- conjunct_test_linter()
```
MichaelChirico, line 1: this is bad IMO. this object will persist to other test files in the suite if it's not captured inside a `test_that()`. that may be OK if it's a rare variable name / not referenced elsewhere, but `linter` is used in almost all test files. Thus this is prone to errors, and really-hard-to-debug issues as each test file is no longer hermetic.

MichaelChirico:
TBH I'm not sure the utility of changes to do `lintr <- foo(); msg <- "..."`. There are some drawbacks to that approach (user-friendliness depends on the editor you're using) -- it's not an unambiguous best practice we're choosing here.

I think if we're refactoring the tests _while_ they're being touched for other reasons, that's OK. Otherwise, I don't see a reason for a clean-up like this at the moment. I'm going to close, but do thank you @IndrajeetPatil for the effort here... if you want to invest effort in cleaning up the test suite (which is definitely needed), I think it would be better spent elsewhere.

PR #1522, opened 2022-08-29
GHA workflow `lint-changed-files` fails when a new lint is found in changed files

### LINKED ISSUE #1489: Shouldn't the "lint-changed-files" GHA fail for PRs that add new lints?
Currently, when working on a PR, the contributor needs to keep manually checking if they have added any new lints. 
Even if they check that they haven't, something might change during code review, and a lint might be added. Now if they forget to re-check, the reviewer needs to point this out, which is additional work for the reviewer.

So I was wondering if we can instead modify the "lint-changed-files" GHA so that it **fails** when any new lints are detected. Build failures will ensure that neither the contributor nor the reviewer will have to do this manually, and it'd be hard to miss the newly introduced lint. WDYT?

IndrajeetPatil:
Closes #1489

This seems to be working: https://github.com/r-lib/lintr/runs/8070424277?check_suite_focus=true


MichaelChirico (approved):
Awesome!

PR #1523, opened 2022-08-29
`object_usage_linter()` sets `useFancyQuotes=FALSE`

### LINKED ISSUE #1498: `object_usage_linter()` should set `useFancyQuotes=FALSE`
Compare:

```r
str <- "foo <- function() {\n  x\n}\n"

options(useFancyQuotes = FALSE)
lintr::lint(text = str, linters = lintr::object_usage_linter())
# <text>:2:3: warning: [object_usage_linter] no visible binding for global variable 'x'.
#   x
#   ^
options(useFancyQuotes = TRUE)
lintr::lint(text = str, linters = lintr::object_usage_linter())
# <text>:2:3: warning: [object_usage_linter] no visible binding for global variable ‘x’.
#   x
#   ^
```

I think we should follow suit of `testthat` and turn off quotes here -- makes the lint appear differently depending on global options & it's harder to manipulate the message with non-ASCII characters.

The fix is pretty trivial to implement; filing here first as a gut check that this is the right approach.

IndrajeetPatil:
Closes #1498

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -198,12 +198,17 @@ parse_check_usage <- function(expression, known_used_symbols = character(), decl
     vals[[length(vals) + 1L]] <<- x
   }
 
-  try(codetools::checkUsage(
-    expression,
-    report = report,
-    suppressLocalUnused = known_used_symbols,
-    suppressUndefined = declared_globals
-  ))
+  withr::with_options(
+    list(useFancyQuotes = FALSE),
+    code = {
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Marking for review by @AshesITR given his feedback on the issue

AshesITR:
Still not 100% convinced this is worth it but if you two agree, I'm fine with the change.

AshesITR (approved):

PR #1526, opened 2022-08-30
Export get_r_string

### LINKED ISSUE #1493: Export get_r_string()?
Is there any reason not to export this helper? Seems useful to share this logic with downstreams trying to write their own XPath-based linters.

MichaelChirico:
Closes #1493 

I think we have enough indirect tests of the function for now, so didn't add anything new.


R/utils.R
```diff
@@ -210,9 +210,21 @@ release_bullets <- function() {
...
+#'  so whenever your linter logic requires testing a `STR_CONST` node's value, use this
+#'  function.
+#' NB: this is also properly vectorized on `s`, and accepts a variety of inputs. Empty inputs
+#'  will become `NA` outputs, which helps ensure that `length(get_r_string(s)) == length(s)`.
+#'
+#' @param s An input string or strings. If `s` is an `xml_node` or `xml_nodeset` and `xpath` is `NULL`,
+#'   extract its string value with [xml2::xml_text()]. If `s` is an `xml_node` or `xml_nodeset`
+#'   and `xpath` is specified, it is extracted with [xml2::xml_find_chr()].
+#' @param xpath An XPath, passed on to [xml2::xml_find_chr()] after wrapping with `string()`.
+#' @export
```
IndrajeetPatil, line 227: Now that this is being exported, should we add an example to the docs?

MichaelChirico: I agree, but off the top of my head it's a bit of a pain to construct examples (or at least, I'll have to sit down with this which I won't have time for quickly)...

Let's mark that as a follow-up bug for before release, but get this into `main` in the meantime?

IndrajeetPatil: Sounds good to me!

MichaelChirico: #1553

IndrajeetPatil (approved):

IndrajeetPatil (approved):
Thanks!

PR #1528, opened 2022-09-01
Add sort_linter()

### LINKED ISSUE #1512: New linter: recommend `sort(x)` over `x[order(x)]`
I think it would be nice to have a linter to recommend using `sort(x)` to sort a vector instead of `x[order(x)]`:

- it is more readable
- it is faster, especially for longer vectors

``` r
library(microbenchmark)

x <- sample(LETTERS, 1e3, replace = TRUE)

microbenchmark(
  sort(x),
  x[order(x)],
  check = "identical"
)
#> Unit: microseconds
#>         expr     min       lq     mean   median       uq      max neval
#>      sort(x) 529.036 544.6405 585.4106 559.5975 597.4515  907.008   100
#>  x[order(x)] 811.534 836.0540 899.0072 865.4155 904.8205 1453.345   100

x_long <- sample(LETTERS, 1e5, replace = TRUE)

microbenchmark(
  sort(x_long),
  x_long[order(x_long)],
  check = "identical"
)
#> Unit: milliseconds
#>                   expr       min        lq      mean    median        uq
#>           sort(x_long)  59.78242  61.85322  64.85112  63.69153  66.59794
#>  x_long[order(x_long)] 150.68092 160.44464 168.89652 164.85925 175.08799
#>        max neval
#>   86.98432   100
#>  207.92772   100
```

<sup>Created on 2022-08-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

## Possible issues

`sort()` and `order()` are both generic and it is theoretically possible that some packages only provide S3 methods for one of them. In this case, they could not be strictly equivalent. However, I would say it's likely a bug in the package.

Bisaloo:
Fix #1512 

I'm happy to receive feedback on this PR and revisit as much as needed. I'm not super familiar with XPath so I did the best I could.

Also, should I add myself as a contributor?


MichaelChirico:
> Also, should I add myself as a contributor?

So far, the Contributor list is only recording folks with a "substantial" contribution history

R/sort_linter.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' `sort()` is the dedicated option to sort a list or vector. It is more legible
+#' and around twice as fast as `.[order(.)]`, with the gap in performance
+#' growing with the vector size.
+#'
+#' @evalRd rd_tags("sort_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+sort_linter <- function() {
+  xpath <- "//expr[
```
MichaelChirico, line 11: It's better to use `//OP-LEFT-BRACKET` (or possibly `//SYMBOL_FUNCTION_CALL`? the goal is to first narrow down to the least common node; I guess the difference between `SYMBOL_FUNCTION_CALL` and `OP-LEFT-BRACKET` will not matter as much as avoiding `//expr`); see #1358. Try re-expression the path that way. (and sorry that the codebase is still full of misleading examples for you to imitate :) )

R/sort_linter.R
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+
+    xml_nodes_to_lints(
+      bad_expr,
+      source_expression = source_expression,
+      lint_message = "sort(x) is better than x[order(x)].",
```
MichaelChirico, line 35: The official message should include a strict equivalent like `sort(x, na.last = TRUE)`.

Or perhaps better yet, we should recommend setting `na.last` to whatever's set in `order()` itself. If `na.last` is not provided, recommend `sort(x, na.last = TRUE)`. If linting `x[order(x, na.last = X)]`, we recommend `sort(x, na.last = X)`.

tests/testthat/test-sort_linter.R
```diff
@@ -0,0 +1,21 @@
...
+})
+
+
+test_that("sort_linter blocks simple disallowed usages", {
+  lint_message <- rex::rex("sort(x) is better than x[order(x)].")
+
+  expect_lint("x[order(x)]", lint_message, sort_linter())
+
+  # Works with extra args in order()
+  expect_lint("x[order(x, decreasing = TRUE)]", lint_message, sort_linter())
```
MichaelChirico, line 17: This is a good point too -- our recommendation should more broadly endeavour to match other arguments in the shared signatures of `sort()` and `order()`, namely `method`, `decreasing`, and `na.last`.

Bisaloo: How should I go about this? I tried to keep using the xmlparsedata output but I'm not sure it's the best way:

```r
extra_args <- xml2::xml_text(xml2::xml_find_all(bad_expr, "//SYMBOL_SUB"))
extra_arg_values <- xml2::xml_text(xml2::xml_find_all(bad_expr, "//SYMBOL_SUB/following-sibling::expr"))
```

Two problems:

- this wouldn't work with unnamed arguments as far as I can tell
- since the arg name and the arg value are captured separately, I would have to reconstruct the call with `paste()` or `glue()`

An alternative would be to start from `xml2::xml_text(bad_expr)`. It returns something like `order(x,decreasing=TRUE,method='radix')` with the current XPath. I would then have to extract the arguments and values so it seems to defeat to purpose of using xmlparsedata earlier.

Bisaloo: I think I'm not clear on what you're asking exactly:

- should I only return a warning when the `order()` call includes one of `method`, `decreasing`, `na.last`?
- should I customize the warning message to match the extra arguments in the call? (kind of like what you describe in https://github.com/r-lib/lintr/pull/1528#discussion_r973968141 but for all arguments)

MichaelChirico: Thanks for your work here Hugo. I our north star is we need to recommend a correct alternative in the lint message.

At a minimum, I think that means being careful about `na.last`, since `sort()` and `order()` differ on their defaults.

(I don't have R in front of me, but I seem to recall all the other arguments share the same default, so it's not as important to capture that in the recommendation -- doing so can be an aspirational follow-up feature).

More specifically addressing your questions
 - No, I think the logic to recommend `sort()` when user is doing `x[order(x)]` applies regardless of the other arguments. The only caveat is that they need to be careful about `na.last` when making the switch.
 - If you can work out how to do so, that's great. It's much friendlier to see a lint message that closely matches your actual code. However, it's not required for approving the PR, so don't bend over backward to accomplish it, we can put it in the TODO queue for future work.

Bisaloo: Thanks for your answer. It clears things up for me.

It should be implemented in [`28a3b51` (#1528)](https://github.com/r-lib/lintr/pull/1528/commits/28a3b5143b39ac5d32cab58e461b7187b2a87ff2).

MichaelChirico:
please update to `main` and re-document to resolve the conflict, thanks!

R/sort_linter.R
```diff
@@ -0,0 +1,86 @@
+#' Require usage of `sort()` over `.[order(.)]`
+#'
+#' `sort()` is the dedicated option to sort a list or vector. It is more legible
```
MichaelChirico, line 3: Link here: `[sort()]`

R/sort_linter.R
```diff
@@ -0,0 +1,86 @@
...
+#' and around twice as fast as `.[order(.)]`, with the gap in performance
+#' growing with the vector size.
```
MichaelChirico, line 5: Something worth mentioning here, a use case like:

```r
y <- y[order(x)]
x <- x[order(x)]
```

Where we are sorting multiple objects according to the order in one "master" reference. In that case, the recommendation is something like:

```r
index <- order(x)
y <- y[index]
x <- x[index]

# above is better than
y <- y[order(x)]
x <- sort(x)
```

WDYT? Should we make some reference to this use case in the lint message too?

R/sort_linter.R
```diff
@@ -0,0 +1,86 @@
...
+#' lint(
+#'   text = "y[order(x)]",
+#'   linters = sort_linter()
+#' )
+#'
+#' @evalRd rd_tags("sort_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+sort_linter <- function() {
+  xpath <- "//OP-LEFT-BRACKET/following-sibling::expr[
```
MichaelChirico: can we add a specific index to any of these `expr`? e.g. `following-sibling::expr[1]`?

R/sort_linter.R
```diff
@@ -0,0 +1,89 @@
...
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+
+    var <- xml2::xml_text(xml2::xml_find_first(bad_expr, ".//SYMBOL"))
+
+    args <- ".//SYMBOL_SUB[text() = 'method' or
```
MichaelChirico, line 58: for non-trivial XPaths, we usually put them in the factory environment (alongside `xpath` above)

R/sort_linter.R
```diff
@@ -0,0 +1,89 @@
...
+                           text() = 'decreasing' or
+                           text() = 'na.last']"
+
+    arg_names <- xml2::xml_text(xml2::xml_find_all(bad_expr, args))
```
MichaelChirico, line 62: I think this will run into trouble on expressions with `>1` lint, try an example like:

```r
c(
  x[order(x, decreasing = TRUE)],
  y[order(y, decreasing = TRUE, na.last = FALSE)]
)
```

because `xml_find_all()` can return any number of results per input, there's no way to may between input & output when `length(bad_expr) > 1`

R/sort_linter.R
```diff
@@ -0,0 +1,89 @@
...
+    arg_values <- xml2::xml_text(
+      xml2::xml_find_all(bad_expr, glue::glue("{args}/following-sibling::expr[1]"))
```
MichaelChirico, line 64: similar to comment above. `glue()` in particular we like to run in the factory environment -- the linter gets evaluated once for `length(expressions)+1` times for every file, so `glue()` adds a bit of overhead that's preferable to avoid

R/sort_linter.R
```diff
@@ -0,0 +1,89 @@
...
+  Linter(function(source_expression) {
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+
+    var <- xml2::xml_text(xml2::xml_find_first(bad_expr, ".//SYMBOL"))
```
MichaelChirico, line 56: I think this will not work on a case like:

```r
foo(x)[order(foo(x))]
```

And there need not even be a `SYMBOL`:

```r
foo()[order(foo())]
```

inst/lintr/linters.csv
```diff
@@ -66,6 +66,7 @@ semicolon_linter,style readability default configurable
 semicolon_terminator_linter,style readability deprecated configurable
 seq_linter,robustness efficiency consistency best_practices default
 single_quotes_linter,style consistency readability default
+sort_linter,style readability best_practices efficiency
```
MichaelChirico, line 69: I believe `style` is reserved for things in the official Tidyverse guide, the other tags look fine though :)

tests/testthat/test-sort_linter.R
```diff
@@ -0,0 +1,54 @@
...
+
+  # If another function is intercalated, don't fail
+  expect_lint("x[c(order(x))]", NULL, sort_linter())
+})
+
+
+test_that("sort_linter blocks simple disallowed usages", {
+  lint_message <- "sort\\(.*\\) is better than"
+
+  expect_lint("x[order(x)]", lint_message, sort_linter())
```
MichaelChirico, line 14: elsewhere we've adopted a "style" of assigning the linter for re-use:

```r
linter <- sort_linter()

expect_lint(..., linter)
```

tests/testthat/test-sort_linter.R
```diff
@@ -0,0 +1,54 @@
...
+
+  # ...even in disorder
+  expect_lint("x[order(decreasing = TRUE, x)]", lint_message, sort_linter())
+})
+
+test_that("sort_linter produces customized warning message", {
+
+  expect_lint(
+    "y[order(y)]",
+    "sort\\(y, na\\.last = TRUE\\) is better than y\\[order\\(y\\)\\]\\.",
```
MichaelChirico, line 27: prefer `rex::rex()` for readability to avoid escaping everything :)

tests/testthat/test-sort_linter.R
```diff
@@ -0,0 +1,54 @@
...
+  expect_lint(
+    "y + x[order(x)]",
+    "sort\\(x\\, na\\.last = TRUE\\) is better than x\\[order\\(x\\)\\]\\.",
+    sort_linter()
+  )
+
+  # Default na.last = TRUE is overwritten if na.last is already provided
+  expect_lint(
+    "x[order(x, na.last = FALSE)]",
+    paste("sort\\(x, na\\.last = FALSE\\) is better than",
```
MichaelChirico, line 41: this can be in one line -- this package uses 120 as the max line width

MichaelChirico:
Looking great! Thanks again for your work here. Getting much closer to the finish line! 🏁

R/sort_linter.R
Bisaloo: I've moved `glue()` in the factory environment.

Not sure about `xml_find_all()` since it uses `bad_expr` which is only defined later.

R/sort_linter.R
Bisaloo: Yes, you're right and I don't know how to resolve this :disappointed:

R/sort_linter.R
Bisaloo: You're right, I've added an example but I believe the lint message is already complicated enough so we should not add more information.

R/sort_linter.R
MichaelChirico: Could you add some failing tests that should pass if this is resolved?

I'll have a go at fixing it, but it would help a lot to start with a more tangible target.

Bisaloo: I've added a failing test in 266ddb65ec2ec90552f07b0f49f078fe0b48dc2b. Thanks!

Bisaloo: Should be fixed by [`64a88b4` (#1528)](https://github.com/r-lib/lintr/pull/1528/commits/64a88b49c24d72afcf84fd2e3ea758d9f7ad1b99).

Bisaloo:
Oops, sorry for the spam, I was trying to request another round of review from both reviewers but looks like I can only select one at a time, hence the confusion :disappointed:

AshesITR (approved):

PR #1534, opened 2022-09-13
Get rid of version label and tooltip

IndrajeetPatil:
This is the reason why the package version in navigation bar looks so dull:

Screenshot 2022-09-13 at 16 09 16

It's either this, or we add the following to `pkgdown/extra.scss`:

```css
.nav-text.text-default.me-auto {
    color: rgba(0,0,0,0.9) !important;
}
```

I prefer the first solution, which I have implemented in this PR. No strong opinions, though. Happy to also add the aforementioned CSS.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -8,8 +8,6 @@ template:
 
 development:
   mode: "auto"
-  version_label: "default"
-  version_tooltip: "Version"
 
 reference:
 - title: Linting


IndrajeetPatil:
Sure, with `pkgdown::build_site()`, this is what I see locally for the development version:

Screenshot 2022-09-19 at 09 26 36

MichaelChirico:
OK great, that definitely looks better.

The live site currently uses the released version, will this PR make it so the site shows the dev version?

IndrajeetPatil:
Nope. 

The stable and development versions have different URLs, and current PR won't change anything about this:

- https://lintr.r-lib.org
- https://lintr.r-lib.org/dev/

It just changes the colour of the tooltip users see when their mouse hovers over the package version in the navigation bar.

MichaelChirico:
OK great, thanks for clarifying!

MichaelChirico (approved):

PR #1541, opened 2022-09-20
New unnecessary_lambda_linter

### LINKED ISSUE #1531: New linter: discourage the use of anonymous functions for trivial cases
# Example

## Discouraged

```r
lapply(l, function(x) mean(x)))
```

## Recommended

```r
lapply(l, mean)
```

# Scope

- to limit the risk of false positive, I'm thinking to limit this to the `apply()` family function (+ `replicate()`), at least as a first step. Or do you think it wouldn't be much more difficult and riskier to do this in all contexts?
- ideally, this should detect also both tidyverse (`~ mean(.x)`) and base (`\(x) mean(x)`) lambda functions
- I'm thinking to limit it to cases with a single argument, i.e. exclude:

  ```r
  lapply(l, function(x) mean(x, na.rm = TRUE))
  ``` 

  even though the anonymous function could be removed here as well by forwarding the `...` because users might find it clearer to not forward them.

MichaelChirico:
Part of #884. Closes #1531.

cc @Bisaloo for viz.

WIP: I will have to add extra logic for customizing the lint message. Filing now for early feedback.


MichaelChirico:
Looks like we'll need a new name -- 37 characters is too long 🫣

open to ideas

R/unnecessary_anonymous_function_linter.R
```diff
@@ -0,0 +1,94 @@
...
+#' Using an anonymous function in, e.g., [lapply()] is not always necessary,
+#'   e.g. `lapply(DF, sum)` is the same as `lapply(DF, function(x) sum(x))` and
+#'   the former is more readable.
+#'
+#' @evalRd rd_tags("unnecessary_anonymous_function_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+unnecessary_anonymous_function_linter <- function() {
+  # TODO(michaelchirico): is there any way to extend this to all (base?) functions
+  #   accepting FUN as an argument? Reduce, Filter, mapply, outer, combn, ...
```
IndrajeetPatil, line 12:
``` r
library(parallel)

is_arg_present <- function(.f, arg_name) {
  fun_args <- as.list(formals(.f))
  is_arg_present <- any(names(fun_args) == arg_name)
  return(is_arg_present)
}

funs_with_arg_name <- function(env, arg_name) {
  objs <- mget(ls(env, all = TRUE), inherits = TRUE)
  funs <- Filter(is.function, objs)
  names(purrr::keep(funs, is_arg_present, arg_name = arg_name))
}

purrr::map(
  list("package:base", "package:utils", "package:stats", "package:graphics", "package:methods", "package:parallel"),
  funs_with_arg_name,
  arg_name = "FUN"
) |>
  purrr::simplify()
#>  [1] ".kronecker"           ".mapply"              "apply"               
#>  [4] "by"                   "by.data.frame"        "by.default"          
#>  [7] "eapply"               "kronecker"            "lapply"              
#> [10] "mapply"               "match.fun"            "outer"               
#> [13] "sapply"               "sweep"                "tapply"              
#> [16] "vapply"               "Vectorize"            "combn"               
#> [19] "maphash"              "addmargins"           "aggregate.data.frame"
#> [22] "aggregate.ts"         "ave"                  "dendrapply"          
#> [25] "kronecker"            "mclapply"             "mcmapply"            
#> [28] "parApply"             "parCapply"            "parRapply"           
#> [31] "parSapply"            "parSapplyLB"          "pvec"
```

<sup>Created on 2022-09-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
> Looks like we'll need a new name -- 37 characters is too long

How about `unnecessary_lambda_linter()`?

R/unnecessary_anonymous_function_linter.R
MichaelChirico: thanks! I think we also need to remove things like `.mapply` that don't support accepting args in `...` form

MichaelChirico:
> How about unnecessary_lambda_linter()?

makes sense to me, but i wonder if less-experienced users will be confused by this name

IndrajeetPatil:
> makes sense to me, but i wonder if less-experienced users will be confused by this name

IMO, given the kind of users that use linter or know what lints are, I think we can safely assume that they either know what this means or have the wherewithal to figure it out on their own. Plus, the doc should help as well.

R/unnecessary_anonymous_function_linter.R
IndrajeetPatil: True. Updated the reprex accordingly.

MichaelChirico: I also removed S3 methods and glancing through the manuals, I also exclude `ave()`, `combn()`, `sweep()`, `kronecker()`, and `by()`. With those, the signature is not exactly amenable because `...` should come directly after `FUN` (or be required, like for `vapply()`), otherwise user may have to specify some odd argument in order to add position arguments to the lambda.

That leaves us with these to add for now (as well as any others where the argument is not named `FUN`, but we can get to that later)

```
mapply
outer
parApply
pvec
```

IndrajeetPatil: Sounds good. 

Thanks for painstakingly going through the docs to decide which ones are appropriate to include here!

R/unnecessary_lambda_linter.R
```diff
@@ -0,0 +1,104 @@
...
+  purrr_fun_xpath <- glue::glue("
+  //SYMBOL_FUNCTION_CALL[ {xp_text_in_table(purrr_mappers)} ]
+  /parent::expr
+  /following-sibling::expr[
+    OP-TILDE
+    and expr[OP-LEFT-PAREN/following-sibling::expr[1][not(preceding-sibling::*[2][self::SYMBOL_SUB])]/{purrr_symbol}]
+    and not(expr/OP-LEFT-PAREN/following-sibling::expr[position() > 1]//{purrr_symbol})
+  ]")
+
+  return(Linter(function(source_expression) {
```
AshesITR, line 64: redundant return

R/unnecessary_lambda_linter.R
```diff
@@ -0,0 +1,104 @@
...
+      purrr_fun_expr,
+      source_expression = source_expression,
+      lint_message = paste(
+        "Avoid unnecessary anonymous functions in purrr iterator function calls, e.g.,",
+        "prefer purrr::map(DF, sum) to purrr::map(DF, ~sum(.x))."
+      ),
+      type = "warning"
+    )
+
+    return(c(default_fun_lints, purrr_fun_lints))
```
AshesITR, line 96: redundant return

AshesITR (changes requested):
I think we should test multi-line functions and braces more thoroughly, especially the false negative case.

e.g.
`lapply(1:10, function(x) { print(x) })`, which should lint

and

`lapply(1:10, function(x) { print(x); x^2 })`, which should not lint.

R/unnecessary_lambda_linter.R
MichaelChirico: nice catch, thanks!

MichaelChirico:
> I think we should test multi-line functions and braces more thoroughly

Great idea! Indeed caught some false negatives. Let's leave them for follow-up (#1567) -- false negatives are more tolerable than false positives.

AshesITR:
Am I missing the positive braced test?
e.g. `function(x) { is.data.frame(x) }`.

Otherwise looks good so far

MichaelChirico:
that's marked for follow-up (#1567), is that a blocker to merge?

AshesITR:
Ah sorry I missed that issue. Is it referenced somewhere in the code or tests?

MichaelChirico:
There wasn't! Just added, it slipped through the cracks.

IndrajeetPatil (approved):

MichaelChirico:
@AshesITR need your approval now since the `Changes requested` is still pending

PR #1547, opened 2022-09-21
`fixed_regex_linter()` doesn't fail with `"\\;"`

### LINKED ISSUE #1545: fixed_regex_linter() fails when encountering "\\;"
```r
lintr::lint('strsplit(x, split = "\\\\;")\n', lintr::fixed_regex_linter())
# Error: '\;' is an unrecognized escape in character string starting ""\;"
```

But the code itself is fine:

```r
strsplit('a;b', '\\;')
# [[1]]
# [1] "a" "b"
```

IndrajeetPatil:
Closes #1545

``` r
lintr::lint(
  'strsplit("a;b", split = "\\\\;")\n', 
  lintr::fixed_regex_linter()
)
#> <text>:1:25: warning: [fixed_regex_linter] This regular expression is static, i.e., its matches can be expressed as a fixed substring expression, which is faster to compute. Here, you can use ";" with fixed = TRUE.
#> strsplit("a;b", split = "\\;")
#>                         ^~~~~
```

<sup>Created on 2022-09-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # lintr (development version)
 
+## Bug fixes
+
+* `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
+
 # lintr 3.0.1
 
 * Skip multi-byte tests in non UTF-8 locales (#1504)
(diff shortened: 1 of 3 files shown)


IndrajeetPatil:
Is there any other character we should anticipate that we are currently missing here?

```r
rex::rex("\\", one_of("^${}().*+?|[]\\<>:;")))
```

MichaelChirico (approved):
I haven't seen any other similar failures.

they seem rare enough that we can just address any new ones incrementally as they arise

MichaelChirico:
thanks!

PR #1550, opened 2022-09-23
Add exceptions for `duplicate_argument_linter()`

### LINKED ISSUE #1345: duplicate_argument_linter should skip on mutate() calls
`dplyr::mutate()` allows sequential updates to variables, e.g.

```
x %>%
  dplyr::mutate(
    col = gsub("t", "", col),
    col = gsub("\\s+$", "xxx", col)
  )
```

but `duplicate_argument_linter()` sees `col` used twice and throws.

IndrajeetPatil:
Closes #1345

Formatting changes coming from `{styler}`.


R/duplicate_argument_linter.R
```diff
@@ -14,6 +14,11 @@ duplicate_argument_linter <- function(except = character()) {
   xpath_call_with_args <- "//expr[EQ_SUB]"
   xpath_arg_name <- "./EQ_SUB/preceding-sibling::*[1]"
 
+  # Functions that allow sequential updates to variables (#1345)
+  allow_list <- c("mutate", "transmute")
```
IndrajeetPatil, line 18: Instead of this, we can also update default argument for `except` parameter: `except = c("mutate", "transmute")`

I feel this might be better because then users, if they wish to, have the choice to not skip this linter for `dplyr::mutate()` calls.

AshesITR: I like the argument solution more. Especially since I remember the repeated updates to cause invalid generated SQL when using dbplyr as a backend in the past, and sequential updates could still reasonably be discouraged in a code style even if they don't error.

IndrajeetPatil: Sounds good. I have made this change.

tests/testthat/test-duplicate_argument_linter.R
```diff
@@ -1,82 +1,144 @@
...
-  expect_lint("function(arg = 1, arg = 1) {}",
+  expect_lint(
+    "function(arg = 1, arg = 1) {}",
     list(message = rex("Repeated formal argument 'arg'.")),
-    duplicate_argument_linter())
+    duplicate_argument_linter()
+  )
+})
+
+test_that("doesn't lint duplicated arguments in allowed functions", {
```
IndrajeetPatil, line 114: Newly added tests start from here on.

NEWS.md
```diff
@@ -6,6 +6,10 @@
 
 ## Changes to defaults
 
+* `duplicate_argument_linter()` no longer produces warnings for functions that allow
```
MichaelChirico, line 9: I would reword the news item to more closely match what's done -- `mutate()` and `transmute()` are added the default of `exclude=`

R/duplicate_argument_linter.R
```diff
@@ -22,11 +24,8 @@ duplicate_argument_linter <- function(except = character()) {
     xml <- source_expression$full_xml_parsed_content
 
     calls <- xml2::xml_find_all(xml, xpath_call_with_args)
-
-    if (length(except)) {
```
MichaelChirico, line 26: we still need the `length(except)` test right? user can still pass `except = character()`, perhaps by accident.

tests/testthat/test-duplicate_argument_linter.R
```diff
@@ -1,82 +1,144 @@
...
     NULL,
-    duplicate_argument_linter(except = "list"))
+    duplicate_argument_linter(except = "list")
+  )
 
   expect_lint(
     "fun(`
-` = 1, `
-` = 2)",
+     ` = 1, `
```
MichaelChirico, line 101: is this a styler bug? the nature of this test has changed -- on `main`, it's testing that the variable named `\n` gets caught when used as a duplicate. Now the test still checks whitespace-only names, but adds spaces...

https://github.com/r-lib/lintr/blob/e993d4d37fc3016aefade264b4547ea76dc40f06/tests/testthat/test-duplicate_argument_linter.R#L73-L75

IndrajeetPatil: No, not styler bug. I unintentionally changed this. Reverted.

R/duplicate_argument_linter.R
IndrajeetPatil: Ah, true. When I opened the PR, I was creating the exception vector inside the function, so the length could never be zero, but in the current implementation, it can indeed be 0.

Fixed this.

NEWS.md
IndrajeetPatil: Okay, I have changed this item. Now it reads:

```md
* `duplicate_argument_linter()` defaults to `except = c("mutate", "transmute")`.
  This is because these functions allow sequential updates to variables (#1345, @IndrajeetPatil).
```

MichaelChirico (approved):

PR #1551, opened 2022-09-23
Check with the oldest supported R version

IndrajeetPatil:
The oldest R version we can check is `R 3.4` (https://github.com/r-lib/pak/issues/417#issuecomment-1257005490).

Given that we can't check on older R versions (or is there a way?) and thus can't confirm support for that R version either way, should we be bumping the minimum needed R version to `3.4`? (Currently, we support `R >= 3.2`.)


IndrajeetPatil:
Hmm, why does it say `{pak}` not available when it [supports](https://cran.r-project.org/web/packages/pak/index.html) `R >= 3.2`?

MichaelChirico (approved):

IndrajeetPatil:
> Given that we can't check on older R versions (or is there a way?) and thus can't confirm support for that R version either way, should we be bumping the minimum needed R version to 3.4? (Currently, we support R >= 3.2.)

@MichaelChirico Any thoughts on this?

MichaelChirico:
Is `pak` a requirement for GHA or one of our upstream dependencies?

IndrajeetPatil:
It's a requirement for GHA, but, IINM, that's the only way we have to know if things are working as expected on the older versions. Since CRAN doesn't test past oldrel-1. So I am not sure how we can confirm that we indeed support `R 3.2`, in the sense that most features work as expected for this version.

MichaelChirico:
It's not too difficult to install 3.2 locally and test :)

I am basically fine keeping the old version until it's proven _not_ to work.

OTOH, for lintr, its main use case is package development, and people submitting to CRAN are technically supposed to be testing their packages with r-devel. So it seems more reasonable to force newer versions of R, and an alignment to GHA makes sense if folks are relying on GHA to test versions of R unavailable to them locally during development.

So, I'm fine either way. @AshesITR any thoughts?

IndrajeetPatil:
> I am basically fine keeping the old version until it's proven not to work.

Okay, I am fine with that. 

I was just curious to hear your thoughts on this, apropos testing older R versions on GHA.

Issue #797, opened 2021-05-10
Linting code chunks in Rmd not working consistently with knitr


renkun-ken:
Consider the following Rmd (saved to e.g. `lintr-test.Rmd`):

````
```{r}
x<-1
```

```{R}
x<-2
```

```{therom, name='test'}
x<-3
```
````

Linting the file has the following output:

```
> lintr::lint("lintr-test.Rmd")                                                                                                                             
/home/renkun/Workspaces/testR/lintr-test.Rmd:2:2: style: Put spaces around all infix operators.
x<-1
~^~~
/home/renkun/Workspaces/testR/lintr-test.Rmd:10:2: style: Put spaces around all infix operators.
x<-3
~^~~
```

Two my understanding, it should lint the first two R chunks and ignores the third chunk.

However, the current chunk extraction does not seem to work consistently with knitr in that knitr treats `r` and `R` as the same. The `theorem` is brought up at https://github.com/REditorSupport/languageserver/issues/421 where bookdown seems to provide some markdown extensions where additional engines are implemented (as introduced at https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html).

I think the expected behavior should be linting the `R` or `r` chunks and ignores all other chunks.

**Comment by russHyde** [comment]
Reproduced this:

```{R}
# "R" engine is a synonym for "r"
a<-1
a
```

This code block, when added to `./tests/testthat/knitr_formats/test.Rmd` throws no assignment lints
But, the code block is evaluated by knitr when knitting to .html.

russHyde:
Also, lintr should attack any code blocks that have engine="Rscript"

````
```{Rscript}
# "Rscript" engine wraps the code in a block in a randomScript and calls
# `Rscript randomScript.R 2.&1`
#
# So, we couldn't access the `a` defined above from an `Rscript` block

b<-1
b
```
````

russHyde:
One area that I'm not sure how to handle are custom knitr engines

Some of these are used to run R code, and so should be linted, eg, [details](https://cran.r-project.org/web/packages/details/vignettes/engine.html) and [targets](https://books.ropensci.org/targets/markdown.html)

Other packages add custom knitr engines that don't run R code (I believe this is the case for the tufte "marginfigure" engine mentioned in #796) and so should be ignored.

Perhaps the lintr config should allow the user to add additional engines whose code should be linted.

russHyde:
I agree re the bug for {therom} example:

````
```{notAnEngine}
[...]

PR #1552, opened 2022-09-24
Anticipate more knitr engines in code blocks

### LINKED ISSUE #796: Support tufte package arbitrary margin content
The [tufte](https://github.com/rstudio/tufte) package provides a special "marginfigure" chunk, which leads to false lintr errors (it is fine to use HTML in there):

````r
```{marginfigure}
"The tidyverse is an opinionated collection of R packages designed for data science. 
All packages share an underlying design philosophy, grammar, and data structures."

<span style="float: right">--- Hadley Wickham</span>
```
````

Produces this error message:

```
02-tidyverse.Rmd:26:1: error: Unexpected '<'
<span style="float: right">--- Hadley Wickham</span>
^
```

I hope this can be fixed. Anyway, thank you very much for your great package!

### PR DESCRIPTION
**Pull request by IndrajeetPatil**
Closes #796

Related to https://github.com/r-lib/lintr/issues/797

Plus styling changes.

### CODE REVIEW

File: tests/testthat/test-knitr_formats.R [hunk 1]
```diff
@@ -169,3 +179,14 @@ test_that("it does lint with malformed input", {
     expect_lint(contents[[i]], expected[[i]], linters = list())
   }
 })
+
+test_that("marginfigure engine doesn't cause problems", {
```
IndrajeetPatil, line 183: Newly added test

R/extract.R
```diff
@@ -98,8 +98,25 @@ filter_chunk_end_positions <- function(starts, ends) {
...
+  "example",     "exm",
+  "exercise",    "exr",
+  "hypothesis",  "hyp"
+)
+
 defines_knitr_engine <- function(start_lines) {
-  engines <- names(knitr::knit_engines$get())
+  engines <- c(
+    names(knitr::knit_engines$get()), # knitr
+    bookdown_engines,                 # bookdown
```
IndrajeetPatil, line 117: Not directly related to the current issue, but still important to be included here for #797

NEWS.md
```diff
@@ -4,6 +4,9 @@
 
 * `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
 
+* `lint()` no longer produces error on encountering raw HTML included in `marginfigure` code blocks 
+  from `{tufte}` package (#796, @IndrajeetPatil).
```
MichaelChirico, line 8: i think "from _the_ {tufte} package" or just "from {tufte}"

R/extract.R
```diff
@@ -98,8 +98,25 @@ filter_chunk_end_positions <- function(starts, ends) {
   code_ends
 }
 
+bookdown_engines <- c(
```
MichaelChirico, line 101: does bookdown export anything we could use instead of maintaining this?

IndrajeetPatil: Nothing I could find, unfortunately. Should we request it?

But I'd prefer if this doesn't block the current PR.

MichaelChirico: Yes, please file an issue, and mark a comment here referring to the "upstream" issue #

.lintr
```diff
@@ -10,6 +10,7 @@ exclusions: list(
   "tests/testthat/dummy_packages",
   "tests/testthat/dummy_projects",
   "tests/testthat/exclusions-test",
+  "tests/testthat/knitr_extended_formats",
```
MichaelChirico, line 13: inspired #1554

R/extract.R
IndrajeetPatil: Actually, looks like the additional engines are registered automatically if the library is loaded, but we don't want to gain dependence on `{bookdown}`, so not sure what to do here. 

Should I still create an issue?

``` r
sort(names(knitr::knit_engines$get()))
#>  [1] "asis"      "asy"       "awk"       "bash"      "block"     "block2"   
#>  [7] "bslib"     "c"         "cat"       "cc"        "coffee"    "comment"  
#> [13] "css"       "ditaa"     "dot"       "embed"     "eviews"    "exec"     
#> [19] "fortran"   "fortran95" "gawk"      "glue"      "glue_sql"  "gluesql"  
#> [25] "go"        "groovy"    "haskell"   "highlight" "js"        "julia"    
#> [31] "lein"      "mysql"     "node"      "octave"    "perl"      "php"      
#> [37] "psql"      "python"    "R"         "Rcpp"      "Rscript"   "ruby"     
#> [43] "sas"       "sass"      "scala"     "scss"      "sed"       "sh"       
#> [49] "sql"       "stan"      "stata"     "targets"   "tikz"      "verbatim" 
#> [55] "zsh"

library(bookdown)
sort(names(knitr::knit_engines$get()))
#>  [1] "asis"        "asy"         "awk"         "bash"        "block"      
#>  [6] "block2"      "bslib"       "c"           "cat"         "cc"         
#> [11] "coffee"      "comment"     "conjecture"  "corollary"   "css"        
#> [16] "definition"  "ditaa"       "dot"         "embed"       "eviews"     
#> [21] "example"     "exec"        "exercise"    "fortran"     "fortran95"  
#> [26] "gawk"        "glue"        "glue_sql"    "gluesql"     "go"         
#> [31] "groovy"      "haskell"     "highlight"   "hypothesis"  "js"         
#> [36] "julia"       "lein"        "lemma"       "mysql"       "node"       
#> [41] "octave"      "perl"        "php"         "proof"       "proposition"
#> [46] "psql"        "python"      "R"           "Rcpp"        "remark"     
#> [51] "Rscript"     "ruby"        "sas"         "sass"        "scala"      
#> [56] "scss"        "sed"         "sh"          "solution"    "sql"        
#> [61] "stan"        "stata"       "targets"     "theorem"     "tikz"       
#> [66] "verbatim"    "zsh"
```

<sup>Created on 2022-09-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: Does that mean the issue is solved by running `library(bookdown)` before running the linter?

IndrajeetPatil: Probably, but we shouldn't count on it, since it's unusual to explicitly load this library in `{bookdown}` projects. 

For static analysis, it would just be easier to have a vector of supported engines, like we are doing for `{knitr}`. Note that users don't need to do `library(knitr)`, so it'd be surprising for them to have to do so for `{bookdown}`.

MichaelChirico: Hmm, still not a fan of maintaining a long list of things which looks likely to change in the future (and the list likely to grow for other knitr engines), at least we should be getting the names supplied by bookdown:

https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/R/utils.R#L551-L556

https://github.com/rstudio/bookdown/blob/dfe92a227b1dce5b8f0a2f758fb545e7df1b7e62/R/zzz.R#L5

> Note that users don't need to do library(knitr), so it'd be surprising for them to have to do so for {bookdown}.

That's not _too_ surprising IMO, `knitr` is the default / core engine for the Rmd format, and so _it_ is the exception rather than bookdown. All other engines would have the same advice: run `library(some.engine)` first to register the chunk engines; this rule also applies to `bookdown`.

IndrajeetPatil: Okay, I have reverted this change to rely on `{bookdown}` being loaded.

tests/testthat/test-knitr_extended_formats.R
```diff
@@ -0,0 +1,32 @@
+regexes <- list(
```
MichaelChirico, line 1: this is overly complicated... please just use the text in the test directly, i.e. make these tests more DAMP:

https://stackoverflow.com/q/6453235/3576984

IndrajeetPatil: I was just trying to be internally consistent. This is what we do in `test-knitr_formats.R`, so thought it is expected that `test-knitr_extended_formats.R` be similar.

Will change that.

MichaelChirico: even in the other file I'm not sure it's the best idea... but even less so in this new file. thanks.

IndrajeetPatil:
@MichaelChirico This should be ready for a review.

AshesITR:
Not convinced that we need to hardcode the fix. Is the bug even triggered if `library(tufte)` is run before linting?

Maybe we should try to pull the package used in `output:` and auto-load its and its dependencies' knitr engines to avoid having the user load them manually before linting?

MichaelChirico:
Agree with Alexander and it's why I haven't reviewed yet -- I want to play around with this a bit on my machine but don't have access to my personal machine where it's easier to load CRAN packages.

Alexander also offers a great suggestion -- can we glean from the Rmd what engine is needed and attempt `library(<engine>)` before linting? User will need to run the linter from a machine where the engine is installed, or else get the false positives, but that doesn't seem like too strict a requirement.

AshesITR:
NB `loadNamespace()` might be sufficient to load knitr engines (not sure). If it is, we should avoid unnecessarily polluting `search()`.

MichaelChirico:
`loadNamespace()` will work for bookdown:

https://github.com/rstudio/bookdown/blob/7b53f2dd7ddb57b7e78ee1575c06b892f549c4e8/R/zzz.R#L4-L6

But I don't see any `.onLoad()` hook for tufte. e.g. `marginfigure` is registered only when calling `tufte_html()`?

https://github.com/rstudio/tufte/blob/044e2c0016cba8cd48e28c72e14fadc239096f00/R/html.R#L159-L165

AshesITR:
Sounds like we should file an issue with tufte then, suggesting they adapt their registration routine to make marginfigure "discoverable".

MichaelChirico:
> Sounds like we should file an issue with tufte then

That was my first thought, but I also don't see any guidance on this being "best practice", e.g. https://bookdown.org/yihui/rmarkdown-cookbook/custom-engine.html

Maybe we should file an issue with `knitr` as well?

AshesITR:
> Maybe we should file an issue with `knitr` as well?

To ask for guidance?

MichaelChirico:
> To ask for guidance?

basically, yes. do they recommend package authors register engines in .onLoad? how else might we detect engines in our use case?

MichaelChirico:
Filed: https://github.com/rstudio/tufte/issues/117

And: https://github.com/yihui/knitr/issues/2181

IndrajeetPatil:
Thanks for filing the issues, @MichaelChirico!

MichaelChirico:
https://github.com/rstudio/tufte/issues/117 now closed: https://github.com/rstudio/tufte/commit/a0970268efe8d5f4af3d36e4866bad388896706f

What remains to be done on this PR?

IndrajeetPatil:
> What remains to be done on this PR?

I am no longer hardcoding `"marginfigure"` engine. I've updated tests to rely on the devel-`{tufte}`.

I don't think there is anything else that needs to be done here.

DESCRIPTION
```diff
@@ -37,6 +37,7 @@ Imports:
     xml2 (>= 1.0.0),
     xmlparsedata (>= 1.0.5)
 Suggests:
+    bookdown,
```
MichaelChirico, line 40: is there any way for us to avoid these Suggests?

e.g. could we detect namespaces in the file linted, and attempt `requireNamespace()` on them, thereby obviating a Suggested dependency, as long as the calling user has the package available?

Doesn't have to be done in this PR (e.g. if it requires some substantial machinery changing in `get_source_expressions()` or `lint()`), just wondering if it's possible.

MichaelChirico:
Yihui notes in https://github.com/yihui/knitr/issues/2181 there's no strong recommendation for engines to be registered `.onLoad()`; I still think it best to assume that's done, and tell users to encourage package authors to do so otherwise if they want `lintr` to work properly.

If there's demand, we could also consider, e.g. a `.lintr` setting to control custom engines.

DESCRIPTION
IndrajeetPatil: This might work, but not always. 

Because, more often than not, people won't use namespace while specifying output formats. That is, they are more likely to write 

```yaml
---
output: tufte_html
---
```

instead of 

```yaml
---
output: tufte::tufte_html
---
```

This is a result of habits formed while using the more standard output formats where users write 

```yaml
---
output: html_document
---
```

instead of 

```yaml
---
output: rmarkdown::html_document
---
```

MichaelChirico: I'm that case, how does knitr detect the namespace?

NEWS.md
```diff
@@ -4,6 +4,9 @@
 
 * `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
 
+* `lint()` no longer produces error on encountering raw HTML included in `marginfigure` code blocks 
+  from the `{tufte}` package (#796, @IndrajeetPatil).
```
MichaelChirico, line 8: can we add some note about what's going on more generally? it may help for users experiencing a similar issue but with other engines

IndrajeetPatil: I have completely rewritten the NEWS item. See if the new version is more apt.

DESCRIPTION
MichaelChirico: could you file a follow-up issue for someone to look into this eventually? thanks

IndrajeetPatil: > [In] that case, how does knitr detect the namespace?

Yeah, I was also curious about the same and went down the rabbit hole of `rmarkdown::render()` and `knitr::knit()`, but gave up given the complexity of these functions. Might revisit later when I have a bit more time on my hands.

> could you file a follow-up issue for someone to look into this eventually? thanks

Done: https://github.com/r-lib/lintr/issues/1617

NEWS.md
```diff
@@ -44,6 +44,25 @@
 * `boolean_arithmetic_linter()` for identifying places where logical aggregations are more appropriate, e.g.
   `length(which(x == y)) == 0` is the same as `!any(x == y)` or even `all(x != y)` (@MichaelChirico)
 
+## Notes
+
+* `lint()` continues to support Rmarkdown documents. For users of custom .Rmd engines, e.g.
```
MichaelChirico, line 49: @IndrajeetPatil PTAL. I moved it from "bug fixes" because there's no longer any code change by lintr. I still don't feel like I have a 100% grasp of the situation, so feel free to refine the item further.

NEWS.md
```diff
@@ -48,6 +44,25 @@
...
+  `marginformat` from {tufte} or `theorem` from {bookdown}, note that those engines must be registered
+  in {knitr} prior to running `lint()` in order for {lintr} to behave as expected, i.e., they should be
+  shown as part of `knitr::knit_engines$get()`.
+  
+  For {tufte} and {bookdown} in particular, one only needs to load the package namespace to accomplish
+  this (i.e., minimally `loadNamespace("tufte")` or `loadNamespace("bookdown")`, respectively, will
+  register those packages' custom engines; since `library()` also runs `loadNamespace()`, running
+  `library()` will also work). Note further that {tufte} only added this code to their `.onLoad()` recently
+  after our request to do so, see https://github.com/rstudio/tufte/issues/117 & ensure you're using a
```
IndrajeetPatil, line 58:
```suggestion
  after our request to do so (see https://github.com/rstudio/tufte/issues/117). Therefore, ensure you're using a
```

MichaelChirico (approved):

NEWS.md
IndrajeetPatil: Thanks, Michael. This looks good to me.

I think we can revise this further as our understanding of the issue evolves while working on #1617.

PR #1549, opened 2022-09-22
`unused_import_linter()` detects datasets from imports

IndrajeetPatil:
Closes #1508

PR #1555, opened 2022-09-24
Slightly improve dataset symbol detection

MichaelChirico:
Follow-up to #1549, the `Items` column has a funny format for some objects, per `?data`:

```
Where the datasets have a different name from the
argument that should be used to retrieve them the index will have
an entry like ‘beaver1 (beavers)’ which tells us that dataset
‘beaver1’ can be retrieved by the call ‘data(beaver)’.
```

From experimenting, it looks like `beaver1` is the one we want to allow.


IndrajeetPatil (approved):
Great catch! Thanks.

PR #1556, opened 2022-09-25
New redundant_equals_linter

### LINKED ISSUE #1500: New linter for `if(x == TRUE)` and `if(x == FALSE)`
It may be useful for beginners if this is pointed out by lintr, since it may be more appropriate to rewrite `if(x == TRUE) {...}`  as follows

```r
if(x) {...}
if(isTRUE(x)) {...}
```

MichaelChirico:
Part of #884; Closes #1500.


R/redundant_equals_linter.R
```diff
@@ -0,0 +1,36 @@
...
+    bad_expr <- xml2::xml_find_all(xml, xpath)
+    op <- xml2::xml_text(xml2::xml_find_first(bad_expr, "*[2]"))
+
+    xml_nodes_to_lints(
+      bad_expr,
+      source_expression = source_expression,
+      lint_message = paste(
+        "Using", op, "on a logical vector is redundant.",
+        "Well-named logical vectors can be used directly in filtering.",
+        "For data.table's `i` argument, wrap the column name in (), like DT[(is_treatment)]."
```
IndrajeetPatil, line 31:
```suggestion
        "For data.table's `i` argument, wrap the column name in (), like `DT[(is_treatment)]`."
```

R/redundant_equals_linter.R
```diff
@@ -0,0 +1,36 @@
+#' Block usage of ==, != on logical vectors
```
IndrajeetPatil, line 1:
```suggestion
#' Block usage of `==`, `!=` on logical vectors
```

R/redundant_equals_linter.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' Testing `x == TRUE` is redundant if `x` is a logical vector. Wherever this
+#'   is used to improve readability, the solution should instead be to improve
+#'   the naming of the object to better indicate that its contents are logical.
+#'
```
IndrajeetPatil, line 6:
```suggestion
#' This can be done using prefixes (is, has, can, etc.). For example, `is_child`, `has_parent_supervision`,
#' `can_watch_horror_movie` clarify their logical nature, while `child`, `parent_supervision`,
#' `watch_horror_movie` don't.
```

tests/testthat/test-redundant_equals_linter.R
```diff
@@ -0,0 +1,30 @@
...
+  linter <- redundant_equals_linter()
+  eq_msg <- rex::rex("Using == on a logical vector is redundant.")
+  ne_msg <- rex::rex("Using != on a logical vector is redundant.")
+  expect_lint("x == TRUE", eq_msg, linter)
+  expect_lint("x != TRUE", ne_msg, linter)
+  expect_lint("x == FALSE", eq_msg, linter)
+  expect_lint("x != FALSE", ne_msg, linter)
+
+  # order doesn't matter
+  expect_lint("TRUE == x", eq_msg, linter)
```
IndrajeetPatil, line 18: For the sake of completeness:

```suggestion
  expect_lint("TRUE == x", eq_msg, linter)
  expect_lint("TRUE != x", eq_msg, linter)
```

tests/testthat/test-redundant_equals_linter.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_lint("x == 1", NULL, redundant_equals_linter())
+  # comparison to TRUE as a string
+  expect_lint("x != 'TRUE'", NULL, redundant_equals_linter())
+})
+
+test_that("redundant_equals_linter blocks simple disallowed usages", {
+  linter <- redundant_equals_linter()
+  eq_msg <- rex::rex("Using == on a logical vector is redundant.")
+  ne_msg <- rex::rex("Using != on a logical vector is redundant.")
+  expect_lint("x == TRUE", eq_msg, linter)
```
IndrajeetPatil, line 12: These are definitely edge cases, but should these lint?
``` r
library(lintr)

lint(
  text = "x == c(TRUE)", 
  linters = redundant_equals_linter()
)

lint(
  text = "x != c(TRUE)", 
  linters = redundant_equals_linter()
)
```

<sup>Created on 2022-09-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I am fine with this being covered in a future PR. Up to you.

IndrajeetPatil (changes requested):
Few nitpicks. But, overall, looks good to me! :)

tests/testthat/test-redundant_equals_linter.R
MichaelChirico: I'll do this with patrick if we're going to be exhaustive :)

tests/testthat/test-redundant_equals_linter.R
MichaelChirico: I don't think it's worth adding :)

those will lint with unneeded_concatenation_linter(), and I suspect it's pretty uncommon to begin with

IndrajeetPatil: Fair enough. 

I was just imagining a scenario where the user has turned off `unneeded_concatenation_linter()`, but the combined probability of both happening is probably minuscule.

MichaelChirico: It's definitely a good thing to keep in mind -- we shouldn't rely on ppl having certain combinations of linters applied to get max value from our linters. It helps that `unneeded_concatenation_linter()` is on by default, so it's not like users have to go searching for 2 custom linters for that case.

But there is also a maintenance overhead tradeoff, and in this case I just don't see the edge case as worthwhile. I just searched our code history (20 years of commits) and nobody ever used `== c(TRUE)` (ditto `FALSE` and `!=`).

IndrajeetPatil (approved):
Thanks!!

Issue #854, opened 2021-08-27
Does lintr-bot still work for github actions?


tonyk7440:
Hi,

I have the [lintr workflow](https://github.com/r-lib/actions/blob/master/examples/lint.yaml) in my repo and it runs quite well.
The only thing that is strange is that when an annotation pops up the lintr-bot doesn't comment in the pull request.
Is this still a feature that is working or are there steps to activate it in a repo?

Thanks.

MichaelChirico:
See some discussion on #1558. We recommend imitating this repo's GHA:

https://github.com/r-lib/lintr/blob/bed54c55523e96f7c89154ae31702cd74bb209d8/.github/workflows/lint-changed-files.yaml

It (1) runs `lint_package()` restricted to the files in the `diff` of a given PR (2) the action fails if any new lint is found, and new lints are printed in the `Files` pain inline with where they're found in the source.

tonyk7440:
Ok, thanks all!

Issue #364, opened 2019-01-16
post github comments as inline code review


mattyb:
The github commenting is a neat feature of lintr. Since the comments aren't tied to commits however, it can be difficult to tell when if a lint comment is still relevant. It would be more powerful if the comments were posted as code review comments with a position value, so they would show up inline with the code and be marked as outdated if the code changes.
https://developer.github.com/v3/pulls/comments/#create-a-comment

jimhester:
I agree! The code review API did not exist when this was implemented in lintr, which is why this has not the way it currently works.

MichaelChirico:
Check the discussion on #1558. We think the current setup is even more convenient (lints are posted as annotations in the `Files` page of a PR). See #854 as well for further reference.

PR #1558, opened 2022-09-25
DO NOT MERGE: Check lintr comment bot

IndrajeetPatil:
#854

### DIFF
--- a/.github/workflows/lint-changed-files.yaml
+++ b/.github/workflows/lint-changed-files.yaml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Extract and lint files changed by this PR
         run: |
-          Sys.setenv("LINTR_ERROR_ON_LINT" = TRUE)
+          Sys.setenv("LINTR_COMMENT_BOT" = TRUE)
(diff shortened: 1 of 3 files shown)


IndrajeetPatil:
@jimhester, @MichaelChirico Does this mean that lintr bot on GitHub no longer posts comments in the PR? 

P.S. Annotations work just fine. This question and experimental PR is inspired by #854.

IndrajeetPatil:
*bump*

MichaelChirico:
I never really knew how the bot worked, so i can't confirm or deny :)

jimhester:
Likely the PAT got invalidated, which is why the bot is not working.

I guess there is some question as to whether we should just get rid of the commenting, as it is probably better to use annotations instead now that they exist.

MichaelChirico:
Agree, the current status quo on this repo using annotations & failing the CI if lints are found works well for us. Let's recommend that instead.

IndrajeetPatil:
I think this issue can also be closed for similar reasons? 
https://github.com/r-lib/lintr/issues/364

PR #1559, opened 2022-09-25
Sync up with lint workflow in `r-lib/actions`

IndrajeetPatil:


.github/workflows/lint.yaml
```diff
@@ -19,16 +15,15 @@ jobs:
...
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: |
-            any::gh
-            any::purrr
-          needs: check
-
-      - name: Install package
-        run: R CMD INSTALL .
+            r-lib/lintr
```
IndrajeetPatil, line 24: This is the only deviation from what's in the standard [example workflow](https://github.com/r-lib/actions/blob/v2-branch/examples/lint.yaml) for linting. 

I think we should always be linting with the latest GitHub version of `{lintr}`.

.github/workflows/lint.yaml
```diff
@@ -1,12 +1,8 @@
 on:
   push:
-    branches:
-      - main
-      - master
+    branches: [main, master]
```
MichaelChirico, line 3: any reason to keep `master` here? just to keep perfectly synced with the upstream source?

IndrajeetPatil: Yes, no other reason. Can also remove it. Fine either way.

MichaelChirico: SGTM, thanks for clarifying

MichaelChirico (approved):

PR #1560, opened 2022-09-25
Remove `{covr}` unneeded from `Suggests`

IndrajeetPatil:


tests/testthat/test-with.R
```diff
@@ -18,7 +18,7 @@ test_that("all default linters are tagged default", {
   # covr modifies package functions causing differing deparse() results even for identical anonymous functions.
   # This happens because default_linters is generated at build time and thus not modifiable by covr, whereas
   # linters_with_tags() constructs the linters at runtime.
-  skip_if(requireNamespace("covr", quietly = TRUE) && covr::in_covr())
+  skip_on_covr()
 
   expect_equal(linters_with_tags("default"), linters_with_defaults())
```
MichaelChirico, line 23: @IndrajeetPatil oh, whoops, we should remove `expect_identical_linter()` since we didn't apply that internally...

MichaelChirico: #1561

MichaelChirico (approved):

MichaelChirico:
Any reason for removing covr in particular?

To me it's part of Suggests since it should technically be part of our CI suite, though that job appears to be broken for quite some time.

IndrajeetPatil:
>  it should technically be part of our CI suite

It is part of our CI suite:

https://github.com/r-lib/lintr/blob/9f7e4ce2eccbe8601ae411604286ef2e5429745f/.github/workflows/test-coverage.yaml#L26

It just doesn't need to be a soft dependency, since we are not using any functions from it, IINM.

MichaelChirico:
I think you're technically right, but I still think it's a good idea to keep it in Suggests. This will help the covr maintainers discover its downstreams / report any issues that might affect us, for example.

IndrajeetPatil:
> discover its downstreams / report any issues that might affect us

But this is what I don't understand. How can it affect us if we use no functionality from it?

We rely on it only via `testthat::skip_on_covr()`, so it will affect `{testthat}`, but not us. Am I mistaken about this?

MichaelChirico:
Filed https://github.com/r-lib/covr/issues/515 to help clarify `covr` maintainers' preferences. Let's merge this for now.

PR #1568, opened 2022-09-29
New lengths_linter

MichaelChirico:
Part of #884


tests/testthat/test-lengths_linter.R
```diff
@@ -0,0 +1,35 @@
+test_that("lengths_linter skips allowed usages", {
+  # TODO(michaelchirico): this probably should be lengths(), but let it pass for now.
+  #   Seen twice as a usage like lapply(x, length) > 0 which does as.integer()
+  #   conversion and thus is equivalent to sapply(x, length). Check whether
+  #   there are any real use cases for lapply(x, length)...
+  expect_lint("lapply(x, length)", NULL, lengths_linter())
```
AshesITR, line 6: Could safely lint this (and map) suggesting `as.list(lengths(...))` instead. Can be a follow-up though.

R/lengths_linter.R
```diff
@@ -0,0 +1,33 @@
...
+#'   `sapply(x, length)`, but faster and more readable.
+#'
+#' @evalRd rd_tags("lengths_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+lengths_linter <- function() {
+  loop_funs <- c("sapply", "vapply", "map_int", "map_dbl")
+  xpath <- glue::glue("
+  //SYMBOL_FUNCTION_CALL[ {xp_text_in_table(loop_funs)} ]
+  /parent::expr/parent::expr[expr[position() = 3 and SYMBOL[text() = 'length']]]
```
AshesITR, line 14: Do you think it's worth to not assume positional arguments here?

MichaelChirico: I think this is kind of a general thing in `lintr`... a few thoughts:

 - Using arguments in their positional order is good practice in general, especially for the most common/simple functions. `sapply(FUN = length, x = x)` would be 🚨 for me. Probably we should document this "philosophy" somewhere.
 - We should catch the case when the argument is named, `FUN = length`, regardless of position
 - We should always keep pipes in mind (`|>` / `%>%`)

For this particular case, I only really caught usages of `sapply(x, length)`, the others are just there because "why not". Even at that it was pretty rare.

tests/testthat/test-lengths_linter.R
MichaelChirico: True, and I'm impressed just how much faster it is than even `vapply(x, length)`. That evaluation overhead is a killer.

```
library(microbenchmark)
l = replicate(1e5, sample(10, sample(10, 1)), simplify=FALSE)
microbenchmark(as.list(lengths(l)), sapply(l, length), vapply(l, length, 1L))
# Unit: milliseconds
#                   expr       min        lq      mean    median        uq       max neval
#    as.list(lengths(l))  4.846123  5.423464  7.144133  5.637852  5.988355  29.90146   100
#      sapply(l, length) 37.334026 42.108947 50.622672 49.059130 50.951532 113.20739   100
#  vapply(l, length, 1L) 33.147643 36.679405 44.902168 43.278030 45.429923 110.50852   100

l = replicate(1e6, sample(10, sample(10, 1)), simplify=FALSE)
microbenchmark(as.list(lengths(l)), sapply(l, length), vapply(l, length, 1L))
# Unit: milliseconds
#                   expr       min       lq     mean    median       uq      max neval
#    as.list(lengths(l))  53.26755  58.4670 127.9654  90.16496 127.2416 436.8011   100
#      sapply(l, length) 431.83550 474.3372 573.0207 523.75971 628.0669 854.3293   100
#  vapply(l, length, 1L) 367.80554 403.4373 454.7605 421.15209 455.9736 748.4410   100
```

MichaelChirico: #1570

R/lengths_linter.R
IndrajeetPatil: I have created an issue to track this : https://github.com/r-lib/lintr/issues/1573

I agree that these would be good improvements to implement at some point, but they shouldn't be the blocking factor for this PR.

IndrajeetPatil (dismissed):

NEWS.md
```diff
@@ -28,6 +28,7 @@
   `lapply(x, function(xi) sum(xi))` can be `lapply(x, sum)` and `purrr::map(x, ~quantile(.x, 0.75, na.rm = TRUE))`
   can be `purrr::map(x, quantile, 0.75, na.rm = TRUE)`. Naming `probs = 0.75` can further improve readability.
 * `redundant_equals_linter()` for redundant comparisons to `TRUE` or `FALSE` like `is_treatment == TRUE` (#1500, @MichaelChirico)
+* `lengths_linter()` for encouraging usage of `lengths(x)` instead of `sapply(x, length)` (and similar)
```
IndrajeetPatil, line 31: Is this sentence complete? 

> (and other similar functions from apply and map families).

IndrajeetPatil (approved):
One minor comment, but ready otherwise.

NEWS.md
MichaelChirico: no need for a complete sentence IMO :)

PR #1569, opened 2022-09-29
New return_assignment_linter

MichaelChirico:
Part of #884


NEWS.md
```diff
@@ -26,6 +26,8 @@
 
 * `redundant_equals_linter()` for redundant comparisons to `TRUE` or `FALSE` like `is_treatment == TRUE` (#1500, @MichaelChirico)
 
+* `return_assignment_linter()` for assignments within the `return()` clause of a function, e.g. `return(x <- foo())` (@MichaelChirico)
+
```
IndrajeetPatil, line 30: Nice!

Do you think implicit return assignments should also lint? I have seen code like this in the wild:

``` r
foo <- function() {
  x <- 4
}
y <- foo() 
y
#> [1] 4
```

<sup>Created on 2022-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Not meant to be resolved in this PR, but looking at it reminded me of this.

AshesITR:
Wondering whether we should centralize this to `return_linter` instead of making this a separate linter. WDYT?

NEWS.md
MichaelChirico: Good point that outside Google this form will be a lot more common (internally, we require explicit `return()` from functions).

Let's mark this as TODO & wait until we get to `explicit_return_linter()`. The logic to identify the last expression in XPath is a bit involved when the last expression is branched. Better to bootstrap off of that.

MichaelChirico:
Good point -- do we have any other linters in the package yet under this umbrella?

Smaller point -- what about `function_return_linter()`, is that too much?

MichaelChirico:
Just glancing over the list, I also see we _could_ put this under `assignment_linter` as well. But I the emphasis here is much more on the `return()` side, so sticking with that.

tests/testthat/test-function_return_linter.R
```diff
@@ -0,0 +1,68 @@
...
+
+  expect_lint(
+    trim_some("
+      foo <- function(x) {
+        return(x <<- x + 1)
+      }
+    "),
+    lint_msg,
+    linter
+  )
```
IndrajeetPatil, line 45: Also add this one for the sake of completeness?

``` r
library(lintr)

lint(
  text = "foo <- function(x) {
        return(x + 1 ->> x)
      }
    ",
  linters = function_return_linter()
)
#> <text>:2:16: warning: [function_return_linter] Move the assignment outside of the return() clause, or skip assignment altogether.
#>         return(x + 1 ->> x)
#>                ^~~~~~~~~~~
```

<sup>Created on 2022-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

NEWS.md
IndrajeetPatil: Sounds good. Created a new issue to track this: https://github.com/r-lib/lintr/issues/1574

IndrajeetPatil (approved):

PR #1576, opened 2022-09-30
Improve docs about how to list and use all linters

### LINKED ISSUE #1482: How to list and include all available linters?
# Synopsis

Correct me if I missed this, but I don't think currently the docs/functions make it easy to:

- list names of all available linters
- include all available linters while linting (a package, e.g.)

# Listing all available linters

It should be possible for the users to easily generate a list of names of available linters. 

Currently, we make this easy only for default linters:

``` r
library(lintr)

names(default_linters) # or `names(default_settings$linters)` will also work
#>  [1] "assignment_linter"                "brace_linter"                    
#>  [3] "commas_linter"                    "commented_code_linter"           
#>  [5] "cyclocomp_linter"                 "equals_na_linter"                
#>  [7] "function_left_parentheses_linter" "infix_spaces_linter"             
#>  [9] "line_length_linter"               "no_tab_linter"                   
#> [11] "object_length_linter"             "object_name_linter"              
#> [13] "object_usage_linter"              "paren_body_linter"               
#> [15] "pipe_continuation_linter"         "semicolon_linter"                
#> [17] "seq_linter"                       "single_quotes_linter"            
#> [19] "spaces_inside_linter"             "spaces_left_parentheses_linter"  
#> [21] "T_and_F_symbol_linter"            "trailing_blank_lines_linter"     
#> [23] "trailing_whitespace_linter"       "vector_logic_linter"
```

<sup>Created on 2022-07-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

But we have many other linters, which are not included in `default_linters` (why?!), and can be seen only* in the [reference](https://lintr.r-lib.org/reference/index.html#individual-linters).

This means that if I wanted to include all available linters, I need to manually inspect the list of default linters and the website, note that down which are not included and then add them to my `.lintr` file. That is cumbersome.

(* There is `lintr:::rd_linterlist()` to list all available linters, but it is not in a form that is useful, and it is not exported.)

So should we implement and export a new `all_linters` list, similar to `default_linters`?

# Linting a package with all available linters

If we implement and export `all_linters`, it also solves the second issue, which is to lint a package will all available linters. Because a user can just do:

```r
with_defaults(..., default = all_linters)
```

I also think that this should be the default, but I am guessing there are good reasons why only certain - and not available - linters were chosen for the default list.

IndrajeetPatil:
Closes #1482


NEWS.md
```diff
@@ -12,16 +12,16 @@
...
-
 ## New and improved features
 
 * New `get_r_string()` helper to get the R-equivalent value of a string, especially useful for R-4-style raw strings.
   Previously an internal `lintr` helper, now exported to facilitate writing custom linters (#1493, @MichaelChirico).
 
 * `object_usage_linter()` improves lint metadata when detecting undefined infix operators, e.g. `%>%` or `:=` (#1497, @MichaelChirico)
 
+* `unused_import_linter()` can detect datasets from imported packages and no longer
+  warns when a package is imported only for its datasets (#1545, @IndrajeetPatil).
```
IndrajeetPatil, line 23: Not related, but minor enough that I decided to include it here instead of making another PR.

README.md
```diff
@@ -38,6 +38,16 @@ usethis::use_github_action("lint")
 lintr::lint_package()
```
IndrajeetPatil, line 38: Do we want to retain this in `README`, or can we get rid of it? Not sure if this is still active.

Screenshot 2022-09-30 at 18 49 48

MichaelChirico: I think we can remove it, AFAIK neither @AshesITR nor I are checking that so not sure how helpful it could be.

vignettes/lintr.Rmd
```diff
@@ -192,6 +192,32 @@ defaults_table <- data.frame(
 knitr::kable(defaults_table)
 ```
 
+#### Using all available linters
+
+The default lintr configuration includes only linters relevant to the tidyverse style guide, but there are many other linters available in `{lintr}`. You can see a list of all available linters using-
```
MichaelChirico, line 197: `-` at the end, is that intentional?

MichaelChirico:
Small things, but LGTM.

README.md
IndrajeetPatil: I have removed it.

MichaelChirico (approved):

PR #1582, opened 2022-10-01
Improve `namespace_imports()` logic

### LINKED ISSUE #1397: namespace_imports logic is wrong for import(foo, except=c(bar, baz))
We currently assume that elements of `$imports` returned by `parseNamespaceFile()` that are length>1 are `importFrom` directives; actually, they might be `import(., except=)` directives.

For example, here's dplyr ([permalink](https://github.com/tidyverse/dplyr/blob/55dfc1c5b49eae3f1b144ebe7d9fa2c3d79b1835/NAMESPACE#L473-L474)):

```r
parseNamespaceFile("dplyr", "..")$imports[2:3]
# [[1]]
# [[1]][[1]]
# [1] "vctrs"
# [[1]]$except
# [1] "data_frame"
# [[2]]
# [[2]][[1]]
# [1] "R6"
# [[2]][[2]]
# [1] "R6Class"
```

 And indeed:

```r
lintr:::namespace_imports("dplyr")[435,]
#       pkg        fun
# 435 vctrs data_frame
```

IndrajeetPatil:
Closes #1397


R/namespace.R
```diff
@@ -1,15 +1,54 @@
...
+  }
+
+  if (is.null(pkg) || length(ns) == 0L) {
     return(empty_namespace_data())
   }
 
-  do.call(rbind, lapply(namespace_data$imports, safe_get_exports))
+  ns[["base"]] <- NULL
+  x <- lapply(ns, named_list_to_df)
+  df <- do.call(rbind.data.frame, c(x, stringsAsFactors = FALSE))
```
MichaelChirico, line 39: just use `rbind` and let dispatch work?

R/namespace.R
```diff
@@ -1,15 +1,54 @@
...
+  df$pkg <- rownames(df)
+  rownames(df) <- NULL
+  transform(df, pkg = gsub(".[[:digit:]]+", "", pkg))
```
MichaelChirico, line 42: what's this step doing?

MichaelChirico (commented):
Missing a test

IndrajeetPatil:
Yeah, this is actually not working yet. Marking it as draft. I don't know how all tests successfully ran locally.

IndrajeetPatil:
Can't seem to get this test to pass. 

```r
expect_lint_free(test_path("dummy_packages", "clean_subdir", "r"))
```

The `NAMESPACE` file is indeed in the "clean_subdir/r" folder, so not sure why this doesn't work. 

@MichaelChirico Any ideas?

R/namespace.R
```diff
@@ -1,15 +1,71 @@
...
-  do.call(rbind, lapply(namespace_data$imports, safe_get_exports))
+  ns_imports[["base"]] <- NULL
+  pkg_names <- names(ns_imports)
+  df <- do.call(
+    rbind.data.frame,
+    c(
+      mapply(ns_imports_list_to_df, ns_imports, pkg_names, SIMPLIFY = FALSE),
+      stringsAsFactors = FALSE
+    )
+  )
```
IndrajeetPatil, line 45: I am fairly certain that the test failures on `R < 4.0` are due to `stringsAsFactors` not being set to `FALSE` here. Something not right the way I am passing `stringsAsFactors` to `rbind()` or `rbind.data.frame()` via `do.call()`.

tests/testthat/test-expect_lint.R
```diff
@@ -60,7 +60,9 @@ test_that("expect_lint_free works", {
   )
   withr::local_envvar(c(NOT_CRAN = "true", R_COVR = "false"))
 
-  expect_lint_free(test_path("dummy_packages", "clean"))
-  expect_lint_free(test_path("dummy_packages", "clean_subdir", "r"))
+  #' expect_lint_free(test_path("dummy_packages", "clean"))
+  #' expect_lint_free(test_path("dummy_packages", "clean_subdir", "r"))
```
IndrajeetPatil, line 64: Just to see if other tests pass. Then need to come back to this issue.

tests/testthat/test-expect_lint.R
```diff
@@ -59,7 +59,23 @@ test_that("expect_lint_free works", {
   )
   withr::local_envvar(c(NOT_CRAN = "true", R_COVR = "false"))
 
-  expect_lint_free(test_path("dummy_packages", "clean"))
-  expect_lint_free(test_path("dummy_packages", "clean_subdir", "r"))
+  withr::with_dir(
+    test_path("dummy_packages", "clean"),
+    {
+      callr::rcmd("INSTALL", ".", echo = FALSE, show = FALSE)
```
IndrajeetPatil, line 65: No need to skip if not installed because we already rely on `{callr}`.

```r
> pak::pkg_deps_explain("lintr", "callr")
✔ Updated metadata database: 4.80 MB in 10 files.                         
ℹ R 4.2 aarch64-apple-darwin20 packages are missing from Bioconductor      
✔ Updating metadata database ... done                                     
lintr -> cyclocomp -> callr 
```

IndrajeetPatil:
@MichaelChirico, @AshesITR See https://github.com/r-lib/lintr/issues/1397#issuecomment-1265234630 with an example showing that this works as expected.

The only thing is I can't seem to get this test to pass. 

```r
expect_lint_free(test_path("dummy_packages", "clean_subdir", "r"))
```

The `NAMESPACE` file is indeed in the "clean_subdir/r" folder, so not sure why this doesn't work.

AshesITR:
Why are you installing the test packages into the actual R library of the testing R version?
This will pollute local development environments with our test packages.

tests/testthat/test-expect_lint.R
MichaelChirico: that may be true now but not guaranteed to hold forever. safer to keep it conditional as long as it's in our own Suggests

IndrajeetPatil: Okay, running the tests conditionally now.

IndrajeetPatil:
> Why are you installing the test packages into the actual R library of the testing R version?
This will pollute local development environments with our test packages.

@AshesITR Yeah, I agree that this is not ideal. But, as opposed to the`parseNamespaceFile()` approach, `getNamespaceImports()` - and in turn `loadNamespace()` - expect the package to be installed. Otherwise, there is **"there is no package called"** error.

I am not sure if the `getNamespaceImports()`-approach taken in this PR is itself doomed, or I am missing something about how to use it with text fixtures. I am open to ideas.

AshesITR:
> > Why are you installing the test packages into the actual R library of the testing R version?
> > This will pollute local development environments with our test packages.
> 
> @AshesITR Yeah, I agree that this is not ideal. But, as opposed to the`parseNamespaceFile()` approach, `getNamespaceImports()` - and in turn `loadNamespace()` - expect the package to be installed. Otherwise, there is **"there is no package called"** error.

Also note that `loadNamespace()` causes `.onLoad()` of said package to run, which can do all kinds of (expensive) things, e.g. start a JVM.

> I am not sure if the `getNamespaceImports()`-approach taken in this PR is itself doomed, or I am missing something about how to use it with text fixtures. I am open to ideas.

Due to the possible downsides, I'd favor an approach that works with `parseNamespaceFile()`. I think it's worth keeping this PR open in case we come to the conclusion that the `parseNamespaceFile()` based approach is not feasible.

AshesITR:
Removing myself for now. Please let me know if further input is necessary.

IndrajeetPatil:
> Also note that loadNamespace() causes .onLoad() of said package to run, which can do all kinds of (expensive) things, e.g. start a JVM.

Efficiency comes secondary to accuracy. The current approach is inaccurate because it completely disregards the exclude directives, while the approach in the current PR respects them. 

But maybe this issue is not worth the hassle because very few packages in the wild use this directive anyway.

Issue #1529, opened 2022-09-06
Object not found error in `lint_package()`


IndrajeetPatil:
## Issue

Encountered this while linting the [dm](https://github.com/cynkra/dm/) package. 

Is it because of [this](https://github.com/cynkra/dm/blob/bcdfa3f76cc74e502ce9f30d5452e9a5e1a59cb8/NAMESPACE#L348) directive? Is this something we should support/anticipate?

```r
import(glue, except = c(collapse))
```

## Reprex

``` r
library(lintr)
lint_package("/Users/indrajeetpatil/Documents/GitHub/dm")
#> Error in get(ns_imports$fun[i], envir = asNamespace(ns_imports$pkg[i])): object 'collapse' not found
```

<sup>Created on 2022-09-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Traceback:

```r
> rlang::last_trace()
<error/rlang_error>
Error:
! object 'collapse' not found
---
Backtrace:
     ▆
  1. └─lintr::lint_package()
  2.   └─lintr::lint_dir(...) at lintr/R/lint.R:220:2
  3.     ├─lintr:::flatten_lints(...) at lintr/R/lint.R:147:2
  4.     │ ├─base::structure(flatten_list(x, class = "lint"), class = "lints") at lintr/R/utils.R:18:2
  5.     │ └─lintr:::flatten_list(x, class = "lint") at lintr/R/utils.R:18:2
  6.     │   └─lintr (local) assign_item(x) at lintr/R/utils.R:38:2
  7.     └─base::lapply(...) at lintr/R/lint.R:147:2
  8.       └─lintr (local) FUN(X[[i]], ...)
  9.         └─lintr::lint(file, ..., parse_settings = FALSE, exclusions = exclusions) at lintr/R/lint.R:151:6
 10.           └─lintr:::get_lints(...) at lintr/R/lint.R:81:8
 11.             ├─lintr:::flatten_lints(linter_fun(expr)) at lintr/R/lint.R:256:4
 12.             │ ├─base::structure(flatten_list(x, class = "lint"), class = "lints") at lintr/R/utils.R:18:2
 13.             │ └─lintr:::flatten_list(x, class = "lint") at lintr/R/utils.R:18:2
 14.             │   └─lintr (local) assign_item(x) at lintr/R/utils.R:38:2
 15.             └─lintr (local) linter_fun(expr) at lintr/R/lint.R:256:4
[...]

PR #1583, opened 2022-10-01
friendlier failure of lint()

### LINKED ISSUE #1459: lintr usage of `glue` failing on `lint_package()`
`lint_package()` is currently failing, as used here:

https://github.com/r-lib/actions/blob/v2-branch/examples/lint.yaml

The error message:

```
Run lintr::lint_package()
Error in glue_data(.x = NULL, ..., .sep = .sep, .envir = .envir, .open = .open,  : 
  Expecting '}'
Calls: <Anonymous> ... extract_glued_symbols -> eval -> eval -> <Anonymous> -> glue_data
Execution halted
```

Potentially related to some recent changes in glue (the `.literal` argument):

https://github.com/tidyverse/glue/issues/259#issuecomment-1035205742

MichaelChirico:
Inspired by #1459 

This adds some small overhead to `lint()` (O(microseconds) per linter per file), but I think not enough to be noticed in typical usage.


tests/testthat/test-lint_package.R
```diff
@@ -134,3 +134,13 @@ test_that("lint_package returns early if no package is found", {
...
   )
 })
+
+test_that("lint_package gives helpful failure when package has syntax issues", {
+  pkg_path <- test_path("dummy_packages", "err_pkg")
+  withr::local_options(lintr.linter_file = "lintr_test_config")
+
+  expect_error(
+    lint_package(pkg_path, linters = list(object_usage_linter(), assignment_linter())),
+    "Linter object_usage_linter\\(\\) failed in.*glue_error.R: Expecting '}'"
```
IndrajeetPatil, line 144: Is it possible to get a cleaner filename? E.g.

> "Linter object_usage_linter\\(\\) failed in glue_error.R: Expecting '}'"

AshesITR:
Can we make `extract_glued_symbols()` more resilient instead? I think syntax errors within linted code should at the most produce lints or warnings and not stop lintr.

Also note this tryCatch will destroy the stack trace and make debugging more cumbersome when we encounter new problems. Not a big fan of this solution.

MichaelChirico:
But this is not a problem limited to #1459... off the top of my head #1529, #1443, #1340. All would have been helped by getting a more helpful failure from `lint()`.

Of course we don't expect the linters to fail, but when they do, it becomes a mess.

> Can we make extract_glued_symbols() more resilient instead?

I think we can do that separately, in addition to the fix suggested here

> Also note this tryCatch will destroy the stack trace and make debugging more cumbersome when we encounter new problems.

Hmm I think the most important thing is knowing which linter & file is responsible, then we can `debug()` that linter in a subsequent call.

However, we can also consider using `withCallingHandlers()` instead; quickly, the `sys.calls()` from swapping `tryCatch()` to `withCallingHandlers()` for the issue in #1459 is:

```r
[[1]]
lint("tests/testthat/dummy_packages/err_pkg/R/glue_error.R", 
    object_usage_linter())

[[2]]
withCallingHandlers(get_lints(expr, linter, linters[[linter]], 
    lint_cache, source_expressions$lines), error = function(cond) {
    browser()
    stop("Linter ", linter, "() failed in ", filename, ": ", 
        conditionMessage(cond), call. = FALSE)
})

[[3]]
get_lints(expr, linter, linters[[linter]], lint_cache, source_expressions$lines)

[[4]]
flatten_lints(linter_fun(expr))

[[5]]
structure(flatten_list(x, class = "lint"), class = "lints")

[[6]]
flatten_list(x, class = "lint")

[[7]]
assign_item(x)

[[8]]
linter_fun(expr)

[[9]]
lapply(fun_assignments, function(fun_assignment) {
    code <- get_content(lines = source_expression$content, fun_assignment)
    fun <- try_silently(eval(envir = env, parse(text = code, 
        keep.source = TRUE)))
    if (inherits(fun, "try-error")) {
        return()
    }
    known_used_symbols <- get_used_symbols(fun_assignment, interpret_glue = interpret_glue)
    res <- parse_check_usage(fun, known_used_symbols = known_used_symbols, 
        declared_globals = declared_globals, start_line = as.integer(xml2::xml_attr(fun_assignment, 
            "line1")), skip_with = skip_with)
    res$name <- rex::re_substitutes(res$name, rex::rex("<-"), 
        "")
    lintable_symbols <- xml2::xml_find_all(fun_assignment, "\n        descendant::SYMBOL\n        | descendant::SYMBOL_FUNCTION_CALL\n        | descendant::SPECIAL\n        | descendant::LEFT_ASSIGN[text() = ':=']\n        ")
    lintable_symbol_names <- gsub("^`|`$", "", xml2::xml_text(lintable_symbols))
    lintable_symbol_lines <- as.integer(xml2::xml_attr(lintable_symbols, 
        "line1"))
    matched_symbol <- vapply(seq_len(nrow(res)), function(i) {
        match(TRUE, lintable_symbol_names == res$name[i] & lintable_symbol_lines >= 
            res$line1[i] & lintable_symbol_lines <= res$line2[i])
    }, integer(1L))
    nodes <- unclass(lintable_symbols)[matched_symbol]
    nodes[is.na(matched_symbol)] <- list(fun_assignment)
    xml_nodes_to_lints(nodes, source_expression = source_expression, 
        lint_message = res$message, type = "warning")
})

[[10]]
FUN(X[[i]], ...)

[[11]]
get_used_symbols(fun_assignment, interpret_glue = interpret_glue)

[[12]]
extract_glued_symbols(expr)

[[13]]
eval(parsed_cl)

[[14]]
eval(parsed_cl)

[[15]]
glue::glue("Groups found: {glue::glue_collapse(purrr::map_chr(remote_groups$results, ~ .x$name), sep = \", \")", 
    .transformer = function (text, envir) 
    {
        parsed_text <- tryCatch(parse(text = text, keep.source = TRUE), 
            error = function(...) NULL, warning = function(...) NULL)
        parsed_xml <- safe_parse_to_xml(parsed_text)
        if (length(parsed_xml) == 0L) {
            return("")
        }
        symbols <- xml2::xml_text(xml2::xml_find_all(parsed_xml, 
            "//SYMBOL"))
        for (sym in symbols) {
            assign(sym, NULL, envir = glued_symbols)
        }
        ""
    })

[[16]]
glue_data(.x = NULL, ..., .sep = .sep, .envir = .envir, .open = .open, 
    .close = .close, .na = .na, .null = .null, .comment = .comment, 
    .literal = .literal, .transformer = .transformer, .trim = .trim)

[[17]]
.handleSimpleError(function (cond) 
{
    browser()
    stop("Linter ", linter, "() failed in ", filename, ": ", 
        conditionMessage(cond), call. = FALSE)
}, "Expecting '}'", base::quote(glue_data(.x = NULL, ..., .sep = .sep, 
    .envir = .envir, .open = .open, .close = .close, .na = .na, 
    .null = .null, .comment = .comment, .literal = .literal, 
    .transformer = .transformer, .trim = .trim)))

[[18]]
h(simpleError(msg, call))
```

There's still some noise in there, but all the important bits are there (esp., the final failure on `glue_data()`). Do you think that's worth pursuing?

MichaelChirico:
OK, for now I simply swapped `tryCatch`-->`withCallingHandlers` -- seems the simplest route / won't require too much pecking around in the stack trace to remove less-relevant bits.

R/lint.R
```diff
@@ -85,7 +85,12 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
   if (!is_tainted(source_expressions$lines)) {
     for (expr in source_expressions$expressions) {
       for (linter in names(linters)) {
-        lints[[length(lints) + 1L]] <- get_lints(expr, linter, linters[[linter]], lint_cache, source_expressions$lines)
+        lints[[length(lints) + 1L]] <- withCallingHandlers(
+          get_lints(expr, linter, linters[[linter]], lint_cache, source_expressions$lines),
+          error = function(cond) {
+            stop("Linter ", linter, "() failed in ", filename, ": ", conditionMessage(cond), call. = FALSE)
```
AshesITR, line 91: Would drop the (), because custom linter naming would cause misleading error messages.

AshesITR:
I think this is wort a NEWS bullet. Otherwise mostly good to merge. Thanks!

tests/testthat/test-lint_package.R
MichaelChirico: `basename()` will work, but may obscure in which directory the file is found. Within the linter, I'm not sure it's easy to establish what the package root was. Leaving the full name for now.

R/lint.R
MichaelChirico: Good point, thanks

MichaelChirico:
PTAL at the NEWS bullet, in particular if you agree that it's usually a bug if a linter fails

IndrajeetPatil (approved):

PR #1590, opened 2022-10-02
Move more XPaths away from //expr logic

MichaelChirico:
Towards #1358 

Most files have one (or at most two) left, so starting to batch them together

### DIFF
--- a/R/conjunct_test_linter.R
+++ b/R/conjunct_test_linter.R
@@ -13,19 +13,27 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 conjunct_test_linter <- function(allow_named_stopifnot = TRUE) {
+  expect_true_assert_that_xpath <- "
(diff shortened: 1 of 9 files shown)


R/conjunct_test_linter.R
```diff
@@ -13,19 +13,27 @@
...
+  stopifnot_xpath <- glue::glue("
+  //SYMBOL_FUNCTION_CALL[text() = 'stopifnot']
+  /parent::expr
+  /following-sibling::expr[1][AND2 {named_stopifnot_condition}]
+  ")
+  expect_false_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'expect_false']
+  /parent::expr
+  /following-sibling::expr[1][OR2]
+  "
```
IndrajeetPatil, line 31: Very nice! 

I wish we'd use such descriptive names more often for XPath variables than just the generic `xpath`, but that's for a later date.

IndrajeetPatil (approved):

Issue #1492, opened 2022-08-01
Adding examples to documentation of each linter


IndrajeetPatil:
I think docs for each linter should have an examples section with at least one example of code that does produce lint and another that doesn't.

I think the style guide is good to provide the underlying rationale for the lint (provided that the reader actually clicks on that link!), but it doesn't lead to much exploration on the reader's part. Giving concrete examples means that the reader can just copy-paste examples and start toying around with them to see what does and doesn't lint.

Note to self: Make sure all of them follow the style guide, and so use `"` instead of `'`, e.g.

## Progress tracker

 - [x] absolute_path_linter
 - [x] any_duplicated_linter
 - [x] any_is_na_linter
 - [x] assignment_linter
 - [x] backport_linter
 - [x] boolean_arithmetic_linter
 - [x] brace_linter
 - [x] class_equals_linter
 - [x] commas_linter
 - [x] commented_code_linter
 - [x] condition_message_linter
 - [x] conjunct_test_linter
 - [x] consecutive_stopifnot_linter
 - [x] cyclocomp_linter
 - [x] duplicate_argument_linter
 - [x] equals_na_linter
 - [x] expect_comparison_linter
 - [x] expect_identical_linter
 - [x] expect_length_linter
 - [x] expect_named_linter
 - [x] expect_not_linter
 - [x] expect_null_linter
 - [x] expect_s3_class_linter
 - [x] expect_s4_class_linter
 - [x] expect_true_false_linter
 - [x] expect_type_linter
 - [x] extraction_operator_linter
 - [x] fixed_regex_linter
 - [x] function_argument_linter
 - [x] function_left_parentheses_linter
 - [x] function_return_linter
 - [x] ifelse_censor_linter
 - [x] implicit_integer_linter
 - [x] infix_spaces_linter
 - [x] inner_combine_linter
 - [x] lengths_linter
 - [x] line_length_linter
 - [x] literal_coercion_linter
 - [x] missing_argument_linter
 - [x] missing_package_linter
 - [x] namespace_linter
 - [x] nested_ifelse_linter
 - [x] no_tab_linter
 - [ ] nonportable_path_linter
 - [x] numeric_leading_zero_linter
 - [x] object_length_linter
 - [x] object_name_linter
 - [x] object_usage_linter
 - [x] outer_negation_linter
 - [x] package_hooks_linter
 - [x] paren_body_linter
 - [x] paste_linter
 - [x] pipe_call_linter
[...]

PR #1591, opened 2022-10-02
Add examples to documentation: Part-1 (a-b)

IndrajeetPatil:
Part of #1492

### DIFF
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -5,7 +5,7 @@
 #' @examples
 #' library(lintr)
 #'
-#' # will produce a lint
+#' # will produce lints
 #' lint(
 #'   text = "x <- T; y <- F",
 #'   linters = T_and_F_symbol_linter()
@@ -16,7 +16,7 @@
 #'   linters = T_and_F_symbol_linter()
 #' )
 #'
-#' # recommended
+#' # okay
 #' lint(
 #'   text = "x <- c(TRUE, FALSE)",
 #'   linters = T_and_F_symbol_linter()
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -3,8 +3,6 @@
 #' Avoid the symbols `T` and `F`, and use `TRUE` and `FALSE` instead.
 #'
 #' @examples
-#' library(lintr)
-#'
 #' # will produce lints
 #' lint(
 #'   text = "x <- T; y <- F",
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -1,6 +1,31 @@
 #' `T` and `F` symbol linter
 #'
-#' Avoid the symbols `T` and `F` (for `TRUE` and `FALSE`).
+#' Avoid the symbols `T` and `F`, and use `TRUE` and `FALSE` instead.
+#'
+#' @examples
+#' library(lintr)
+#'
+#' # will produce a lint
+#' lint(
+#'   text = "x <- T; y <- F",
+#'   linters = T_and_F_symbol_linter()
+#' )
+#'
+#' lint(
+#'   text = "T = 1.2, F = 2.4",
+#'   linters = T_and_F_symbol_linter()
+#' )
+#'
+#' # recommended
+#' lint(
+#'   text = "x <- c(TRUE, FALSE)",
+#'   linters = T_and_F_symbol_linter()
+#' )
+#'
+#' lint(
+#'   text = "t = 1.2; f = 2.4",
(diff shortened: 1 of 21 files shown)


R/any_is_na_linter.R
```diff
@@ -4,6 +4,31 @@
...
+#' )
+#'
+#' lint(
+#'   text = "any(is.na(foo(x)))",
+#'   linters = any_is_na_linter()
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "anyNA(x)",
```
MichaelChirico, line 23: I would use an example of any(!is.na(x))

R/assignment_linter.R
```diff
@@ -6,6 +6,32 @@
 #'   If `FALSE`, [`<<-`][base::assignOps] and `->>` are not allowed.
 #' @param allow_right_assign Logical, default `FALSE`. If `TRUE`, `->` and `->>` are allowed.
 #' @param allow_trailing Logical, default `TRUE`. If `FALSE` then assignments aren't allowed at end of lines.
+#'
+#' @examples
+#' library(lintr)
+#'
+#' # will produce lints
+#' lint(
```
MichaelChirico, line 14: for ones with parameters I think it's more important to showcase their effects

R/brace_linter.R
```diff
@@ -12,6 +12,36 @@
...
+#' library(lintr)
+#'
+#' # will produce lints
+#' lint(
+#'   text = "f <- function() { 1 }",
+#'   linters = brace_linter()
+#' )
+#'
+#' lint(
+#'   text = "if (TRUE) {
```
MichaelChirico, line 25: feels _slightly_ misleading since the indentation of the actual code as parsed won't match what you see visually, but I don't have a good answer for how to fix it... WDYT

R/T_and_F_symbol_linter.R
```diff
@@ -1,6 +1,31 @@
 #' `T` and `F` symbol linter
 #'
-#' Avoid the symbols `T` and `F` (for `TRUE` and `FALSE`).
+#' Avoid the symbols `T` and `F`, and use `TRUE` and `FALSE` instead.
+#'
+#' @examples
+#' library(lintr)
+#'
```
MichaelChirico, line 7: I wouldn't use library(lintr) in the examples, it shouldn't be necessary, is that recommended somewhere?

IndrajeetPatil: No, not recommended anywhere. 

My thinking was that the users can basically copy-paste the examples and run it with no addition.
Otherwise, they run it, realize they forgot to `library(lintr)`.

But, if there is some downside to doing so, can easily remove these calls.

MichaelChirico: The same's true of any package, though, so including `library(lintr)` may seem puzzling.

Users running examples _via_ `example()` will have no such issue -- `example()` will always call `library()`:

https://github.com/r-devel/r-svn/blob/e8334d998b8a3185e490f488629a7a2cc7e3825d/src/library/utils/R/example.R#L83-L84

R/any_is_na_linter.R
IndrajeetPatil: Done.

R/T_and_F_symbol_linter.R
IndrajeetPatil: Okay, fair enough. 

Also, at least in RStudio, you can run the examples in the IDE itself, so they won't even need to copy-paste anything.

I have removed these calls.

R/brace_linter.R
IndrajeetPatil: Yeah, that's a good point.

How about we first show them what the code looks like and then use that text in the `lint()` call?

``` r
library(lintr)

cat("if (TRUE) { \nreturn(1) }")
#> if (TRUE) { 
#> return(1) }

lint(
  text = "if (TRUE) { \nreturn(1) }",
  linters = brace_linter()
)
#> <text>:2:11: style: [brace_linter] Closing curly-braces should always be on their own line, unless they are followed by an else.
#> return(1) }
#>           ^
```

MichaelChirico: good idea, LGTM

IndrajeetPatil: Done.

R/assignment_linter.R
IndrajeetPatil: I have added examples of parameter usage, except `absolute_path_linter(lax = FALSE)`. Can't come up with a good example for that one.

MichaelChirico: SGTM, I think of this one as more peripheral anyway :sweat_smile:

MichaelChirico (approved):

PR #1596, opened 2022-10-03
Get rid of a few lints

IndrajeetPatil:


R/lint.R
```diff
@@ -640,16 +640,16 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
     as.character(utils::packageVersion("lintr"))
   sarif$runs[[1L]]$originalUriBaseIds$ROOTPATH$uri <- ""
   rule_index_exists <- FALSE
-  root_path_uri <- gsub("\\\\", "/", package_path)
+  root_path_uri <- gsub("\\\\", "/", package_path, fixed = TRUE)
```
MichaelChirico, line 643: please double-check it's doing the same thing, I think the fixed equivalent should be just `"\\"`

R/lint.R
```diff
@@ -640,16 +640,16 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
...
 
   if (startsWith(root_path_uri, "/")) {
-    root_path_uri <- paste("file://", root_path_uri, sep = "")
```
MichaelChirico, line 646: wow i'm amazed these survived so long :flushed:

R/lint.R
```diff
@@ -691,7 +691,7 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
       append(one_result, list(message = list(text = lint$message)))
     one_location <- list(physicalLocation = list(
       artifactLocation = list(
-        uri = gsub("\\\\", "/", lint$filename),
+        uri = gsub("\\\\", "/", lint$filename, fixed = TRUE),
```
MichaelChirico, line 694: ditto

R/backport_linter.R
```diff
@@ -13,7 +13,7 @@
 backport_linter <- function(r_version = getRversion(), except = character()) {
   r_version <- normalize_r_version(r_version)
   backport_blacklist <- backports[r_version < R_system_version(names(backports))]
-  backport_blacklist <- lapply(backport_blacklist, function(nm) setdiff(nm, except))
+  backport_blacklist <- lapply(backport_blacklist, function(nm) setdiff(nm, except)) # nolint: unnecessary_lambda_linter
```
IndrajeetPatil, line 16: The suggested, no lint version is less readable, IMO. Better to just skip here.

``` r
library(lintr)

r_version <- lintr:::normalize_r_version("4.0.0")
except <- character()
backports <- lintr:::backports
backport_blacklist <- backports[r_version < R_system_version(names(backports))]

lint(
  text = "lapply(backport_blacklist, function(nm) setdiff(nm, except))",
  linters = unnecessary_lambda_linter()
)
#> <text>:1:28: warning: [unnecessary_lambda_linter] Pass setdiff directly as a symbol to lapply() instead of wrapping it in an unnecessary anonymous function. For example, prefer lapply(DF, sum) to lapply(DF, function(x) sum(x)).
#> lapply(backport_blacklist, function(nm) setdiff(nm, except))
#>                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

lint(
  text = "lapply(backport_blacklist, setdiff, y = except)",
  linters = unnecessary_lambda_linter()
)

identical(
  lapply(backport_blacklist, function(nm) setdiff(nm, except)),
  lapply(backport_blacklist, setdiff, y = except)
)
#> [1] TRUE
```

MichaelChirico: `lapply(backport_blacklist, setdiff, except)` looks fine to me. But I'm indifferent.

R/lint.R
IndrajeetPatil: Haha, true. This is why I am adding more linters to our lint changed GHA workflow. 
We should be able to catch a few more of such instances in the future.

MichaelChirico (approved):

PR #1597, opened 2022-10-03
Improve resilience of object_usage_linter() to glue syntax issues

### LINKED ISSUE #1459: lintr usage of `glue` failing on `lint_package()`

MichaelChirico:
Closes #1459

This invalidated the earlier test added in #1583, but we can test it more directly / without a fake package, as done here.

### DIFF
--- a/tests/testthat/test-lint_file.R
+++ b/tests/testthat/test-lint_file.R
@@ -225,3 +225,16 @@ test_that("Deprecated positional usage of cache= works, with warning", {
   )
   expect_identical(l, lint("a = 2\n", assignment_linter(), cache = FALSE))
 })
+
+test_that("Linters throwing an error give a helpful error", {
+  tmp_file <- withr::local_tempfile(lines = "a <- 1")
+  linter <- function() Linter(function(source_expression) stop("a broken linter"))
+  expect_error(
+    lint(tmp_file, linter()),
+    paste0("Linter 'linter' failed in ", tmp_file, ": a broken linter")
+  )
+  expect_error(
+    lint(tmp_file, list(broken_linter = linter())),
+    paste0("Linter 'broken_linter' failed in ", tmp_file, ": a broken linter")
+  )
+})
--- a/tests/testthat/test-lint_file.R
+++ b/tests/testthat/test-lint_file.R
@@ -7,7 +7,6 @@ lint_assignments <- function(filename) {
 }
 
 test_that("lint() results do not depend on the working directory", {
-
   # a dummy package for use in the test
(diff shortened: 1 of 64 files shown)


tests/testthat/test-lint_file.R
```diff
@@ -225,3 +225,18 @@ test_that("Deprecated positional usage of cache= works, with warning", {
...
 })
+
+test_that("Linters throwing an error give a helpful error", {
+  tmp_file <- withr::local_tempfile(lines = "a <- 1")
+  linter <- function() Linter(function(source_expression) stop("a broken linter"))
+  # NB: Some systems/setups may use e.g. symlinked files when creating under tempfile();
+  #   we don't care much about that, so just check basename()
+  expect_error(
+    lint(tmp_file, linter()),
+    paste0("Linter 'linter' failed in .*", basename(tmp_file), ": a broken linter")
```
AshesITR, line 236: `basename(...)` could need regex escaping. Use `rex::rex()` instead of `paste0()`?

MichaelChirico:
Hey @IndrajeetPatil another reason to please resist being trigger-happy with the `Update branch` button -- it makes interacting with the remote more tedious. I would need to remember `git pull origin $BRANCH` every time I try to push if the branch is regularly getting merges from main.

I'd rather handle merges to main locally if needed, then push directly to remote.

Basically as long as there's no merge conflict, I don't see any need to do an Update until just before the PR is merged.

tests/testthat/test-lint_file.R
MichaelChirico: I don't think it's true because R generates temp files from alphanumeric characters, but it can't hurt to use `rex()` instead. done.

IndrajeetPatil:
Oops, sorry about that.

I thought this was ready for a merge since you made a commit to address Alexander's comment. But I was mistaken about that.

I will refrain from updating anyone else's PR in the future to avoid this.

AshesITR (approved):

PR #1599, opened 2022-10-03
Add additional test for `unnecessary_lambda_linter()`

IndrajeetPatil:


MichaelChirico:
oh, I'm surprised this passes. can you check TODOs in the linter code, I seem to recall there's a TODO about exactly this test case...

IndrajeetPatil:
This one?

https://github.com/r-lib/lintr/blob/75882b5a46ebacdbe0a2ec898f726fe29e7bc1f5/R/unnecessary_lambda_linter.R#L31-L34

MichaelChirico:
yes, shouldn't this condition be failing your new test 😵‍💫

https://github.com/r-lib/lintr/blob/75882b5a46ebacdbe0a2ec898f726fe29e7bc1f5/R/unnecessary_lambda_linter.R#L45

IndrajeetPatil:
The test should fail, right? 

Here `count(SYMBOL_FORMALS)` is `2`, and so `count(SYMBOL_FORMALS) = 1` condition is not met and xpath doesn't catch anything. So no lint.

tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -12,6 +12,9 @@ test_that("unnecessary_lambda_linter skips allowed usages", {
   # at any nesting level
   expect_lint("parLapply(cl, x, function(xi) foo(xi, 2, bar(baz(xi))))", NULL, linter)
 
+  # multi-argument case
+  expect_lint("lapply(l, function(x, y) setdiff(x, y))", NULL, linter)
```
MichaelChirico, line 16: oh I see now. I wouldn't add this test. the code as written can't run (y has no default), and any working version can be replaced without lambda.

no sense adding an incorrect test

IndrajeetPatil: Can replace it with this one:

``` r
l <- list("x" = 1:3, "y" = c(1:3, NA))

lapply(l, function(x, na.rm = TRUE) mean(x, na.rm = na.rm))
#> $x
#> [1] 2
#> 
#> $y
#> [1] 2
```

tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -13,7 +13,7 @@ test_that("unnecessary_lambda_linter skips allowed usages", {
   expect_lint("parLapply(cl, x, function(xi) foo(xi, 2, bar(baz(xi))))", NULL, linter)
 
   # multi-argument case
-  expect_lint("lapply(l, function(x, y) setdiff(x, y))", NULL, linter)
+  expect_lint("lapply(l, function(x, na.rm = TRUE) mean(x, na.rm = na.rm))", NULL, linter)
```
MichaelChirico, line 16: I still don't see why you'd need a lambda here.

lapply(l, mean, na.rm=TRUE) does the same.

IndrajeetPatil: Fair enough.

PR #1602, opened 2022-10-03
Move deprecated linters to one file and hide them on website

### LINKED ISSUE #1601: Move all deprecated linters to a single file and hide them on the website
Move deprecated linters to `lintr-deprecated.R`.

Leave them out of the website using the roxygen tag:

```r
#' @keywords internal
```

IndrajeetPatil:
Closes #1601


R/lintr-deprecated.R
```diff
@@ -0,0 +1,186 @@
+#' Deprecated functions in lintr
+#'
+#' These functions have either been deprecated from lintr.
```
MichaelChirico, line 3: grammar?

R/lintr-deprecated.R
```diff
@@ -0,0 +1,186 @@
...
+#' an else, a comma, or a closing bracket.
+#'
+#' @param allow_single_line if `TRUE`, allow an open and closed curly pair on the same line.
+#' @evalRd rd_tags("closed_curly_linter")
+#' @seealso
+#'   [linters] for a complete list of linters available in lintr. \cr
+#'   <https://style.tidyverse.org/syntax.html#indenting>
+#' @keywords internal
+#' @export
+closed_curly_linter <- function(allow_single_line = FALSE) {
```
MichaelChirico, line 21: We should put them in the same .Rd as well

IndrajeetPatil: Sounds good.

What about tests? Should they also be moved to a single file?

MichaelChirico: That part's not as important, I'd leave them where they are to make the git history cleaner.

IndrajeetPatil: Okay, I personally like to keep `test-*.R` files in sync with `.R` files so that if one is removed, then the test file can be removed as well. But I think this is definitely a matter of personal taste.

Retaining the tests in separate files as before.

MichaelChirico (approved):

IndrajeetPatil:
I forgot about `with_defaults()`, but I am not sure if it actually fits with the rest of the deprecated functions in this file. Maybe better to leave it where it is now?

PR #1603, opened 2022-10-03
Add examples to documentation: Part-2 (c-d)

IndrajeetPatil:


R/class_equals_linter.R
```diff
@@ -5,7 +5,30 @@
...
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "is_lm <- inherits(x, 'lm')",
+#'   linters = class_equals_linter()
+#' )
+#'
+#' lint(
+#'   text = "if (inherits(x, 'character')) stop('no')",
```
MichaelChirico, line 29: I wouldn't use this, `inherits(x, "character")` is itself an antipattern... use `is.character()` instead?

R/commas_linter.R
```diff
@@ -2,6 +2,34 @@
...
+#' )
+#'
+#' lint(
+#'   text = "mean(x,trim = 0.2,na.rm = TRUE)",
+#'   linters = commas_linter()
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "switch(op, x = foo, y = bar)",
```
MichaelChirico, line 19: since we're using `switch()` already, please add a case with an empty switch argument to show how it's done (`switch(op, x = , y = foo)`)

R/commas_linter.R
```diff
@@ -2,6 +2,34 @@
...
+#'   linters = commas_linter()
+#' )
+#'
+#' lint(
+#'   text = "mean(x, trim = 0.2, na.rm = TRUE)",
+#'   linters = commas_linter()
+#' )
+#'
+#' lint(
+#'   text = "a[1, , 2, , 3]",
```
MichaelChirico, line 29: `[]` is a funny one with this linter, would help to add a "bad" example as well, e.g. `x[,, drop=TRUE]`

R/condition_message_linter.R
```diff
@@ -1,9 +1,36 @@
 #' Block usage of `paste()` and `paste0()` with messaging functions using `...`
 #'
-#' `stop(paste0(...))` is strictly redundant -- `stop(...)` is equivalent.
-#'   `stop(...)` is also preferable to `stop(paste(...))`. The same applies to
-#'   all default condition functions, i.e., [stop()], [warning()], [message()],
-#'   and [packageStartupMessage()].
+#' Both `stop(paste0(...))` and `stop(paste(...))` are strictly redundant -- `stop(...)` is
```
MichaelChirico, line 3: Don't like the rewording... the point is that `stop(paste0(...))` is _exactly_ equivalent to `stop(...)`, but it's not quite so for `stop(paste(...))` because we have to add extra whitespace to account for `sep = " "`, so "strictly redundant" is too strong.

Maybe reword with that in mind?

R/condition_message_linter.R
```diff
@@ -1,9 +1,36 @@
...
+#' )
+#'
+#' lint(
+#'   text = "warning(paste0('a string', ' another'))",
+#'   linters = condition_message_linter()
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "stop('a string', 'another')",
```
MichaelChirico, line 21: add the extra space here as mentioned so that the bad/good calls are more correspondent

R/cyclocomp_linter.R
```diff
@@ -4,6 +4,20 @@
 #'
 #' @param complexity_limit Maximum cyclomatic complexity, default 15. Expressions more complex
 #' than this are linted. See [cyclocomp::cyclocomp()].
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "if(TRUE) 1 else 2",
```
MichaelChirico, line 11: nit: isn't the spacing rule `if (TRUE) 1 else 2`?

tests/testthat/test-condition_message_linter.R
```diff
@@ -38,6 +44,25 @@ test_that("condition_message_linter blocks simple disallowed usages", {
...
   )
 
+  # `sep` argument allowed, but only if it is different from default
+  expect_lint(
+    "stop(paste(x, sep = ' '))",
+    rex::rex("Don't use paste to build stop strings."),
+    condition_message_linter()
+  )
+
+  expect_lint(
```
MichaelChirico, line 54: i don't think we need three tests for this, unless it's being done with a `patrick` parameterized test

R/condition_message_linter.R
IndrajeetPatil: I have reworded this, but not sure if I have stuck the landing with wording. Happy to give it another try.

R/condition_message_linter.R
```diff
@@ -1,9 +1,39 @@
 #' Block usage of `paste()` and `paste0()` with messaging functions using `...`
 #'
-#' `stop(paste0(...))` is strictly redundant -- `stop(...)` is equivalent.
-#'   `stop(...)` is also preferable to `stop(paste(...))`. The same applies to
-#'   all default condition functions, i.e., [stop()], [warning()], [message()],
-#'   and [packageStartupMessage()].
+#' This is because
```
MichaelChirico, line 3: I'd avoid a pronoun (`This`) to start the documentation, instead spell out what `This` is (note that in the actual Rd, this sentence is not connected to the title):

> This linter discourages combining condition functions like [stop()] with string concatenation functions [paste()] and [paste0()]. This is because.

Then below we don't need to link [stop()]

MichaelChirico (dismissed):

tests/testthat/test-condition_message_linter.R
IndrajeetPatil: I have added parametrized tests now.

MichaelChirico (dismissed):

MichaelChirico (approved):

PR #1610, opened 2022-10-03
switch to using a hanging indent for extended XPaths

MichaelChirico:
Switch from extended xpaths like

```xpath
//NODE[ conditions]
/parent::expr[ conditions ]
/parent::expr[ conditions ]
```

To use hanging indent for changing axes:

```xpath
//NODE[ conditions]
  /parent::expr[ conditions ]
  /parent::expr[ conditions ]
```

I think it's more readable this way, especially when there's multiple XPaths combined with `|`, or there's a `//` sub-path (as in `package_hooks_linter()`).

### DIFF
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -33,25 +33,25 @@
 T_and_F_symbol_linter <- function() { # nolint: object_name.
   xpath <- paste0(
     "//SYMBOL[",
-    "(text() = 'T' or text() = 'F')", # T or F symbol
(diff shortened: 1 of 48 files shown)


MichaelChirico:
In case it helps the reviewer, here's a link to the diff ignoring whitespace differences:

https://github.com/r-lib/lintr/pull/1610/files?w=1

IndrajeetPatil (approved):
Definitely improves the readability! Thanks.

Btw, do you format the `XPath` strings manually, or is there something equivalent of `{styler}` for `XPath`?

MichaelChirico:
Manually, I don't know of any XPath style guide and a quick search didn't turn up anything either.

PR #1612, opened 2022-10-04
improve handling of glued object extraction

MichaelChirico:
1. Pull declaration of the `.transformer` function outside the `for (cl in glue_calls)` loop
2. Use the `.envir` environment to remove the dependency of the `.transformer` on globals inside `extract_glued_symbols`
3. Operate directly on the `getParseData()` output for this symbol case rather than converting to XML and then searching an XPath
4. Use `names(env)` over `ls(env, all.names=TRUE)` which I found is somehow [20x faster](https://twitter.com/michael_chirico/status/1577129104515530753)

I tried testing the performance on `dplyr`, but it's mostly using `glue()` unqualified, so I haven't had any luck yet. We might try loosening that restriction.


MichaelChirico:
Benchmark-ish. Did a find-and-replace in one dplyr file for `glue(` --> `glue::glue(` to ensure `object_usage_linter()` is actually evaluating `glue()` calls. Then ran `lint()` 40 times:

```r
# on main
system.time(replicate(40, simplify=FALSE, lint("dplyr/R/join-by.R", object_usage_linter())))
#     user  system elapsed
#   63.096   0.331  63.701

# on this branch
system.time(replicate(40, simplify=FALSE, lint("dplyr/R/join-by.R", object_usage_linter())))
#     user  system elapsed
#   59.484   0.204  59.891
```

So a small speed-up (5-6%) in typical usages. It should be a lot better in pathological cases (especially a `{...}` expression with a lot of variables)

R/object_usage_linter.R
```diff
@@ -196,7 +182,28 @@ extract_glued_symbols <- function(expr) {
...
+    warning = function(...) NULL
+  )
+  if (is.null(parsed_text)) {
+    return("")
+  }
+  parse_data <- utils::getParseData(parsed_text)
+  # covers NULL & NA cases
+  if (nrow(parse_data) == 0L) {
+    return("")
+  }
```
IndrajeetPatil, line 201: Are there existing tests indirectly covering the two conditionals here?

R/object_usage_linter.R
```diff
@@ -196,7 +182,28 @@ extract_glued_symbols <- function(expr) {
...
+  symbols <- parse_data$text[parse_data$token == "SYMBOL"]
```
AshesITR, line 202: Does this approach differentiate between `x` and `` `x` ``?

MichaelChirico: Hmm I think it doesn't, but neither did the old approach.

AshesITR (approved):

PR #1615, opened 2022-10-04
Turn on codecov comments

IndrajeetPatil:
This makes it easy to see how much of the newly added code in a PR is being covered by tests.


MichaelChirico:
weird, that's been turned off for 3 years, it seems, but we were definitely getting those comments more recently than that...

PR #1618, opened 2022-10-05
Add examples to the docs for utilities

IndrajeetPatil:


R/get_source_expressions.R
```diff
@@ -51,8 +51,11 @@
 #'   \item{error}{A `Lint` object describing any parsing error.}
 #'   \item{lines}{The [readLines()] output for this file.}
 #' }
+#'
+#' @examples
+#' tmp <- withr::local_tempfile(lines = c("x <- 1", "y <- x + 1"))
```
MichaelChirico, line 56: could you confirm that withr works under example() to clean up the tempfile afterwards?

MichaelChirico (approved):

R/get_source_expressions.R
IndrajeetPatil: I _think_ so. Otherwise, [these](https://github.com/r-lib/withr/blob/35301f6439fd7f675325a105e33db8d158405c1f/R/tempfile.R#L13-L14) examples in `{withr}` would have been flagged by CRAN.

PR #1621, opened 2022-10-05
Error on a `NOTE` in R CMD check

IndrajeetPatil:
`NOTE`s can create problems during the CRAN submission, so we might as well catch them early on.

Also, consistently use `.yaml` instead of `.yml`.


MichaelChirico (approved):
happy to test this out, we may revert it slows us down a lot

PR #1624, opened 2022-10-06
Add more missing tests

IndrajeetPatil:


MichaelChirico (approved):

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,10 @@
+test_that("set_lang and reset_lang work as expected", {
+  old_lang <- Sys.getenv("LANGUAGE")
+  new_lang <- "en"
+
+  lintr:::set_lang(new_lang)
+  expect_identical(Sys.getenv("LANGUAGE"), new_lang)
+
+  lintr:::reset_lang(old_lang)
+  expect_identical(Sys.getenv("LANGUAGE"), old_lang)
```
IndrajeetPatil, line 9: @MichaelChirico Is it not strange that even with these tests merged into the `main`, [codecov](https://app.codecov.io/gh/r-lib/lintr/blob/main/R/utils.R) still says that `reset_lang()` is not covered by any tests?

MichaelChirico: try a test commit where you add a stop() inside reset_lang(), to ensure it's actually being run

PR #1626, opened 2022-10-06
Add snapshot tests for testing generated Rd fragments

IndrajeetPatil:
The only issue I see with this is that the snapshots will need to be updated

- any time a new linter is added
- tags are updated
- tags for a given linter are changed
- etc.

Which makes them fragile. But, OTOH, they are quite easy to update as well (`testthat::snapshot_accept()`).

P.S. These are not run on CRAN by default.

### DIFF
--- a/tests/testthat/_snaps/linter_tags.md
+++ b/tests/testthat/_snaps/linter_tags.md
@@ -1,0 +1,254 @@
+# Generated Rd fragments are as expected for rd_tags
+
+    Code
+      rd_tags("assignment_linter")
+    Output
+      [1] "\\section{Tags}{"                                                                                           
+      [2] "\\link[=consistency_linters]{consistency}, \\link[=default_linters]{default}, \\link[=style_linters]{style}"
+      [3] "}"                                                                                                          
+
+# Generated Rd fragments are as expected for rd_linters
+
+    Code
+      rd_linters("robustness")
+    Output
+       [1] "\\section{Linters}{"                                
+       [2] "The following linters are tagged with 'robustness':"
+       [3] "\\itemize{"                                         
+       [4] "\\item{\\code{\\link{absolute_path_linter}}}"       
+       [5] "\\item{\\code{\\link{backport_linter}}}"            
+       [6] "\\item{\\code{\\link{class_equals_linter}}}"        
+       [7] "\\item{\\code{\\link{equals_na_linter}}}"           
+       [8] "\\item{\\code{\\link{missing_package_linter}}}"     
+       [9] "\\item{\\code{\\link{namespace_linter}}}"           
+      [10] "\\item{\\code{\\link{nonportable_path_linter}}}"    
+      [11] "\\item{\\code{\\link{seq_linter}}}"                 
+      [12] "\\item{\\code{\\link{strings_as_factors_linter}}}"  
+      [13] "\\item{\\code{\\link{T_and_F_symbol_linter}}}"      
+      [14] "\\item{\\code{\\link{undesirable_function_linter}}}"
+      [15] "\\item{\\code{\\link{undesirable_operator_linter}}}"
+      [16] "}"                                                  
+      [17] "}"                                                  
+
+# Generated Rd fragments are as expected for rd_taglist
+
+    Code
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
> I'd rather just nocov these functions

I like this option. Would make a separate PR for that.

> I really dislike snapshot tests (or at least think they have a very limited scope)

I had a feeling that you did 😅

If you find any free time ever, please let me know if the "Headaches" section in my [slide deck](https://indrajeetpatil.github.io/intro-to-snapshot-testing/#/title-slide) does justice to your worries about snapshot testing. Always looking for a critical feedback on this topic 🙃

MichaelChirico:
> If you find any free time ever, please let me know if the "Headaches" section in my [slide deck](https://indrajeetpatil.github.io/intro-to-snapshot-testing/#/title-slide) does justice to your worries about snapshot testing. Always looking for a critical feedback on this topic 🙃

Sorry, I'd missed this comment.

Had a look, I think you have the basic idea. But the headaches IME are much more pervasive then is let on.

Basically, we tried turning on CRAN tests for snapshots for a while, but the tests would break constantly.

The crux of it I think is that you're testing behavior that you don't actually own. e.g. {pillar} has a bunch of snapshot tests, but there's all sorts of markup done by {rlang}, {cli}, {crayon}, etc, all of which have relatively volatile interfaces on these user-facing aspects (which is basically fine, since those things are intended for interactive use anyway), but every time _your upstreams_ make some tiny change, _your_ tests break.

It's one thing for fragility to be easily anticipated as you are doing dev work, it's another to be at the mercy of upstreams breaking you all the time. So even in the case of output testing, I prefer tests like `expect_output()` that allow me to focus on the aspects of the output that I own myself in my package.

In my mind, snapshots are most useful as a sort of end-to-end test (this gets close to the full-file test concept you covered), where one-by-one unit tests are too focused, and you want to run some test on the full output of a complicated workflow. The point (which you touch on) is to use them as sparingly as possible -- only a very small percentage of your package's tests should be snapshots.

And IMO they shouldn't be used at all for warnings/errors, `expect_warning()`/`expect_error()` do pretty much fine there; the only possible exception is code that produces several warnings/messages/maybe an error, where the `expect_*` API is a bit awkward and snapshots may be cleaner (but again, there's been a lot of volatility here as {rlang} evolves, adding `abort()` and now more and more classed errors, just extra sources of fragility).

IndrajeetPatil:
> The crux of it I think is that you're testing behavior that you don't actually own. 

That's a really neat way to put it. Stealing it for my slides! :)

I completely agree with you, esp. because I've first-hand experience with such failures as I, too, rely a lot on _tidyverse_ packages.

But I am wondering to what extent this is the feature of the golden testing itself versus a reflection of the rapidly changing API of _tidyverse_. E.g., let's say in 10 years, the _rlang_, _vctrs_, _tibble_, etc. low-level deps are stable and feature-complete and there are no more changes, at least none that can cause break tests. Maybe in such contexts, maintaining snapshots wouldn't require a lot of work. But, as you mention, you still need to depend on _others_ to not make changes that cause test failures in _your_ tests.

You must've already seen it, Kirill has built some cool [infrastructure](https://github.com/krlmlr/actions-sync) to deal with this: actions with cron jobs that make PRs when upstream changes and snapshots need to be updated. I think that's _a_ way to deal with the maintenance headache associated with volatile upstream deps. 

At any rate, I'll update my slides to provide this additional context. Thanks, as usual, for your insightful comments!

MichaelChirico:
> I am wondering to what extent this is the feature of the golden testing itself versus a reflection of the rapidly changing API of tidyverse

I think it's a mix of both... e.g. the .Rout version of golden tests offered by base R doesn't seem as fragile to me. And `tools::Rdiff()` gets at what I think is a good approach -- add some post-processing layer to the output that diffs out some unimportant aspects of the output. e.g. if there was an option in {testthat} to activate some "minified" version of the CLI/pillar/etc decorative output, e.g. via an option, that would help a lot in day-to-day use cases. There's of course still a need for the "end-to-end" type test I mentioned, where including the "UX" aspect of output formatting is important.

PR #1628, opened 2022-10-06
Always qualify rex in tests, make private API calls clearer

MichaelChirico:
I did a run of running our tests directly with `source()` rather than through `testthat` functions, so we can tell where private functions are used.

I think in general we should try and replace these tests with tests on the public API, and resort to private API tests only for crucial internal code that's too gnarly to test the "proper" way.

But for now, happy to just make clear where these private API tests are by qualifying `:::`.

Also ensured `rex::rex()` is used everywhere for consistency.


IndrajeetPatil (approved):
LGTM, thanks!

I will also remember to do this for any new tests I might add.

PR #1629, opened 2022-10-07
New for_loop_index_linter

MichaelChirico:
Part of #884 

This one is not violated often in our experience, but it's quite bad when it is.

```
x <- 1:10
for (x in x) x
x
# [1] 10
```

Seems like undesirable behavior.


IndrajeetPatil (dismissed):
LGTM. 

The only nitpick I have is using `` `for()` `` instead of _for_ in the docs and message. But don't feel strongly about it.

MichaelChirico:
"for loop" is a much more general programming construct, `for()` doesn't look right to me. I'll keep it as is. Thanks!

IndrajeetPatil:
Sure, but I find it easier to parse the following 

> This is true for `for()` loops.

over this

> This is true for for loops.

IndrajeetPatil (approved):

MichaelChirico:
Sure, I would disambiguate with "for `for` loops", but that's not used here anyway?

PR #1638, opened 2022-10-08
Recognize |> in pipe_continuation_linter

### LINKED ISSUE #1609: Extend pipe_continuation_linter to cover native pipes (|>)
https://github.com/r-lib/lintr/blob/ee764f2cace1026b497ed14ec9fafb39ac0924cd/R/pipe_continuation_linter.R#L30

Doesn't cover `//PIPE` yet

MichaelChirico:
Closes #1609

Also cleaned up/standardized the R & test scripts a bit


IndrajeetPatil:
This correctly doesn't lint, but I'm wondering if it's worth including in the test case because I've seen code like this in the wild.

``` r
library(magrittr)

1:4 %>% {
 (.) %>% sum()
}
#> [1] 10

library(lintr)

lint(
  text = "1:4 %>% {
    (.) %>% sum()
  }",
  linters = pipe_continuation_linter()
)
```

<sup>Created on 2022-10-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico:
Sure, can't hurt. It's related to our `UnnecessaryNestingLinter()` which would by default try and remove the `{` here, but things are complicated in pipe world where it's not always trivial to refactor `%>% { ... }` expressions

IndrajeetPatil (approved):
LGTM, thanks!

PR #1640, opened 2022-10-08
add examples to get_r_string

### LINKED ISSUE #1553: Add examples for get_r_string
Follow-up to #1526

MichaelChirico:
Closes #1553

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -233,13 +233,12 @@ platform_independent_sort <- function(x) x[platform_independent_order(x)]
 #' get_r_string(expr_as_xml, "expr[3]") # "b"
 #'
 #' # more importantly, extract strings under R>=4 raw strings
-#' if (getRversion() >= "4.0.0") {
-#'   tmp4.0 <- withr::local_tempfile(lines = "c(R'(a\\b)', R'--[a\\\"\'\"\\b]--')")
-#'   expr_as_xml4.0 <- get_source_expressions(tmp4.0)$expressions[[1L]]$xml_parsed_content
-#'   writeLines(as.character(expr_as_xml4.0))
-#'   get_r_string(expr_as_xml4.0, "expr[2]") # "a\\b"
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -224,6 +224,23 @@ platform_independent_sort <- function(x) x[platform_independent_order(x)]
...
+#'
+#' @examples
+#' tmp <- withr::local_tempfile(lines = "c('a', 'b')")
+#' expr_as_xml <- get_source_expressions(tmp)$expressions[[1L]]$xml_parsed_content
+#' writeLines(as.character(expr_as_xml))
+#' get_r_string(expr_as_xml, "expr[2]") # "a"
+#' get_r_string(expr_as_xml, "expr[3]") # "b"
+#'
+#' # more importantly, extract strings under R>=4 raw strings
+#' if (getRversion() >= "4.0.0") {
```
IndrajeetPatil, line 236: Instead of `if()` block, should we use `@examplesIf` instead?

Due to [this](https://github.com/r-lib/pkgdown/issues/2175) issue, only the output for `get_r_string(expr_as_xml4.0, "expr[3]")` will be shown on the website, but not `get_r_string(expr_as_xml4.0, "expr[2]")`.

MichaelChirico: Good idea for simplicity, but does it actually help with printing? Since roxygen is just hiding the `if()` statement?

IndrajeetPatil: I checked this with my [minimalReprex](https://github.com/IndrajeetPatil/minimalReprex) package:

# Using `if()`

Screenshot 2022-10-08 at 07 29 51

# Using `@examplesIf`

Screenshot 2022-10-08 at 07 34 11

So this definitely helps.

IndrajeetPatil (approved):

R/utils.R
MichaelChirico: awesome, nice find!

PR #1641, opened 2022-10-08
Add "list" to S3 class of lint(), Lint() output

### LINKED ISSUE #1494: lint() output should be class c("lints", "list") for S3 dispatch
I am working on something where I need to edit the line numbers of the `lint()` output; a natural way to do that is like so:

```r
l <- lint(...)
l <- lapply(l, within, line_number <- line_number + 1)
```

But this fails because `within` attempts to dispatch on `lints` class. For now, we need to call the method explicitly:

```r
l <- lapply(l, within.list, line_number <- line_number + 1)
```

We could also define a `within.lints()` method, but that only works for this specific case, I think the problem is more general (it will affect other generics used too process the `lint()` output, too).

MichaelChirico:
Closes #1494

### DIFF
--- a/tests/testthat/test-methods.R
+++ b/tests/testthat/test-methods.R
@@ -148,3 +148,9 @@ test_that("split.lint works as intended", {
   l <- lint(tmp, seq_linter())
   expect_true(all(vapply(split(l), inherits, logical(1L), "lints")))
 })
+
+test_that("within.list is dispatched", {
+  l <- lint(text = "a=1\nb=2", linters = infix_spaces_linter())
+  expect_silent(l <- lapply(l, within, line_number <- line_number + 1L))
+  expect_identical(vapply(l, `[[`, integer(1L), "line_number"), 2:3)
+})
--- a/tests/testthat/test-methods.R
+++ b/tests/testthat/test-methods.R
@@ -31,6 +31,7 @@ test_that("as.data.frame.lints", {
     ),
     "lint"
   )
+  expect_type(l1, "list")
 
   # A larger lint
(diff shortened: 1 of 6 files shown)


tests/testthat/test-methods.R
```diff
@@ -31,6 +31,7 @@ test_that("as.data.frame.lints", {
     ),
     "lint"
   )
+  expect_type(l1, "list")
```
IndrajeetPatil, line 34: Nit: Add a similar new test for `lint()`?

MichaelChirico: thanks for calling it out... i didn't see a "nice" place to add it on a quick glance. but i'll have another go.

tests/testthat/test-methods.R
```diff
@@ -147,3 +148,9 @@ test_that("split.lint works as intended", {
   l <- lint(tmp, seq_linter())
   expect_true(all(vapply(split(l), inherits, logical(1L), "lints")))
 })
+
+test_that("within.list is dispatched", {
+  l <- lint(text = "a=1\nb=2", linters = infix_spaces_linter())
+  expect_silent(l <- lapply(l, within, line_number <- line_number + 1L))
+  expect_identical(vapply(l, `[[`, integer(1L), "line_number"), 2:3)
```
IndrajeetPatil, line 155: I know, this is a bit annoying.

```suggestion
  expect_identical(vapply(l, `[[`, integer(1L), "line_number"), 2L:3L)
```

IndrajeetPatil (approved):

PR #1642, opened 2022-10-08
Handle <<reference_chunks>> in Rnw,Rmd

### LINKED ISSUE #779: Linting of Sweave "macros" << >> fails
```
download.file("https://raw.githubusercontent.com/cran/psychotree/master/inst/doc/raschtree.Rnw", tmp <- tempfile())
lintr::lint(tmp)
# ...
# /tmp/RtmpisKACL/file4f6512692723:242:1: error: unexpected input
# <<fit_raschtree>>
# ^
```

I'm not sure the full syntax for these "macro" chunks / if there's a way to insert them statically, or what we should do about these (maybe just skip?); filing the issue in case anyone has a better idea.

MichaelChirico:
Closes #779

I didn't see any `<<examples>>` included with {knitr}, but the solution is simple enough & basically re-uses stuff directly from {knitr}, so I didn't feel too compelled to add any more tests.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,8 @@
 
 * `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
 
-* `get_source_expressions()` can handle reference chunks in Sweave and Rmd files like `<<ref_file>>` (#779, @MichaelChirico)
+* `get_source_expressions()` can handle Sweave/Rmarkdown documents with reference chunks like `<<ref_file>>` (#779, @MichaelChirico).
+  Note that these are simply skipped, rather than attempting to retrieve the reference & also lint it.
 
 ## Changes to defaults
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -4,7 +4,8 @@
 
 * `fixed_regex_linter()` no longer fails with regular expression pattern `"\\;"` (#1545, @IndrajeetPatil).
 
-* `get_source_expressions()` can handle reference chunks in Sweave and Rmd files like `<<ref_file>>` (#779, @MichaelChirico)
+* `get_source_expressions()` can handle Sweave/Rmarkdown documents with reference chunks like `<<ref_file>>` (#779, @MichaelChirico).
+  Note that these are simply skipped, rather than attempting to retrieve the reference & also lint it.
```
IndrajeetPatil, line 8:
```suggestion
  Note that these are simply skipped, rather than attempting to retrieve the reference and also lint it.
```

R/extract.R
```diff
@@ -20,13 +20,17 @@ extract_r_source <- function(filename, lines, error = identity) {
     return(output)
   }
 
+  output_env <- environment()
```
IndrajeetPatil, line 23: `# nolint: object_usage_linter` here?

MichaelChirico: oh, that's annoying. it's used just below!

IndrajeetPatil (approved):

PR #1646, opened 2022-10-08
Use {withr} more comprehensively in tests to manage state

MichaelChirico:
I couldn't remove all `on.exit()` usages, here's what remains:

```r
tests/testthat/test-error.R:29:  on.exit(lintr:::reset_lang(old_lang), add = TRUE)
tests/testthat/test-get_source_expressions.R:5:    on.exit(close(con))
tests/testthat/test-object_usage_linter.R:312:  on.exit(utils::globalVariables(old_globals, package = globalenv(), add = FALSE))
tests/testthat/test-settings.R:61:  on.exit(if (is.na(old)) Sys.unsetenv(test_env) else sym_set_env(test_env, old))
```

Of those

 - test-error.R: can't be removed unless we add a custom unexported `local_lang()`
 - test-get_source_expressions.R: see https://github.com/r-lib/withr/issues/210
 - test-object_usage_linter.R: I support we could use `withr::defer()`, but `on.exit()` seems fine here
 - test-settings.R: this _may_ be replaced by `withr::local_envvar()`, but I was too lazy to play around with it & ensure it works

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,6 +1,9 @@
 library(testthat)
 library(lintr)
-loadNamespace("rex")
-loadNamespace("withr")
+# suppress printing environment name (noisy)
+invisible({
+  loadNamespace("rex")
+  loadNamespace("withr")
+})
 
 test_check("lintr")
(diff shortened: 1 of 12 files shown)


IndrajeetPatil (approved):
Wow, that's a thorough clean-up. Thanks a lot, @MichaelChirico! 

As for the remaining instances of `on.exit()`, we can revisit them later. 🧹

PR #1649, opened 2022-10-08
Additional test for `namespace_linter()`+ clean-up

IndrajeetPatil:
I also wanted to add a test for Case-2 here, which currently isn't tested. But I couldn't come up with one. 
Is it even possible for a package to be installed with a broken `NAMESPACE`?

https://github.com/r-lib/lintr/blob/0af31f53a1007e464cf975bf9d62e79b484a7f35/R/namespace_linter.R#L54-L71

### DIFF
--- a/tests/testthat/test-namespace_linter.R
+++ b/tests/testthat/test-namespace_linter.R
@@ -1,35 +1,25 @@
-test_that("returns the correct linting", {
+test_that("namespace_linter skips allowed usages", {
   linter <- namespace_linter()
 
-  expect_lint(
-    "stats::sd",
-    NULL,
-    linter
-  )
+  expect_lint("stats::sd", NULL, linter)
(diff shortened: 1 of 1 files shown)


MichaelChirico:
> Is it even possible for a package to be installed with a broken NAMESPACE?

no, I was thinking it would run on packages under development? maybe I got mixed up when writing here. should we simplify the code instead?

IndrajeetPatil:
> should we simplify the code instead?

Yeah, I agree. I think Case 2 doesn't need to be there. 

But I think we should handle it in a separate issue and PR, since the current one is quite straightforward to review and merge.

IndrajeetPatil:
Created a new issue to track this: https://github.com/r-lib/lintr/issues/1650

MichaelChirico (approved):

Issue #1546, opened 2022-09-21
function_argument_linter() could look for missing() handling and mention it in the lint message


MichaelChirico:
I think the simplest fix for existing violations of `function_argument_linter()` is to set a default of `NULL` for functions violating the design principle.

This doesn't always align the functions to the intent of the design guide, but it is safer for functions which may cause breaking changes when changing the argument order, and from what I've seen it's usually a good choice.

One case I can think of where it's _not_ safe to blithely make this switch is when such arguments are handled with `missing()`. Consider:

```r
foo <- function(x, y = 2, z) {
  if (missing(z)) z <- 3
  x * y * z
}
```

And existing usage `foo(1, 2)`, which will break if changing to `z = NULL` by default -- there was an implicit default value of `z` set inside the function. We can of course accommodate this by changing `missing(z)` to `is.null(z)` or even making the default `z = 3` explicit, but it is more complicated, at least.

OTOH, it should be straightforward for the linter logic to look, when identifying parameter `zed` as violating `function_argument_linter()`, for usages of `missing(zed)` in the applicable function body, and if found, mention this caveat in the linter message.

PS I'm also curious if anyone can think of other edge cases where setting this default can break existing code.

IndrajeetPatil:
Note to ourselves: 

Once this is implemented, also update the docs for `function_argument_linter()` to mention this caveat and to add an example of a function that uses `missing()`.

MichaelChirico:
I'm realizing my example here is flawed -- `foo(1, 2)` will continue to give 6 after changing the default:

```r
foo <- function(x, y = 2, z = NULL) {
  if (missing(z)) z <- 3
  x * y * z
}
foo(1, 2)
```

Maybe I'm mis-remembering, but I recall breaking some test suites by neglecting to update a `missing()` usage when changing to `NULL` by default in some cases, but the example here doesn't capture that.

Or maybe it's just that mixing `z = NULL` and `missing(z)` like here is an odd design choice & it's usually preferable to switch `missing()` to `is.null()` in the process of refactoring -- so we include the note in the lint message not necessarily as a guard for back-compatibility, but to facilitate _better_ refactoring.

Issue #1155, opened 2022-05-20
implicit_integer_linter shouldn't lint 1:10 (optionally?)


MichaelChirico:
```
lintr::lint("1:10\n", lintr::implicit_integer_linter())
# <text>:1:2: style: Integers should not be implicit. Use the form 1L for integers or 1.0 for doubles.
# 1:10
# ~^
# <text>:1:5: style: Integers should not be implicit. Use the form 1L for integers or 1.0 for doubles.
# 1:10
#   ~~^
```

but `typeof(1:10)` is `"integer"`, and the performance is essentially identical:

```
microbenchmark::microbenchmark(
  1L:10L,
  1L:10,
  1:10L,
  1:10,
  times = 1e6
)

# Unit: nanoseconds
#    expr min  lq     mean median  uq      max neval
#  1L:10L 199 242 479.8983    257 284 89303201 1e+06
#   1L:10 197 237 376.0418    251 278 10444161 1e+06
#   1:10L 198 239 354.6013    252 280 10678523 1e+06
#    1:10 190 233 405.9966    246 274 14666175 1e+06
```

AshesITR:
implicit_integer_linter isn't really about performance, more about explicit typing.

MichaelChirico:
right -- the more important point is that `identical(1:10, 1L:10L)`. the performance bit is just demonstrating that there's no performance reason to prefer one or the other either.

AshesITR:
Is there a philosophical difference to `identical(1.0, 1)`?

MichaelChirico:
to me, it's different -- almost everywhere in R, `1` actually means `1:double`. but with `1:10` it's actually `1:integer`.

for `1.0` v `1`, as I understand it we only like `1.0` as a way of signalling "yes, this is a double, it doesn't need to be `1L`" to the linter.

put another way -- the linter wants `1` to be `1L` whenever possible. it aims to set the default numeric literal to integer, and only allow doubles with decimal points. but for `1:10`, `1` is already (in effect) integral, so there's no need to intervene.

that's how I see it at least.

PS I used to be a stickler about writing 1L:10L but got sick of the extra characters and especially at the cost of typing difficulty -- typing `1L:10L` requires some deft coordination to get the shift key pressed in exactly the right places

AshesITR:
I tend to rarely need the `:` with literals, usually `seq_*()` functions provide a nicer solution.

To me the improved clarity of intent is worth the extra effort spent typing.

[...]

PR #1651, opened 2022-10-08
Add examples to documentation: Part-4 (f-i)

IndrajeetPatil:
Except `fixed_regex_linter()`; I will come back to it later.


R/function_argument_linter.R
```diff
@@ -1,6 +1,30 @@
 #' Function argument linter
 #'
-#' Check that arguments with defaults come last in all function declarations, as per the tidyverse design guide.
+#' Check that arguments with defaults come last in all function declarations,
+#' as per the tidyverse design guide.
+#'
```
MichaelChirico, line 5: Might be worth noting how we approached a fix to be maximally backwards-compatible for existing code: set these arguments to be `NULL` by default. See also #1546.

Maybe (1) mention this in exposition here and (2) suggest a `x=NULL` version (etc) under the `okay` examples

WDYT?

R/function_left_parentheses_linter.R
```diff
@@ -4,6 +4,29 @@
...
+#'   text = "mean (x)",
+#'   linters = function_left_parentheses_linter()
+#' )
+#'
+#' lint(
+#'   text = "stats::sd(c (x, y, z))",
+#'   linters = function_left_parentheses_linter()
+#' )
+#'
+#' # okay
```
MichaelChirico, line 19: suggestion: other `okay` examples like an `if ()` statement where ` (` is allowed

MichaelChirico: also an example with `function()` since rules are slightly different for signatures

R/function_return_linter.R
```diff
@@ -4,6 +4,19 @@
 #'   confusing (because assigning to `x` has some side effect that is muddled
 #'   by the dual-purpose expression).
 #'
+#' @examples
+#' # will produce lints
+#' lint(
```
MichaelChirico, line 9: suggestion: an example using a `return()` assignment done for side effects

e.g.

https://github.com/r-lib/lintr/blob/854b2c7442008b4d091b72fe52e16995d86d333b/tests/testthat/test-function_return_linter.R#L40

and

https://github.com/r-lib/lintr/blob/854b2c7442008b4d091b72fe52e16995d86d333b/tests/testthat/test-function_return_linter.R#L69-L72

R/implicit_integer_linter.R
```diff
@@ -2,6 +2,24 @@
 #'
 #' Check that integers are explicitly typed using the form `1L` instead of `1`.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "x <- 1",
```
MichaelChirico, line 8: Given this comment, an example showing extraction would also be illustrative:

https://github.com/r-lib/lintr/issues/1155#issuecomment-1231709614

R/infix_spaces_linter.R
```diff
@@ -79,6 +79,40 @@ infix_overload <- data.frame(
...
+#' )
+#'
+#' lint(
+#'   text = "1:4 %>% sum()",
+#'   linters = infix_spaces_linter()
+#' )
+#'
+#' lint(
+#'   text = "x <-      2.4",
+#'   linters = infix_spaces_linter(allow_multiple_spaces = TRUE)
```
MichaelChirico, line 108: probably better to give an example where this is done for multi-expression alignment

R/inner_combine_linter.R
```diff
@@ -6,6 +6,19 @@
...
 #'   is applied only once.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "c(log10(x), log10(y), log10(z))",
+#'   linters = inner_combine_linter()
+#' )
+#'
+#' # okay
```
MichaelChirico, line 16: suggestion: an `okay` example like `c(as.Date(x, "%Y/%m/%d"), as.Date(z, "%Y%m%d"))`

MichaelChirico: or keep the theme here and use `log()` with different `base=` arguments

tests/testthat/test-function_left_parentheses_linter.R
```diff
@@ -1,10 +1,11 @@
-test_that("returns the correct linting", {
+test_that("function_left_parentheses_linter skips allowed usages", {
   linter <- function_left_parentheses_linter()
-  msg <- rex::rex("Remove spaces before the left parenthesis in a function call.")
 
   expect_lint("blah", NULL, linter)
   expect_lint("print(blah)", NULL, linter)
+  expect_lint('"print"(blah)', NULL, linter)
```
MichaelChirico, line 6: nice tests! glad it already works

R/inner_combine_linter.R
IndrajeetPatil: Excellent suggestion!

I have used an example with `log(base = )`.

R/infix_spaces_linter.R
IndrajeetPatil: I was tiptoeing around it because the issue is again the code that we show won't exactly be the code that is parsed. But maybe we can make an exception once.

I have changed the example.

R/implicit_integer_linter.R
IndrajeetPatil: Good idea.

R/function_return_linter.R
IndrajeetPatil: I have included these examples. But should I also include versions of these that don't lint? These will be multiline expressions and a bit difficult to show in `text = `.

R/function_argument_linter.R
IndrajeetPatil: Excellent points. I have both updated the description and added an example for this.

R/function_left_parentheses_linter.R
IndrajeetPatil: > also an example with `function()` since rules are slightly different for signatures

I have added an example, but not sure if this is what you had in mind. What is different for signatures?

R/function_return_linter.R
MichaelChirico: How about like

```
code_lines <- "
e <- new.env()
foo <- function(x) {
  e$val <- x + 1
  return(e$val)
}
"
cat(code_lines)
lint(text = code_lines, function_return_linter())
```

R/implicit_integer_linter.R
```diff
@@ -9,6 +9,11 @@
 #'   linters = implicit_integer_linter()
 #' )
 #'
+#' lint(
+#'   text = "x[c(1, 2)]",
```
MichaelChirico, line 13: I would stick with a simpler example, `x[1]` -> `x[1L]`, the `c()` part here makes it look like an example for `c()`, not for extraction

R/infix_spaces_linter.R
MichaelChirico: how does it look like

```
code_lines <- "
ab     <- 1L
abcdef <- 2L
"
cat(code_lines)
lint(text = code_lines, ...)
```

R/function_return_linter.R
IndrajeetPatil: Looks good.

R/function_left_parentheses_linter.R
MichaelChirico: that's what i had in mind, just function calls vs. function signatures

IndrajeetPatil: Ah, I see.

R/function_return_linter.R
```diff
@@ -28,6 +28,24 @@
 #'   linters = function_return_linter()
 #' )
 #'
+#' lint(
+#'   text = "foo <- function(x) return(x + 1); x <- x + 1",
```
MichaelChirico, line 32: confused by this, maybe a typo? this is meant to be the "ok" equivalent of the <<- case right?

IndrajeetPatil: Typo indeed. Fixed it.

Just wanted to show equivalent of achieving the same thing that doesn't produce a lint:

``` r
x <- 1
foo <- function(x) return(x <<- x + 1)
foo(x)
x
#> [1] 2

x <- 1
foo <- function(x) return(x + 1)
x <- foo(x)
x
#> [1] 2
```

<sup>Created on 2022-10-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico (approved):

MichaelChirico:
Thanks for all the work here, I know it's a slog

IndrajeetPatil:
Thanks for so thoroughly reviewing these changes! Find it very useful to clarify my own understanding of things.

PR #1653, opened 2022-10-09
Remove code to deal with broken namespaces

### LINKED ISSUE #1650: Re-evaluate if code to deal with a broken `NAMESPACE` is needed by `namespace_linter()`
Is it even possible for a package to be installed with a broken `NAMESPACE`?

@MichaelChirico thinks no, but this might be useful for packages under development? It's unclear and needs to be reassessed.

https://github.com/r-lib/lintr/blob/0af31f53a1007e464cf975bf9d62e79b484a7f35/R/namespace_linter.R#L54-L71

IndrajeetPatil:
Closes #1650


R/namespace_linter.R
```diff
@@ -51,27 +51,10 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
       return(lints)
     }
 
-    ## Case 2 (rare?): pkg namespace is broken in pkg::foo
+    ## Case 2/3/4: problems with foo in pkg::foo / pkg:::foo
 
     # run here, not in the factory, to allow for run- vs. "compile"-time differences in package structure
     namespaces <- lapply(packages, function(package) tryCatch(getNamespace(package), error = identity))
-    failed_namespace <- vapply(namespaces, inherits, "condition", FUN.VALUE = logical(1L))
```
MichaelChirico, line 58: We should either

1 . also remove `tryCatch()` and just do `lapply(packages, getNamespace)`
2. keep the `failed_namespace` check and throw an error `if (any(failed_namespace))`

IndrajeetPatil: Good idea. Done. See if the error message is okay or needs to be modified.

MichaelChirico: sorry that was a 1 or 2. how can namespaces inherit from error if we remove tryCatch?

IndrajeetPatil: 🤦 

Sorry for the autopilot mode. 

I think option 2 is better, so going with that one.

MichaelChirico (approved):

PR #1655, opened 2022-10-09
Add examples to documentation: Part-5 (l-n)

IndrajeetPatil:
Except `nonportable_path_linter()`; I will come back to it later.


R/missing_argument_linter.R
```diff
@@ -4,6 +4,35 @@
...
+#'
+#' lint(
+#'   text = "mean(x, )",
+#'   linters = missing_argument_linter(except = "mean")
+#' )
+#'
+#' lint(
+#'   text = "mean(x, )",
+#'   linters = missing_argument_linter(allow_trailing = TRUE)
+#' )
```
MichaelChirico, line 34: not a good example I think, since this may be erroneous to use. better to use a list2() example, maybe filter() or tibble() or mutate(), to demonstrate a case where the lint is more subjective as opposed to avoiding errors

R/missing_package_linter.R
```diff
@@ -1,6 +1,19 @@
 #' Missing package linter
 #'
-#' Check for missing packages in `library()`, `require()`, `loadNamespace()` and `requireNamespace()` calls.
+#' Check for missing packages in `library()`, `require()`, `loadNamespace()`, and `requireNamespace()` calls.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "library(statts)",
```
MichaelChirico, line 8: I worry statts is too close to something that could conceivably be a real package name (if not now, then eventually). maybe pick something more obscure?

R/namespace_linter.R
```diff
@@ -6,6 +6,35 @@
 #'
 #' @param check_exports Check if `symbol` is exported from `namespace` in `namespace::symbol` calls.
 #' @param check_nonexports Check if `symbol` exists in `namespace` in `namespace:::symbol` calls.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "statts::sd(c(1, 2, 3))",
```
MichaelChirico, line 13: similar issue here with statts

R/missing_argument_linter.R
IndrajeetPatil: I have changed the example, but still wanted to understand your comment.

Under what circumstances will this be the case?

> this may be erroneous to use

``` r
x <- c(1, 2, 3)
mean(x, )
#> [1] 2
```

<sup>Created on 2022-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/missing_package_linter.R
IndrajeetPatil: Can't think of something that I don't see being a package name, so chose `xyzxyz`. Happy to change if you can think of something better.

MichaelChirico: SGTM, I think a gibberish name is OK too...

R/missing_argument_linter.R
MichaelChirico: You're right, I was speaking a bit more generally (hadn't check if `mean(x, )` works)...  point being it's very hard to tell at a glance whether code like that will run or not (and hence a lint is really good), whereas I thing `allow_trailing=TRUE` is intended to allow a common source of false positives through (namely, `list2()` usages)

MichaelChirico (approved):

R/missing_argument_linter.R
IndrajeetPatil: Gotcha!

PR #1656, opened 2022-10-10
New unnecessary_placeholder_linter

MichaelChirico:
Part of #884

Originally named `MagrittrDotLinter()`, I think the new name is an improvement (and paves the way for base's `_` to be included), but open to further refinement here.


MichaelChirico:
Agree we could add those, thanks for the suggestion! let's file follow-ups for those

IndrajeetPatil (approved):

PR #1660, opened 2022-10-10
small simplification for glue, add coverage test

MichaelChirico:
`length(parsed_text) == 0` also covers the case like `glue("{}")` where an empty expression is parsed (`parsed_text` is `expression()`). I don't see how `nrow(parse_data) == 0` is possible after that (even a comment generates parse data).


R/object_usage_linter.R
```diff
@@ -165,8 +165,7 @@ extract_glued_symbols <- function(expr) {
       parse(text = xml2::xml_text(cl)),
       error = function(...) NULL,
       warning = function(...) NULL
-    )[[1L]]
-    if (is.null(parsed_cl)) next
```
MichaelChirico, line 169: I can't think of how we could end up with multiple expressions (`[[1]]`) or a parsing error here... if there's a flaw in our thinking that might lead to that happening, we should let it cause an error & users can report it so we can tighten the logic elsewhere.

R/object_usage_linter.R
```diff
@@ -160,13 +160,16 @@ extract_glued_symbols <- function(expr) {
   }
   glued_symbols <- new.env(parent = emptyenv())
 
+  unexpected_error <- function(cond) {
+    stop("Unexpected failure to parse glue call, please report: ", conditionMessage(cond))
```
IndrajeetPatil, line 164: Can we add a test to make sure that we get this expected error message?

MichaelChirico: no -- I don't know how to trigger the error. it should be nocov but covr doesn't separate out tryCatch branches

IndrajeetPatil (approved):

PR #1661, opened 2022-10-10
coverage: skip a region we think can't be reached

MichaelChirico:
I believe we don't know of any way to trigger this region in a test, please confirm

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -265,6 +265,7 @@ parse_check_usage <- function(expression,
   )
 
   missing <- is.na(res$message)
+  # nocov start
   if (any(missing)) {
     res[missing, ] <- re_matches(
       vals[missing],
@@ -278,6 +279,7 @@ parse_check_usage <- function(expression,
       )
     )
   }
+  # nocov end
 
   res <- res[!is.na(res$message), ]
 


AshesITR (commented):
Checking the code of `R/codetools.R`, all issues are collected using the format `signalUsageIssue(message)`. The regex - if I read it correctly - implicitly assumes there always is an `sQuote(...)` in the message.

This is not the case for the following messages, scouring the source code:

```
w$signal("... may be used in an incorrect context", w) # l. 416
w$signal(paste(a, "may be used in an incorrect context:", pasteExpr(e)), w) # l. 436
w$signal(paste("wrong number of arguments to '.Internal':", pasteExpr(e)), w) # l. 676
w$signal(paste("bad argument to '.Internal':", pasteExpr(e[[2]])), w) # l. 682
```

``` r
tst <- function() { a <- ...; a }
codetools::checkUsage(tst)
#> <anonymous>: ... may be used in an incorrect context (<text>:1)
```

<sup>Created on 2022-10-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

AshesITR:
Closing since we've uncovered a bug that needs fixing. I'll add the nocov in the patch for #1714

PR #1664, opened 2022-10-10
Add examples to documentation: Part-6 (o-p)

IndrajeetPatil:

### DIFF
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -44,6 +44,40 @@ object_name_xpath <- local({
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
 #'   match at least one of these styles.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "my_var <- 1L",
+#'   linters = object_name_linter(styles = "CamelCase")
+#' )
+#'
+#' lint(
+#'   text = "xYz <- 1L",
+#'   linters = object_name_linter(styles = "UPPERCASE")
+#' )
+#'
+#' lint(
+#'   text = "MyVar <- 1L",
+#'   linters = object_name_linter(styles = "dotted.case")
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "my_var <- 1L",
+#'   linters = object_name_linter(styles = "snake_case")
+#' )
+#'
+#' lint(
+#'   text = "xyz <- 1L",
+#'   linters = object_name_linter(styles = "lowercase")
+#' )
+#'
+#' lint(
+#'   text = "my.var <- 1L",
+#'   linters = object_name_linter(styles = "dotted.case")
+#' )
+#'
 #' @evalRd rd_tags("object_name_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
@@ -172,6 +206,25 @@ regexes_rd <- toString(paste0("\\sQuote{", names(style_regexes), "}"))
 #'   see the detailed note in [object_name_linter()] for more details.
 #'
 #' @param length maximum variable name length allowed.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "very_very_long_variable_name <- 1L",
+#'   linters = object_length_linter(length = 10L)
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "very_very_long_variable_name <- 1L",
+#'   linters = object_length_linter(length = 30L)
+#' )
+#'
+#' lint(
+#'   text = "var <- 1L",
+#'   linters = object_length_linter(length = 10L)
+#' )
+#'
 #' @evalRd rd_tags("object_length_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
--- a/R/object_name_linters.R
+++ b/R/object_name_linters.R
@@ -54,7 +54,7 @@ object_name_xpath <- local({
 #'
 #' lint(
 #'   text = "xYz <- 1L",
-#'   linters = object_name_linter(styles = "UPPERCASE")
+#'   linters = object_name_linter(styles = c("UPPERCASE", "lowercase"))
 #' )
 #'
 #' lint(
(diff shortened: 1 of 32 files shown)


R/object_name_linters.R
```diff
@@ -44,6 +44,40 @@ object_name_xpath <- local({
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
 #'   match at least one of these styles.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "my_var <- 1L",
+#'   linters = object_name_linter(styles = "CamelCase")
```
MichaelChirico, line 52: maybe add an example with multiple `styles`?

R/object_name_linters.R
```diff
@@ -172,6 +206,25 @@ regexes_rd <- toString(paste0("\\sQuote{", names(style_regexes), "}"))
 #'   see the detailed note in [object_name_linter()] for more details.
 #'
 #' @param length maximum variable name length allowed.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "very_very_long_variable_name <- 1L",
```
MichaelChirico, line 213: does `object_name_linters` work on `assign()` variables or `setGeneric()`?

If so, an example would be good, otherwise, maybe a follow-up issue?

R/paste_linter.R
```diff
@@ -21,6 +21,40 @@
...
 #'   `paste()` and `paste0()` with `collapse = ", "` is not linted.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "paste('a', 'b', sep = '')",
+#'   linters = paste_linter()
+#' )
+#'
+#' lint(
```
MichaelChirico, line 32: please add examples for cases 3 & 4, e.g. `paste0("a", "b", sep = " ")` and `paste(rep("x", 20L), collapse = "")`

R/pipe_continuation_linter.R
```diff
@@ -3,14 +3,56 @@
 #' Check that each step in a pipeline is on a new line, or the entire pipe fits on one line.
 #'
 #' @evalRd rd_tags("pipe_continuation_linter")
+#'
+#' @examples
+#' # will produce lints
+#' code_lines <- "1:3 %>%\n mean() %>% as.character()"
+#' cat(code_lines)
```
MichaelChirico, line 10: oh, actually this should be `writeLines()`, come to think of it, since `code_lines` doesn't have a trailing `\n`... let's fix more broadly in a follow-up?

R/pipe_continuation_linter.R
```diff
@@ -3,14 +3,56 @@
...
+#' lint(
+#'   text = code_lines,
+#'   linters = pipe_continuation_linter()
+#' )
+#'
+#' code_lines <- "1:3 |> mean() |>\n     as.character()"
```
MichaelChirico, line 16: is this example a mistake? it won't lint right?

MichaelChirico: oh NVM, I am in the wrong linter in my head :)

R/object_name_linters.R
IndrajeetPatil: You mean this should lint?

``` r
library(lintr)

lint(
  text = assign("myVar", 2L),
  linters = object_name_linter(styles = c("dotted.case", "lowercase"))
)
```

<sup>Created on 2022-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/pipe_continuation_linter.R
IndrajeetPatil: The change was easy enough, so included in this PR itself.

R/pipe_continuation_linter.R
IndrajeetPatil: It does lint:

``` r
library(lintr)

code_lines <- "1:3 |> mean() |>\n as.character()"
writeLines(code_lines)
#> 1:3 |> mean() |>
#>  as.character()
lint(
  text = code_lines,
  linters = pipe_continuation_linter()
)
#> <text>:1:15: style: [pipe_continuation_linter] `|>` should always have a space before it and a new line after it, unless the full pipeline fits on one line.
#> 1:3 |> mean() |>
#>               ^~
```

<sup>Created on 2022-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/object_name_linters.R
MichaelChirico: i think so, shouldn't it?

R/pipe_continuation_linter.R
MichaelChirico: i meant that we also used `cat()` elsewhere in the sister PRs, e.g.

https://github.com/r-lib/lintr/blob/ab8b77d006b10d56af5054aea288f9f9d35f1bf7/R/brace_linter.R#L22

so we should do a sweep to use `writeLines()` more broadly

R/paste_linter.R
```diff
@@ -34,6 +34,16 @@
 #'   linters = paste_linter()
 #' )
 #'
+#' lint(
+#'   text = "paste0(c('a', 'b'), sep = ' ')",
+#'   linters = paste_linter()
+#' )
+#'
+#' lint(
+#'   text = "paste0(rep('*', 10L), collapse='')",
```
MichaelChirico, line 43: let's use `collapse = ''` to be closer to the style guide

(I guess that means maybe we should switch to using `"` in the code too? WDYT?)

IndrajeetPatil: Ah, yes, that's a good point. 

I have changed this particular example here, but I will make a separate PR to change all the existing examples to make sure that they all follow the style guide.

R/object_name_linters.R
IndrajeetPatil: Yes, I agree. Created https://github.com/r-lib/lintr/issues/1665 to track this.

R/pipe_continuation_linter.R
IndrajeetPatil: Yes, done this more broadly in this PR itself.

MichaelChirico (approved):

PR #1670, opened 2022-10-11
jsonlite is a Suggests dependency

MichaelChirico:


IndrajeetPatil:
I think there also needs to be a skip for this in the tests for sarif output function.

IndrajeetPatil (approved):
LGTM, thanks!

PR #1671, opened 2022-10-11
crayon is a Suggested dependency

MichaelChirico:


IndrajeetPatil (approved):
LGTM. Thanks.

At some point, we should think if we want to switch to cli for formatting and pretty-printing condition messages.

Issue #1542, opened 2022-09-20
seq_linter() gives the wrong message for n:1 lints


MichaelChirico:
The result message should mention `rev()`, but currently it doesn't:

```
lint(text = ".N:1", linters = seq_linter())
# <text>:1:1: warning: [seq_linter] .N:1 is likely to be wrong in the empty edge case. Use seq_len(1) instead.
# .N:1
# ^~~~
```

It also says `seq_len(1)` instead of `seq_len(.N)`

IndrajeetPatil:
The same issue also holds for all other relevant functions I think:

``` r
library(lintr)

lint(text = "length(x):1", linters = seq_linter())
#> <text>:1:1: warning: [seq_linter] length(...):1 is likely to be wrong in the empty edge case. Use seq_along(...) instead.
#> length(x):1
#> ^~~~~~~~~~~

length(BOD):1
#> [1] 2 1

rev(seq_along(BOD))
#> [1] 2 1
```

<sup>Created on 2022-09-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

We have a TODO to track this:

https://github.com/r-lib/lintr/blob/a8d5d1f68b824fc335f69654bc62a7646bdad8ef/R/seq_linter.R#L67-L68

MichaelChirico:
Also for viz: trying to get r-core to give me more confidence in recommending a replacement in the `rev()` case:

https://bugs.r-project.org/show_bug.cgi?id=18406

Issue #531, opened 2020-10-07
seq_linter gives wrong message


MichaelChirico:
```
writeLines('mtcars[1:min(8, nrow(mtcars)), ]', tmp <- tempfile())
lintr::lint(tmp, seq_linter)
# /tmp/RtmpWXD4Q3/file4c4595f9321d2:1:8: warning: Avoid 1:min expressions, use seq_len. Disable this with a comment: `# nolint` (one line) or `# lintr: disable` (multiple lines). See go/tidy-style#linting for more.
# mtcars[1:min(8, nrow(mtcars)), ]
```

I think this is incorrect because `! min %in% bad_funcs`, i.e. it's not the `1:min` that's triggering the lint, it's the use of `nrow` in `1:`.

AshesITR:
Can you check if this issue occurs in the master branch?
I can't reproduce with that.

MichaelChirico:
Same for me on current master. `sessionInfo()` in details:

<details>

```
R version 4.0.0 (2020-04-24)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Linux Mint 19.3

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:
 [1] LC_CTYPE=en_SG.UTF-8       LC_NUMERIC=C               LC_TIME=en_SG.UTF-8       
 [4] LC_COLLATE=en_SG.UTF-8     LC_MONETARY=en_SG.UTF-8    LC_MESSAGES=en_SG.UTF-8   
 [7] LC_PAPER=en_SG.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_SG.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] lintr_2.0.1.9000

loaded via a namespace (and not attached):
 [1] ps_1.4.0           crayon_1.3.4       withr_2.3.0        rprojroot_1.3-2    assertthat_0.2.1  
 [6] R6_2.4.1           backports_1.1.10   magrittr_1.5       stringi_1.5.3      lazyeval_0.2.2    
[...]

PR #1674, opened 2022-10-11
Add examples to documentation: Part-7 (q-s)

IndrajeetPatil:


MichaelChirico:
can you check the coverage report?

IndrajeetPatil:
If I am reading this correctly, the coverage drop in `methods.R`

Screenshot 2022-10-11 at 19 11 06

is due to change here

Screenshot 2022-10-11 at 19 11 27

But this change wasn't made in this PR, so it's confusing why the coverage drop is associated with this PR.

IndrajeetPatil:
Okay, this one is surprising:

Screenshot 2022-10-11 at 19 19 33

R/redundant_equals_linter.R
```diff
@@ -6,6 +6,32 @@
...
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "if (any(x == TRUE)) 1",
+#'   linters = redundant_equals_linter()
+#' )
+#'
+#' lint(
+#'   text = "if (any(x != TRUE)) 0",
```
MichaelChirico, line 18: use FALSE on this one?

R/redundant_ifelse_linter.R
```diff
@@ -9,6 +9,35 @@
...
+#'   linters = redundant_ifelse_linter()
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "x >= 2.5",
+#'   linters = redundant_ifelse_linter()
+#' )
+#'
+#' lint(
```
MichaelChirico, line 31: maybe mention in a comment that this is the strict equivalent, but often just using the logical vector will "just work", something along those lines?

R/semicolon_linter.R
```diff
@@ -6,6 +6,42 @@
...
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "a <- 1;",
+#'   linters = semicolon_linter()
+#' )
+#'
+#' lint(
+#'   text = "a <- 1; b <- 1",
```
MichaelChirico, line 18: WDYT about replacing this with an online function example, or adding a third example? that's the other place this is tempting to use in ad hoc code....

R/seq_linter.R
```diff
@@ -9,6 +9,39 @@
 #' These often cause bugs when the right-hand side is zero.
 #' It is safer to use [base::seq_len()] or [base::seq_along()] instead.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "seq(length(x))",
```
MichaelChirico, line 15: let's do an n:1 example to show the rev() recommendation?

R/single_quotes_linter.R
```diff
@@ -2,13 +2,34 @@
 #'
 #' Check that only double quotes are used to delimit string constants.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "c('a', 'b')",
```
MichaelChirico, line 8: let's add an example where ' is ok (when " is in the string)

R/spaces_left_parentheses_linter.R
```diff
@@ -2,6 +2,19 @@
 #'
 #' Check that all left parentheses have a space before them unless they are in a function call.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
```
MichaelChirico, line 7: in an earlier linter (function_left_parentheses_linter? is that the name?) we showed if/function signature examples, should we do the reverse here?

R/sprintf_linter.R
```diff
@@ -1,7 +1,25 @@
...
 #'
-#' Check for an inconsistent number of arguments or arguments with incompatible types (for literal arguments) in
-#' `sprintf()` calls.
+#' Check for an inconsistent number of arguments or arguments with incompatible types
+#' (for literal arguments) in `sprintf()` calls.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "sprintf('hello %1$s %1$s %3$d', x, y, z)",
```
MichaelChirico, line 9: I find it odd to be using the 1$ redirections, those are rarely seen in practice IME. Let's just stick with plain templates which should be easier to read

R/string_boundary_linter.R
```diff
@@ -21,6 +21,35 @@
...
+#'   linters = string_boundary_linter()
+#' )
+#'
+#' lint(
+#'   text = 'endsWith(x, "z")',
+#'   linters = string_boundary_linter()
+#' )
+#'
+#' lint(
+#'   text = 'grepl("z$", x)',
```
MichaelChirico, line 49: this is not very illustrative of _why_ this argument exists. would be more instructive to mention about missing behavior

R/system_file_linter.R
```diff
@@ -3,6 +3,19 @@
 #' [system.file()] has a `...` argument which, internally, is passed to
 #'   [file.path()], so including it in user code is repetitive.
 #'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = 'system.file(file.path("path", "to", "data"), package = "foo")',
+#'   linters = system_file_linter()
```
MichaelChirico, line 10: missing example: when file.path() is the outer call

R/seq_linter.R
IndrajeetPatil: This gives wrong warning message right now. Related to #531 and #1542. So maybe we hold off on including an example? 

``` r
library(lintr)

lint(
  text = "ncol(x):1L",
  linters = seq_linter()
)
#> <text>:1:1: warning: [seq_linter] ncol(...):1L is likely to be wrong in the empty edge case. Use seq_len(1L) instead.
#> ncol(x):1L
#> ^~~~~~~~~~
```

<sup>Created on 2022-10-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: OK good point. we can add such an example along with closing the other issues.

R/spaces_left_parentheses_linter.R
IndrajeetPatil: Not sure what you mean by "reverse" here.

R/redundant_ifelse_linter.R
MichaelChirico: nit: use complete sentence syntax (capitalization, terminal `.`)

R/single_quotes_linter.R
```diff
@@ -15,6 +15,11 @@
 #'   linters = single_quotes_linter()
 #' )
 #'
+#' lint(
+#'   text = "\"'this is fine'\"",
```
MichaelChirico, line 19: do `writeLines()` with this one as well, it's hard to grasp the layers of escaping happening

MichaelChirico: also, maybe surround this with some other code? it's a little bit hard to grok when the entire expression is just a string. Even just `paste0(x, '"this is fine"')` would be better

PS isn't this example reversed, shouldn't `'` be on the outside?

R/spaces_left_parentheses_linter.R
MichaelChirico: Complementary examples to these:

https://github.com/r-lib/lintr/blob/615f72adc8f299d1265a072e5dbe21be8486f120/R/function_left_parentheses_linter.R#L21-L40

IndrajeetPatil: So including the okay examples but with a space to show that this doesn't lint with spaces_left linter?

``` r
library(lintr)

lint(
  text = "mean (x)",
  linters = spaces_left_parentheses_linter()
)

lint(
  text =  "stats::sd (c (x, y, z))",
  linters = spaces_left_parentheses_linter()
)

lint(
  text =  "if (TRUE) x else y",
  linters = spaces_left_parentheses_linter()
)

lint(
  text =  "foo <- function (x) (x + 1)",
  linters = spaces_left_parentheses_linter()
)
```

<sup>Created on 2022-10-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: Hmm, now I'm doubting everything :)

I definitely don't think examples like `mean (x)` are helpful... but maybe the better thing to do is just remove the `if()` example from `function_left_parentheses_linter()`? Since `if()` is irrelevant to that linter, and is covered here instead.

And then add cross-links between these two linters' help pages since they're so closely related.

WDYT?

IndrajeetPatil: SGTM. I have made the necessary changes. See if this is better.

MichaelChirico (approved):

PR #1680, opened 2022-10-11
Split S3 and S4 linters (and tests) into their own files

IndrajeetPatil:
This always made me do double-take if I was working in the right file.


MichaelChirico (approved):
I don't really see a need for this but sure :)

MichaelChirico (approved):

Issue #1657, opened 2022-10-10
`unnecessary_placeholder_linter()` could cover other magrittr operators too


MichaelChirico:
Follow-up to #1656 

IINM, `%<>%` and `%T>%` should work out of the box if we expand `text()` in XPath.

The ` %$%` is a bit more tricky and can possibly be dealt with separately.  

WDYT?

``` r
library(magrittr)

y <- 1
y %<>% sum(., 1) # lint
y
#> [1] 2

z <- 1
z %<>% sum(1)
z
#> [1] 2

# N.B. removing plots output manually to keep the reprex clean
1 %T>% plot %>% sum(1)
#> [1] 2
1 %T>% plot() %>% sum(1)
#> [1] 2
1 %T>% plot(.) %>% sum(1) # lint
#> [1] 2

mtcars %$% cor(wt, mpg)
#> [1] -0.8676594
mtcars %$% cor(.$wt, .$mpg) # lint
#> [1] -0.8676594

<sup>Created on 2022-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
```

_Originally posted by @IndrajeetPatil in https://github.com/r-lib/lintr/issues/1656#issuecomment-1272836124_

MichaelChirico:
This was done in #2046

PR #1681, opened 2022-10-11
`unnecessary_placeholder_linter()` covers a few other ops

IndrajeetPatil:
Part of #1657

### DIFF
--- a/R/unnecessary_placeholder_linter.R
+++ b/R/unnecessary_placeholder_linter.R
@@ -33,7 +33,7 @@
 unnecessary_placeholder_linter <- function() {
   # TODO(michaelchirico): handle R4.2.0 native placeholder _ as well
   xpath <- "
-  //SPECIAL[text() = '%>%']
+  //SPECIAL[text() = '%>%' or text() = '%T>%' or text() = '%<>%']
     /following-sibling::expr[
       expr/SYMBOL_FUNCTION_CALL
       and not(expr[
--- a/R/unnecessary_placeholder_linter.R
+++ b/R/unnecessary_placeholder_linter.R
@@ -4,6 +4,8 @@
 #'   passed as the first positional argument; using it can cause confusion
 #'   and impacts readability.
 #'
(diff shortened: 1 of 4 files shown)


tests/testthat/test-unnecessary_placeholder_linter.R
```diff
@@ -1,25 +1,40 @@
...
-  # . used in position other than first --> ok
-  expect_lint("x %>% foo(y, .)", NULL, unnecessary_placeholder_linter())
-  # ditto for nested usage
-  expect_lint("x %>% foo(y, bar(.))", NULL, unnecessary_placeholder_linter())
-  # . used twice --> ok
-  expect_lint("x %>% foo(., .)", NULL, unnecessary_placeholder_linter())
-  # . used as a keyword argument --> ok
-  expect_lint("x %>% foo(arg = .)", NULL, unnecessary_placeholder_linter())
-})
+skip_if_not_installed("patrick")
```
MichaelChirico, line 1: hmm at this point I think we have enough of our test suite using Patrick that we should make it required & skip this overhead...

MichaelChirico: #1682

R/unnecessary_placeholder_linter.R
```diff
@@ -33,7 +33,7 @@
 unnecessary_placeholder_linter <- function() {
   # TODO(michaelchirico): handle R4.2.0 native placeholder _ as well
   xpath <- "
-  //SPECIAL[text() = '%>%']
+  //SPECIAL[text() = '%>%' or text() = '%T>%' or text() = '%<>%']
```
MichaelChirico, line 36: Note which are covered in the documentation?

IndrajeetPatil: Good point. Done.

We can revise this once the native pipe is also supported.

MichaelChirico (approved):

PR #1683, opened 2022-10-12
keep "configurable" tag accurate

### LINKED ISSUE #1668: Add a test to ensure all configurable linters get the 'configurable' tag
I doubt it holds true even now, so we should fix that & add a regression test.

MichaelChirico:
Closes #1668


IndrajeetPatil (approved):
LGTM, thanks.

IndrajeetPatil:
Btw, out of curiosity, locally, if I run the tests (either using `devtools::test()` or `devtools::test_active_file()`) in `test-linter_tags.R`, I see:

```r
── Skip (test-linter_tags.R:89): lintr help files are up to date ───────────────
Reason: dir.exists(helper_db_dir) is not TRUE
```

How can I change that?

P.S. To test changes in this PR, I manually created `lintr_db` (via `lintr_db <- available_linters(exclude_tags = NULL)`).

MichaelChirico:
You need to run it from the installed package, not `devtools::test()` (which I guess runs `pkgload::load_all()`), since we need the package help file to be available, which is retrieved from the installed location.

So to test it locally:

 1. `R CMD INSTALL .`
 2. `library(lintr); testthat::test_file('tests/testthat/test-linter_tags.R')`

IndrajeetPatil:
Got it. Thanks!

Should we note this down in the test file as well?

PR #1690, opened 2022-10-12
Further reading for no tabs linter

### LINKED ISSUE #1672: Add more explanation to `no_tab_linter()`
Currently, [it](https://lintr.r-lib.org/dev/reference/no_tab_linter.html) only states that one should be preferred over the other, but gives no reason as to why. And, no, Richard Hendricks did not make me write this 🙃

Screenshot 2022-10-11 at 16 33 52

IndrajeetPatil:
Closes #1672

### DIFF
--- a/R/no_tab_linter.R
+++ b/R/no_tab_linter.R
@@ -23,7 +23,6 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #'
 #' @references
-#' - https://softwareengineering.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e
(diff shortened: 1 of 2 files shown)


R/no_tab_linter.R
```diff
@@ -18,6 +21,10 @@
 #'
 #' @evalRd rd_tags("no_tab_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
+#'
+#' @references
+#' - https://www.jwz.org/doc/tabs-vs-spaces.html
```
MichaelChirico, line 26: my only concern is that blogs like this tend to wind up as a maintenance headache in the long term as redirects evolve / pages die.

but fine to leave it in for now

MichaelChirico (approved):

PR #1691, opened 2022-10-13
1:10 doesnt lint, optionally

### LINKED ISSUE #1155: implicit_integer_linter shouldn't lint 1:10 (optionally?)

MichaelChirico:
Closes #1155

### DIFF
--- a/R/implicit_integer_linter.R
+++ b/R/implicit_integer_linter.R
@@ -2,6 +2,8 @@
 #'
 #' Check that integers are explicitly typed using the form `1L` instead of `1`.
 #'
+#' @param allow_colon Logical, default `FALSE`. If `TRUE`, expressions involving `:`
+#'   won't throw a lint regardless of whether the inputs are implicitly integers.
 #' @examples
 #' # will produce lints
 #' lint(
@@ -11,6 +13,11 @@
 #'
 #' lint(
 #'   text = "x[2]",
+#'   linters = implicit_integer_linter()
+#' )
+#'
+#' lint(
+#'   text = "1:10",
 #'   linters = implicit_integer_linter()
 #' )
 #'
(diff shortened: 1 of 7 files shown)


R/implicit_integer_linter.R
```diff
@@ -30,18 +37,28 @@
...
+#' lint(
+#'   text = "1:10",
+#'   linters = implicit_integer_linter(allow_colon = TRUE)
+#' )
+#'
 #' @evalRd rd_tags("implicit_integer_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-implicit_integer_linter <- function() {
+implicit_integer_linter <- function(allow_colon = FALSE) {
```
IndrajeetPatil, line 48: Shouldn't the default argument be `TRUE` for this parameter? 
Which cases that should produce a lint will be missed out on, if we default to `TRUE`?

MichaelChirico: Done for back-compatiblity... current HEAD implicitly uses `FALSE`.

I'd be open to changing it later, but at least for this release we should use this default.

IndrajeetPatil: SGTM.

Can you please file a new issue to remember to do this? And maybe also choose the appropriate milestone for the issue?

IndrajeetPatil (approved):

R/implicit_integer_linter.R
MichaelChirico: #1694. Let's hold off on choosing a milestone, but maybe 3.1.0? Let's see if we get any feedback after 3.0.2 (I guess it's really 3.0.3 now) is released with this PR.

Issue #1692, opened 2022-10-13
Use public interface in tests wherever possible (don't use :::)


MichaelChirico:
e.g. this suite can be replaced by `expect_lint()` tests:

https://github.com/r-lib/lintr/blob/bed54c55523e96f7c89154ae31702cd74bb209d8/tests/testthat/test-implicit_integer_linter.R#L2-L35

Here's a list of tests using `:::`, generated with `grep -Flr ":::" tests/testthat | awk -F/ '{print "- [ ] " $3}' | sort`, we can cross some off if they are just used for coverage & testing with public API is difficult:

- [ ] test-absolute_path_linter.R #2801
- [ ] test-cache.R #2801
- [x] test-ci.R
- [x] test-commented_code_linter.R #1726
- [x] test-comments.R 
- [x] ~test-dir_linters.R~
- [x] test-error.R
- [ ] test-exclusions.R #1720
- [ ] test-get_source_expressions.R #2805
- [x] test-implicit_integer_linter.R #1693 
- [ ] test-linter_tags.R #2801
- [ ] test-lint_dir.R #2801
- [x] test-lint_package.R
- [ ] test-make_linter_from_regex.R #2801
- [ ] test-methods.R #2801
- [x] test-namespace_linter.R
- [x] test-namespace.R #1721
- [ ] test-normalize_exclusions.R #2801
- [ ] test-object_name_linter.R #2801
- [ ] test-parse_exclusions.R #2801
- [ ] test-settings.R #2801
- [x] test-undesirable_operator_linter.R
- [x] test-use_lintr.R #1774
- [x] test-utils.R #1720

IndrajeetPatil:
I think `test-namespace_linter.R` is a false positive here. It's picking up `:::` from tests where they are deliberate. E.g.

```r
  expect_lint(
    "stats:::sdd(c(1,2,3))",
    list(message = rex::rex("'sdd' does not exist in {stats}")),
    linter
  )
```

PR #1693, opened 2022-10-13
use parameterized test on public API for implicit_integer_linter()

MichaelChirico:
Part of #1692

### DIFF
--- a/tests/testthat/test-implicit_integer_linter.R
+++ b/tests/testthat/test-implicit_integer_linter.R
@@ -1,39 +1,56 @@
-# styler: off
-test_that("single numerical constants are properly identified ", {
-  # Test single numerical constants
-  is_implicit <- lintr:::is_implicit_integer
+skip_if_not_installed("tibble")
+local({
+  # Note: cases indicated by "*" are whole numbers, but don't lint because the user has
+  #   effectively declared "this is a double" much as adding '.0' is otherwise accepted.
+  cases <- tibble::tribble(
+            ~num_value_str, ~should_lint,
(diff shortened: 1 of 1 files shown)


IndrajeetPatil (approved):
LGTM. Thanks!

Btw, why do you use `local()` with `{patrick}` tests? 
I didn't use it for the tests I wrote, and I can change that if I understand the reasoning here.

MichaelChirico:
> Btw, why do you use local() with {patrick} tests?

because of the extra variables associated with the test (`cases`, `int_max`, `cases_int_max`)... those are needed for the test but make the tests less "hermetic" by polluting the local environment (can possibly be seen by other tests, by mistake). `local()` ensures those are removed "on.exit" from the `local()` execution frame.

There's another approach that could work with {rlang} but I prefer the `local()` one for now...

https://github.com/r-lib/withr/issues/205

MichaelChirico:
(with plain {testthat} tests, this is encapsulated in `test_that()`, but we need this workaround for {patrick}... hmm maybe that could be improved upstream...)

IndrajeetPatil:
> make the tests less "hermetic" by polluting the local environment (can possibly be seen by other tests, by mistake)

I see. Thanks for clarifying!

My understanding was that objects created in each test file live in their own local environments and won't be accessible from other test files. But maybe there are contexts (like `{patrick}` tests) where this is possible?

MichaelChirico:
> My understanding was that objects created in each test file live in their own local environments

I think that's right, but there are also the other tests in the same file

PS filed: https://github.com/google/patrick/issues/13

PR #1702, opened 2022-10-13
Fix assignment_linter when comment further on in brace

### LINKED ISSUE #1701: `assignment_linter(allow_trailing = FALSE)` fails when comment in brace
I have found an issue with the new allow_trailing argument for `assignment_linter` where if there is a comment somewhere in a set of braces, then it will flag all of the assignments in the braces

```r
text <- 'testthat::test_that("app ui", {
  ui <- appUI()
  golem::expect_shinytaglist(ui)
  # Check that formals have not been removed
  fmls <- formals(appUI)
    
  for (i in "request") {
    testthat::expect_true(i %in% names(fmls))
  }
})

testthat::test_that("app server", {
  server <- appServer
  testthat::expect_type(server, "closure")
  # Check that formals have not been removed
  fmls <- formals(appServer)
  for (i in c("input", "output", "session")) {
    testthat::expect_true(i %in% names(fmls))
  }
})'

lint(
  text = text,
  linters = assignment_linter(allow_trailing = FALSE)
)
```

Expected: no lints
Actual: 4 lints

```
<text>:2:6: style: [assignment_linter] Assignment <- should not be trailing at end of line
  ui <- appUI()
     ^~
<text>:5:8: style: [assignment_linter] Assignment <- should not be trailing at end of line
  fmls <- formals(appUI)
       ^~
<text>:13:10: style: [assignment_linter] Assignment <- should not be trailing at end of line
  server <- appServer
         ^~
<text>:16:8: style: [assignment_linter] Assignment <- should not be trailing at end of line
  fmls <- formals(appServer)
       ^~
```

ashbaldry:
Resolves #1701

### DIFF
--- a/R/assignment_linter.R
+++ b/R/assignment_linter.R
@@ -67,7 +67,7 @@ assignment_linter <- function(allow_cascading_assign = TRUE, allow_right_assign 
       "//EQ_SUB",
       "//EQ_FORMALS"
     ),
-    "[@line1 < following-sibling::*[1]/@line1 or @line1 = following-sibling::*[1][self::COMMENT]/@line1]"
+    "[@line1 < following-sibling::expr[1]/@line1]"
   )
 
   xpath <- paste(collapse = " | ", c(
--- a/R/assignment_linter.R
+++ b/R/assignment_linter.R
@@ -67,7 +67,7 @@ assignment_linter <- function(allow_cascading_assign = TRUE, allow_right_assign 
       "//EQ_SUB",
       "//EQ_FORMALS"
     ),
-    "[@line1 < following-sibling::*[1]/@line1 or //COMMENT]"
+    "[@line1 < following-sibling::*[1]/@line1 or @line1 = following-sibling::*[1][self::COMMENT]/@line1]"
   )
 
   xpath <- paste(collapse = " | ", c(
--- a/tests/testthat/test-assignment_linter.R
+++ b/tests/testthat/test-assignment_linter.R
@@ -107,7 +107,10 @@ test_that("arguments handle trailing assignment operators correctly", {
     ),
     assignment_linter(allow_trailing = FALSE)
   )
+})
 
+test_that("allow_trailing interacts correctly with comments in braced expressions", {
+  linter <- assignment_linter(allow_trailing = FALSE)
   expect_lint(
     trim_some("
     {
@@ -117,7 +120,7 @@ test_that("arguments handle trailing assignment operators correctly", {
     }
     "),
     NULL,
-    assignment_linter(allow_trailing = FALSE)
+    linter
   )
 
   expect_lint(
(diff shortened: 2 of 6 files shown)


tests/testthat/test-assignment_linter.R
```diff
@@ -107,4 +107,22 @@ test_that("arguments handle trailing assignment operators correctly", {
     ),
     assignment_linter(allow_trailing = FALSE)
   )
+
+  expect_lint(
+    "{x <- 'foo' \n #foo \n y <- 'bar'}",
```
MichaelChirico, line 112: could you use trim_some() here for readability? thanks

tests/testthat/test-assignment_linter.R
```diff
@@ -107,4 +107,22 @@ test_that("arguments handle trailing assignment operators correctly", {
     ),
     assignment_linter(allow_trailing = FALSE)
   )
+
+  expect_lint(
+    "{x<-1\n#blah\ny<-2}",
```
MichaelChirico, line 112: sorry again for the confusion, here's an example:

```r
expect_lint(
  trim_some("
    {
      x <- '#x'
      y <- '#y'
    }
  "),
  NULL,
  assignment_linter(allow_trailing = FALSE)
)
```

ashbaldry: Okay, now I really understand! Thanks for the clarification :)

R/assignment_linter.R
```diff
@@ -67,7 +67,7 @@ assignment_linter <- function(allow_cascading_assign = TRUE, allow_right_assign
       "//EQ_SUB",
       "//EQ_FORMALS"
     ),
-    "[@line1 < following-sibling::*[1]/@line1 or //COMMENT]"
+    "[@line1 < following-sibling::*[1]/@line1 or @line1 = following-sibling::*[1][self::COMMENT]/@line1]"
```
MichaelChirico, line 70: would this work as one condition instead?

```
@line1 < following-sibling::*[not(self::COMMENT)][1]/@line1
```

MichaelChirico: or perhaps even

```
@line1 < following-sibling::expr[1]/@line1
```

(less confident)

ashbaldry: They both pass all of the tests, and I've added one more to test that it doesn't detect trailing assignment in strings. They're both better than the original query. I've updated with the second suggestion; as my xpath knowledge has come from looking at the lintr queries, I'm happy to go with the first one if you're more comfortable with it

MichaelChirico:
Thanks, LGTM! Please add a NEWS entry & then we can merge

R/assignment_linter.R
MichaelChirico: as long as it passes tests it should be ok

MichaelChirico:
Thanks!

MichaelChirico (approved):

PR #1704, opened 2022-10-14
unnecessary_lambda_linter catches functions with braces

### LINKED ISSUE #1703: unnecessary_lambda_linter should work with explicit return()
Split off from #1567, since handling `return()` as well is another layer of complication

Relevant test cases:

```
 test_that("cases with braces are caught", {
   linter <- unnecessary_lambda_linter()
  
   expect_lint(
     "lapply(x, function(xi) { return(print(xi)) })",
     "xxx",
     linter
   )
   
   expect_lint(
     trim_some("
       lapply(x, function(xi) {
         return(print(xi))
       })
     "),
     "xxx",
     linter
   )
 })
```

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/issues/1567#issuecomment-1260456007_

MichaelChirico:
Closes #1703

### DIFF
--- a/tests/testthat/test-unnecessary_lambda_linter.R
+++ b/tests/testthat/test-unnecessary_lambda_linter.R
@@ -81,3 +81,34 @@ test_that("purrr-style anonymous functions are also caught", {
     unnecessary_lambda_linter()
   )
 })
+
+test_that("cases with braces are caught", {
+  linter <- unnecessary_lambda_linter()
+  print_msg <- rex::rex("Pass print directly as a symbol to lapply()")
+
+  expect_lint(
+    "lapply(x, function(xi) { print(xi) })",
+    print_msg,
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      lapply(x, function(xi) {
+        print(xi)
+      })
+    "),
+    print_msg,
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      lapply(x, function(xi) {
+        print(xi)
+      })
+    "),
+    print_msg,
+    linter
+  )
+})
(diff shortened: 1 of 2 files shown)


tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -81,3 +81,34 @@ test_that("purrr-style anonymous functions are also caught", {
     unnecessary_lambda_linter()
   )
 })
+
+test_that("cases with braces are caught", {
+  linter <- unnecessary_lambda_linter()
+  print_msg <- rex::rex("Pass print directly as a symbol to lapply()")
```
IndrajeetPatil, line 87: Nitpick: Everywhere else we are using _lint_msg_ as a variable name

tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -81,3 +81,34 @@ test_that("purrr-style anonymous functions are also caught", {
...
+
+  expect_lint(
+    trim_some("
+      lapply(x, function(xi) {
+        print(xi)
+      })
+    "),
+    print_msg,
+    linter
+  )
```
IndrajeetPatil, line 113: Is the second test here a duplicate of the first one?

tests/testthat/test-unnecessary_lambda_linter.R
MichaelChirico: used print_ in this case because the message is customized for the print() function in the code, in case there are other tests w different messages

tests/testthat/test-unnecessary_lambda_linter.R
MichaelChirico: oh, appears so, thanks

MichaelChirico: turned it into a test with no lint + braces now, PTAL

IndrajeetPatil: LGTM, thanks.

IndrajeetPatil (approved):

PR #1708, opened 2022-10-14
Include `stringsAsFactors = FALSE` where relevant

IndrajeetPatil:
Also add the corresponding linter to our workflow

### DIFF
--- a/.dev/revdep_get_repos.R
+++ b/.dev/revdep_get_repos.R
@@ -50,12 +50,20 @@ git_urls <- extract_github_repo(urls)
 matched <- nzchar(git_urls)
 
 utils::write.csv(
-  data.frame(package = lintr_pkg[!matched], repo = file.path("https://github.com/cran", lintr_pkg[!matched])),
+  data.frame(
+    package = lintr_pkg[!matched],
+    repo = file.path("https://github.com/cran", lintr_pkg[!matched]),
+    stringsAsFactors = FALSE
+  ),
   "revdep-no-repos",
   row.names = FALSE, quote = FALSE
 )
 utils::write.csv(
-  data.frame(package = lintr_pkg[matched], repo = git_urls[matched]),
+  data.frame(
+    package = lintr_pkg[matched],
+    repo = git_urls[matched],
+    stringsAsFactors = FALSE
+  ),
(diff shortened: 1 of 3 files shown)


.dev/revdep_get_repos.R
```diff
@@ -50,12 +50,20 @@ git_urls <- extract_github_repo(urls)
 matched <- nzchar(git_urls)
 
 utils::write.csv(
-  data.frame(package = lintr_pkg[!matched], repo = file.path("https://github.com/cran", lintr_pkg[!matched])),
+  data.frame(
```
MichaelChirico, line 53: we don't need to do this in .dev scripts. I hope all our maintainers are on 4.0 by now :)

tests/testthat/test-linter_tags.R
```diff
@@ -13,7 +13,11 @@ test_that("validate_linter_db works as expected", {
   )
   expect_false(suppressWarnings(lintr:::validate_linter_db(df_empty, "mypkg")))
 
-  df <- data.frame(linter = "absolute_path_linter", tags = "robustness")
+  df <- data.frame(
```
MichaelChirico, line 16: we should rather get rid of this as part of #1692 :)

.dev/revdep_get_repos.R
IndrajeetPatil: Yes, possibly. But is there any harm in doing so? 🙃

Without it, if a future PR changes anything about this file, the build will fail because of this lint.

tests/testthat/test-linter_tags.R
IndrajeetPatil: True, but maybe still worth including this change until that issue is closed?

MichaelChirico: agree

.dev/revdep_get_repos.R
MichaelChirico: we should probably just exclude .dev from .lintr_new

MichaelChirico: (another good use case for context-specific config... would be great to turn off that specific linter in this directory)

MichaelChirico (approved):

Issue #1698, opened 2022-10-13
Investigate further warning message seen in lint GHA workflow


IndrajeetPatil:
cf. https://github.com/r-lib/lintr/actions/runs/3240547631/jobs/5311303504

```r
Warning message:
In add_exclusions(exclusions, excluded_lines, linters_string, exclude_linter_sep,  :
  Could not find linter named ‘one_linter’ in the list of active linters. Make sure the linter is uniquely identified by the given name or prefix.
```

MichaelChirico:
It's coming from here, not sure why it's only happening in that workflow though:

https://github.com/r-lib/lintr/blob/4b37029f989629287e910fa3aba46ab15d1975cb/tests/testthat/test-unreachable_code_linter.R#L124-L138

IndrajeetPatil:
There is a new warning [now](https://github.com/r-lib/lintr/actions/runs/3298818644/jobs/5441425380), in addition to the existing one:

```
1: In parse_check_usage(fun, known_used_symbols = known_used_symbols,  :
  Possible bug in lintr: Couldn't parse usage message ‘<anonymous> : <anonymous> : <anonymous>: local variable 'col2' assigned but may not be used
’. Ignoring 2 usage warnings. Please report an issue at https://github.com/r-lib/lintr/issues.
```

AshesITR:
> cf. https://github.com/r-lib/lintr/actions/runs/3240547631/jobs/5311303504
> 
> 
> 
> ```r
> 
> Warning message:
> 
> In add_exclusions(exclusions, excluded_lines, linters_string, exclude_linter_sep,  :
> 
>   Could not find linter named ‘one_linter’ in the list of active linters. Make sure the linter is uniquely identified by the given name or prefix.
> 
> ```

That should easily be fixed by changing the nolint directive like we do in other tests.

AshesITR:
> There is a new warning [now](https://github.com/r-lib/lintr/actions/runs/3298818644/jobs/5441425380), in addition to the existing one:
> 
> 
> 
> ```
> 
> 1: In parse_check_usage(fun, known_used_symbols = known_used_symbols,  :
> 
>   Possible bug in lintr: Couldn't parse usage message ‘<anonymous> : <anonymous> : <anonymous>: local variable 'col2' assigned but may not be used
> 
> ’. Ignoring 2 usage warnings. Please report an issue at https://github.com/r-lib/lintr/issues.
> 
> ```

[...]

PR #1709, opened 2022-10-14
Check if #1698 warning disappears with the other workflow

IndrajeetPatil:


IndrajeetPatil:
Also seen with the same workflow as the lint changed workflow.

PR #1711, opened 2022-10-14
Make lint workflow more comprehensive

IndrajeetPatil:
Prompted by https://github.com/r-lib/lintr/pull/1710.

Currently, we are at the mercy of files being changed in order for lints to be detected by the lints-changed workflow.

Instead, we can make our lint workflow itself more comprehensive. This way, we can catch and clean them once in a while. For lint workflow, we don't even error when lints are found, so this shouldn't be disruptive.

WDYT?


MichaelChirico:
Removed myself since you marked as draft afterwards, please add me back when you want some eyes on this

IndrajeetPatil:
Yeah, sorry, forgot to remove the review request.

I saw just how many lints were turning up and decided to do some clean up first and then mark it ready for a review. The output will then be more approachable and actionable.

PR #1712, opened 2022-10-14
Get rid of `expect_identical_linter()` lints

IndrajeetPatil:


tests/testthat/test-methods.R
```diff
@@ -71,10 +71,7 @@ test_that("as.data.frame.lints", {
     stringsAsFactors = FALSE
   )
 
-  expect_equal(
-    df,
-    exp
-  )
+  expect_equal(df, exp)
```
MichaelChirico, line 74: what's going on here? 🙃

IndrajeetPatil: `df` columns are not of integer type here:

```r
── Failure (test-methods.R:74): as.data.frame.lints ────────────────────────────
`df` (`actual`) not identical to `exp` (`expected`).

`actual$line_number` is a double vector (1, 2)
`expected$line_number` is an integer vector (1, 2)

`actual$column_number` is a double vector (1, 6)
`expected$column_number` is an integer vector (1, 6)
```

IndrajeetPatil: Was going to fix it in a different PR, but might as well do it here.

MichaelChirico: thanks, I would have suggested just leaving the style fix for that other PR

IndrajeetPatil: So should I undo this here?

MichaelChirico: no it's fine, minor point here

MichaelChirico (approved):

MichaelChirico:
madman, you actually did it! 🤟🤟

PR #1713, opened 2022-10-15
Get rid of `expect_identical_linter()` lints: Part-2

IndrajeetPatil:
Missed a few in https://github.com/r-lib/lintr/pull/1712

### DIFF
--- a/tests/testthat/test-linter_tags.R
+++ b/tests/testthat/test-linter_tags.R
@@ -73,7 +73,7 @@ test_that("warnings occur only for deprecated linters", {
       }
     )
   })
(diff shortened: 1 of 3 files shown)


MichaelChirico (approved):
could you edit the PR description (also for #1712)?

"get rid of" doesn't sound right. expect_equal() isn't "wrong" or harmful, the phrasing doesn't sound right. we are just preferring a stronger default test.

More importantly, we should make clear that this is only a partial fix (right?), and that we are simply addressing the cases where a drop-in replacement works "out of the box", leaving cases that require a refactor (etc) for a more intensive future PR

IndrajeetPatil:
Good point. Done for both PRs.

PR #1715, opened 2022-10-15
Improve object_usage_linter()

### LINKED ISSUE #1714: Not all usage issues reported back to user
``` r
tst <- function() { a <- ...; a; b }
codetools::checkUsage(tst)
#> <anonymous>: ... may be used in an incorrect context (<text>:1)
#> <anonymous>: no visible binding for global variable 'b' (<text>:1)
```

<sup>Created on 2022-10-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Lintr only reports one of the problems:

``` r
lintr::lint(text = "tst <- function() { a <- ...; a; b }", linters = lintr::object_usage_linter())
#> <text>:1:34: warning: [object_usage_linter] no visible binding for global variable 'b'
#> tst <- function() { a <- ...; a; b }
#>                                  ^
```

<sup>Created on 2022-10-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

___

Checking the code of `R/codetools.R`, all issues are collected using the format `signalUsageIssue(message)`. The regex - if I read it correctly - implicitly assumes there always is an `sQuote(...)` in the message.

This is not the case for the following messages, scouring the source code:

```
w$signal("... may be used in an incorrect context", w) # l. 416
w$signal(paste(a, "may be used in an incorrect context:", pasteExpr(e)), w) # l. 436
w$signal(paste("wrong number of arguments to '.Internal':", pasteExpr(e)), w) # l. 676
w$signal(paste("bad argument to '.Internal':", pasteExpr(e[[2]])), w) # l. 682
```

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/1661#pullrequestreview-1143188740_

AshesITR:
fixes #1714


R/object_usage_linter.R
```diff
@@ -268,30 +275,26 @@ parse_check_usage <- function(expression,
...
-        capture(
-          name = "message",
-          "possible error in ", capture(name = "name", anything), ": ", anything
-        ),
-        line_info
-      )
-    )
+    warning("Possible bug in lintr: Couldn't parse usage message ", sQuote(vals[missing][[1L]]), ". ",
+            "Ignoring ", sum(missing), " usage warnings.")
+    res <- res[!missing, ]
```
AshesITR, line 295: Left this in as a safety net. We could also go the offensive programming route of deleting the `is.na(res$message)` related part alltogether. WDYT?

tests/testthat/test-object_usage_linter.R
```diff
@@ -545,7 +545,8 @@ test_that("fallback works", {
     "),
     list(
       message = rex::rex("no visible global function definition for ", anything, "non_existing_assign<-"),
-      column_number = 6L
+      line_number = 2L,
+      column_number = 3L
```
AshesITR, line 549: Added benefit of the improved lint locations is we actually find a better way to place this one :)

R/object_usage_linter.R
```diff
@@ -122,9 +122,16 @@ object_usage_linter <- function(interpret_glue = TRUE, skip_with = TRUE) {
...
       )
-
       nodes <- unclass(lintable_symbols)[matched_symbol]
-      nodes[is.na(matched_symbol)] <- list(fun_assignment)
+
+      # fallback to line based matching if no symbol is found
+      nodes[is.na(matched_symbol)] <- lapply(which(is.na(matched_symbol)), function(i) {
+        line_based_match <- xml2::xml_find_first(
+          fun_assignment,
+          glue::glue("descendant::expr[@line1 = {res$line1[i]} and @line2 = {res$line2[i]}]")
```
AshesITR, line 131: Not sure if `find_first` will always find the outermost expression, i.e. the one covering the complete line range, but no additional lines.
I'm 90% sure though, because all other exprs will be descendants of that specific one.

R/object_usage_linter.R
```diff
@@ -268,30 +275,26 @@ parse_check_usage <- function(expression,
...
       capture(
         name = "message",
-        anything,
-        one_of(quote, "\u2018"),
-        capture(name = "name", anything),
-        one_of(quote, "\u2019"),
-        anything
+        zero_or_more(any, type = "lazy"),
+        maybe(
+          one_of(quote, "\u2018"),
```
AshesITR, line 280: Since we use `withr::with_options()` to disable fancy quotes now, this could be simplified to `"'"`. WDYT?

MichaelChirico: SGTM, let's do it

R/object_usage_linter.R
MichaelChirico: I like this approach. Maybe leave a TODO to revisit this in "a while" if we don't hear any user feedback?

R/object_usage_linter.R
MichaelChirico: Hmm, something like is `xml_find_first()` doing a greedy search under `descendant::expr`? FWIW here's a quick example:

```r
xml = read_xml('<x><a><a><a><a></a></a></a></a></x>')
xml |> xml_find_first('//x/descendant::a') |> as.character() |> writeLines()
<a>
  <a>
    <a>
      <a/>
    </a>
  </a>
</a>
```

does appear to find the outermost one. Not sure if that's generalizable, or if it's specifically what you had in mind.

MichaelChirico, line 131: maybe a good place to use `glue_data()`?

`glue_data(res[i, ], "descendant::expr[@line1 = {line1} and @line2 = {line2}]")`

or maybe `res[i, ]` has a lot of overhead (I'm always wary of creating a lot of `data.frames` / running `[.data.frame` a lot), and `glue_data(res, "... {line1[i]} ... {line2[i]}")` would be better...

MichaelChirico:
High-level feedback -- the NEWS bullet is great, but I really couldn't figure out what's going on from the original issue description (#1714), consider going back and adding a quick summary reprex at the top for posterity

R/object_usage_linter.R
AshesITR: Res will be quite small because it is `nrow(lints)`, so no worries. Also, if we don't use `res[i, ]`, glue_data saves 0 bytes, displeasing my inner code-golfer 😉

AshesITR:
Moved the reprex around and expanded the checkUsage one so both have the same code. That should make it much easier to see the problem.

MichaelChirico (approved):

PR #1720, opened 2022-10-16
Use public interface to test internal functions: Part-1

IndrajeetPatil:
Part of #1692


tests/testthat/test-exclusions.R
```diff
@@ -1,18 +1,3 @@
-test_that("line_info works as expected", {
-  expect_identical(
-    lintr:::line_info(integer(), type = "end"),
```
IndrajeetPatil, line 3: This is tested indirectly via `exclude()`.

AshesITR (approved):
Thanks!

PR #1722, opened 2022-10-16
customize lint message in literal_coercion_linter()

### LINKED ISSUE #1439: customize literal_coercion_linter message
we should customize this to match what's actually linted. but it predates this PR, so we can put that as a follow-up issue

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/1437#discussion_r908656360_

MichaelChirico:
Closes #1439


R/literal_coercion_linter.R
```diff
@@ -84,13 +84,18 @@ literal_coercion_linter <- function() {
 
     bad_expr <- xml2::xml_find_all(xml, xpath)
 
+    coercion_str <- xml2::xml_text(bad_expr)
+    # the linter logic should ensure that it's safe to run eval() here
+    literal_equivalent_str <- vapply(str2expression(coercion_str), function(expr) deparse1(eval(expr)), character(1L))
```
MichaelChirico, line 89: Oh, hmm, for the `rlang::` coercions, this `eval()` will induce a dependency on {rlang}.

I see two options here:

 1. Attempt to translate `rlang::lgl()` into `as.logical()` and run that `eval()` instead
 2. Check `requireNamespace("rlang")`, using the original generic message if it's not available and `eval()` otherwise. Would require us (IINM) to add `rlang` to Suggests

The latter is easier to implement (in particular, inputs like `lgl(1,)` make things more complicated for option 1), and avoids the possibility that `as.logical()` and `lgl()` differ sometimes.

Any opinions?

AshesITR: I'm fine with option 2. Option 1 seems like bugs waiting to happen.

R/literal_coercion_linter.R
```diff
@@ -84,13 +82,31 @@ literal_coercion_linter <- function() {
...
+      #   but this seems like overkill. Just use a generic message and move on.
+      lint_message <- paste(
+        "Use literals directly where possible, instead of coercion.",
+        "c.f. 1L instead of as.integer(1) or rlang::int(1), or NA_real_ instead of as.numeric(NA).",
+        "NB: this message can be improved to show a specific replacement if 'rlang' is installed."
+      )
+    } else {
+      coercion_str <- xml2::xml_text(bad_expr)
+      # the linter logic & rlang requirement should ensure that it's safe to run eval() here
+      literal_equivalent_str <- vapply(str2expression(coercion_str), function(expr) deparse1(eval(expr)), character(1L))
```
AshesITR, line 104: NB: this will return `1` instead of `1.0` for `as.numeric(1)`.
```r
deparse1(eval(quote(as.numeric(1))))
#> [1] "1"
```
Feel free to make this a follow-up though.

MichaelChirico: Hmm using `1.0` assumes `implicit_integer_linter()` is active though right? not sure what our default message should be in that case...

AshesITR: Well, I think the general advice to use `1.0` for `as.numeric(1)` seems very good, even if implict integers aren't linted. The generic method seems fine as-is.

MichaelChirico: Added a TODO for now

MichaelChirico: > It could be more geared towards rlang seeing that the base functions will never cause it, though.

Yes and no... the generic will be thrown twice here, for example (if `rlang` is missing):

```r
c(
  as.integer(1),
  rlang::int(1)
)
```

AshesITR: Oh, I didn't notice that. Is there harm in vectorizing the message code?

MichaelChirico: Just maintenance overhead. Started writing it like that but code was already getting ugly / requiring careful refactor to handle the two branches without too much duplication.

Stopped because it doesn't seem that likely to occur in practice IMO. Seems rare that user (1) uses `rlang`, but doesn't have it installed in the linting environment and (2) uses `rlang` inconsistently, i.e., `chr()` in one place but `as.integer()` in another

AshesITR: Okay, seems fair.

NEWS.md
```diff
@@ -64,6 +64,8 @@
 
 * `object_name_linter()` gains parameter `regexes` to allow custom naming conventions (#822, #1421, @AshesITR)
 
+* `literal_coercion_linter()` reports a replacement in the lint message (#1439, @MichaelChirico)
```
AshesITR, line 67: It's not immediately clear to me what this means. Maybe "`literal_coercion_linter()` reports the equivalent literal as a replacement in the lint message"?

AshesITR (commented):
Basically LGTM

NEWS.md
MichaelChirico: Gave some concrete examples, PTAL

AshesITR (approved):

PR #1723, opened 2022-10-17
Move path linters to their own files

IndrajeetPatil:
To have a one-to-one correspondence between linter and its test file


AshesITR (commented):
WDYT about calling the path utilities file `path_utils.R` instead? `utils_*` seems odd from a grammar point of view.

Content-wise it's good to merge.

R/absolute_path_linter.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' Check that no absolute paths are used (e.g. "/var", "C:\\System", "~/docs").
+#'
+#' @param lax Less stringent linting, leading to fewer false positives.
+#' If `TRUE`, only lint path strings, which
+#'
+#' * contain at least two path elements, with one having at least two characters and
+#' * contain only alphanumeric chars (including UTF-8), spaces, and win32-allowed punctuation
+#'
+#' @include utils_path_linters.R
```
AshesITR, line 11: This shouldn't be necessary since there is no build-time dependency on the file?

IndrajeetPatil: Ah, yes, you are right. I thought we were taking the same approach here as `make_linter_from_regex()`, but we aren't.

Will change that.

IndrajeetPatil:
> WDYT about calling the path utilities file `path_utils.R` instead? `utils_*` seems odd from a grammar point of view.

I read it like "utilities for *", and having that prefix clusters all utilities separately from the "main" functions.

But this is just a personal preference. I've renamed the file to `path_utils.R`.

AshesITR (approved):

PR #1724, opened 2022-10-17
Use `xp_text_in_table()` to create XPath variables

IndrajeetPatil:
Before I do this more widely, here is a POC PR.

I think using `xp_text_in_table()` to create XPath variables is more readable and easily extendable than listing each possible option with `text() = ` in the XPath itself.

Additionally, just looking at the values, it's sometimes difficult to gather what do they represent. But when stored in a vector and assigned to a variable with a meaningful name, this is no longer the case.  

We do this in [some places](https://github.com/r-lib/lintr/blob/080613e29644290a835658d7019ebad05b4b60f3/R/literal_coercion_linter.R#L48-L53) already. 

WDYT?

### DIFF
--- a/R/yoda_test_linter.R
+++ b/R/yoda_test_linter.R
@@ -45,8 +45,10 @@ yoda_test_linter <- function() {
     or (STR_CONST and not(OP-DOLLAR))
     or ((OP-PLUS or OP-MINUS) and count(expr[NUM_CONST]) = 2)
   "
+  expectations <- c("expect_equal", "expect_identical", "expect_setequal")
+
   xpath <- glue::glue("
-  //SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical' or text() = 'expect_setequal']
+  //SYMBOL_FUNCTION_CALL[ { xp_text_in_table(expectations) } ]
     /parent::expr
     /following-sibling::expr[1][ {const_condition} ]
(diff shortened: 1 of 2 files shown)


R/yoda_test_linter.R
```diff
@@ -45,8 +45,11 @@ yoda_test_linter <- function() {
     or (STR_CONST and not(OP-DOLLAR))
     or ((OP-PLUS or OP-MINUS) and count(expr[NUM_CONST]) = 2)
   "
+  expectations <- c("expect_equal", "expect_identical", "expect_setequal")
+  expectations_xpath <- xp_text_in_table(expectations)
+
   xpath <- glue::glue("
-  //SYMBOL_FUNCTION_CALL[text() = 'expect_equal' or text() = 'expect_identical' or text() = 'expect_setequal']
+  //SYMBOL_FUNCTION_CALL[ { expectations_xpath } ]
```
MichaelChirico, line 52: i don't agree this is clearer. `expectations` is too vague and makes us read elsewhere to understand.

Whereas the current version is very clear, since all of the relevant functions are right in front of us.

So I would revert

MichaelChirico:
I don't think it's always worth it. I don't have a clear rule for when to use `xp_text_in_table()` vs. not, but typically 2-3 entries I would typically just write out. There is an overhead to needing to remember which variable has which functions in it, vs. just having them plainly written.

Happy to review PRs where you think it's an improvement, but I would focus effort elsewhere.

PR #1727, opened 2022-10-17
call clear_settings

MichaelChirico:
Feel like I'm missing something here, shouldn't this have been caught earlier? 🤔


AshesITR:
This chunk trips all builds.
Can you debug?

```r
library(lintr)
linter_name <- "assignment_linter"

withr::with_tempfile("tmp", {
  writeLines("a = 1", tmp)

  # linter column is just 'get'
  print(as.data.frame(lint(tmp, linters = get(linter_name)())))

  this_linter <- get(linter_name)()
  attr(this_linter, "name") <- linter_name
  # linter column is 'assignment_linter'
  print(as.data.frame(lint(tmp, linters = this_linter)))

  # more concise alternative: use eval(call(.))
  print(as.data.frame(lint(tmp, linters = eval(call(linter_name)))))
})
```

MichaelChirico:
> This chunk trips all builds. Can you debug?
> 
> ```r
> library(lintr)
> linter_name <- "assignment_linter"
> 
> withr::with_tempfile("tmp", {
>   writeLines("a = 1", tmp)
> 
>   # linter column is just 'get'
>   print(as.data.frame(lint(tmp, linters = get(linter_name)())))
> 
>   this_linter <- get(linter_name)()
>   attr(this_linter, "name") <- linter_name
>   # linter column is 'assignment_linter'
>   print(as.data.frame(lint(tmp, linters = this_linter)))
> 
>   # more concise alternative: use eval(call(.))
>   print(as.data.frame(lint(tmp, linters = eval(call(linter_name)))))
> })
> ```

Finally got around to this :)

`clear_settings()` deletes `settings$encoding`, so `read_lines()` was trying `iconv(from = NULL)` hence the bug. Didn't look into why it only happens there.

R/lint.R
```diff
@@ -58,7 +58,7 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
 
   if (isTRUE(parse_settings)) {
```
MichaelChirico, line 59: I see, actually there's a non-trivial issue. We only run `read_settings()` here, but `get_lines()` above relies on some settings. Not immediately clear how to switch things around. If `encoding=` is the only issue, then we can just allow `encoding=NULL` to work in a few places. Reading more carefully...

MichaelChirico:
Subsumed by #2180

PR #1730, opened 2022-10-18
Add more linters to "lint-changed" workflow

IndrajeetPatil:


.lintr_new
```diff
@@ -25,6 +29,7 @@ linters: linters_with_defaults(
     strings_as_factors_linter(),
     unnecessary_lambda_linter(),
     unneeded_concatenation_linter(allow_single_expression = FALSE),
+    unused_import_linter(),
```
AshesITR, line 32: Why is that useful to us as a package? It only lints `library()` and friends.

AshesITR:
Also: Is there a reason to let `.lintr` and `.lintr_new` diverge? The split had the goal to eventually use all appropriate linters in both workflows, but to avoid flooding out too many lints that had crept into our code.

.lintr_new
IndrajeetPatil: It's useful for us to make sure that contributors don't include these lints in their contributions, and the CI/CD catches them even if the reviewers miss it for some reason.

IndrajeetPatil:
> Is there a reason to let .lintr and .lintr_new diverge?

No reason in particular. It's just that running the whole suite of linters included in `.lintr_new` for _all_ files (which is what will happens if we make `.lintr `similar to `.lintr_new`) will be a bit more expensive and time-consuming. OTOH, `.lintr_new` runs all its linters only for the few changed files, and so isn't as expensive. 

At some point, I'd like us to switch to using `linters_with_tags(NULL)`, i.e., include all linters. But this is going to require some groundwork to handle the existing lints.

.lintr_new
AshesITR: I expect `tests/testthat.R` to produce these lints because it loads some packages without actively using it. Can you confirm this?

PR #1735, opened 2022-10-18
Modify a few more examples

IndrajeetPatil:


R/with.R
```diff
@@ -11,13 +11,13 @@
...
-#' #    add print (unnamed, i.e. with no accompanying message)
-#' #    add return (as taken from all_undesirable_functions)
+#' #    remove `sapply` (using `NULL`)
+#' #    add `cat` (with an accompanying message),
+#' #    add `print` (unnamed, i.e. with no accompanying message)
+#' #    add `source` (as taken from `all_undesirable_functions`)
 #' my_undesirable_functions <- modify_defaults(
 #'   defaults = default_undesirable_functions,
-#'   sapply = NULL, "cat" = "No cat allowed", "print", all_undesirable_functions[["return"]]
+#'   sapply = NULL, "cat" = "No cat allowed", "print", all_undesirable_functions[["source"]]
```
IndrajeetPatil, line 20: Gets rid of a warning in the [docs](https://lintr.r-lib.org/dev/reference/modify_defaults.html):

> Trying to remove ‘return’, which is not in `defaults`.

AshesITR: Oh that warning looks like a bug, if it appears as you commented?!
At least the intention seemed to be to select return in addition to some other stuff.

R/lint.R
```diff
@@ -28,11 +28,10 @@
 #' @return An object of class `c("lints", "list")`, each element of which is a `"list"` object.
 #'
 #' @examples
-#' \dontrun{
-#'   lint("some/file-name.R") # linting a file
-#'   lint("a = 123\n")        # linting inline-code
-#'   lint(text = "a = 123")   # linting inline-code
-#' }
+#' f <- withr::local_tempfile(lines = "a=1", fileext = "R")
```
MichaelChirico, line 31: Hmm, I just remembered that `withr` is only a suggested dependency. So we shouldn't be running this example unless we check withr is installed. that makes things more complicated (this applies to other examples too)

IndrajeetPatil: You are right! Forgot about that.

I can just `tempfile()` then, instead of `withr::local_tempfile()`.

IndrajeetPatil: Turns out, we are using `{withr}` in many examples. So I have instead chosen the `#' @examplesIf` directive here and elsewhere.

R/with.R
IndrajeetPatil: > Oh that warning looks like a bug, if it appears as you commented?!

Yes: https://lintr.r-lib.org/dev/reference/modify_defaults.html#ref-examples

> At least the intention seemed to be to select return in addition to some other stuff.

I think the warning makes sense, no? `return()` is indeed not included in the default list: 

``` r
library(lintr)

all_undesirable_functions[["return"]]
#> NULL
```

<sup>Created on 2022-10-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The code should instead look like the following?

``` r
library(lintr)

my_undesirable_functions <- modify_defaults(
  defaults = default_undesirable_functions,
  sapply = NULL, "cat" = "No cat allowed", "print", "return"
)

names(my_undesirable_functions)
#>  [1] ".libPaths"     "attach"        "browser"       "cat"          
#>  [5] "debug"         "debugcall"     "debugonce"     "detach"       
#>  [9] "library"       "mapply"        "options"       "par"          
#> [13] "print"         "require"       "return"        "setwd"        
#> [17] "sink"          "source"        "Sys.setenv"    "Sys.setlocale"
#> [21] "trace"         "undebug"       "untrace"
```

<sup>Created on 2022-10-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/lint.R
AshesITR: Why not `cat()` and `lint(text=)` everywhere?

R/with.R
AshesITR: ... the code should use a function from `setdiff(names(all_undesirable_functions), names(default_undesirable_functions))`.

R/lint.R
MichaelChirico: this example is specifically about showing the different ways of using lint()

several other places are examples for get_source_expressions()

I only saw one (trailing_blank_lines_linter()) that would be better suited by the text= approach l

AshesITR: Oh, yes, sorry. How does `@examplesIf` render in the actual docs? As long as they show up, I'd say it's fine for those examples that really need a tempfile.

IndrajeetPatil: > How does @examplesIf render in the actual docs? 

We are already using it [here](https://github.com/r-lib/lintr/blob/a32f2d23d07d734358924e456b10f3d375fc70b3/R/absolute_path_linter.R#L11), and you can see how this is rendered in the docs [here](https://lintr.r-lib.org/dev/reference/absolute_path_linter.html).

AshesITR: And in the R 3.6 docs?

IndrajeetPatil: You mean, what they would like in pkgdown website, PDF manual, or in RStudio IDE help page?

I can at least answer the first one: they will just show up as unevaluated lines of code.

For example, this will render 

```r
#' @examplesIf getRversion() > '5.0'
#' sd(cars$speed)
#' sd(cars$dist)
```

to this

Screenshot 2022-10-19 at 20 50 24

AshesITR: Fair enough. I was thinking about the interactive R help pages displayed in RStudio.

R/with.R
IndrajeetPatil: Oh, I see. Should we create a new issue to track this?

AshesITR: You changed it to `"source"` here already, so this PR would close the issue, right?

IndrajeetPatil: Yeah, true that. I misunderstood your comment. Resolving this.

R/with.R
```diff
@@ -11,14 +11,17 @@
...
 #' )
+#'
+#' # enlist names of functions specified as undesirable
```
AshesITR, line 23: enlist -> list

AshesITR (approved):

PR #1738, opened 2022-10-19
Sync up GHA workflows with `r-lib/actions`

IndrajeetPatil:
cf. https://github.com/r-lib/actions/commit/cc4293a2b89c8a04e869c75fdb2701035bef502a

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -56,7 +56,7 @@ jobs:
           sudo locale-gen zh_CN
           echo "LC_ALL=zh_CN" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


AshesITR:
What's the benefit of checkout@v3 vs. v2?

IndrajeetPatil:
> What's the benefit of checkout@v3 vs. v2?

GitHub actions now recommend using Node 16 (over 12; cf. https://github.blog/changelog/2021-12-10-github-actions-github-hosted-runners-now-run-node-js-16-by-default/), and checkout@v3 supports Node 16 (cf. https://github.com/actions/checkout/releases/tag/v3.0.0).

And, in general, we should try to keep in sync with the upstream (`r-lib/actions`), which has also made these changes (both to checkout@v3 and cache@v3).

AshesITR (approved):

PR #1739, opened 2022-10-19
Add "no suggests" R CMD check workflow

IndrajeetPatil:
This can help us catch instances of soft dependency usage in examples or in tests without the necessary skip.

### DIFF
--- a/.github/workflows/check-no-suggests.yaml
+++ b/.github/workflows/check-no-suggests.yaml
@@ -1,0 +1,66 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+#
+# NOTE: This workflow only directly installs "hard" dependencies, i.e. Depends,
+# Imports, and LinkingTo dependencies. Notably, Suggests dependencies are never
+# installed, with the exception of testthat, knitr, and rmarkdown. The cache is
+# never used to avoid accidentally restoring a cache containing a suggested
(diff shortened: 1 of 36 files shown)


.github/workflows/check-no-suggests.yaml
```diff
@@ -0,0 +1,66 @@
...
+# NOTE: This workflow only directly installs "hard" dependencies, i.e. Depends,
+# Imports, and LinkingTo dependencies. Notably, Suggests dependencies are never
+# installed, with the exception of testthat, knitr, and rmarkdown. The cache is
+# never used to avoid accidentally restoring a cache containing a suggested
+# dependency.
+#
+# Customization for `{lintr}`
+#
+# The following soft dependencies are not optional for testing:
+# `{patrick}`, `{rex}`, `{withr}`
```
AshesITR, line 13: Why is rex a soft dep? We use it in our package code, no?

MichaelChirico: shall we delete rex from the loadNamespace calls? I believe it was added by mistake

IndrajeetPatil: Yeah, I think it was introduced accidentally. Made a separate PR to fix that.

IndrajeetPatil:
We will need to keep this PR updated and pending until all relevant issues to get the new check to succeed are fixed in separate PRs.

IndrajeetPatil:
The new check is green now.

AshesITR (approved):
Nice work, thanks!

PR #1742, opened 2022-10-20
Embed gifs in vignettes

IndrajeetPatil:
No need to use still images.

If you knit this locally, you can see that the gifs work just fine.

If you want to see an example of a package on CRAN that includes gif in this way, see [this](https://cran.r-project.org/web/packages/datawizard/readme/README.html).


.github/workflows/R-CMD-check.yaml
```diff
@@ -74,5 +74,5 @@ jobs:
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
-        with:
-          error-on: '"note"'
+        # with:
+        #   error-on: '"note"'
```
IndrajeetPatil, line 78: 😬


        happy to test this out, we may revert it slows us down a lot

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/1621#pullrequestreview-1132006330_

AshesITR:
I remember we did exactly that (static images) to reduce the size at basically no loss of information.

AshesITR:
#1325

IndrajeetPatil:
> at basically no loss of information

Not sure what you mean. Vignettes don't have the gifs on CRAN, so there is a loss of information, no?

With the current approach, it's a win-win. Both for CRAN vignette and pkgdown website.

AshesITR:
Yes, but the gif's don't contain any important information, so I'd rather not violate installed size guidelines because of them. If you manage to get them to < 1MB in total, it'd be cool though 😊

IndrajeetPatil:
Hmm, in that case, should we just get rid of the gifs and only include stills? 

If there is no loss of information, what's the added value of including the gifs?

AshesITR:
It's prettier on lintr.r-lib.org where there is no file size restriction to 1MB. That's what the current code does, conditionally show the animated versions online.

IndrajeetPatil:
Yes, I get that, but it comes at the cost of an additional soft dependency (note that this is why "noSuggests" check is currently [failing](https://github.com/r-lib/lintr/pull/1739)) and added complexity of maintaining different versions of the docs.

IndrajeetPatil:
We don't have gifs for RStudio, VS Code, and Atom, either; just the screenshots. So we might as well be consistent and do the same for other IDEs.

AshesITR:
> Error:   Failed to download Pandoc 2.14.2: Error: connect ETIMEDOUT 185.199.109.133:443

What's the problem root cause of the failure?

IndrajeetPatil:
> Error: Failed to download Pandoc 2.14.2: Error: connect ETIMEDOUT 185.199.109.133:443

@AshesITR Where are you seeing these errors? Based on the error message, looks like a transitory issue?

AshesITR:
In the PR you linked that's the only build failure.

IndrajeetPatil:
Oh, that was a spurious build failure; I've retriggered it. The actual issue you can see in the first commit [build](https://github.com/r-lib/lintr/actions/runs/3281713362/jobs/5404112752) as well:

```r
Error: --- re-building ‘continuous-integration.Rmd’ using rmarkdown
--- finished re-building ‘continuous-integration.Rmd’
--- re-building ‘creating_linters.Rmd’ using rmarkdown
--- finished re-building ‘creating_linters.Rmd’
--- re-building ‘editors.Rmd’ using rmarkdown
Quitting from lines 26-32 (editors.Rmd) 
```

AshesITR:
Seems like we should copy `pkgdown::in_pkgdown()` then?

IndrajeetPatil:
I still think it'd be better to consistently have gifts or stills for all editors, not gifs for some, while still for others. But punting that for now.

Closing in favour of https://github.com/r-lib/lintr/pull/1747.

PR #1743, opened 2022-10-20
Move utilities for package and project root to their own file

IndrajeetPatil:
Since `lint.R` is already quite bloated.


DESCRIPTION
```diff
@@ -143,6 +143,7 @@ Collate:
     'redundant_ifelse_linter.R'
     'regex_subset_linter.R'
     'routine_registration_linter.R'
+    'rprojroot_utils.R'
```
MichaelChirico, line 146: not a big fan of the name (too focused on .Rproj)... how about 'package_metadata_utils.R'?

IndrajeetPatil: Done!

I had named it so originally because most of these helpers functions were taken from `{rprojroot}` package, but, yeah, we don't necessarily need that to be reflected in the file name.

AshesITR:
The functions are related to `config.R`. Is there room in that file? Otherwise I'd call it `config_utils.R`. `package_metadata_utils.R` sound like it has something to do with finding linter tags in other packages...

IndrajeetPatil:
@AshesITR We currently don't have `config.R` file, so I've created `config_utils.R` instead.

AshesITR:
Ah I meant `settings.R`

IndrajeetPatil:
Gotcha. `settings.R` is already quite busy, so moved this to `settings_utils.R` instead.

AshesITR:
WDYT about moving some utilities from `settings.R` as well (`find_config()` to `is_directory()`)?

IndrajeetPatil:
@AshesITR Moved a couple more utilities.

AshesITR (approved):
Thanks!

AshesITR (approved):

PR #1747, opened 2022-10-22
Port over `pkgdown::in_pkgdown()`

IndrajeetPatil:
For #1739

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,6 @@ Suggests:
     jsonlite,
     mockery,
     patrick,
-    pkgdown,
     rlang,
     rmarkdown,
     rstudioapi (>= 0.2),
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -41,7 +41,6 @@ Suggests:
     jsonlite,
     mockery,
     patrick,
-    pkgdown,
```
AshesITR, line 44: Is putting pkgdown in Suggests not the usual way of signaling we use it for development?

IndrajeetPatil: Nope.

IndrajeetPatil: You can sample any of the r-lib, tidyverse, etc. organizations packages, all of which use _pkgdown_, but none of them have _pkgdown_ in `Suggests`. As far as CRAN is concerned, it's irrelevant.

AshesITR: Okay, thanks for clarifying.

AshesITR (approved):

PR #1748, opened 2022-10-23
Use `nolint` directives for `undesirable_operator_linter()` lints

IndrajeetPatil:
Waiting for #1751


AshesITR:
NB you can use `# nolint: undesirable_operator.` as well.

R/lint.R
```diff
@@ -737,7 +733,7 @@ highlight_string <- function(message, column_number = NULL, ranges = NULL) {
   line <- fill_with(" ", maximum)
 
   lapply(ranges, function(range) {
-    substr(line, range[1L], range[2L]) <<-
+    substr(line, range[1L], range[2L]) <<- # nolint: undesirable_operator_linter.
       fill_with("~", range[2L] - range[1L] + 1L)
```
IndrajeetPatil, line 737: Is the indentation lint here a false positive? AFAICT, the indentation should be 6 spaces here.

> Indentation should be 4 spaces but is 6 spaces.

Btw, if I try to inline this, I create `line_length_linter()` lint.

IndrajeetPatil: cc @AshesITR

AshesITR: It is, seems like the comment throws off the XPath for indentation changes.

IndrajeetPatil: Thanks for confirming. I will wait for the related issue to be resolved. Then we can revisit this PR.

AshesITR (approved):

PR #1756, opened 2022-10-27
Add nolint for `ids_with_token()` definition

IndrajeetPatil:
Triggers `function_argument_linter()`:

> Arguments without defaults should come before arguments with defaults.

### DIFF
--- a/R/ids_with_token.R
+++ b/R/ids_with_token.R
@@ -27,7 +27,7 @@
 #' entry of the list of source expressions. Indices correspond to the
 #' *rows* where `fun` evaluates to `TRUE` for the `value` in the *token* column.
 #' @export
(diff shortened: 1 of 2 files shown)


R/ids_with_token.R
```diff
@@ -27,7 +27,7 @@
 #' entry of the list of source expressions. Indices correspond to the
 #' *rows* where `fun` evaluates to `TRUE` for the `value` in the *token* column.
 #' @export
-ids_with_token <- function(source_expression, value, fun = `==`, source_file) {
+ids_with_token <- function(source_expression, value, fun = `==`, source_file) { # nolint: function_argument_linter.
```
MichaelChirico, line 30: Instead, set the default to `source_file = NULL`

IndrajeetPatil: Thanks, done.

AshesITR (approved):

PR #1757, opened 2022-10-28
Check for .lintr file in extra subdirectory

### LINKED ISSUE #1746: Modify `.lintr` file search algorithm
We have a multi language team and I am trying to roll out a common linter for the team instead of a different structure for each language. I've been testing [super-linter](https://github.com/github/super-linter) with R and I am coming into a few issues for R.

One is that super-linter expects configurations files all live in the folder `.github/linters` but it seems from [here](https://lintr.r-lib.org/articles/lintr.html#other-configuration-options) that lintr doesn't check that path meaning that running functions in RStudio don't detect the above file.

I currently have two `.lintr` files, one in the project root and one in `.github/linters` that are just a copy of one another to get around this issue.

Wondering if `.github/linters` could be added to the search list?

tonyk7440:
Fixes #1746 

Added another condition to check for the `.lintr` file in the `.github/linters` subdirectory to improve compatibility with Super-Linter.

Let me know if there's anything I missed, happy to update


AshesITR:
If we consider this, we might also want #460

tonyk7440:
I can see there is a lint error raised for the cyclomatic complexity linter. I'm not sure what would be the best way to resolve. I don't see an obvious way to make the code simpler. I could increase the threshold to be 16 but not sure if that is an advised route

AshesITR:
While the current search order is non-breaking, it seems very counter-intuitive to give a "global" config at `~/.lintr` priority over a local config at `.github/linters`.

AshesITR:
Regarding the cyclomatic complexity: You could refactor the existence check into a `for` loop that returns as soon as one of the candidate files exists.

tonyk7440:
@AshesITR Thanks for the pointers, I have updated the code, it looks to be passing that and other tests now. Let me know if you suggest other changes

R/settings_utils.R
```diff
@@ -68,30 +68,24 @@ find_config <- function(filename) {
...
+    find_config2(path),
+    # Next check for a file in the user directory
+    # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
+    file.path(Sys.getenv("HOME", unset = "~"), linter_file)
+  )
+
+  # Search through locations, return first valid result
+  for (loc in file_locations) {
+    if (isTRUE(file.exists(eval(loc)))) {
+      return(eval(loc))
```
AshesITR, line 87: why are you using `eval(loc)`? `loc` should already be a character, so no need to `eval()` it.

R/settings_utils.R
```diff
@@ -68,30 +68,24 @@ find_config <- function(filename) {
...
-  if (isTRUE(file.exists(linter_config))) {
-    return(linter_config)
-  }
-
-  ## next check for a file in the .github/linters directory
-  linter_config <- file.path(path, ".github/linters", linter_file)
-  if (isTRUE(file.exists(linter_config))) {
-    return(linter_config)
+  # List possible .lintr file locations
+  file_locations <- list(
```
AshesITR, line 72: use `c()` instead of `list()` so that `file_locations` is a character vector, not a list.

AshesITR (commented):
@tonyk7440 thanks for your work so far!

@jimhester, @MichaelChirico, @IndrajeetPatil: WDYT about expanding the config search paths?

R/settings_utils.R
tonyk7440: Updated, thanks!

jimhester:
Sounds good to me

NEWS.md
```diff
@@ -70,6 +70,9 @@
 
 * `literal_coercion_linter()` reports a replacement in the lint message, e.g. code like `as.integer(1)` will
   suggest using `1L` instead, and code like `as.numeric(NA)` will suggest using `NA_real_` instead (#1439, @MichaelChirico)
+  
+* Added check for `.lintr` configuration file in the `.github/linters` project sub folder. This is to improve
```
AshesITR, line 74: This is a change in behavior and thus potentially a breaking change: previously if a repo contained `.github/linters/.lintr` and no other lintr configs and a local `~/.lintr` is present, lintr 3.0.2 would pull the one from $HOME instead of the one in `.github`. Please move this to a new "breaking changes" section.
Also not sure about the wording @MichaelChirico any thoughts from you as a native?

vignettes/lintr.Rmd
```diff
@@ -95,6 +95,7 @@ The first one found, if any, will be used.
 2.  A linter file (that is a file named like `lintr.linter_file`) in the currently searched directory, i.e. the directory of the file passed to `lint()`. When run from `lint_package()`, this directory can differ for each linted file.
 3.  A linter file in a parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time.
 4.  A linter file in the user's `HOME` directory.
+5.  A linter file in the `.github/linters` child directory of the currently-search directory.
```
AshesITR, line 98: This priority order is incorrect now.

AshesITR (commented):
Nice work so far, thanks!
Now that we are sure we'll accept this change, please add a test package where the `.github/linters/.lintr` config takes effect.
Let me know if you need hints on how to do that.

vignettes/lintr.Rmd
tonyk7440: Corrected priority order, marking as resolved

AshesITR:
Thanks for trying to cook up a test.
Due to the R CMD check notes, you'll likely have to programmatically `dir.create()` the `.github` path.
Also note we usually use a different `lintr.lintr_file` option to circumvent said R CMD check note regarding the `.lintr` filename.

Finally, to test that the config is pulled from `.github/linters`, you should make sure the test package does not contain a linter config at its root.
That way, if the config is not found, no exclusions would take effect and the test would fail when comparing to the lints generated with the linter config file present at the package root.

Right now, IINM, the `lint_package()` calls in your test always find `.lintr` at the package root.

NEWS.md
tonyk7440: I've updated the file from the pointer above, also changed the wording, not sure if I should mark this as resolved

tonyk7440:
Thanks @AshesITR!

I've updated the code, seems like the tests are passing now, let me know if I should make further changes.

R/settings_utils.R
```diff
@@ -68,24 +68,24 @@ find_config <- function(filename) {
...
-  if (isTRUE(file.exists(linter_config))) {
-    return(linter_config)
+  # List possible .lintr file locations
+  file_locations <- c(
+    # Check for a file in the current directory
+    file.path(path, linter_file),
+    # Next check for a file in the .github/linters directory
+    file.path(path, ".github/linters", linter_file),
+    # Next check for a file in higher directories
+    find_config2(path),
```
AshesITR, line 78: `find_config2()` should also look for `.github/linters` (i.e. as a sibling directory to `getwd()`)
This will ensure e.g. code running in `tests/testthat` from within a package will use a package-level `.github/linters/.lintr`

tests/testthat/test-lint_package.R
```diff
@@ -132,3 +132,43 @@ test_that("lint_package returns early if no package is found", {
...
+    # In `github/linters`add a `.lintr` file that excludes the whole of `abc.R`
+    # and the first line of `jkl.R` (and remove it on finishing this test)
+    local_config(file.path(pkg_path, ".github/linters"), "exclusions: list('R/abc.R', 'R/jkl.R' = 1)")
+
+    lints_using_github_lintr <- withr::with_dir(
+      pkg_path,
+      lint_package(".", linters = list(assignment_linter()))
+    )
+
+    # Now add the same file to the package root
```
AshesITR, line 157: Delete the `.github/linters/.lintr` file here for more safety

AshesITR (commented):
Excellent work so far, thank you!
Just a few more adjustments.

Regarding the `NEWS` wording, I'd like to have input from @MichaelChirico.

tests/testthat/test-lint_package.R
tonyk7440: Code updated, marking as resolved

R/settings_utils.R
tonyk7440: Hi @AshesITR ,

Sorry I'm not sure what you mean here. is it to combine lines 76 & 78 into one using `find_config2(file.path(path, ".github", "linters"))`
Or something different as you think there is a bug?

AshesITR: I'm thinking that the code of `find_config2()`, which currently looks for a `.lintr` file in a parent directory, should be modified to also look for a `.github/linters/.lintr` file. This would make it so `lint("pkg/tests/testthat/test-my_stuff.R")` could find a config at `pkg/.github/linters/.lintr` as well.

AshesITR:
@tonyk7440 do you need assistance with refactoring `find_config2()`?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # lintr (development version)
 
+## Breaking changes
+
+* Updated the location priority order that `lintr` checks to find a `.lintr` settings file. It now checks in the `.github/linters` project sub folder after the currently searched directory. This is to improve
```
MichaelChirico, line 5: I would start with ".lintr configs can now be kept in directory .github/linters' for better compatibility with Super-Linter. Note that this may be a breaking change if you already have a config in both such a directory and in your R project's root since the former will now be discovered first where it was ignored before." and also adding something like "Please see vignette(...) for details on how configs are discovered."

OTOH @AshesITR it seems pretty unlikely anyone's affected by this (why have a .lintr file in that particular place if it's not actually doing anything?), am I missing something?

AshesITR: It's more like if you want super-linter to be less critical than your personal `~/.lintr`, e.g. for performance reasons.
I agree this is very unlikely to occur. Technically, it's still a breaking change so might be best to alert users.

R/settings_utils.R
```diff
@@ -69,24 +69,24 @@ find_config <- function(filename) {
...
-    return(linter_config)
-  }
-
-  ## next check for a file in the user directory
-  # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
-  home_dir <- Sys.getenv("HOME", unset = "~")
-  linter_config <- file.path(home_dir, linter_file)
-  if (isTRUE(file.exists(linter_config))) {
-    return(linter_config)
+  # List possible .lintr file locations
```
MichaelChirico, line 72: I think we can be a bit more concise w the comments on the list and move the "meat" to the top where it's a little less distracting.

the first comment can add "NB: this vector specifies a priority order for where to find the configs, i.e. the first location where a config exists is chosen and configs which may exist in subsequent directories are ignored"

then in the item comments we can e.g. change "check for a file in the current directory" to just "current directory" etc

vignettes/lintr.Rmd
```diff
@@ -93,8 +93,9 @@ The first one found, if any, will be used.
 
 1.  If `options(lintr.linter_file)` is an absolute path, this file will be used. The default for this option is `".lintr"`.
 2.  A linter file (that is a file named like `lintr.linter_file`) in the currently searched directory, i.e. the directory of the file passed to `lint()`. When run from `lint_package()`, this directory can differ for each linted file.
-3.  A linter file in a parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time.
-4.  A linter file in the user's `HOME` directory.
+3.  A linter file in the `.github/linters` child directory of the currently search directory.
```
MichaelChirico, line 96: "currently search directory"? is that just "current directory"?

MichaelChirico (commented):
sorry for the delay, I didn't realize I was pinged (email notifications don't call that out separately from general issue updates)

NEWS.md
MichaelChirico: OK, makes sense, in that case we should mention how to accomplish moving your preferred config to the front on an ad hoc basis (e.g. set that config as env). unless we already do so in the vignette, in which case we're covered by "see vignette for details"

AshesITR: That's already documented in the vignette, but as a workaround I'd recommend copying `~/.lintr` to `./.lintr` instead.

AshesITR:
No worries, thanks for your feedback and good work so far.

klmr:
@tonyk7440 Hi, just wondering: are there any updates on this PR?

MichaelChirico:
Gentle ping @tonyk7440 as we're moving towards release and it would be ideal to include this breaking change alongside a few others at the bump to 3.1.0.

If you're currently short on time, please ping & we can try and crowdsource getting this across the finish line.

Thanks again!

R/settings_utils.R
tonyk7440: Thanks @MichaelChirico , updated the code per the comment above

NEWS.md
tonyk7440: Updated file as per comments

tonyk7440:
Hi @AshesITR @MichaelChirico @klmr , sorry for the delay in responding.
I've resolved most of the comments above. There is one outstanding ask from @AshesITR, due to my track record above I'll ask to crowdsource this change. Or else the above changes could be merged for the upcoming release and postpone that change to a later release?
Thanks!

R/settings_utils.R
```diff
@@ -61,24 +61,26 @@ find_config <- function(filename) {
...
+    # Higher directories
+    find_config2(path),
+    # User directory
+    # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
+    file.path(Sys.getenv("HOME", unset = "~"), linter_file)
+  )
+
+  # Search through locations, return first valid result
+  for (loc in file_locations) {
+    if (isTRUE(file.exists(loc))) {
```
klmr, line 81: Why the `isTRUE()` call? As far as I can tell `file.exists()` should always return exactly one logical value here so this code doesn’t need to guard against non-logical values or vectors of length > 1.

klmr: Ugh, this comment (along with others that seem to be gone?!) was written months ago but was somehow part of a pending review without my noticing.

R/settings_utils.R
klmr: @AshesITR Were you thinking of something along these lines?

```diff
--- a/R/settings_utils.R
+++ b/R/settings_utils.R
@@ -90,7 +90,10 @@ find_config2 <- function(path) {
   config <- basename(getOption("lintr.linter_file"))
   path <- normalizePath(path, mustWork = FALSE)

-  while (!has_config(path, config)) {
+  while (
+    !has_config(path, config)
+    && !has_config(file.path(path, ".github", "linters"), config)
+  ) {
     path <- dirname(path)
     if (is_root(path)) {
       return(character())

```

Making this change would mean that we could remove the corresponding line from the `file_locations` definition in `find_config` (we can also remove the “current directory” logic, which is also covered by `find_config2` if I am not mistaken!).

This avoids having to hard-code the path `.github/linters` in two separate locations. — In fact, it would mean that `find_config` would not need to be changed at all (although that change removes redundancy and is probably a good idea regardless).

AshesITR: Yes, something like that looks about right.

klmr: Alright, I have a candidate implementation which fixes the unit test and implements this.

The question is: what is the easiest/cleanest way of adding it to this PR? Or should I create a new PR? Either way it is based off @tonyk7440’s initial changes, so these would be incorporated when merging a new PR instead of this one.

AshesITR: @tonyk7440 can you give edit permissions to @klmr on your fork (or at least on this branch)?

tonyk7440: Hi @klmr, I think you should have access to edit the fork now, let me know if you don't

klmr: I do, thanks! I will  push the changes as soon as I find time outside of work hours, which should happen in the next 48 hours.

klmr:
Alright, I’ve pushed my suggested changes. Happy for a review.

vignettes/lintr.Rmd
```diff
@@ -93,8 +93,9 @@ The first one found, if any, will be used.
 
 1.  If `options(lintr.linter_file)` is an absolute path, this file will be used. The default for this option is `".lintr"`.
 2.  A linter file (that is a file named like `lintr.linter_file`) in the currently searched directory, i.e. the directory of the file passed to `lint()`. When run from `lint_package()`, this directory can differ for each linted file.
-3.  A linter file in a parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time.
-4.  A linter file in the user's `HOME` directory.
+3.  A linter file in the `.github/linters` child directory of the currently-searched directory.
+4.  A linter file in a parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time.
```
AshesITR, line 97: This could be made more clear, stating that 2. and 3. are searched from the parent.

NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 ## Deprecations & Breaking Changes
 
+* .lintr configs can now be kept in the directory .github/linters for better compatibility with Super-Linter. Note that this may be a breaking change if you already have a config in both such a directory and in your R project's root since the former will now be discovered first where it was ignored before. Please see vignette("lintr") for details on how configs are discovered. (#1746, @tonyk7440)
```
AshesITR, line 5: IINM, the only breaking constellation is if `.github/linter/.lintr` is in a subdirectory, right?

i.e.
```
.lintr
tests/.github/linters/.lintr
```

will find a different config when run from `tests/`. Is that correct?

klmr: You’re absolutely right! Will amend.

vignettes/lintr.Rmd
AshesITR: Aren't we checking 2. and 3. for the current directory and only if none match, the parent is searched recursively?

This description suggests we first search for `.lintr` in any parent and only look for `.github/linters/.lintr` if there is no `.lintr` anywhere in the path.

AshesITR: @klmr this is the only blocker I see. Good to merge once the docs match the behavior.

klmr: Updated. Please take a look. I think we won’t get around making nested bullet points, since that reflects how the algorithm works. Is this clear enough?

vignettes/lintr.Rmd
```diff
@@ -89,11 +89,13 @@ exclusions: list(
...
 
-1.  If `options(lintr.linter_file)` is an absolute path, this file will be used. The default for this option is `".lintr"`.
-2.  A linter file (that is a file named like `lintr.linter_file`) in the currently searched directory, i.e. the directory of the file passed to `lint()`. When run from `lint_package()`, this directory can differ for each linted file.
-3.  A linter file in a parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time.
+1.  If `options("lintr.linter_file")` is an absolute path, this file will be used. The default for this option is `".lintr"`.
+2. A project-local linter file; that is, either
+    1. a linter file (that is, a file named like `lintr.linter_file`) in the currently-searched directory, i.e. the directory of the file passed to `lint()`; or
+    2. a linter file in the `.github/linters` child directory of the currently-searched directory.
+3. A project-local linter file in the closest parent directory of the currently-searched directory, starting from the deepest path, moving upwards one level at a time. When run from `lint_package()`, this directory can differ for each linted file.
```
AshesITR, line 98: nit: Whitespace is inconsistent across top-level bullets.

AshesITR (commented):
LGTM. Just a minor nitpick left.

vignettes/lintr.Rmd
klmr: Oops. Fixed.

AshesITR (approved):
Thanks a lot, you two!

MichaelChirico:
@AshesITR it looked like this is ready to merge, LMK if that was in error.

PR #1761, opened 2022-11-04
Address `undesirable_function_linter()` lints

IndrajeetPatil:


R/utils.R
```diff
@@ -145,12 +145,16 @@ quote_wrap <- function(x, q) paste0(q, x, q)
 # interface to work like options() or setwd() -- returns the old value for convenience
 set_lang <- function(new_lang) {
   old_lang <- Sys.getenv("LANGUAGE", unset = NA)
-  Sys.setenv(LANGUAGE = new_lang)
+  Sys.setenv(LANGUAGE = new_lang) # nolint: undesirable_function_linter.
   old_lang
 }
 # handle the logic of either unsetting if it was previously unset, or resetting
 reset_lang <- function(old_lang) {
-  if (is.na(old_lang)) Sys.unsetenv("LANGUAGE") else Sys.setenv(LANGUAGE = old_lang)
```
IndrajeetPatil, line 153: Can't add comment inline because that will trigger `line_length()` linter.

AshesITR: NB wondering why is `Sys.unsetenv()` not undesirable.
Maybe a follow up?

IndrajeetPatil: SGTM.

IndrajeetPatil: Created issue: https://github.com/r-lib/lintr/issues/1762

AshesITR:
Just so I understand, this is to enable `undesirable_function_linter()` in `.lintr_new`?

If so, why not add it immediately?
IINM, it's currently disabled.

IndrajeetPatil:
@AshesITR I was going to do this later in one go, but we can do this PR-by-PR.

I have updated `.lintr_new`.

AshesITR:
What we did in the last cleanup was

1. add all linters to `.lintr_new` to ensure no new "dirty" code is produced and then
3. PR-by-PR fix one type of lints and add the fixed linter to `.lintr`.

IndrajeetPatil:
Yeah, that's what I've also been doing, but wasn't sure if we want to continue to let `.lintr_new` diverge from `.lintr`.

At any rate, I have updated `.lintr_new`.

AshesITR:
Let's file an issue for this cleanup and discuss first what `.lintr_new`, and eventually `.lintr`, should look like in the end.
Once we've settled on that, I think we can implement the necessary PRs with more confidence.

For this discussion, the current tally of `lint_package(linters = linters_with_tags(NULL))` is probably a good starting point.

IndrajeetPatil:
@AshesITR We need to clean these lints regardless of what we decide here. So instead of keeping this PR blocked, I'd suggest that I revert the change to `.lintr_new`, and we can address how it should be updated in a different PR. 

The current PR will then be solely about address `undesirable_function_linter()` lints. WDYT?

AshesITR:
SGTM. Should also revert the nolint comments then.

AshesITR (changes requested):
Waiting for reverts to `.lintr_new` and the `# nolint: undesirable_function_linter.` comments.
The rest is ready to approve.

AshesITR (approved):

PR #1765, opened 2022-11-09
use normalizePath() in find_package()

### LINKED ISSUE #1759: lint_package() and find_package(".") does not work anymore if depth >= 2
In my package I have a code style test so it calls `lint_package()` from the `tests/testthat` folder.
The new logic in find_package fails at finding the root of the package ( I guess from this commit https://github.com/r-lib/lintr/commit/3009505944493acdaeeec6ff819acede1f39250a )

so If I call lint_package from a deeper level than 2 it fails at finding the package actually...
```
test_that("check that package has code style", {
        skip_on_covr()
        lints <- lintr::lint_package()
        ....
        }
```
The above code warns/fails with:

```
Warning message:
In lintr::lint_package() :
  Didn't find any R package searching upwards from '.'.
  ```

### PR DESCRIPTION
**Pull request by AshesITR**
fixes #1759

Also fixed a (newly generated) warning

```
Warning (test-settings.R:108): it has a smart default for encodings
path[1]="dummy_packages/cp1252/R/metropolis-hastings-rho.R": No such file or directory
Backtrace:
 1. testthat::expect_identical(...)
      at test-settings.R:108:2
 6. lintr:::find_package(pkg_file)
 7. base::normalizePath(path)
      at lintr/R/settings_utils.R:7:2
```

Previously, existence of the path provided to `lint_package()` was never tested so it slipped by.

### CODE REVIEW

File: tests/testthat/dummy_packages/assignmentLinter/DESCRIPTION [hunk 1]
```diff
@@ -7,3 +7,6 @@ Author: Who wrote it
 Maintainer: Who to complain to <yourfault@somewhere.net>
 Description: More about what it does (maybe more than one line)
 License: What license is it under?
+Suggests: 
+    testthat (>= 3.0.0)
+Config/testthat/edition: 3
```
AshesITR, line 12: I used `usethis::use_testthat()` inside the `assignmentLinter` package to create a realistic case for doubly nested files.

tests/testthat/test-lint_package.R
```diff
@@ -91,7 +91,11 @@ test_that(
     )
     lints_from_a_subdir <- withr::with_dir(
       file.path(pkg_path, "R"),
-      lint_package("..", linters = list(assignment_linter()))
+      lint_package(".", linters = list(assignment_linter()))
```
AshesITR, line 94: This now tests two things: That exclusions are relative to the `.lintr` config and that `lint_package()` finds the package from within as intended.

AshesITR:
@IndrajeetPatil there are currently 135 more undesirable_operator lints in this branch (ignoring the nolinted ones).
All of them lint a `lintr:::` expression in our test suite.
Maybe we should consider removing that one?

Note these are the only lints generated by `withr::with_options(list(lintr.linter_file = ".lintr_new"), lint_package())` at the moment.

IndrajeetPatil:
> Maybe we should consider removing that one?

Definitely. We can add it back once https://github.com/r-lib/lintr/issues/1692 is resolved. 

Feel free to make the change in this PR.

AshesITR:
Okay. Let's do that in a follow-up to keep this PR focused.

tests/testthat/dummy_packages/assignmentLinter/tests/testthat.R
```diff
@@ -0,0 +1,12 @@
+# This file is part of the standard setup for testthat.
+# It is recommended that you do not modify it.
+#
+# Where should you do additional test configuration?
+# Learn more about the roles of various files in:
+# * https://r-pkgs.org/tests.html
+# * https://testthat.r-lib.org/reference/test_package.html#special-files
+
```
IndrajeetPatil, line 8: Nit: We don't need to include this. It is unnecessarily distracting.

IndrajeetPatil (approved):
LGTM, thanks!

Just one nitpick, but I am fine either way it is resolved.

tests/testthat/dummy_packages/assignmentLinter/tests/testthat.R
AshesITR: removed. We can always recreate this if we want to.

IndrajeetPatil (approved):

tests/testthat/dummy_packages/assignmentLinter/tests/testthat.R
IndrajeetPatil: Sorry for the misunderstanding. 

I meant that we don't need that comment in the `testthat.R` file. It would still be good to have the `testthat.R` file without those comments. Otherwise, it'd be confusing to have tests without nothing executing those tests.

IndrajeetPatil (approved):

PR #1780, opened 2022-11-29
Include only names of linters in example outputs

IndrajeetPatil:
Including definitions of functions makes the output on the website quite overwhelming.

cf. https://lintr.r-lib.org/dev/reference/linters_with_tags.html#ref-examples

### DIFF
--- a/R/with.R
+++ b/R/with.R
@@ -80,11 +80,11 @@ modify_defaults <- function(defaults, ...) {
 #' # `linters_with_defaults()` and `linters_with_tags("default")` are the same:
 #' all.equal(linters_with_defaults(), linters_with_tags("default"))
 #'
-#' # Get names of all linters useful for package development
+#' # Get all linters useful for package development
 #' linters <- linters_with_tags(tags = "package_development")
 #' names(linters)
 #'
-#' # Get names of all linters provided by lintr
+#' # Get all linters provided by lintr
 #' linters <- linters_with_tags(tags = NULL)
 #' names(linters)
 #'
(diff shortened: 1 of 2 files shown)


R/with.R
```diff
@@ -80,11 +80,11 @@ modify_defaults <- function(defaults, ...) {
 #' # `linters_with_defaults()` and `linters_with_tags("default")` are the same:
 #' all.equal(linters_with_defaults(), linters_with_tags("default"))
 #'
-#' # Get all linters useful for package development
-#' linters_with_tags(tags = "package_development")
+#' # Get names of all linters useful for package development
+#' names(linters_with_tags(tags = "package_development"))
```
AshesITR, line 84: I think it would be better to store the actual object of interest in a variable and then (maybe) show names(...)

R/with.R
```diff
@@ -80,11 +80,13 @@ modify_defaults <- function(defaults, ...) {
 #' # `linters_with_defaults()` and `linters_with_tags("default")` are the same:
 #' all.equal(linters_with_defaults(), linters_with_tags("default"))
 #'
-#' # Get all linters useful for package development
-#' linters_with_tags(tags = "package_development")
+#' # Get names of all linters useful for package development
+#' linters <- linters_with_tags(tags = "package_development")
+#' names(linters)
 #'
-#' # Get all linters provided by lintr
```
AshesITR, line 86: This can be reverted then, right?

IndrajeetPatil: Good catch, thanks! Reverted.

AshesITR (approved):
LGTM, thanks!

PR #1783, opened 2022-11-30
WIP: Implement `unnecessary_nested_if_linter()`

### LINKED ISSUE #1778: Introduce `unnecessary_nested_if_linter()`
In the wild ([e.g.](https://github.com/gadenbuie/starwarsdb/blob/b6339df02e2a2394120ddf36d74b746d1f141f33/R/connect.R#L63-L67)), I have seen this pattern multiple times: there are nested `if()` statements, when a single conditional with the right conditional expression would do.

For example, this

``` r
x <- TRUE
y <- TRUE

# actual
if (x) {
  if (y) {
    print(1.1)
  }
}
#> [1] 1.1
```

can be rewritten as:

```r
if (x && y) print(1.1)
#> [1] 1.1
```

<sup>Created on 2022-11-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The benefit is, of course, reduction in complexity and improved readability.

Currently, this doesn't produce a lint:

``` r
library(lintr)

code <- "if (x) {
  if (y) {
    print(1.1)
  }
}"

lint(
  text = code,
  linters = linters_with_tags(NULL)
)
```

<sup>Created on 2022-11-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

---

P.S. Needless to say, this is relevant only if there isn't some additional code in the outer scope. E.g.

```r
if (x) {
  y <- x + y
  if (y) {
    ...
  }
}
```

IndrajeetPatil:
Closes #1778 

``` r
library(lintr)

# will produce lints
writeLines("if (x) { \n  if (y) { \n   return(1L) \n  } \n}")
#> if (x) { 
#>   if (y) { 
#>    return(1L) 
#>   } 
#> }
lint(
  text = "if (x) { \n  if (y) { \n   return(1L) \n  } \n}",
  linters = unnecessary_nested_if_linter()
)
#> <text>:2:3: warning: [unnecessary_nested_if_linter] Don't use nested `if` statements, where a single `if` with the combined conditional expression will do. For example, instead of `if (x) { if (y) { ... }}`, use `if (x && y) { ... }`.
#>   if (y) { 
#>   ^~

# okay
writeLines("if (x && y) { \n  return(1L) \n}")
#> if (x && y) { 
#>   return(1L) 
#> }
lint(
  text = "if (x && y) { \n  return(1L) \n}",
  linters = unnecessary_nested_if_linter()
)

writeLines("if (x) { \n  y <- x + 1L\n  if (y) { \n   return(1L) \n  } \n}")
#> if (x) { 
#>   y <- x + 1L
#>   if (y) { 
#>    return(1L) 
#>   } 
#> }
lint(
  text = "if (x) { \n  y <- x + 1L\n  if (y) { \n   return(1L) \n  } \n}",
  linters = unnecessary_nested_if_linter()
)
```

<sup>Created on 2022-11-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


tests/testthat/test-unnecessary_nested_if_linter.R
```diff
@@ -0,0 +1,66 @@
...
+      if (x) {
+        return(1L)
+      } else {
+        if (y) {
+          return(2L)
+        }
+      }"),
+    NULL,
+    unnecessary_nested_if_linter()
+  )
```
IndrajeetPatil, line 38: This is currently failing. Need to update XPath to fix this test.

tests/testthat/test-unnecessary_nested_if_linter.R
```diff
@@ -0,0 +1,157 @@
...
+        }
+      }"),
+    lint_message,
+    linter
+  )
+
+  expect_lint(
+    "if (x) if (y) 1L",
+    lint_message,
+    linter
```
IndrajeetPatil, line 155: Rare, but may worth be covering as an edge case.

tests/testthat/test-unnecessary_nested_if_linter.R
```diff
@@ -0,0 +1,157 @@
...
+  expect_lint(
+    trim_some("
+      if ((x && y) || (if (x) TRUE else FALSE)) {
+        1L
+      }"),
+    NULL,
+    linter
+  )
+
+  # if there is any additional code between the inner and outer scopes, no lint
```
IndrajeetPatil, line 67: Stumped by this one; can't seem to find XPath that covers these test cases.

AshesITR: Just a hunch, but maybe you need to use `count(*)` somewhere to ensure the first `if ()` only has one child expression?

R/unnecessary_nested_if_linter.R
```diff
@@ -0,0 +1,57 @@
...
+#'   text = "if (x) { \n  y <- x + 1L\n  if (y) { \n   return(1L) \n  } \n}",
+#'   linters = unnecessary_nested_if_linter()
+#' )
+#'
+#' @evalRd rd_tags("unnecessary_nested_if_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+unnecessary_nested_if_linter <- function() {
+  xpath <- glue::glue("
+                    //*[1][IF and not(ELSE)]
```
AshesITR, line 29: Avoid `//*` using `//IF/parent::expr[not(ELSE)]/...` instead.

tests/testthat/test-unnecessary_nested_if_linter.R
IndrajeetPatil: Thanks! That was helpful. Seems to be working now.

IndrajeetPatil:
There is one false negative remaining that I can't seem to fix. If you can suggest a solution, it can be resolved in this PR, or we can file an issue and punt it to later. I have never seen code like this in the wild, and this is just for the sake of completeness.

``` r
library(lintr)

lint(
  text = "if (x) if (y) 1L",
  linters = unnecessary_nested_if_linter()
)
```

AshesITR:
Try `"//IF/parent::expr[not(ELSE)]/parent::expr[not(OP-LEFT-BRACE) or count(expr) = 1]/ancestor-or-self::expr[IF and not(ELSE)]/IF"`.

Also note: `glue::glue()` is unnecessary here.

IndrajeetPatil:
> It might be easier to work inward from the outer IF, though. The ancestor-or-self axis has the risk of recursing up more levels, so a test case like the following should be included.

You were right. I included this test, and it does fail with the current approach.

AshesITR:
Here is my shot at the edge case.
I haven't found a good way to limit `descendant-or-self` or `ancestor-or-self` to just one level, so coding out the two likely ways.
This won't find `if (x) { { if (y) { code } } }`, though.

For that, we'd need all intermediate nodes to satisfy `expr[OP-LEFT-BRACE and count(expr) = 1]`.

```
//IF/parent::expr[not(ELSE)]/OP-RIGHT-PAREN/following-sibling::expr[IF and not(ELSE)] |
//IF/parent::expr[not(ELSE)]/OP-RIGHT-PAREN/following-sibling::expr[OP-LEFT-BRACE and count(expr) = 1]/expr[IF and not(ELSE)]
```

AshesITR:
Some more suggested tests because this is so hard to get right:

```r
if (x) {
  {
    some_code
    if (y) {
      trip_me
    }
  }
}

if (x) {
  {
    if (y) {
      trip_me
    }
    some_code
  }
}


if (x) {
  {
    some_code
    {
      if (y) {
        trip_me
      }
    }
  }
}
```

AshesITR (approved):

AshesITR:
Cool, nice work 😁

PR #1785, opened 2022-12-01
Add GHA workflow to compute code coverage via examples

### LINKED ISSUE #1675: Add GHA for examples coverage?
TIL covr supports coverage for examples. This will be good to get a sense of where there's major gaps in what's in examples.

https://covr.r-lib.org/reference/package_coverage.html

Has anyone seen a package using this feature in GHA? does codecov support something like having two coverage layers? or maybe we need to do something like gh-pages where we have one branch used just to calculate examples coverage?

IndrajeetPatil:
Closes #1675


.github/workflows/test-coverage-examples.yaml
```diff
@@ -0,0 +1,45 @@
...
+            local::.
+
+      - name: Test example coverage
+        run: |
+          options(crayon.enabled = TRUE)
+          library(covr)
+          coverage <- covr::package_coverage(type = "examples", quiet = TRUE)
+          print(coverage)
+          percent_coverage <- covr::percent_coverage(coverage)
+          threshold <- 95
```
IndrajeetPatil, line 37: Maybe 90% is a more reasonable threshold.

.github/workflows/test-coverage-examples.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Test example coverage
+        run: |
+          options(crayon.enabled = TRUE)
+          library(covr)
+          files_to_exclude <- c(
+            # examples present but not run
+            "R/lint.R",
+            "R/use_lintr.R",
+            # mostly internal utilities
```
IndrajeetPatil, line 38: We can work on removing some of these exceptions in the future.

MichaelChirico:
Thanks for keeping at this PR.

I'm not sure at a high level how to think of this PR... are we expecting ourselves to achieve (asymptotically) 100% coverage, twice? (once through `tests/` code and again through `\examples{}`?

Would it be better if we just focused on getting (asymptotically) 100% coverage between `tests/` ∪ `\examples{}`? i.e., including examples for coverage is just used as supplemental coverage to what we get from tests?

IndrajeetPatil:
> I'm not sure at a high level how to think of this PR... are we expecting ourselves to achieve (asymptotically) 100% coverage, twice?

No, I think the goals here are quite different-

- code coverage through **tests**: targeted at **developers** and used as a check that most LOC are covered by unit tests
- code coverage through **examples**: targeted at **users** and used as a check that the key parts of the API are covered well in examples

I don't think there is much to gain from achieving 100% code coverage in the latter case, and this is why we have such an extensive exceptions list. The goal is to make sure that we have enough examples for key functions, and code coverage is just a way to check that; maybe there is another way to check this, but I am not aware of one.

Does that sound reasonable?

MichaelChirico:
Does it make sense to have such a test in CI? It feels like something we should monitor occasionally, but is not high importance to match on every PR.

In particular the long list of `files_to_exclude` seems like a maintenance headache to me.

Next, what does `# nocov` look like for examples? Would we have to hind them in `\dontshow{}` blocks to avoid cluttering the user-facing code?

IndrajeetPatil:
> Does it make sense to have such a test in CI? 

I think so, yes. We have added linters for which either there were no examples or not sufficient number of examples.

> It feels like something we should monitor occasionally, but is not high importance to match on every PR.

Sure, we don't need to run it on every PR. We could set up a cron job to run this workflow on a weekly or monthly basis on the default branch. 

> In particular the long list of files_to_exclude seems like a maintenance headache to me.

This is going to be a problem only if we frequently add new files that contain utility or internal-only functions. But, based on my experience in the last 2 years, this is rarely the case, and it may become increasingly rare going forward.

> Next, what does # nocov look like for examples?

Can you give me an example where we will need to do this? 
(Note that the 90% threshold is set for the entire codebase, and not on a file-by-file basis.)

MichaelChirico:
> Can you give me an example where we will need to do this?

Sorry, I was confusing myself. I guess what I had in mind is a separate sentinel like `# nocovExamples` we could define for this CI, that might also get us out of all the `except=` logic, WDYT?

IndrajeetPatil:
> I guess what I had in mind is a separate sentinel like # nocovExamples we could define for this CI, that might also get us out of all the files_to_exclude logic, WDYT? 

Yes, that's something we can slowly work on.

MichaelChirico (approved):
LGTM to try it out, thanks for your persistence!

Issue #1476, opened 2022-07-27
Release lintr 3.1.0


MichaelChirico:
I think the fixes we have for our CRAN warnings/errors, as well as the fix for #1405, are enough to warrant a patch release.

So filing this issue (along with the created milestone) for us to earmark some issues/PRs as important enough for inclusion, then we can submit a new release.

Prepare for release:

* [x] [`3.1.0` milestone](https://github.com/r-lib/lintr/milestone/3)
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()` [done](https://win-builder.r-project.org/s1DCNgaQf871)
* [x] `rhub::check_for_cran()` [1](https://builder.r-hub.io/status/lintr_3.0.2.9000.tar.gz-c4c8067bd211471799147fcc5d125cb0) [2](https://builder.r-hub.io/status/lintr_3.0.2.9000.tar.gz-8f097dd080944b8482d2a2031dfe62e8) [3](https://builder.r-hub.io/status/lintr_3.0.2.9000.tar.gz-2faea2d3d2054a8f865fbdce8e2cd413)
* [x] Run [`revdepcheck::revdep_check()`](https://revdepcheck.r-lib.org/reference/revdep_check.html).
* [x] Run `.dev/revdep_get_repos.R` and `.dev/revdep_compare_releases.R`.

Submit to CRAN:

* [x] `usethis::use_version('patch')`
* [x] `devtools::submit_cran()`
* [x] Approve email

Wait for CRAN...

* [x] Update `cran-comments.md`
* [x] Accepted :tada:
* [x] `usethis::use_github_release()`
* [x] `usethis::use_dev_version()`
* [ ] Tweet

IndrajeetPatil:
I think the version for this issue should be changed, since `3.0.1` is already under review at CRAN. 

Maybe `3.1.0`?

MichaelChirico:
renamed the milestone as 3.0.2 for now. we're free to rename it if we decide for a bump to 3.1

IndrajeetPatil:
This needs to be bumped again since the latest CRAN is 3.0.2.

jimhester:
Since there are new features and linters in this release I think it should probably be 3.1.0

IndrajeetPatil:
Shouldn't we make a new release soon? There is a lot of new functionality already that users can immediately benefit from.

MichaelChirico:
Yes, I think we're about due...

[...]

Issue #763, opened 2021-02-17
Bad error handling for invalid `Lint()`s


AshesITR:
When a linter returns a `Lint` with an invalid (`NA`) `column_number`, as has recently happened in some regex-based linters, the error message is extremely confusing:

```
Error in rep.int(character, length) : invalid 'times' value
Calls: <Anonymous> ... print.lint -> cat -> highlight_string -> fill_with -> paste0
Execution halted
```

IMO `Lint()` should do some mild verification during construction so that invalid `Lint()`s can't be created successfully.

Here are a bunch of checks we could add:

 1. No `NA`s are allowed anywhere.
 2. `1L <= column_number <= nchar(line) + 1L`.
 3. `line_number >= 1L`.
 4. `ranges` is a list of length 2 integer(ish) vectors, or `NULL`.
 5. `1L <= ranges[[i]][1L] <= ranges[[i]][2L] <= nchar(line) + 1L` for all `i`.

WDYT?

MichaelChirico:
Would `NULL` be allowed for all of these? I'm thinking of lints that aren't as specific as the line/column level (e.g. file-level lints or line-level lints), which might make sense to leave one/both of `line_number` / `column_number` blank.

For `ranges`, don't we have cases where `ranges[2L]` might apply to a different line?

AshesITR:
IINM NULLs are allowed in some places, but ranges are constrained to a single line only (or NULL), as we can currently only flag one line.

Line and column numbers could default to 1 as that would always be valid.

PR #1788, opened 2022-12-03
safer Lint(), xml_nodes_to_lints() and lint()

### LINKED ISSUE #1427: Invalid parentheses can cause lintr to error or hang
### Steps to reproduce
Run `lintr::lint()` on the following snippets.

This one throws `Error in rep.int(character, length) : invalid 'times' value`:
```r
function() {)
```

This one hangs forever:
```r
{ if (x) x()
```

### Expected behavior
The function should exit cleanly and report a human-readable error.

### Notes
Tested on version `3.0.0` released to CRAN as well as the development version at commit `6c3f0fd`.

AshesITR:
fixes #1427, and ensures an easier debugging experience because `Lint()` errors immediately on `NA` in `ranges`.
`xml_nodes_to_lints()` uses default values if the location XPaths fail to return a number.
This could also be made an error. WDYT?

Note that the current fallback solution doesn't guarantee col1 <= column_number <= col2.
Do we want to enforce this as well?


R/get_source_expressions.R
```diff
@@ -83,7 +83,7 @@ get_source_expressions <- function(filename, lines = NULL) {
   source_expression$content <- get_content(source_expression$lines)
   parsed_content <- get_source_expression(source_expression, error = function(e) lint_parse_error(e, source_expression))
 
-  if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line))) {
+  if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line) || e$message == "unexpected end of input")) {
```
AshesITR, line 86: This fixes the `"{ if (x) x()"` case.

R/indentation_linter.R
```diff
@@ -181,7 +181,7 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
       }
       change_begin <- as.integer(xml2::xml_attr(change, "line1")) + 1L
       change_end <- xml2::xml_find_num(change, xp_block_ends)
-      if (change_begin <= change_end) {
+      if (isTRUE(change_begin <= change_end)) {
```
AshesITR, line 184: This fixes `"function() {)"` for `indentation_linter()`.

R/xml_nodes_to_lints.R
```diff
@@ -61,17 +61,29 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
...
 
   lines <- source_expression[["lines"]]
   if (is.null(lines)) lines <- source_expression[["file_lines"]]
 
   if (xml2::xml_attr(xml, "line2") == line1) {
     col2 <- xp_find_location(xml, range_end_xpath)
+    if (is.na(col2)) {
+      warning("Could not find range end for lint. Defaulting to width 1.")
+      col2 <- col1
+    }
```
AshesITR, line 77: This fixes `"function() {)"` for `function_left_parentheses_linter()`.

AshesITR:
Related to #763 as well.
Might do this together.

AshesITR:
Funny how many bugs this change uncovered in our own codebase 😅

R/indentation_linter.R
```diff
@@ -231,7 +231,10 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
       lint_lines <- unname(as.integer(names(source_expression$file_lines)[bad_lines]))
       lint_ranges <- cbind(
         pmin(expected_indent_levels[bad_lines] + 1L, indent_levels[bad_lines]),
-        pmax(expected_indent_levels[bad_lines], indent_levels[bad_lines])
+        pmin(
+          pmax(expected_indent_levels[bad_lines], indent_levels[bad_lines]),
+          nchar(source_expression$file_lines[bad_lines]) + 1L
+        )
```
AshesITR, line 237: If the expected indent is larger than the current line width, the lint range would become invalid.

R/indentation_linter.R
```diff
@@ -241,7 +244,12 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
         type = "style",
         message = lint_messages,
         line = unname(source_expression$file_lines[bad_lines]),
-        ranges = apply(lint_ranges, 1L, list, simplify = FALSE)
+        ranges = lapply(
+          seq_along(bad_lines),
+          function(i) {
+            list(lint_ranges[i, ])
+          }
+        )
```
AshesITR, line 252: `apply(..., simplify = FALSE)` is R >= 4.1, so we need to use `lapply()` instead.

tests/testthat/test-methods.R
```diff
@@ -42,7 +42,7 @@ test_that("as.data.frame.lints", {
       type = "error",
       message = "Under no circumstances is the use of foobar allowed.",
       line = "a <- 1",
-      ranges = list(c(1L, 2L), c(10L, 20L))
+      ranges = list(c(1L, 2L), c(6L, 7L))
```
AshesITR, line 45: Previous second range was invalid.

tests/testthat/test-rstudio_markers.R
```diff
@@ -40,7 +40,7 @@ test_that("it returns markers which match lints", {
       Lint(
         filename = "test_file2",
         line_number = 10L,
-        column_number = 5L,
+        column_number = 1L,
```
AshesITR, line 43: line defaults to `""`, so only `column_number = 1L` is valid.

AshesITR:
@IndrajeetPatil @MichaelChirico this is ready for review now, I think.

R/get_source_expressions.R
```diff
@@ -152,6 +152,19 @@ lint_parse_error <- function(e, source_expression) {
   lint_parse_error_nonstandard(e, source_expression)
 }
 
+fixup_line <- function(line) {
+  # helper to ensure a valid string is provided as line where necessary handles tow cases:
```
MichaelChirico, line 156: i'm not understanding this comment, some typos?

> handles tow cases

R/get_source_expressions.R
```diff
@@ -152,6 +152,19 @@ lint_parse_error <- function(e, source_expression) {
...
+  #  * NA lines, as generated by running outside of code blocks in Rmd documents
+  #  * invalid string lines, as generated by invalid encoding settings
+  if (is.na(line)) {
+    return("")
+  }
+  tryCatch({
```
MichaelChirico, line 162: Nit: for readability I might prefer s.t. like

```r
nchar_fails <- inherits(tryCatch(nchar(line, type = "chars"), error = identity), "error")
if (nchar_fails) {
  ""
} else {
  line
}
```

Or something along those lines. Now it might be tricky/require some mental cache to figure out what's actually being returned, and the construct of having two statements in the `tryCatch()` when only the first is expected to cause issues might be confusing (though I often use it myself and I'm sure it's used elsewhere in our sources...)

WDYT?

R/get_source_expressions.R
MichaelChirico, line 156: nit: I often write internal helpers' docs as roxygen2 style with `@noRd` tag

R/get_source_expressions.R
```diff
@@ -172,23 +185,27 @@ lint_parse_error_r43 <- function(e, source_expression) {
...
-  if (line < 1L || line > length(source_expression$lines)) {
+  if (line_number < 1L || line_number > length(source_expression$lines)) {
     # Safely handle invalid location info
-    line <- 1L
+    line_number <- 1L
     column <- 1L
   }
 
+  line <- fixup_line(source_expression$lines[[line_number]])
+  if (column == 0L || column > nchar(line) + 1L) {
```
MichaelChirico, line 198: nit: it gives me some pause to see `nchar()` run 3 times, but I think it's fine b/c there's only one line here and this is only run once/file, is that right?

R/indentation_linter.R
MichaelChirico: worth leaving as a comment?

R/indentation_linter.R
MichaelChirico: leave a comment for several years later when we can depend on R>=4.1?

R/lint.R
```diff
@@ -402,6 +402,18 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
     )
   }
 
+  max_col <- max(nchar(line) + 1L, 1L, na.rm = TRUE)
+  if (is.na(column_number) || column_number < 0L || column_number > max_col) {
```
MichaelChirico, line 406: since `Lint()` is exported, let's check `length(column_number) != 1L` as well?

R/lint.R
```diff
@@ -402,6 +402,18 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
     )
   }
 
+  max_col <- max(nchar(line) + 1L, 1L, na.rm = TRUE)
```
MichaelChirico, line 405: Also check `length(line) > 0` to ensure `length(max_col) > 0`

R/lint.R
```diff
@@ -418,6 +430,12 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
   obj
 }
 
+is_valid_range <- function(range, max_col) {
+  0L <= range[[1L]] &&
```
MichaelChirico, line 434: maybe a `stopifnot(length(ranges) == 2L)`?

also why `range` vs `ranges` difference?

MichaelChirico: oh nvm, I confused this with being the `check_ranges()` body

R/lint.R
```diff
@@ -463,6 +481,23 @@ rstudio_source_markers <- function(lints) {
   out
 }
 
+check_ranges <- function(max_col, ranges) {
+  if (is.null(ranges)) {
+    return()
+  }
+  if (!is.list(ranges)) {
+    stop("`ranges` must be NULL or a list.")
+  } else if (!all(lengths(ranges) == 2L) || !all(vapply(ranges, is.numeric, logical(1L)))) {
```
MichaelChirico, line 490: looks like we run 5-6 loops over `ranges`, how about

```r
if (!is.list(ranges)) {
  stop(...)
}
for (ii in seq_along(ranges)) check_range(ranges[[ii]])
```

and an individual range checker for each range that checks `length()`, `is.numeric()`, `anyNA()`, `is_valid_range()`?

tests/testthat/test-xml_nodes_to_lints.R
```diff
@@ -86,6 +86,48 @@ test_that("it handles multi-line lints correctly", {
...
 
+test_that("it doesn't produce invalid lints", {
+  # Part of #1427
+  code <- "before   %+%   after"
+  tmpfile <- withr::local_tempfile(lines = code)
+  expr <- get_source_expressions(tmpfile)$expressions[[2L]]
+  xml <- expr$full_xml_parsed_content
+  node <- xml2::xml_find_first(xml, "//SPECIAL")
+
+  expect_warning(l_invalid_loc1 <- xml_nodes_to_lints(
```
MichaelChirico, line 97: these tests require a lot of staring to figure out what's going on, any ideas for improvement? o/w a comment is needed

MichaelChirico:
Thanks for this! Mostly nit-picky comments, sorry about that. 95% of the way there :crossed_fingers:

R/get_source_expressions.R
AshesITR: Yeah, should be "two cases"
The problem is some situations can wind up with `line` being either `NA_character_` or containing an invalid multibyte character. Both of these will mess up the `Lint()` object, so need to be addressed.

IndrajeetPatil:
> The linked issues are milestone 3.0.3, so moving this PR to 3.0.3 as well.

I'm confused now. From https://github.com/r-lib/lintr/issues/1476, I understood that there isn't going to be a `3.0.3` release; the next release will be `3.1.0`.

R/get_source_expressions.R
AshesITR: Rewrote to 
```r
nchars <- tryCatch(nchar(line, type = "chars"), error = function(e) NA_integer_)
if (is.na(nchars)) {
  ""
} else {
  line
}
```

R/get_source_expressions.R
AshesITR: This is only called once per file and only if a parse error occurred in that file.
Also, one of the `nchar()` calls could only be avoided by complicating the return value of `fixup_line()` to also return `nchar(line)`, or 0.

All in all, I don't really think this is worth the effort.

MichaelChirico: Agreed, thanks for confirming.

R/get_source_expressions.R
```diff
@@ -152,6 +152,30 @@ lint_parse_error <- function(e, source_expression) {
...
+#'
+#' [nchar()] can detect both cases: 1. returns `NA_integer_`, 2. errors
+#'
+#' @param line Possibly misencoded or NA line
+#'
+#' @return `line` if it's a valid non-NA string, otherwise an empty string.
+#'
+#' @noRd
+fixup_line <- function(line) {
+  nchars <- tryCatch(nchar(line, type = "chars"), error = function(e) NA_integer_)
```
MichaelChirico, line 171: tempting to use `NA` here, since we're only checking `is.na()`, since technically `logical` type _could_ be smaller than integer... OTOH maybe `NULL` would allocate even less space. OTOOH this doesn't really matter. Moving on.

R/lint.R
```diff
@@ -382,7 +382,7 @@ reorder_lints <- function(lints) {
 #' Create a `lint` object
 #' @param filename path to the source file that was linted.
 #' @param line_number line number where the lint occurred.
-#' @param column_number column number where the lint occurred.
+#' @param number column number where the lint occurred.
```
MichaelChirico, line 385: is that intentional? this would be a user-facing API change? I guess this is by mistake...

R/lint.R
```diff
@@ -418,6 +433,35 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
   obj
 }
 
+is_number <- function(number, n = 1L) {
+  length(number) == n && is.numeric(number) && !anyNA(number)
```
MichaelChirico, line 437: TIL `length()` is ever-so-slightly faster & so is better to test first. nice!

R/get_source_expressions.R
AshesITR:
``` r
pryr::object_size(NA)
#> 56 B
pryr::object_size(NA_integer_)
#> 56 B
```

<sup>Created on 2022-12-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'd prefer type safety (`nchars` is always an integer) instead of code brevity here.

PR #1799, opened 2022-12-05
Add test for native pipe in `commented_code_linter()`

IndrajeetPatil:


tests/testthat/test-commented_code_linter.R
```diff
@@ -84,13 +84,15 @@ test_that("commented_code_linter can detect operators in comments and lint corre
 
   test_ops <- c(
     "+", "=", "==", "!=", "<=", ">=", "<-", "<<-", "<", ">", "->",
-    "->>", "%%", "/", "^", "*", "**", "|", "||", "&", "&&", "%>%",
-    "%anything%"
+    "->>", "%%", "%||%", "/", "^", "*", "**", "|", "||", "&", "&&",
```
IndrajeetPatil, line 87: From rlang

tests/testthat/test-commented_code_linter.R
```diff
@@ -84,13 +84,15 @@ test_that("commented_code_linter can detect operators in comments and lint corre
...
+    "%>%", "%<>%", "%T>%", "%anything%"
```
IndrajeetPatil, line 88: From magrittr

AshesITR:
Why would `commented_code_linter()` need more tests regarding the pipe?

IndrajeetPatil:
> Why would commented_code_linter() need more tests regarding the pipe?

Because, currently, we are only testing the magrittr pipe, and not the native pipe.

AshesITR:
But from a code analysis point-of-view, all the SPECIALs are equivalent, i.e. `%anything%` and `%>%` are essentially the same thing for the parser.
I'm fine with adding a `|>` test just-in-case, but the others seem excessive.

AshesITR (approved):
Thanks 😊

PR #1803, opened 2022-12-06
Add contributing guidelines document

IndrajeetPatil:
Via `usethis::use_tidy_contributing()`

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Contributing to lintr
+
+This outlines how to propose a change to lintr. 
+For more detailed info about contributing to this, and other tidyverse packages, please see the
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Some orgs limit members from contributing to projects with contributing guidelines (e.g. Google has an allowlist), could you elaborate on the reasoning to include one?

We should lay out the tradeoffs of this decision in the PR before merging.

IndrajeetPatil:
👀

I had no idea that such restrictions exist! Thanks for letting me know. 

In that case, I don't see this document worth the risk of potentially preventing someone from contributing.

MichaelChirico:
I don't know how restrictive they are in practice (e.g. I have yet to encounter a repo where I can't contribute for this reason), but yes, there might be some legal/liability issues I guess (I don't know details).

PR #1812, opened 2022-12-08
add assignment_as_infix = TRUE to indentation_linter()

### LINKED ISSUE #1800: False positive in `indentation_linter()`?
Possibly related to https://github.com/r-lib/styler/issues/1065.

``` r
library(lintr)

'test_that("bla", {
  x <-
    letters[1:3] %>%
    paste0() %>%
    length()
})' -> code

lint(
  text = code,
  linters = indentation_linter()
)
#> <text>:4:4: style: [indentation_linter] Indentation should be 6 spaces but is 4 spaces.
#>     paste0() %>%
#>    ^~~
```

This is what `{lintr}` expects:

```r
test_that("bla", {
  x <-
    letters[1:3] %>%
      paste0() %>%
      length()
})
```

But `{styler}` changes this to the above-mentioned version that produces lints.

AshesITR:
fixes #1800

Currently, this only suppresses indentation for descendants of `<-`-triggered indents and only if these descendants are

 * infix operators
 * not within a function call

I'm not 100% sure this fixes all cases, but I added a more complex test case to ensure it works as intended.
Let me know if you can think of other test cases that should be included.

Also, let me know if the exception should apply to other assignment-like operators, such as EQ_ASSIGN, EQ_SUB, EQ_FORMALS, ...

One case where we probably still have a false-positive is within `lapply()` and friends. That seems hard if not impossible to fix:

```r
lapply(x,
  function(e) {
    temp_var <-
      e +
      42
  }
)
```


AshesITR:
Link rot check is getting HTTP 500s at the moment. All other checks pass ✔️

R/indentation_linter.R
```diff
@@ -118,6 +140,21 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
     }
   }
 
+  if (isTRUE(assignment_as_infix)) {
+    infix_condition <- glue::glue("
+      and not(
+        ancestor::expr[
+          preceding-sibling::LEFT_ASSIGN[{xp_last_on_line}] or expr[SYMBOL_FUNCTION_CALL]
```
MichaelChirico, line 147: I guess we should include other assignment symbols here too, in general

R/indentation_linter.R
```diff
@@ -186,17 +223,13 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
...
-          } else {
-            expected_indent_levels[to_indent] <- expected_indent_levels[to_indent] + indent
-          }
-          is_hanging[to_indent] <- FALSE
-        }
+        expected_indent_levels[to_indent] <- find_new_indent(
+          expected_indent_levels[to_indent],
+          change_type,
+          indent,
+          as.integer(xml2::xml_attr(change, "col2"))
```
MichaelChirico, line 230: add the argument name here for readability

MichaelChirico (commented):
thanks!

AshesITR:
Thanks for your feedback.
I added `OP-LEFT-BRACE` to the resetting tokens and `EQ_SUB`, `EQ_ASSIGN` and `EQ_FORMALS` to the suppressing tokens.
The new tests cover all the new exceptions.

I hope the comment is more clear on how the XPath works now.

AshesITR:
@MichaelChirico LMK when you have the time to review. I want to avoid spamming merges.

R/indentation_linter.R
```diff
@@ -97,7 +118,8 @@
 #' - <https://style.tidyverse.org/functions.html#long-lines-1>
 #'
 #' @export
-indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "always", "never")) {
+indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "always", "never"),
```
MichaelChirico, line 121: nit: i usually prefer each argument with defaults on its own line once >1 line is used, WDYT? not sure it's a stated rule, but it's similar to the long lines rule for calls.

R/indentation_linter.R
```diff
@@ -261,6 +297,16 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
   })
 }
 
+find_new_indent <- function(current_indent, change_type, indent, hanging_indent) {
+  if (change_type == "hanging") {
```
MichaelChirico, line 301: nit: maybe a `switch()` is more appropriate? `change_type` feels pretty `enum`-like

tests/testthat/test-indentation_linter.R
```diff
@@ -634,6 +634,96 @@ test_that("hanging_indent_stlye works", {
...
+      function(e) {
+        temp_var <-
+          e +
+          42
+      }
+    )
+  ")
+
+  # test brace restorator
+  code_infix_3 <- trim_some("
```
MichaelChirico, line 661: that's some horrible code 😄

MichaelChirico (approved):
Only nits left, great work!

PR #1814, opened 2022-12-08
Implement `implicit_assignment_linter()`

### LINKED ISSUE #1777: New linter for assignments within conditional expressions
In the wild, I have come across cases where variables are assigned within conditional expressions and their value is then immediately used as a condition and/or inside their scope. I think this makes the code difficult to understand, and we should lint it.

WDYT?

Here is a reprex with a simple assignment, but you can imagine far more complex assignments and the difficulty to understand the code increases with the complexity of the assignment.

``` r
library(lintr)

if (x <- 1L) print(x)
#> [1] 1

# currently, this doesn't generate any lints
lint(
  text = "if (x <- 1L) print(x)",
  linters = linters_with_tags(tags = NULL)
)
```

<sup>Created on 2022-11-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The suggested alternative here in the lint message would be the following (i.e. do variable assignment outside the expression and then use the variable instead):

```r
x <- 1L
if (x) print(x)
```

IndrajeetPatil:
Closes #1777

---

``` r
library(lintr)

# will produce lints
lint(
  text = "if (x <- 1L) TRUE",
)
#> <text>:1:7: warning: [implicit_assignment_linter] Avoid implicit assignments in function calls. For example, instead of `if (x <- 1L) { ... }`, write `x <- 1L; if (x) { ... }`.
#> if (x <- 1L) TRUE
#>       ^~

lint(
  text = "mean(x <- 1:4)",
)
#> <text>:1:8: warning: [implicit_assignment_linter] Avoid implicit assignments in function calls. For example, instead of `if (x <- 1L) { ... }`, write `x <- 1L; if (x) { ... }`.
#> mean(x <- 1:4)
#>        ^~

# okay
writeLines("x <- 1L\nif (x) TRUE")
#> x <- 1L
#> if (x) TRUE
lint(
  text = "x <- 1L\nif (x) TRUE",
)

writeLines("x <- 1:4\nmean(x)")
#> x <- 1:4
#> mean(x)
lint(
  text = "x <- 1:4\nmean(x)",
)
```

<sup>Created on 2022-12-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


IndrajeetPatil:
Converting back to draft because found a bunch of false positives.

tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -0,0 +1,103 @@
...
+
+  expect_lint(
+    trim_some("
+    f <- function() {
+      p <- g()
+      p <- if (is.null(p)) x else p
+    }"),
+    NULL,
+    linter
+  )
```
IndrajeetPatil, line 75: This test keeps failing. 🙈

73mb2o

R/implicit_assignment_linter.R
```diff
@@ -0,0 +1,90 @@
...
+#'
+#' @export
+implicit_assignment_linter <- function() {
+  assignments <- c(
+    "LEFT_ASSIGN", # e.g. mean(x <- 1:4)
+    "RIGHT_ASSIGN" # e.g. mean(1:4 -> x)
+  )
+  xpath_fun_call <- paste0(
+    "
+    //SYMBOL_FUNCTION_CALL[not(text() = 'capture.output')]
```
AshesITR, line 44: These exceptions should be configurable.
I can see `test_that()`, `local()` and language functions like `quote()` joining that list by default, but there are probably others.

R/implicit_assignment_linter.R
```diff
@@ -0,0 +1,90 @@
...
+    /parent::expr
+    /following::expr[1]
```
AshesITR, line 46: Is "following" an alias for "following-sibling"?

AshesITR:
I feel like you need `following-sibling` everywhere you use `following`.


axis | description
-- | --
following | Selects everything in the document after the closing tag of the current node
following-sibling | Selects all siblings after the current node

- https://www.w3schools.com/xml/xpath_axes.asp

IndrajeetPatil:
> I feel like you need `following-sibling` everywhere you use `following`.

Thanks! Yes, that was indeed the problem. 
In my mind, I was thinking about sibling, but somehow wrote down just `following`.

R/implicit_assignment_linter.R
```diff
@@ -0,0 +1,112 @@
...
+#' @export
+implicit_assignment_linter <- function(except = c(
+                                         "bquote", "capture.output", "expression", "expr",
+                                         "expect_error", "expect_warning", "expect_message",
+                                         "expect_no_error", "expect_no_warning", "expect_no_message",
+                                         "expect_condition", "expect_no_condition",
+                                         "expect_invisible", "expect_visible",
+                                         "expect_output", "expect_silent",
+                                         "local", "quo", "quos", "quote", "test_that"
+                                       )) {
```
IndrajeetPatil, line 47: @AshesITR I styled the file with `{styler}`, but still can't get rid of the lint from `indentation_linter`:

> Line 40: Indentation should be 2 spaces but is 41 spaces. 

So the builds are going to fail for now.

AshesITR: There is a line length lint as well.
styler seems off here, this is similar to a multiline function call, which should be block-indented.

Compare
```r
lapply(x, FUN = function(i) {
  42L
})
```

Lintr expects this to look like
```r
implicit_assignment_linter <- function(except = c(
  "bquote", "capture.output", "expression", "expr",
  "expect_error", "expect_warning", "expect_message",
  "expect_no_error", "expect_no_warning", "expect_no_message",
  "expect_condition", "expect_no_condition",
  "expect_invisible", "expect_visible",
  "expect_output", "expect_silent",
  "local", "quo", "quos", "quote", "test_that"
)) { 
  # code ... 
}
```

AshesITR (approved):

tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -0,0 +1,144 @@
...
+  linter <- implicit_assignment_linter()
+
+  expect_lint("if (x <- 1L) TRUE", lint_message, linter)
+  expect_lint("if (1L -> x) TRUE", lint_message, linter)
+  expect_lint("while (x <- 0L) FALSE", lint_message, linter)
+  expect_lint("while (0L -> x) FALSE", lint_message, linter)
+  expect_lint("for (x in y <- 1:10) print(x)", lint_message, linter)
+  expect_lint("for (x in 1:10 -> y) print(x)", lint_message, linter)
+
+  expect_lint("mean(x <- 1:4)", lint_message, linter)
```
MichaelChirico, line 132: what happens for multi-line anonymous functions?

```r
lapply(x, function(xi) {
  xi <- xi + 1
  xi
})
```

ditto for purrr anonymous style

PR #1817, opened 2022-12-08
Revise exceptions to `implicit_assignment_linter()`

IndrajeetPatil:
Follow up on #1814


AshesITR:
Can we run this linter on some CRAN packages to see which functions come up and sift for false positives?
e.g. I've not seen snapshot tests used with implicit assignments.

IndrajeetPatil:
You are right, the snapshot variants are not necessary to be included in the exceptions.

I was imagining [a test](https://github.com/cynkra/dm/blob/8f2a5c2867cc4305d498355b8fb91c33f8827b2b/tests/testthat/test-json_nest.R#L2-L7) like the following tripping up the linter, but it doesn't:

``` r
library(lintr)

lint(
  text = "expect_snapshot({
    df <- tibble::tibble(x = c(1, 1, 1, 2, 2, 3), y = 1:6, z = 6:1)
    nested <- json_nest(df, data = c(y, z))
    nested
  })",
)
```

<sup>Created on 2022-12-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
@AshesITR I have reverted the changes. `tryCatch` is the only new addition now. 
We can keep expanding the exceptions vector depending on the user feedback.

MichaelChirico:
I would be in favor of

 1. an argument to toggle allowing assignments in any braced expression; allow by default
 2. an exceptions argument with a very small set of defaults

in my mind this + the curly linters make for readable code since the assignments will always be on their own line, at least.

am I missing something?

IndrajeetPatil:
> an argument to toggle allowing assignments in any braced expression; allow by default

I don't think we even need the argument. Function calls where implicit assignments happen inside braced expressions won't lint, even if there is no exception specified for that function.

For example, here I leave out `local()` from `except` argument, and it still won't lint:

``` r
library(lintr)

lint(
  text = "local({ a <- 1 })",
  linters = implicit_assignment_linter(except = "tryCatch")
)
``` 

It lints only when the braces are left out.

``` r

lint(
  text = "local(a <- 1)",
  linters = implicit_assignment_linter(except = "tryCatch")
)
#> <text>:1:9: warning: [implicit_assignment_linter] Avoid implicit assignments in function calls. For example, instead of `if (x <- 1L) { ... }`, write `x <- 1L; if (x) { ... }`.
#> local(a <- 1)
#>         ^~
```

<sup>Created on 2022-12-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

So the question is are there any valid instances of usages of functions like `local()` where users can leave out the curly braces? 

If no, we can safely reduce the number of functions included in `except` argument.

MichaelChirico:
oh got it, could you add tests for that too? I was also looking for those in my peek earlier & didn't see any.

as long as curly expressions are always linted, I am not sure what to include as exceptions TBH. maybe just the expect_* functions? even capture.output() seems preferable to run as 

```r
capture.output({
  x <- 1
})
```

IndrajeetPatil:
Do you think there is a new linter in order here that suggests to always use the curly braces for functions that take expression as argument? 

E.g. `local()`, `tryCatch()`, `withCallingHandlers()`, etc.

MichaelChirico:
no, because they may not involve assignment -- capture.output(print(x)) is fine.

in fact one of the google linters is unnecessary_nesting_linter() that would throw for cases adding braces unnecessarily. I need to adjust that to allow one-expression assignments...

AshesITR:
The more I think about it, the more I come to the conclusion that

```r
fun({
  x <- 42L
})
```

should _always_ be allowed and that

```r
fun(x <- 42L)
```

should be linted unless `fun` is one of the language functions, i.e. `quote()`, `bquote()`, ` expression()` and friends from `{rlang}`.
These functions actually have different output with / without braces. All other functions should use braces for readability IMO

WDYT?

MichaelChirico:
Good call-out that the low-level functions do indeed behave differently with / without braces. definitely need an exception for those.

so let's use that as the principle. in the end we have just one argument (except) with no defaults -- the low-level cases are always excepted, but users have the option to add more.

MichaelChirico:
Is there a valid case for removing any of those functions from the exception list?

I had in mind that if you want to add one or two functions, now you have to also copy-paste 5-6 extras to augment the default behavior.

pretty inconvenient & easy to overlook if those functions should always be excepted.

AshesITR:
But if you do that and encounter a false positive from the defaults, you'll be able to help yourself.

OTOH if you want it linted, you can't.

One possible scenario for `bquote()` could be to enforce generated code blocks to be enclosed by braces.

MichaelChirico:
I guess that's the reason for offering a constant in cases like `undesirable_{function,operator}_linter()`... I'm OK to include everything in the default, but let's monitor if we should offer a constant suitable for `modify_defaults()` in the next release.

AshesITR:
So we want `except = c("expression", "quote", "bquote")` and {rlang} friends, right?

MichaelChirico:
> So we want `except = c("expression", "quote", "bquote")` and {rlang} friends, right?

Works for me

IndrajeetPatil:
> So we want `except = c("expression", "quote", "bquote")` and {rlang} friends, right?

But that would mean the `{testthat}` expectations will lint, and none of the following tests will pass:

https://github.com/r-lib/lintr/blob/e0ebf6b6a3b86ff26c84a76eaf5cebfd1e27ce8b/tests/testthat/test-implicit_assignment_linter.R#L230-L241

MichaelChirico:
> But that would mean the {testthat} expectations will lint, and none of the following tests will pass:

At a glance I think that's fine. Do we have any sense of how common this usage is? Seems weird to me that a package author needs to test `expect_no_message(out <- f(-1))`. `expect_no_message(f(-1))` will be better in the vast majority of cases, no? The only exception I can think of is some regression tests for assignment methods (e.g. `[<-.my_class` was found at some point to issue a warning, the bug is fixed, now we add a test like that). That seems rare enough to warrant a `# nolint` or just use `{}`.

AshesITR:
I would've expected the withr test on line 251 to fail due to missing braces. Why does it pass?

IndrajeetPatil:
> I would've expected the withr test on line 245 to fail due to missing braces. Why does it pass?

Good catch. Prima facie, looks like a bug to me. Will have a closer look soon.

IndrajeetPatil:
> I would've expected the withr test on line 245 to fail due to missing braces. Why does it pass?
> 
> `expect_lint("with_options(list(digits = 3L), x <- getOption('digits'))", NULL, linter)`

@AshesITR I have created a follow-up issue (https://github.com/r-lib/lintr/issues/1893).

IndrajeetPatil:
Is there anything else that needs to be done here? Otherwise, this is ready for a merge.

AshesITR (approved):

PR #1821, opened 2022-12-09
Add tests for multi-line anonymous functions for `implicit_assignment_linter()`

IndrajeetPatil:
Follow-up on #1814

### DIFF
--- a/tests/testthat/test-implicit_assignment_linter.R
+++ b/tests/testthat/test-implicit_assignment_linter.R
@@ -77,19 +77,53 @@ test_that("implicit_assignment_linter skips allowed usages", {
 
   expect_lint(
     trim_some("
-      lapply(x, function(xi) {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -74,6 +74,26 @@ test_that("implicit_assignment_linter skips allowed usages", {
...
+        xi <- xi + 1
+        xi
+      })"),
+    NULL,
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      map(x, function(xi) {
```
MichaelChirico, line 90: I had in mind calls with `~`, could you add one of those?

IndrajeetPatil: Done. Lemme know if you have any other tests in mind.

PR #1833, opened 2022-12-11
Tests for multiple lints in `implicit_assignment_linter()`

IndrajeetPatil:


tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -208,14 +208,41 @@ test_that("implicit_assignment_linter blocks disallowed usages", {
...
 
+  # conditional statements
   expect_lint("if (x <- 1L) TRUE", lint_message, linter)
   expect_lint("if (1L -> x) TRUE", lint_message, linter)
   expect_lint("while (x <- 0L) FALSE", lint_message, linter)
   expect_lint("while (0L -> x) FALSE", lint_message, linter)
   expect_lint("for (x in y <- 1:10) print(x)", lint_message, linter)
   expect_lint("for (x in 1:10 -> y) print(x)", lint_message, linter)
 
+  # nested conditional statements
```
MichaelChirico, line 219: recently I lean more towards "if you need a comment within a test_that(), it should be split into multiple test_that(), WDYT?

AshesITR: No strong opinion but we do use comments a lot because many times it's just a tiny special case being tested and the general expectation is that the linting is correct.

The advantage of a separate test is that in can be skipped independently. Apart from that, I see no clear benefits of one over the other.

IndrajeetPatil: I don't have a stronger opinion either way, but have broken down one large test into a few, more focused, smaller ones.

When the tests fail, I think this will provide a better, more specific error message.

MichaelChirico: right, I think that's the gist of it -- we can get the sense of the test right away in the log instead of after reading through the code.

it's not a big thing (e.g. I'm not going to spend time going through existing tests) but something to think about for new code.

MichaelChirico (approved):

PR #1838, opened 2022-12-12
Additional test for `unnecessary_placeholder_linter()`

IndrajeetPatil:
To make sure that we don't lint valid usage of placeholder here:

``` r
library(magrittr)
1:4 %>% { sum(.) }
#> [1] 10
x <- 1:4
x %<>% { sum(.) }
x
#> [1] 10
```

<sup>Created on 2022-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Without the placeholder, the results are incorrect:

``` r
library(magrittr)
1:4 %>% { sum() }
#> [1] 0

x <- 1:4
x %<>% { sum() }
x
#> [1] 0
```

<sup>Created on 2022-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


MichaelChirico:
sorry, a bit confused at the PR description vs actual update. is the outcome of this not the same?

```r
1:4 %>% sum()
x = 1:4
x %<>% sum()
```

IndrajeetPatil:
Sorry for the confusion: 

The examples in the top-level comment were just to show how a placeholder (`.`) is needed here and that it'd be good to have a test that ensures that we are not producing a false positive in such contexts (we aren't).

It's important that there not be a false positive here because the lint message recommends removing `.`, and that would produce an incorrect result here:

``` r
library(magrittr)
1:4 %>% { sum() }
#> [1] 0

x <- 1:4
x %<>% { sum() }
x
#> [1] 0
```

<sup>Created on 2022-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico (approved):

PR #1839, opened 2022-12-12
Use braces in `tryCatch()` calls

IndrajeetPatil:
Inspired by discussion in https://github.com/r-lib/lintr/pull/1817.


R/get_source_expressions.R
```diff
@@ -484,22 +484,26 @@ get_source_expression <- function(source_expression, error = identity) {
   parse_error <- FALSE
 
   tryCatch(
-    source_expression$parsed_content <- parse(
-      text = source_expression$content,
-      srcfile = source_expression,
-      keep.source = TRUE
-    ),
+    {
+      source_expression$parsed_content <- parse(
```
MichaelChirico, line 488: I think more readable / canonical is to avoid the double assignment altogether, e.g.

```r
parsed_content <- tryCatch(...)
if (inherits(parsed_content, ...)) { ... }

...

source_expression$parsed_content <- parsed_content
```

IndrajeetPatil: Good suggestion. 

I have done this, but I couldn't use the suggested name (`parsed_content`) for the reason mentioned in my comment.

If I am mistaken, let me know.

R/get_source_expressions.R
```diff
@@ -483,19 +483,10 @@ get_single_source_expression <- function(loc,
...
-      srcfile = source_expression,
-      keep.source = TRUE
-    ),
-    error = error
-  )
-
-  # This needs to be done twice to avoid
-  #   https://bugs.r-project.org/bugzilla/show_bug.cgi?id=16041
+  # This object should be named `e` because, if invalid characters are present,
+  # the error needs to be assigned in the parent frame, where it is named `e`.
```
IndrajeetPatil, line 487: For this reason, the content can't be assigned to `parsed_content` or some other, more informative name.

MichaelChirico: isn't the parent's name given by the first argument to assign()?

so assign("e", parsed_content, ...) works?

IndrajeetPatil: Ah, yes, true that.

Sorry, I was doing search and replace, which changed both `e` and `"e"`, and that's why it wasn't working.

I have made the necessary change now and the object is called `parsed_content`. 

Thanks!

MichaelChirico (approved):
thanks!

PR #1840, opened 2022-12-12
Allow `except = NULL` to remove all exceptions in `implicit_assignment_linter()`

IndrajeetPatil:
Setting this argument to `NULL` feels like a more natural way to go about removing exceptions than `except = character()`.


R/implicit_assignment_linter.R
```diff
@@ -45,7 +45,7 @@ implicit_assignment_linter <- function(except = c(
   "expect_output", "expect_silent",
   "local", "quo", "quos", "quote", "test_that"
 )) {
-  stopifnot(is.character(except))
+  if (!is.null(except)) stopifnot(is.character(except))
```
AshesITR, line 48: I'd suggest to use `stopifnot(is.null(except) || is.character(except))` so the generated error message reflects the behavior more closely.

AshesITR (approved):

PR #1844, opened 2022-12-13
Improve error message in `namespace_linter()`

IndrajeetPatil:
Specifically, mention where to create issue if the linter fails.


R/namespace_linter.R
```diff
@@ -85,9 +85,14 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
     # run here, not in the factory, to allow for run- vs. "compile"-time differences in package structure
     namespaces <- lapply(packages, function(package) tryCatch(getNamespace(package), error = identity))
     failed_namespace <- vapply(namespaces, inherits, "condition", FUN.VALUE = logical(1L))
+    # nocov start
     if (any(failed_namespace)) {
-      stop("Failed to retrieve namespaces for one or more of the packages. Please report this issue.") # nocov
+      stop(
+        "Failed to retrieve namespaces for one or more of the packages. ",
+        "Please report an issue at https://github.com/r-lib/lintr/issues."
+      )
```
IndrajeetPatil, line 93: Similar to what we do here:

https://github.com/r-lib/lintr/blob/becdbac98bc32c4719539e1d3d5263025b63c738/R/object_usage_linter.R#L295-L298

R/namespace_linter.R
```diff
@@ -85,9 +85,14 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
     # run here, not in the factory, to allow for run- vs. "compile"-time differences in package structure
     namespaces <- lapply(packages, function(package) tryCatch(getNamespace(package), error = identity))
     failed_namespace <- vapply(namespaces, inherits, "condition", FUN.VALUE = logical(1L))
+    # nocov start
     if (any(failed_namespace)) {
-      stop("Failed to retrieve namespaces for one or more of the packages. Please report this issue.") # nocov
+      stop(
+        "Failed to retrieve namespaces for one or more of the packages. ",
```
MichaelChirico, line 91: Let's improve this message too while we're at it -- "the packages" is a bit vague, what packages?

We could either describe it generally "one or more of the packages used with `::` or `:::`", or more specifically by actually naming the packages.

IndrajeetPatil: For now, I have just included the generic message.

I have created a new issue (https://github.com/r-lib/lintr/issues/1855) to remember to improve the message by including problematic package names, since it isn't as trivial as I thought, and don't want to expand the scope of this PR beyond what I originally had in mind.

MichaelChirico (approved):
thanks!

Issue #1847, opened 2022-12-15
`expect_lint_free()` cannot find R package


daniepi:
Hi, it seems like that in version `3.0.2` `lintr::expect_lint_free()` in a test can't locate package.
Works just fine with `3.0.1`.

```
test_that("Lint free code", {
  lintr::expect_lint_free()
})
```

gives

```
Warning (test-lint.R:5:3): Lint free code
Didn't find any R package searching upwards from '.'.
Backtrace:
 1. lintr::expect_lint_free()
      at test-lint.R:5:2
 2. lintr::lint_package(...)
```

Am I doing something wrong? Thanks.

MichaelChirico:
could you share some more information, e.g. the folder structure of your package? e.g. a tree structure like `tree` would give, see https://gist.github.com/jennybc/2bf1dbe6eb1f261dfe60

daniepi:
@MichaelChirico thanks for prompt answer. I would say that package structure is 'standard'

```
.
├── DESCRIPTION
├── LICENSE
├── NAMESPACE
├── NEWS.md
├── R
│   ├── our.package-package.R
│   ├── connections.R
│   ├── data_checks.R
│   ├── ea_writer.R
│   ├── factor_functions.R
│   ├── missing_values.R
│   ├── s3_functions.R
│   ├── sql_functions.R
│   ├── utils.R
│   └── zzz.R
├── README.Rmd
├── README.md
├── our_project.Rproj
└── tests
    ├── testthat
    │   ├── _snaps
    │   │   └── utils.md
    │   ├── helper-schema.R
    │   ├── setup.R
    │   ├── test-connections.R
    │   ├── test-data_checks.R
    │   ├── test-ea_writer.R
    │   ├── test-examples.R
    │   ├── test-factor_functions.R
    │   ├── test-lint.R
    │   ├── test-missing_values.R
    │   ├── test-s3_fuctions.R
    │   ├── test-sql_functions.R
    │   └── test-utils.R
    └── testthat.R
```
Note that I have changed some names and excludes `figures`, `man`and `inst` folders.

MichaelChirico:
Thanks! I think it's possible {testthat} runs in an isolated environment.

One more thing to help debugging, try adding this in your code:

```
test_that("Lint free code", {
  stop("Executing from: ", getwd())
  lintr::expect_lint_free()
})
```

[...]

PR #1850, opened 2022-12-16
friendlier error/warning for failing find_package()

MichaelChirico:
Related: #1847

### DIFF
--- a/R/lint.R
+++ b/R/lint.R
@@ -242,10 +242,16 @@ lint_package <- function(path = ".", ...,
     dots <- dots[-1L]
   }
 
+  if (length(path) > 1L) {
+    stop("Only linting one package at a time is supported.")
+  }
   pkg_path <- find_package(path)
 
   if (is.null(pkg_path)) {
-    warning("Didn't find any R package searching upwards from '", path, "'.")
+    warning(sprintf(
+      "Didn't find any R package searching upwards from '%s'.",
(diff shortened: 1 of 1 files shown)


R/lint.R
```diff
@@ -242,10 +242,16 @@ lint_package <- function(path = ".", ...,
...
+  if (length(path) > 1L) {
+    stop("Only linting one package at a time is supported.")
+  }
   pkg_path <- find_package(path)
 
   if (is.null(pkg_path)) {
-    warning("Didn't find any R package searching upwards from '", path, "'.")
+    warning(sprintf(
+      "Didn't find any R package searching upwards from '%s'.",
+      if (path == ".") normalizePath(path) else path
```
IndrajeetPatil, line 253: Do we have tests for both statements here?

AshesITR: Is there any benefit in only conditionally normalizing?

MichaelChirico: I was thinking lint_package("lintr") would be easier to read than /long/path/to/where/I/have/lintr. but no strong preference.

AshesITR: OTOH in the error condition, the full path avoids any working directory related confusion.

MichaelChirico: OK, removed the conditional

IndrajeetPatil (approved):

PR #1862, opened 2022-12-18
Deprecate `unnecessary_concatenation_linter()`

### LINKED ISSUE #1797: Introduce alias for `unneeded_concatenation_linter()`?
`unneeded_concatenation_linter()` is the odd duck out here, and it keeps tripping us (and I assume `{lintr}` users) up (e.g. see my edit to [this](https://github.com/r-lib/lintr/issues/1787) issue title).  

``` r
library(lintr)
ns <- sort(getNamespaceExports("lintr"))
cat(ns[startsWith(ns, "unn")], sep = "\n")
#> unnecessary_lambda_linter
#> unnecessary_nested_if_linter
#> unnecessary_placeholder_linter
#> unneeded_concatenation_linter
```

Unfortunately, we can't make a breaking change by renaming it, but maybe a good compromise will be to provide an alias? WDYT?

```diff
unnecessary_lambda_linter
unnecessary_nested_if_linter
unnecessary_placeholder_linter
+ unnecessary_concatenation_linter
unneeded_concatenation_linter
```

IndrajeetPatil:
In favour of `unnecessary_concatenation_linter()`

Closes #1797


R/unnecessary_concatenation_linter.R
```diff
@@ -13,45 +13,47 @@
...
+#'   linters = unnecessary_concatenation_linter(allow_single_expression = TRUE)
 #' )
 #'
-#' @evalRd rd_tags("unneeded_concatenation_linter")
+#' @evalRd rd_tags("unnecessary_concatenation_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-unneeded_concatenation_linter <- function(allow_single_expression = TRUE) {
+unnecessary_concatenation_linter <- function(allow_single_expression = TRUE) { # nolint: object_length_linter.
+  stopifnot(is.logical(allow_single_expression) && length(allow_single_expression) == 1L)
```
MichaelChirico, line 55: prefer splitting up the conditions -- stopifnot(a, b) gives better error messages than stopifnot(a && b)

tests/testthat/test-unnecessary_concatenation_linter.R
```diff
@@ -101,15 +101,15 @@ test_that("sequences with : are linted whenever a constant is involved", {
...
     "c(a:foo(b))",
     "Unneeded concatenation of a simple expression",
-    unneeded_concatenation_linter(allow_single_expression = FALSE)
+    unnecessary_concatenation_linter(allow_single_expression = FALSE)
   )
 })
 
 test_that("c(...) does not lint under !allow_single_expression", {
-  expect_lint("c(...)", NULL, unneeded_concatenation_linter(allow_single_expression = FALSE))
+  expect_lint("c(...)", NULL, unnecessary_concatenation_linter(allow_single_expression = FALSE))
```
MichaelChirico, line 114: please add a new test for the deprecation warning

R/unnecessary_concatenation_linter.R
IndrajeetPatil: True that. I've split them now.

IndrajeetPatil:
@MichaelChirico How can I add tests for new validation tests I have added. 

I thought the following would work, but it produces an error:

```r
test_that("invalid allow_single_expression argument produce informative error messages", {
  expect_error(
    expect_lint("c()", NULL, unnecessary_concatenation_linter(allow_single_expression = 1)),
    "is.logical(allow_single_expression) is not TRUE"
  )

  expect_error(
    expect_lint("c()", NULL, unnecessary_concatenation_linter(allow_single_expression = c(TRUE, FALSE))),
    "length(allow_single_expression) == 1L is not TRUE"
  )
})
```

MichaelChirico:
use rex() right?

R/lintr-deprecated.R
```diff
@@ -176,3 +176,21 @@ semicolon_terminator_linter <- function(semicolon = c("compound", "trailing")) {
...
+#' @export
+unneeded_concatenation_linter <- function(allow_single_expression = TRUE) {
+  lintr_deprecated(
+    old = "unneeded_concatenation_linter",
+    new = "unnecessary_concatenation_linter",
+    version = "3.1.0",
+    type = "Linter"
+  )
+
+  stopifnot(
```
MichaelChirico, line 191: isn't this check being duplicated?

IndrajeetPatil: Happy to remove it. 

I deliberately duplicated it because when the user uses `unneeded_concatenation_linter()`, it can be confusing to see that the error message refers to `unnecessary_concatenation_linter()`: 

``` r
library(lintr)

unneeded_concatenation_linter(allow_single_expression = 1.0)
#> Warning: Linter unneeded_concatenation_linter was deprecated in lintr version
#> 3.1.0. Use unnecessary_concatenation_linter instead.
#> Error in unnecessary_concatenation_linter(allow_single_expression = allow_single_expression): is.logical(allow_single_expression) is not TRUE
```

<sup>Created on 2022-12-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: ok fair enough

MichaelChirico (approved):

PR #1863, opened 2022-12-19
Add a list of accepted "misspelled" words

IndrajeetPatil:


MichaelChirico:
at a glance this seems like more of a maintenance burden than it's worth.

Why have this in CI, as opposed to part of the release procedure?

IndrajeetPatil:
I have removed the spelling test to not gain an additional soft dependency.

The idea here is to keep a list of words which we know to be acceptable words, so that we don't need to go through the list every single time we make a release.

Currently, on main, devtools::spell_check() generates a huge list we need to go through to see which are truly misspelled words.

If we add a list like the one in this PR, only the newly introduced, misspelled words will be flagged. This reduces maintainance burden for us.

MichaelChirico:
> we don't need to go through the list every single time we make a release.

right but there will presumably be some churn, every time new documentation/vignettes are added there's likely to be some new words to add.

happy to try it out for a few releases to see just how big that delta is each time though

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+cyclomatic
+dcf
+de
+debian
+dplyr
+dragosmg
+eg
+envvar
+eval
+favour
```
MichaelChirico, line 62: same, we should be consistent

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+XPath
+XPaths
+YAML
+Yihui
+addin
+addins
+alllowercase
+backports
+backticked
+behaviour
```
MichaelChirico, line 42: another

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+SARIF
+SublimeLinter
+Tidyverse
+UpperCamelCase
+Wickham
+XPath
+XPaths
+YAML
+Yihui
+addin
```
MichaelChirico, line 37: is this an accepted spelling? I would have expected add-in

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+covr
+customizable
+customizations
+cyclomatic
+dcf
+de
+debian
+dplyr
+dragosmg
+eg
```
MichaelChirico, line 59: we should use `e.g.` consistently

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+repo
+rlang
+roxygen
+sandboxing
+spilt
+src
+stopifnot
+styler
+subdir
+syntastic
```
MichaelChirico, line 102: is syntastic a word?

IndrajeetPatil: It's name of a Vim plugin: https://github.com/vim-syntastic/syntastic

MichaelChirico: does `spelling` support comments in the `WORDLIST`? it could be good to annotate cases like this so we don't forget

IndrajeetPatil: IINM, it's a debian/control file, so, I don't think so.

inst/WORDLIST
IndrajeetPatil: Yes: https://rstudio.github.io/rstudioaddins/

inst/WORDLIST
```diff
@@ -0,0 +1,122 @@
...
+pre
+programmatically
+qmd
+readably
+regexes
+repo
+rlang
+roxygen
+sandboxing
+spilt
```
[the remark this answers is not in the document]
IndrajeetPatil: This is going to be a tough one for someone like me who comes from the Commonwealth. 😅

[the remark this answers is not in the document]
MichaelChirico: Just be happy we don't need to decide about metric units 😂

TBH I'm comfortable switching to `en-UK` too, but we should be consistent (do we have any function names assuming stateside spelling?).

inst/WORDLIST
```diff
@@ -0,0 +1,118 @@
...
+io
+jwz
+knitr
+labelled
+lang
+languageserver
+linter
+linters
+lintr's
+localisation
```
MichaelChirico, line 75: one more 🇬🇧

inst/WORDLIST
```diff
@@ -0,0 +1,118 @@
...
+lowerCamelCalse
+magrittr
+michaelchirico
+mis
+nd
+nodeset
+nolint
+parameterised
```
MichaelChirico, line 83: one more 🇬🇧

inst/WORDLIST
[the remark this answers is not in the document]
IndrajeetPatil: Thankfully, no, we don't have any function names that assume US spellings.

MichaelChirico (approved):

PR #1866, opened 2022-12-20
Add new map_vec() to purrr_mappers vector

Bisaloo:


MichaelChirico:
thanks! could you add yourself to the NEWS item and add a test for this?

IndrajeetPatil (dismissed):
Oh, purrr 1.0.0 is already on CRAN. Cool!

Thanks.

MichaelChirico:
Should map2_vec and pmap_vec also be included?

Bisaloo:
> Should map2_vec and pmap_vec also be included?

Then all `map2()` and `pmap()` variants should be added, which is currently not the case. But since `mapply()` is covered, I guess it makes sense to put them as well.

MichaelChirico:
ah, nvm, I didn't read the docs enough... actually I'm surprised mapply is there, maybe it shouldn't be. the logic would be different for the >1-arg function case

Bisaloo:
But if I understand correctly, it clashes with 

https://github.com/r-lib/lintr/blob/7657984c6e97b4aa31d608b29de8200474db1925/R/unnecessary_lambda_linter.R#L34

MichaelChirico:
I guess I'd you're using mapply(function(x) x, 1:10) that is indeed a lint but I still assume the logic is off more generally. anyway, that's for a follow-up issue

Bisaloo:
On the same note, `outer()` behaves differently than the others as well

Bisaloo:
> thanks! could you add yourself to the NEWS item and add a test for this?

Should I just expand on the existing news bullet rather than adding a new one since there hasn't been a release since this linter was created?

https://github.com/r-lib/lintr/blob/67576efb1a33196ce422f58cbf5a847beafaff54/NEWS.md?plain=1#L98-L100

MichaelChirico:
yes exactly, I would just add to the bullet with your name and this PR number

MichaelChirico (approved):
thanks!

tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -80,6 +80,11 @@ test_that("purrr-style anonymous functions are also caught", {
     rex::rex("Pass foo directly as a symbol to map_int()"),
     unnecessary_lambda_linter()
   )
+  expect_lint(
+    "purrr::map_vec(x, ~foo(.x, y))",
+    rex::rex("Pass foo directly as a symbol to map_vec()"),
+    unnecessary_lambda_linter()
```
DavisVaughan, line 86: @MichaelChirico is this recommending you use `map_vec(x, foo, y = y)` instead?

We actually recommend using the anonymous function form rather than using `...` to pass named arguments through to `foo()` now. Using the anonymous function form makes it a little easier to see which arguments belong to which function (i.e. does `y` belong to `map_vec()` or `foo()`?), and tends to give slightly better error messages. See the end of the new documentation of `.f` at https://purrr.tidyverse.org/reference/map.html for more.

MichaelChirico: Thanks Davis! Given that recommendation I think we should exclude purrr mappers by default; filed #1871 to handle that before release.

PR #1869, opened 2022-12-21
Add `colsums_rowsums_linter()`

### LINKED ISSUE #1764: New `lint_colsums_rowsums()` linter
I very commonly see the following patterns:

```r
apply(x, 1, sum)
apply(x, 2, sum)
```

It would be great in terms of performance and readability to have a linter recommending the use of the `rowSums()` and `colSums()` base functions instead.

Bisaloo:
Fix #1764 

This is still a work of progress but it can already receive a first round of review if you wish while I'm addressing the remaining TODO items:

To do:
- [ ] Test lint messages more strictly
- [ ] Add messages for simple cases when we have a matrix, or when `l2 = length(dim(x))`
- [ ] Add NEWS item
- [ ] Add linter tags

Open questions for input:
- [ ] I ended up implemented a linter for both `colSums()`/`rowSums()` and `colMeans()`/`rowMeans()`. The function should maybe have a more generic name but I'm not sure what
- [ ] The case when `MARGIN` is passed as an integer is not covered but I believe this is not correct syntax anyways (should there be a linter for this?)
  ```r
  apply(x, 1L, mean)
  ```
- [ ] Detection of `x`, `MARGIN` and `FUN` is purely based on position. I don't believe it should be common to see deviation from this but I'm open to counterpoints

### DIFF
--- a/R/colsums_rowsums_linter.R
+++ b/R/colsums_rowsums_linter.R
@@ -1,0 +1,78 @@
+#' Require usage of `colSums(x)` or `rowSums(x)` over `apply(x, ., sum)`
+#'
+#' [colSums()] and [rowSums()] are clearer and more performant alternatives to
+#' `apply(x, 2, sum)` and `apply(x, 1, sum)` respectively in the case of 2D
+#' arrays, or matrices
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "apply(x, 1, sum)",
+#'   linters = colsums_rowsums_linter()
+#' )
+#'
+#' lint(
+#'   text = "apply(x, 2, sum)",
+#'   linters = colsums_rowsums_linter()
+#' )
+#'
+#' lint(
+#'   text = "apply(x, 2, sum, na.rm = TRUE)",
+#'   linters = colsums_rowsums_linter()
+#' )
+#'
+#' @evalRd rd_tags("expect_named_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+colsums_rowsums_linter <- function() {
+
+  # mean() and sum() have very different signatures so we treat them separately.
+  # sum() takes values to sum over via ..., has just one extra argument and is not a generic
+  # mean() is a generic, takes values to average via a single argument, and can have extra arguments
+  sums_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'apply']
+    /parent::expr
+    /following-sibling::expr[
+      SYMBOL[text() = 'sum']
+      and (position() = 3)
+    ]
+  "
+
+  means_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'apply']
+    /parent::expr
+    /following-sibling::expr[
+      SYMBOL[text() = 'mean']
+      and (position() = 3)
+    ]
+  "
+
+  Linter(function(source_expression) {
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    bad_sums_expr <- xml2::xml_find_all(xml, sums_xpath)
+    sums_lint_message <- sprintf("colSums()")
+    sums_lint <- xml_nodes_to_lints(
+      bad_sums_expr,
+      source_expression = source_expression,
+      sums_lint_message,
+      type = "warning"
+    )
+
+    bad_means_expr <- xml2::xml_find_all(xml, means_xpath)
+    means_lint_message <- sprintf("colMeans()")
+    means_lint <- xml_nodes_to_lints(
+      bad_means_expr,
+      source_expression = source_expression,
+      means_lint_message,
+      type = "warning"
+    )
+
+    c(sums_lint, means_lint)
+
+  })
+}
--- a/R/colsums_rowsums_linter.R
+++ b/R/colsums_rowsums_linter.R
@@ -54,6 +54,13 @@ colsums_rowsums_linter <- function() {
     ]
   "
 
+  xpath <- glue::glue("{sums_xpath} | {means_xpath}")
+
+  # This doesn't handle the case when MARGIN and FUN are named and in a different position
+  # but this should be relatively rate
+  margin_xpath <- "expr[position() = 3]"
+  fun_xpath <- "expr[position() = 4]"
+
   Linter(function(source_expression) {
     if (!is_lint_level(source_expression, "expression")) {
       return(list())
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -98,6 +98,8 @@
 
 ### New linters
 
+* `colsums_rowsums_linter()` recommends use of dedicated `rowSums()`, `colSums()`, `colMeans()`, `rowMeans()` over `apply(., MARGIN, sum)` or `apply(., MARGIN, mean)`. The recommended alternative is much more efficient and more readable (#1869, @Bisaloo) .
```
AshesITR, line 101: There's extra space before full stop.

AshesITR:
Regarding passing `MARGIN` as integer: we actually encourage this with `implicit_integer_linter()`.

Bisaloo:
> Regarding passing `MARGIN` as integer: we actually encourage this with `implicit_integer_linter()`.

Good point! I'm not sure to address that.

Since I'm doing operations on l1 and l2 to generate the custom error message, I can only think of solutions like `eval(parse(text = margin))`, which probably goes beyond the scope of static analysis.

AshesITR:
You an use `xml_text()` instead of `xml_integer()` and then `as.integer(re_substitutes(l1, "L$", ""))` iinm.

Bisaloo:
Done. The `try(eval(parse(text = margin)))` would also support cases like `seq(1, 3)`, `seq_len(3)`, `c(1,2,3)`, etc. But I'm afraid we're stepping outside of the scope of lintr.

R/colsums_rowsums_linter.R
```diff
@@ -0,0 +1,156 @@
...
+    # See #1764 for details about various cases. In short:
+    # - If apply(., 1:l2, sum) -> rowSums(., dims = l2)
+    # - If apply(., l1:l2, sum) -> rowSums(colSums(., dims = l1 - 1), dims = l2 - l1 + 1)
+    # - This last case can be simplified to a simple colSums() call if l2 = length(dim(.))
+    # - dims argument can be dropped if equals to 1. This notably is the case for matrices
+    if (is.null(l2)) {
+      l2 <- l1
+    }
+
+    l1 <- suppressWarnings(as.integer(re_substitutes(l1, "L$", "")))
```
AshesITR, line 89: Which warnings are suppressed here?

R/colsums_rowsums_linter.R
```diff
@@ -0,0 +1,156 @@
...
+    fun <- xml2::xml_text(xml2::xml_find_all(bad_expr, fun_xpath))
+    fun <- tools::toTitleCase(fun)
+
+    margin <- xml2::xml_find_all(bad_expr, margin_xpath)
+
+    narm_val <- xml2::xml_text(
+      xml2::xml_find_first(bad_expr, "SYMBOL_SUB[text() = 'na.rm']/following-sibling::expr")
+    )
+
+    recos <- lapply(seq_along(bad_expr), function(i) {
```
AshesITR, line 143: This could be a `mapply()` call instead.

tests/testthat/test-colsums_rowsums_linter.R
```diff
@@ -0,0 +1,96 @@
...
+})
+
+test_that("colsums_rowsums_linter works even when MARGIN can't be parsed", {
+
+  linter <- colsums_rowsums_linter()
+  lint_message <- rex::rex("l1")
+
+  expect_lint("apply(x, seq(2, 4), sum)", lint_message, linter)
+
+  expect_lint("apply(x, m, sum)", lint_message, linter)
```
AshesITR, line 57: If `m` is a non-contiguous vector (e.g. `m <- c(1, 4)`), this would be a false positive, right?

tests/testthat/test-colsums_rowsums_linter.R
```diff
@@ -0,0 +1,96 @@
...
+
+  expect_lint("apply(x, 1, sum, na.rm = TRUE)", lint_message, linter)
+
+  expect_lint("apply(x, 2, sum, na.rm = TRUE)", lint_message, linter)
+
+  expect_lint("apply(x, 1, mean, na.rm = TRUE)", lint_message, linter)
+
+  expect_lint("apply(x, 2, mean, na.rm = TRUE)", lint_message, linter)
+
+  lint_message <- rex::rex(not("na.rm"))
```
AshesITR, line 73: What regex does this generate? I'm not sure this will test what you want it to test.

R/colsums_rowsums_linter.R
```diff
@@ -0,0 +1,156 @@
...
+
+  xpath <- glue::glue("{sums_xpath} | {means_xpath}")
+
+  # This doesn't handle the case when MARGIN and FUN are named and in a different position
+  # but this should be relatively rate
+  var_xpath  <- "expr[position() = 2]"
+  margin_xpath <- "expr[position() = 3]"
+  fun_xpath <- "expr[position() = 4]"
+
+  craft_msg <- function(var, margin, fun, narm_val) {
```
AshesITR, line 70: Might be easier to read if you move this out of the linter function.

R/colsums_rowsums_linter.R
```diff
@@ -0,0 +1,156 @@
...
+      craft_msg(var[i], margin[i], fun[i], narm_val[i])
+    })
+
+    xml_nodes_to_lints(
+      bad_expr,
+      source_expression = source_expression,
+      lint_message = sprintf("Use %1$s rather than %2$s", recos, get_r_string(bad_expr)),
+      type = "warning"
+    )
+
```
AshesITR, line 153: These line breaks are superfluous.

AshesITR (commented):
Thanks, we're almost there.

tests/testthat/test-colsums_rowsums_linter.R
Bisaloo: Yes, should I limit it to cases where MARGIN is numeric?

R/colsums_rowsums_linter.R
IndrajeetPatil: Don't we prefer `Map()` over `mapply()`?

https://github.com/r-lib/lintr/blob/08098ef77adb1eccdf18f8b44a7053f36297df28/R/zzz.R#L66

tests/testthat/test-colsums_rowsums_linter.R
AshesITR: Since you already have the `l1:l2` logic in your code, I'd refactor the XPath to allow NUM_CONSTANT and OP-COLON with two NUM_CONSTANT children.

R/colsums_rowsums_linter.R
AshesITR: Good catch. Still should be `Map()` then.

MichaelChirico:
> The function should maybe have a more generic name but I'm not sure what

`matrix_moments_linter()` ?

AshesITR:
Future expansion of this linter would probably go towards adding more functions to the list of detected things (the `matrixStats` package has a few high-performance functions for rowwise / columnwise operations, e.g.)

Maybe `matrix_apply_linter()`?

R/colsums_rowsums_linter.R
Bisaloo: I've added a comment inline: [r-lib/lintr@`67087fc` (#1869)](https://github.com/r-lib/lintr/pull/1869/commits/67087fc4a7c6dbc729a87cb5427b89559480999e#diff-a49767522a81382929d9458e0751ea894daf500e64c8b958d57e5bb304718a14R119)

Bisaloo:
I like `matrix_apply_linter()` since "moments" might be unclear to non-statisticians.

Bisaloo:
Hi @AshesITR @IndrajeetPatil @MichaelChirico, is there anything left for me to do in this PR or are features freezed until the new release gets out? https://github.com/r-lib/lintr/issues/1476

AshesITR (changes requested):
Sorry for the delay in reviewing, I haven't had much spare time lately.
@MichaelChirico do we want to leave the name as `matrix_apply_linter()`?
If so, LGTM after fixing the news.

MichaelChirico:
> @MichaelChirico do we want to leave the name as matrix_apply_linter()?

works for me! Thanks @Bisaloo for all the work on this & for your patience!

AshesITR (approved):
Thank you for your effort!

PR #1872, opened 2022-12-21
Separate out tests for allowed assignments with braces

IndrajeetPatil:
For `implicit_assignment_linter()`

### DIFF
--- a/tests/testthat/test-implicit_assignment_linter.R
+++ b/tests/testthat/test-implicit_assignment_linter.R
@@ -162,23 +162,18 @@ test_that("implicit_assignment_linter respects except argument", {
   )
 })
 
+test_that("implicit_assignment_linter skips allowed usages with braces", {
+  linter <- implicit_assignment_linter(except = character(0L))
+
+  expect_lint("output <- capture.output({ x <- f() })", NULL, linter)
+  expect_lint("quote({ a <- 1L })", NULL, linter)
+  expect_lint("bquote({ a <- 1L })", NULL, linter)
+  expect_lint("expression({ a <- 1L })", NULL, linter)
+  expect_lint("local({ a <- 1L })", NULL, linter)
+})
+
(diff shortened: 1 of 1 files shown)


tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -162,23 +162,18 @@ test_that("implicit_assignment_linter respects except argument", {
   )
 })
 
-test_that("implicit_assignment_linter makes exceptions for functions that capture side-effects", {
-  linter <- implicit_assignment_linter()
+test_that("implicit_assignment_linter skips allowed usages with braces", {
+  linter <- implicit_assignment_linter(except = character(0L))
```
IndrajeetPatil, line 166: Note that no exceptions are included here.

tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -162,23 +162,18 @@ test_that("implicit_assignment_linter respects except argument", {
...
 
-  # base
-  expect_lint("output <- capture.output(x <- f())", NULL, linter)
-  expect_lint("quote(a <- 1L)", NULL, linter)
-  expect_lint("bquote(a <- 1L)", NULL, linter)
-  expect_lint("expression(a <- 1L)", NULL, linter)
+  expect_lint("output <- capture.output({ x <- f() })", NULL, linter)
```
MichaelChirico, line 168: looks like these are all side-effect functions... my recollection (CMIIW) was we wanted to allow `<-` within braces no matter the parent function.

If that's the case, could you make some of these tests use a generic function like `foo({ a <- 1L })`?

It may also be preferable to keep the code aligned to the rest of the default linters, which means using

```r
foo({
  a <- 1L
})
```

because of `brace_linter()`.

MichaelChirico (approved):

PR #1873, opened 2022-12-22
Ignore `{purrr}` on `R < 3.5`

IndrajeetPatil:


.github/workflows/R-CMD-check.yaml
```diff
@@ -73,6 +73,8 @@ jobs:
           extra-packages: |
             any::rcmdcheck
             bookdown=?ignore-before-r=3.5.0
+            patrick=?ignore-before-r=3.5.0
```
IndrajeetPatil, line 76: Because `{patrick}` also depends on `{purrr}`:           

```r
> pak::pkg_deps_explain("lintr", "purrr", dependencies = "all")
✔ Updated metadata database: 5.10 MB in 12 files.                         
✔ Updating metadata database ... done                                     
lintr -> patrick -> purrr                                                  
lintr -> pkgdown -> purrr
```

AshesITR: Cool I didn't know of this feature before 👍🏻

AshesITR (approved):

MichaelChirico:
I am missing the context here -- why was this done?

Asking b/c our 3.4 job is broken b/c we use `loadNamespace("patrick")` in the tests. We can fix that with `requireNamespace()` instead of `loadNamespace()`, but I'm trying to figure out why we exclude it to begin with (since {purrr}'s listed R dependency is R>=3.4)

AshesITR:
@MichaelChirico https://github.com/tidyverse/purrr/issues/1025

Issue #580, opened 2020-11-23
How should the `exclusions = ...` argument work?


AshesITR:
Follow up to #539.

How should `lint`, `lint_dir` and `lint_package` handle `exclusions` specified through the manual argument?

russHyde:
I flagged this in #539. When working on a related issue (ensuring that files could be excluded by relative-paths; https://github.com/russHyde/lintr/pull/2), @joaopmatias noted that the exclusions-handling is different for files defined in the .lintr, compared to those defined in the 'exclusions' argument.

The following integration test was added to `test-lint_package.R`:
```
test_that(                                                                                                                                     
   "`lint_package` does not depend on path to pkg - with exclusions argument", {                                                               
     # The test checks the results of lint_package when the excluded regions are                                                               
     # given as an argument of the function and are specified using absolute and                                                               
     # relative paths                                                                                                                          
                                                                                                                                               
      pkg_path <- file.path("dummy_packages", "assignmentLinter")                                                                              
                                                                                                                                               
      # Create list of exclusions relative to the whole of `abc.R` and the first                                                               
     # line of `jkl.R`                                                                                                                         
     exclusions <- list(normalizePath(file.path(pkg_path, 'R/abc.R')),                                                                         
                        'R/jkl.R' = 1)                                                                                                         
                                                                                                                                               
      expected_lines <- c("mno = 789")                                                                                                         
     lints_from_outside <- lint_package(                                                                                                       
       pkg_path, linters = list(assignment_linter), exclusions = exclusions                                                                    
     )                                                                                                                                         
                                                                                                                                               
      expect_equal(                                                                                                                            
       as.data.frame(lints_from_outside)[["line"]], expected_lines                                                                             
     )                                                                                                                                         
 })
```

We added a full-path and a relative-path to the exclusions. A single lint was expected in the non-excluded lines of the file that was excluded using relative paths.

I realise now that we should have had  a `parse_settings = FALSE` argument in the call to `lint_package`; with this modification, the exclusions appear to be handled properly.

Happy to have this issue closed if you agree @AshesITR
----

By 'properly', according to the docs, exclusions should be specified relative to an assumed project-root. They can also be specified as full-paths (though this doesn't appear mentioned in the docs). It doesn't specify what will happen when no project-root can be found (should exclusions be relative to the directory passed to 'lint_dir'?) though that's probably a minor issue

AshesITR:
[...]

PR #1877, opened 2022-12-29
Linter wrapper to exclude execution on pattern files

galachad:
This linter wrapper allow exclude linters from execute on the specific files.

Useful for package development, example:
```
linters_with_defaults(
  line_length_linter(100),
  object_usage_linter = NULL,
  cyclocomp_linter(complexity_limit = 20),
  undesirable_function_linter = exclude_wrapper_linter(
    undesirable_function_linter(),
    exclusions = c("inst/", "vignettes/")
  )
)
```

Currentlyit is no possible to run `lintr::lint_package` or `lintr::lint_dir` to apply rules for specific files/directories.


MichaelChirico:
Would you mind filing a bug, including a bit more motivation for this PR? I think we prefer having a bug first for FRs that might require some design review.

galachad:
From any existing issue this pr can be linked to #580 

The main motivation was to use `lintr::lint_package` and apply rules base on the `.lintr` config. Easy to maintain, not needed to modify any existing ci workflow and  consistent between local and ci run.  

Examples:
`inst` directory contains the R templates scripts or shiny app that will not follow the some rules.
The `exclude_wrapper_linter` allow to exclude  `undesirable_function_linter`.

`vignettes` are rmarkdown file and in many project they presenting the results of script to the end users.
Simple example: `library`

The other rules can be apply to other directories.

I happy to discuss the concept if you think the logic behind that need to be improved.

AshesITR:
Would this be solved if you could create separate `.lintr` files in the directories you want linted?

galachad:
Yes it's also possible to solve this way but the settings need to be rewritten and it's big refactor task. The `lint_dir` is evaluate by the `lint_package` but the `lint_dir` itself gets path recursive with `dir(recursive = TRUE)` (do not loading settings again).

I think the wrapper solution is not bad idea.  Maybe one more `include_wrapper `should be added for inclusion criteria. (what do you think?). It also allow to play with regex for some more custom solution. 

The alternative for this is running custom Rscript for linting. Simple wrapper will apply rules with config.

AshesITR:
It just occurred to me that you should be able to achieve your use case using the built in exclusion system.

```
exclude: list(
    "inst" = "undesirable_function_linter"
  )
```

should do the trick.

galachad:
Are you proposing new solution? This is first time I see this construction in lintr. 

Would be nice to have anything that allow define rules in better way but something like that is not implemented yet.

You can check this with:

```
lintr:::normalize_exclusions(
    list("inst" = "undesirable_function_linter" ),
    root = getwd(),
    pattern = rex::rex(".", one_of("Rr"), or("", "html", "md", "nw", "rst", "tex", "txt"), end),
)
```

`Error: Full line exclusions must be numeric or integer vectors. items: 1 are not!`

There are also tests for normalize_exclusions for check that behaviour.

I'm also not sure if the exclusion handling any values from exclusion as most of the elements was handle by `names2()` of normalized exclusions.

As I mentioned before it's possible to implement better solution but any of them need to refactor the code and logic behind that. 

The two inclusion/exclusion wrapper will cover most of the use cases that any project can have.

galachad:
Hi @AshesITR 

I'm sorry for my mistake  I read the code and documentation more carefully and I notice that it can be defined as numeric linters.

```
list(
  inst = list(undesirable_function_linter)
)
```

so I'm closing it.

PR #1878, opened 2022-12-29
`implicit_assignment_linter()` doesn't produce false positives with walrus operator

### LINKED ISSUE #1827: False positive in `implicit_assignment_linter()` for code with pipe
``` r
library(lintr)

lint(
  text = "mutate(data, a := b)",
)


lint(
  text = "data %>% mutate(a := b)",
)
#> <text>:1:19: warning: [implicit_assignment_linter] Avoid implicit assignments in function calls. For example, instead of `if (x <- 1L) { ... }`, write `x <- 1L; if (x) { ... }`.
#> data %>% mutate(a := b)
#>                   ^~
```

<sup>Created on 2022-12-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
Closes #1827

### DIFF
--- a/R/implicit_assignment_linter.R
+++ b/R/implicit_assignment_linter.R
@@ -56,9 +56,11 @@ implicit_assignment_linter <- function(except = c(
     //SYMBOL_FUNCTION_CALL"
   }
 
+  # The walrus operator `:=` is also `LEFT_ASSIGN`, but is not a relevant operator
+  # to be considered for the present linter.
   assignments <- c(
-    "LEFT_ASSIGN", # e.g. mean(x <- 1:4)
-    "RIGHT_ASSIGN" # e.g. mean(1:4 -> x)
+    "LEFT_ASSIGN[text() = '<-']", # e.g. mean(x <- 1:4)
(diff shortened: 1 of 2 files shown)


R/implicit_assignment_linter.R
```diff
@@ -56,9 +56,11 @@ implicit_assignment_linter <- function(except = c(
     //SYMBOL_FUNCTION_CALL"
   }
 
+  # The walrus operator `:=` is also `LEFT_ASSIGN`, but is not a relevant operator
+  # to be considered for the present linter.
   assignments <- c(
-    "LEFT_ASSIGN", # e.g. mean(x <- 1:4)
-    "RIGHT_ASSIGN" # e.g. mean(1:4 -> x)
+    "LEFT_ASSIGN[text() = '<-']", # e.g. mean(x <- 1:4)
```
MichaelChirico, line 62: this will exclude `<<-` and `->>`, though, would `text() != ':='` be better?

IndrajeetPatil: Good catch!

I forgot about super-assignments because there were no tests for them. I have made the necessary change and also added tests for the super-assignment operators.

MichaelChirico (approved):

PR #1879, opened 2022-12-31
Avoid implicit assignments in expectations

### LINKED ISSUE #1820: Avoid unnecessary implicit assignments in our own test suite
Still not 100% sure this necessitates excepting `expect_warning()`

e.g. even in case the value needs to be re-used in a subsequent test, you can do
```r
# bad
expect_warning(prepared_thing <- my_setup_with_warning(), "setup warning")
expect_equal(funny_function(prepared_thing), funny_value) |>
  expect_warning("funny warning")

# good
my_setup_with_warning() |>
  funny_function() |>
  expect_equal(funny_value) |>
  expect_warning("setup warning") |>
  expect_warning("funny warning")
```

This is also how we structure our tests that expect warnings.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/issues/1777#issuecomment-1343414735_

IndrajeetPatil:
Closes #1820


IndrajeetPatil:
@AshesITR Before I continue with this, just wanted to double-check with you if this is what you had in mind for fixing #1820.

IndrajeetPatil:
I haven't touched the tests for deprecated functions, which also contains some implicit assignment lints.

TBH, I think we should `nolint` all test files for deprecated functions; we are going to remove them at some point anyway.

IndrajeetPatil:
The new warning appearing in tests is unrelated to the current PR. I have created an issue to track this: #1905.

IndrajeetPatil:
@AshesITR This is ready for a review, btw. Tagging just in case you missed the original review request during the holidays.

IndrajeetPatil:
> Looks good. Just wondering if we should commit to the brace-syntax from magrittr or grab the result in an extra `supressWarnings()` for the `xml_nodes_to_lints()` tests.
> 
> The former would not be compatible with the native pipe once we drop support for R versions without it.

Bumping the minimum needed R version to 4.1 should be many years away, so that's not too worrisome.

That said, the brace syntax is not too straightforward to reason about, and so maybe it's better to avoid it. Thus, I have removed the brace-syntax, and have instead used temp variables for the tests in question.

tests/testthat/test-backport_linter.R
```diff
@@ -64,3 +58,13 @@ test_that("backport_linter detects backwards-incompatibility", {
     backport_linter("3.0.0", except = c("numToBits", "R_user_dir"))
   )
 })
+
+test_that("backport_linter generates expected warnings", {
+  skip_if_not_installed("magrittr")
```
MichaelChirico, line 63: we need `library(magrittr)` here though, no? since we're assuming magrittr is not attached, `%>%` won't be available.

options:

 1. just nest the calls
 2. use `` `%>%` <- magrittr::`%>%` ``

IndrajeetPatil: Hmm, why doesn't this test fail? 
Is it because `{testthat}` itself relies on `{magrittr}`, and so the latter namespace is always loaded while running tests?

IndrajeetPatil: I am no longer using magrittr pipes here.

tests/testthat/test-cache.R
```diff
@@ -92,10 +92,19 @@ test_that("load_cache returns an empty environment if reading cache file fails",
   cache_f1 <- file.path(d1, fhash(f1))
   writeLines(character(), cache_f1)
 
-  expect_warning(e2 <- lintr:::load_cache(file = f1, path = d1), "Could not load cache file")
+  expect_warning(
+    {
```
MichaelChirico, line 96: I found myself ready to suggest nesting the expectations instead & then remembered that's basically what we just undid by removing `magrittr` :)

Someday we'll depend on R>4.1 & then we can use `|>`....

In the meantime, there may be some cases where 2-nesting looks fine:

```r
expect_warning(
  expect_identical(ls(lintr:::load_cache(file = f1, path = d1), character()),
  "Could not load cache file",
  fixed = TRUE
)
```

up to you whether / which cases you agree

MichaelChirico (approved):

tests/testthat/test-cache.R
IndrajeetPatil: TBH, I personally prefer to keep each expectation separate. Nested expectations are much harder to reason about.

Issue #1874, opened 2022-12-23
Linter to recommend new expectations to check for absence of exceptions?


IndrajeetPatil:
`{testthat}` has recently introduced new [expectations](https://testthat.r-lib.org/reference/expect_no_error.html) that assert the absence of an error, warning, or message. 

These expectations are (IMO: https://github.com/r-lib/testthat/issues/1642) far more readable than the previous idiom of setting `regexp = NA`.

Should we introduce a new linter that recommends using these new expectations?

```r
# actual (forces reader to figure out what the second argument is and what setting it to NA entails)
expect_message({ ... }, NA)
expect_warning({ ... }, NA)
expect_error({ ... }, NA)
expect_condition({ ... }, NA)

# suggested
expect_no_message({ ... })
expect_no_warning({ ... })
expect_no_error({ ... })
expect_no_condition({ ... })
```

P.S. A fly in the ointment is that the lifecycle for these new expectations is `experimental`. So maybe we should wait until it's stable?

AshesITR:
Fine with me. Probably should wait for the lifecycle to upgrade.

IndrajeetPatil:
In testthat 3.2.0 release, these functions were marked as stable, so we could work on this issue:

<img width="1366" alt="Screenshot 2024-05-24 at 10 10 10" src="https://github.com/r-lib/lintr/assets/11330453/3eb5ec23-ae42-4428-8194-f4ac01ab7c62">

MichaelChirico:
we can do expect_no_linter() to cover these and put expect_no_lint()?

that's very close in name to expect_not_linter(), my first instinct is _not_ to combine them...

AshesITR:
I was thinking `expect_condition_linter()` or `expect_signal_linter()`, but neither of those incorporate `expect_lint()`...

AshesITR:
WDYT about `expect_no_condition_linter()`?

MichaelChirico:
I think it's good enough. a lint is a "condition", right? :smile:

AshesITR:
SGTM. I'm starting to grasp why Pylint uses numbers. for lints 😂

F-Noelle:
expect_absence_linter, perhaps? given that this is how they seem to be described collectively.

Also expect_no_lint, expect_lint do have checks as the secondary parameter which is also equal to NULL and not NA.

Not saying that this linter shouldn't handle expect_no_lint with the other absence functions, but it is distinct separate logic that needs to be handled as well.

PR #1880, opened 2022-12-31
Use new expectations to check for absence of exceptions

IndrajeetPatil:
Related to #1874


DESCRIPTION
```diff
@@ -44,7 +44,7 @@ Suggests:
     rlang,
     rmarkdown,
     rstudioapi (>= 0.2),
-    testthat (>= 3.0.0),
+    testthat (>= 3.1.5),
```
IndrajeetPatil, line 47: `expect_no_warning()` was introduced in this version.

MichaelChirico: I don't have any hard rule here, but something like "wait at least 6 months after CRAN release before depending on that version" seems reasonable, absent any pressing reason to have such a hot-off-the-press dependency.

So let's table this PR until the next release?

IndrajeetPatil: Fine by me. 

`testthat 3.1.5` was released on 2022-10-08. So we should be hitting that 6-month mark pretty soon.

MichaelChirico: good point! in fact at this point by the time we land a release we'll be there I reckon.

MichaelChirico (approved):

IndrajeetPatil:
I have already cleaned the newly found lint in https://github.com/r-lib/lintr/pull/1879

PR #1870, opened 2022-12-21
Error on empty character vector args in undesirable linters

IndrajeetPatil:
Closes #1867

PR #1895, opened 2023-01-08
Automatically add .lintr to .Rbuildignore by the `use_lintr` function

### LINKED ISSUE #1805: `use_lintr()` should add `.lintr` to `.Rbuildignore`

eitsupi:
Close #1805

It would be useful if .lintr files were automatically added to .Rbuildignore when creating R packages.


eitsupi:
Is purrr causing the R 3.4 test to fail, but probably not a problem? https://github.com/r-lib/lintr/pull/1870#issuecomment-1361664719

AshesITR:
The R 3.4 build error can be fixed by ignoring packages that import purrr for R < 3.5.

However, not a big fan of introducing a dependency on usethis and fs just for this feature.
Can you add the line to `.Rbuildignore` using base R?

R/use_lintr.R
```diff
@@ -43,5 +43,33 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
 }
+
+add_build_ignore <- function(path) {
+  path_build_ignore <- ".Rbuildignore"
+  existing_lines <- readLines(path_build_ignore, encoding = "UTF-8", warn = FALSE)
+
+  escape_path <- gsub("^\\./", "", path)
+  escape_path <- gsub("\\.", "\\\\.", escape_path)
+  escape_path <- gsub("/$", "", escape_path)
+  escape_path <- paste0("^", escape_path, "$")
```
AshesITR, line 65: This could be `rex::rex(start, rex::re_substitutes(path, list(start, "./") %or% list("/", end), ""), end)` and it would safely escape any chars in path.

R/use_lintr.R
```diff
@@ -43,5 +43,32 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+      add_build_ignore(config_file),
+      silent = TRUE
+    )
+  }
+
   invisible(config_file)
 }
+
+add_build_ignore <- function(path) {
+  path_build_ignore <- ".Rbuildignore"
```
AshesITR, line 59: did you mean `file.path(dirname(config_file), ".Rbuildignore")`?

eitsupi:
It doesn't seem to pass the test but locally it does so I don't know what is causing it.

R/use_lintr.R
```diff
@@ -43,5 +43,29 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
     )
   )
   write.dcf(the_config, config_file, width = Inf)
+
+  # If this is an R package and if available, add .lintr file to .Rbuildignore
+  if (file.exists("DESCRIPTION") && file.exists(".Rbuildignore")) {
```
AshesITR, line 48: This test looks in the current working directory, not where the config was placed.

eitsupi: Since there is no guarantee that `DESCRIPTION` and `.lintr` will be placed in the same directory, I believe that the current working directory can should be used.

eitsupi:
> Can you add the line to `.Rbuildignore` using base R?

Sorry I could not get this to work without usethis.

Bisaloo:
Noting that this PR will fix https://github.com/r-lib/lintr/issues/1805.

eitsupi:
> Noting that this PR will fix #1805.

Oh, I didn't realize it was already requested. Thank you for pointing that out!

MichaelChirico:
I agree about not adding either dependency.

PR #1899, opened 2023-01-12
Document alternatives to nested `ifelse()` 

IndrajeetPatil:
Based on the feedback I have received from repos where I implemented lint workflow, I have come to realize that the short description we have here that mentions possible alternatives is not enough. It might be worth including some examples in the documentation itself.

I am especially not clear on what the proposed `{base}`-only alternative is supposed to look like. I've mentioned one of the many possibilities I could think of, but happy to change it based on feedback.


Bisaloo:
Is there any reason to not use `switch()`?

```r
switch(
  x,
  "a" = 2L,
  "b" = 3L,
  1L
)
```

IndrajeetPatil:
With `switch()`, you are restricted to a single expression, while the look-up table can deal with multiple expressions.

You _can_ also do that with `switch()`, but the solutions tend to be hacky. 

For example, covert booleans to strings and then use `switch()`:

```r
switch(paste(stability[x] == "sufficient", convergence[x] == "converged"),
  "TRUE TRUE" = paste0(
    text, " and the indices are reliable (ESS = ",
    insight::format_value(diagnostic$ESS, digits = 0), ")"
  ),
  "TRUE FALSE" = paste0(
    text, " even though the indices appear as reliable (ESS = ",
    insight::format_value(diagnostic$ESS, digits = 0), ")"
  ),
  "FALSE TRUE" = paste0(
    text, " but the indices are unreliable (ESS = ",
    insight::format_value(diagnostic$ESS, digits = 0), ")"
  ),
  "FALSE FALSE" = paste0(
    text, " and the indices are unreliable (ESS = ",
    insight::format_value(diagnostic$ESS, digits = 0), ")"
  )
)
```

A look-up table or a data frame provides the most general solution because you can easily use multiple expressions.

```r
dplyr::filter(df, stability == "sufficient", convergence == "converged")
```

Does that make sense?

IndrajeetPatil:
I think we can mention both alternative `switch()` and look-up table, and qualify that with generalization available for each option.

MichaelChirico:
My own experience is the difficulty of making the behavior work w.r.t missing inputs. It can be hard to keep track of which inputs being missing leads to missing output, and converting that to case expressions. Compounded by the behavior w.r.t. missingness is not often well-tested so it's easy to make a breaking change by accident.

R/nested_ifelse_linter.R
```diff
@@ -9,9 +9,52 @@
...
+#'
+#'     ```r
+#'     data.table::fcase(
+#'       x == "a", 2L,
+#'       x == "b", 3L,
+#'       default = 1L
+#'     )
+#'     ```
+#'
+#'   - Use `dplyr::case_when()`
```
AshesITR, line 36: This particular case should actually be handled by `dplyr::recode(x, "a" = 2L,  "b" = 3L, .default = 1L)`

IndrajeetPatil: `recode()` has been superseded in favour of `case_match()`: https://dplyr.tidyverse.org/reference/recode.html

AshesITR: Oh, thanks for the info. Then with `case_match()` accordingly.

R/nested_ifelse_linter.R
```diff
@@ -9,9 +9,53 @@
...
```
AshesITR, line 36: NB this is mismatched to the code.

IndrajeetPatil: Good catch! Thanks. Fixed it.

AshesITR (approved):

PR #1901, opened 2023-01-16
fix: update URLs to GHA workflow examples

dpprdan:
This just adds/fixes two links to the r-lib/actions lintr workflows. 

Does lintr-bot actually work on GHA? I haven't seen it in a long while.


IndrajeetPatil (approved):
Thanks!

IndrajeetPatil:
> Does lintr-bot actually work on GHA? I haven't seen it in a long while (i.e. I don't remember seeing it with GHA, but only when we were still using Travis).

No. For more, see: https://github.com/r-lib/lintr/pull/1558

PR #1902, opened 2023-01-16
remove mentions of lintr-bot

dpprdan:
I gather from #1558 that lintr-bot is retired, so its mentions should be removed.

Re the Travis section in the vignette. {[HIBPwnd](https://github.com/jumpingrivers/HIBPwned/)} has been archived, so that is not a good example anymore. Do you have another one? Or should we just remove the whole Travis section since it is probably not used much anymore anyway?

### DIFF
--- a/vignettes/continuous-integration.Rmd
+++ b/vignettes/continuous-integration.Rmd
@@ -67,8 +67,7 @@ after_success:
 ```
 
 Live example of a package using this setup:
-[`hibpwned`](https://github.com/jumpingrivers/HIBPwned/blob/master/.travis.yml),
-[lintr-bot commenting on a PR](https://github.com/jumpingrivers/HIBPwned/pull/30).
+[`hibpwned`](https://github.com/jumpingrivers/HIBPwned/blob/master/.travis.yml).
 
 ## For projects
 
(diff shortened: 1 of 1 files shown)


vignettes/continuous-integration.Rmd
```diff
@@ -33,11 +33,11 @@ usethis::use_github_action("lint")
...
 This file configures all the steps required to run `lintr::lint_package()` on your package.
 
-[lintr-bot](https://github.com/lintr-bot) will then add comments to the commit or pull request with the lints found and they will also be printed as
+Comments to the commit or pull request will be printed as
 [annotations](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-status-checks#types-of-status-checks-on-github)
 along side the status check on GitHub.
-If you want to disable the commenting you can set the environment variable `LINTR_COMMENT_BOT=false`.
-Additionally, if you want the builds to produce an error instead of just a warning, you can set the environment variable `LINTR_ERROR_ON_LINT=true`.
+If you want the builds to produce an error instead of just a warning, you can set the environment variable `LINTR_ERROR_ON_LINT=true`.
+This is set by default in [r-lib/actions](https://github.com/r-lib/actions/)'s `lint.yaml`.
```
IndrajeetPatil, line 40: Maybe we should also mention [`lint-changed-files.yaml`](https://github.com/r-lib/actions/blob/v2-branch/examples/lint-changed-files.yaml) here?

dpprdan: Makes sense IMO, I just added it.

PR #1915, opened 2023-02-25
ignore symbols in formula when pinpointing undefined variable lints

### LINKED ISSUE #1914: object_usage_linter gets line number wrong in multi-NSE case
Here, note that NSE is used in both the formula and the `data=` argument, and it's throwing off the line number detection:

```r
lintr::lint("
foo <- function(x) {
  lm(
    y ~ z,
    data = x[!is.na(y)]
  )
}
", lintr::object_usage_linter())
# <text>:4:5: warning: [object_usage_linter] no visible binding for global variable 'y'
#     y ~ z,
#     ^
```

This makes it look like `codetools` is confused by the LHS of the formula, but actually what's being marked is the `data=` argument on the next line.

`codetools` itself only provides a range:

```r
codetools::checkUsage(foo)
# <anonymous>: no visible binding for global variable ‘y’ (:2-5)
```

MichaelChirico:
Closes #1914 

I'm not a fan of the `descendant::*[ancestor::*]` XPath, but I'm not sure it can be avoided.

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -101,7 +101,7 @@ object_usage_linter <- function(interpret_glue = TRUE, skip_with = TRUE) {
         fun_assignment,
         "
         descendant::SYMBOL[not(ancestor::expr[OP-TILDE])]
-        | descendant::SYMBOL_FUNCTION_CALL
+        | descendant::SYMBOL_FUNCTION_CALL[not(ancestor::expr[OP-TILDE])]
         | descendant::SPECIAL
         | descendant::LEFT_ASSIGN[text() = ':=']
         "
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -100,7 +100,7 @@ object_usage_linter <- function(interpret_glue = TRUE, skip_with = TRUE) {
       lintable_symbols <- xml2::xml_find_all(
         fun_assignment,
         "
-        descendant::SYMBOL
+        descendant::SYMBOL[not(ancestor::expr[OP-TILDE])]
         | descendant::SYMBOL_FUNCTION_CALL
         | descendant::SPECIAL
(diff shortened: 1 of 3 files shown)


R/object_usage_linter.R
```diff
@@ -100,8 +100,8 @@ object_usage_linter <- function(interpret_glue = TRUE, skip_with = TRUE) {
       lintable_symbols <- xml2::xml_find_all(
         fun_assignment,
         "
-        descendant::SYMBOL
-        | descendant::SYMBOL_FUNCTION_CALL
+        descendant::SYMBOL[not(ancestor::expr[OP-TILDE])]
+        | descendant::SYMBOL_FUNCTION_CALL[not(ancestor::expr[OP-TILDE])]
         | descendant::SPECIAL
```
AshesITR, line 105: Could this also appear in a formula?

MichaelChirico: I think any infix operator used in a formula has to be defined elsewhere, so no. trying to thing of an example...

```r
foo <- function() {
  lm(y ~ X %**% Z, data = X %**% Y)
}
```

If `checkUsage` is dinging `%**%`, it should also be dinged in the formula.

The only thing I can think of is where the operator has some special definition inside the formula, which doesn't happen in base since only `%in%` has special interpretation:

https://stat.ethz.ch/R-manual/R-devel/library/stats/html/formula.html

I guess it's conceivable for other downstreams to define special interpretation methods for formulas, but because `checkUsage()` ignores formulas entirely (IIUC), it would have to be that _both_ (1) `%**%` has a special interpretation _and_ (2) it's being used incorrectly outside of a formula.

That seems pretty out there...

MichaelChirico:
gentle ping @AshesITR

AshesITR (approved):

AshesITR (approved):

PR #1921, opened 2023-02-27
fail gracefully when missing Suggests in tests

MichaelChirico:
Instigated by failures on R 3.4 (https://github.com/r-lib/lintr/pull/1873), but actually this is a better implementation of the rules around Suggests packages anyway which tell us to fail with more aplomb when Suggests are missing.

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,10 +1,10 @@
 library(testthat)
 library(lintr)
 
-# suppress printing environment name (noisy)
-invisible({
-  loadNamespace("patrick")
-  loadNamespace("withr")
-})
-
-test_check("lintr")
+# These packages are used heavily in the suite and are thus required
+if (
+  requireNamespace("patrick", quietly = TRUE) &&
+    requireNamespace("withr", quietly = TRUE)
+) {
+  test_check("lintr")
+}


AshesITR:
Not sure I agree with not testing anything and remaining silent in case a vital test package is missing.
Do I understand correctly that the issue is patrick not supporting R 3.4 anymore?

MichaelChirico:
no, it's due to purrr (on which patrick depends) having an unstated R3.5 dependency.

Not that the package examples still run, at least.

should we go back to skip_if_not_installed for patrick?

we could also bump our R dependency

AshesITR:
Fine with both, `skip_if_not_installed()` and `R >= 3.5`.
@jimhester what would you prefer?

MichaelChirico:
Hmm let's bump the R version. Per https://www.tidyverse.org/blog/2019/04/r-version-support/, all tidyverse packages are now R>=3.5 (even if not stated explicitly), so we'll have to bump sooner or later anyway.

IndrajeetPatil:
This also makes me uneasy. I would much rather see the builds fail if any of these two soft(-ish) deps are not available, which would force us to confront whatever the issue is that is making these deps available and act accordingly.

And, yes, I agree that we should bump the minimum needed R version to `3.5`, in line with the _tidyverse_'s stated policy.

MichaelChirico:
superseded by #1922.

PR #1922, opened 2023-02-27
Bump R dependency to R3.5

MichaelChirico:
Supersedes https://github.com/r-lib/lintr/pull/1921.

The version bump is bigger than I thought, but we've only been testing on R3.4, so that feels to me like our implicit dependency is really R3.4 -- we don't know for sure that the package works on R3.2/R3.3 right? Which makes this a more gradual bump R3.4->R3.5.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,7 +21,7 @@ License: MIT + file LICENSE
 URL: https://github.com/r-lib/lintr, https://lintr.r-lib.org
 BugReports: https://github.com/r-lib/lintr/issues
 Depends:
-    R (>= 3.2)
+    R (>= 3.5)
 Imports:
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -21,7 +21,7 @@ License: MIT + file LICENSE
 URL: https://github.com/r-lib/lintr, https://lintr.r-lib.org
 BugReports: https://github.com/r-lib/lintr/issues
 Depends:
-    R (>= 3.2)
+    R (>= 3.5)
```
IndrajeetPatil, line 24: We should also make a note of this change in NEWS.

IndrajeetPatil (approved):
Thanks!

PR #1923, opened 2023-02-27
Let MBCS warning fall through the first time it's encountered

### LINKED ISSUE #1920: Strict test (warnings=errors) failing because of encoding issue
Observed failure:

```
── Error (test-get_source_expressions.R:111): Warns if encoding is misspecified ──
Error in `grepl(pat, text, perl = TRUE)`: (converted from warning) input string 3 is invalid UTF-8
Backtrace:
    ▆
 1. └─lintr::lint(filename = file, parse_settings = FALSE) at test-get_source_expressions.R:111:2
 2.   └─lintr::get_source_expressions(filename, lines)
 3.     └─lintr:::extract_r_source(...)
 4.       └─lintr:::get_knitr_pattern(filename, lines)
 5.         └─("knitr" %:::% "detect_pattern")(...)
```

That's coming from `knitr:::detect_pattern()`:

https://github.com/yihui/knitr/blob/1a0f2cc269c6a7f86f10b77d12229a80c14d6ee6/R/pattern.R#L155

My guess is this isn't an issue for knitr because it's already handled UTF-8 conversion by the time it runs `grepl()`:

https://github.com/yihui/knitr/blob/4767f7e9cad21df08a6ca46f028fd1bac3065bca/R/parser.R#L798

If we're going to use knitr's internal code, we should do the same sanitization.

MichaelChirico:
Closes #1920. Not an ideal solution but I'd rather just un-break our CI for now.

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -44,11 +44,10 @@ get_knitr_pattern <- function(filename, lines) {
   # suppressWarnings for #1920. TODO(michaelchirico): this is a bit sloppy -- we ignore
   #   warnings here because encoding issues are caught later and that code path handles them
   #   correctly by converting to a lint. It would require some refactoring to get that
-  #   right here as well, but it would avoid the duplication & avoid suppressing
-  #   good warnings. We might also convert this to a withCallingHandlers() to ignore only
-  #   the 'invalidMBCS' condition the parser is throwing.
-  pattern <- suppressWarnings(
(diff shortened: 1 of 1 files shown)


IndrajeetPatil (approved):
Thanks!

Yeah, I agree that we should resolve the CI/CD failure ASAP, and this is good enough to get there.

PR #1934, opened 2023-03-26
detect functional lambdas in object_usage_linter

### LINKED ISSUE #1933: object_usage_linter() misses functions defined with \(x) lambda
```r
lintr::lint("f <- \\() x\n", lintr::object_usage_linter())
lintr::lint("f <- function() x\n", lintr::object_usage_linter())
# Warning message:
# In parse_check_usage(fun, known_used_symbols = known_used_symbols,  :
#   Possible bug in lintr: Couldn't parse usage message ‘<anonymous>: no visible binding for global variable 'x'
# ’. Ignoring 1 usage warnings.
```

(the warning in the second case is #1917)

MichaelChirico:
Closes #1933


MichaelChirico:
Test failure looks spurious / caused by upstream

tests/testthat/test-object_usage_linter.R
```diff
@@ -624,3 +624,17 @@ test_that("messages without a quoted name are caught", {
...
+test_that("functional lambda definitions are also caught", {
+  skip_if_not_r_version("4.1.0")
+
+  expect_lint(
+    trim_some("
+      fun <- \\() {
+        a <- 1
+      }
+    "),
+    rex::rex("local variable", anything, "assigned but may not be used"),
```
IndrajeetPatil, line 637: Why is it `anything` and not `'a'` here?

MichaelChirico: Just copied it from the above test:

https://github.com/r-lib/lintr/blob/4f12a43d525994bf783e037605941d7bbbfd2c03/tests/testthat/test-object_usage_linter.R#L3

The advantage I see is we don't need to bother about whether `a` is quoted, or with which characters. OTOH that above usage may just be to re-use the message across tests where different-named variables are linted.

Happy to change if you'd like.

IndrajeetPatil: I think it's fine as it is then.

IndrajeetPatil (approved):
Thanks!

PR #1940, opened 2023-03-29
Catch consecutive calls to assert_that in renamed consecutive_asserion_linter

### LINKED ISSUE #1604: consecutive_stopifnot_linter should include assert_that (and hence be renamed)
Similar reasoning for `consecutive_stopifnot_linter()` applies for using `assert_that(); assert_that(); assert_that()`; unless a `message` is specified, it's better to use the `...` and condense to one `assert_that()` call.

This will entail a deprecation cycle for `consecutive_stopifnot_linter()`, so let's align on a name before continuing.

I think `consecutive_assertion_linter()` fits well, but differs from `conjunct_test_linter()` (which also covers `assert_that()` and `stopifnot()`).

`consecutive_test_linter()` sounds more like it would apply to `testthat` functions, so I'm not sold on it.

MichaelChirico:
Closes #1604


R/consecutive_stopifnot_linter.R
```diff
@@ -1,48 +0,0 @@
...
-#' lint(
-#'   text = "stopifnot(x, y)",
-#'   linters = consecutive_stopifnot_linter()
-#' )
-#'
-#' @evalRd rd_tags("consecutive_stopifnot_linter")
-#' @seealso [linters] for a complete list of linters available in lintr.
-#' @export
-consecutive_stopifnot_linter <- function() {
-  # match on the expr, not the SYMBOL_FUNCTION_CALL, to ensure
```
MichaelChirico, line 23: NB: this comment intentionally removed as the logic now lints on cases like:

```r
assert_that(x)
assertthat::assert_that(y)
```

Technically it would also lint on

```r
SomePkg::stopifnot(x)
base::stopifnot(y)
```

But that seems unlikely in practice. We could extend the logic to something like "lint if the namespaces are equal or both absent, or if only present once, but not if both present and not equal" but that seems like overkill.

tests/testthat/test-consecutive_assertion_linter.R
```diff
@@ -0,0 +1,109 @@
...
+  ")
+  expect_lint(
+    lines_comment,
+    rex::rex("Unify consecutive calls to stopifnot()."),
+    consecutive_assertion_linter()
+  )
+})
+
+test_that("assert_that usages are handled correctly too", {
+  linter <- consecutive_assertion_linter()
```
IndrajeetPatil, line 60: The `linter` variable is never used in this scope.

MichaelChirico: great catch!

IndrajeetPatil (approved):
LGTM. Thanks!

PR #1949, opened 2023-04-04
Outdent code extracted from indented chunks

### LINKED ISSUE #1945: indentation_linter: fenced code blocks inside enumeration in RMarkdown document
Consider the following RMarkdown document, which has fenced code inside an enumeration:

````rmd
* An enumeration item with code:

  ```{r}
  "Hello"
  ```
````

‘lintr’ has the spurious complaint

> ```
> mwe.rmd:4:2: style: [indentation_linter] Indentation should be 0 spaces but is 2 spaces.
>   "Hello"
> ~^
> ```

When removing the indent (which works, but is unconventional) and knitting the document, ‘knitr’ reinserts the indent in the output Markdown document.

### PR DESCRIPTION
**Pull request by MichaelChirico**
Closes #1945

This handles the issue at hand, not sure if @AshesITR would rather avoid further entrenching ourselves in the current setup :)

### CODE REVIEW

**Comment by MichaelChirico** [comment]
(ignoring the lint that's turned up & covered by #1879)

**Comment by AshesITR** [comment]
Will this make the lints show in the correct position?

**Comment by MichaelChirico** [comment]
> Will this make the lints show in the correct position?
> I'm mostly worried about the column_number field.

Good point. I guess it will be wrong... `print(lints)` will look fine, but plugin tools will highlight the wrong area (IINM).

Is that enough not to merge here? It seems to me like this is an improvement & we could just mark a TODO for now, WDYT?

**Comment by klmr** [comment]
Tough question! Personally I'd prefer if correct code didn't trigger any false positives, since linter warnings are actually *fairly noisy* in most editors. As a consequence this leads users (well, at least me personally) to disable linting entirely rather than having even just a handful of false positives to deal with.

**Comment by MichaelChirico** [comment]
that was my thinking too. bad lint metadata isn't an issue on lint-free code. and indented code blocks should be relatively rare anyway.

OTOH we should throw a warning in the meantime when this is detected

**Comment by AshesITR** [comment]
Is there an easy way to detect this?
Or can we even fix it by simply matching the lints `line` metadata with the actual file lines as a postprocessing step in `lint()`?

**Comment by AshesITR** [comment]
Ah okay fine. The disadvantage of signalling in `get_source_expressions()` is that such a warning would pop up even for code without lints.

File: R/extract.R [hunk 1]
```diff
@@ -22,12 +22,14 @@ extract_r_source <- function(filename, lines, error = identity) {
 
   output_env <- environment() # nolint: object_usage_linter. False positive-ish -- used below.
   Map(
-    function(start, end) {
+    function(start, end, indent) {
       line_seq <- seq(start + 1L, end - 1L)
-      output_env$output[line_seq] <- lines[line_seq]
+      chunk_code <- lines[line_seq]
+      output_env$output[line_seq] <- if (indent > 0L) substr(chunk_code, indent + 1L, nchar(chunk_code)) else chunk_code
```
AshesITR, line 28: Why the branch? `substr(x, 1L, nchar(x)) == x` unless there is some encoding issue, no?

R/extract.R
```diff
@@ -73,7 +75,13 @@ get_chunk_positions <- function(pattern, lines) {
   # only keep those blocks that contain at least one line of code
   keep <- which(ends - starts > 1L)
 
-  list(starts = starts[keep], ends = ends[keep])
+  starts <- starts[keep]
+  ends <- ends[keep]
+
+  # This is _somewhat_ fragile... better may be to do gsub("```.*", "", pattern$chunk.begin),
+  #   but that somehow feels roughly as fragile (and less readable).
+  start_indent_matches <- gregexpr("^[\t >]*", lines[starts], perl = TRUE)
```
AshesITR, line 83: Are there other indents apart from spaces and blockquotes?

tests/testthat/test-get_source_expressions.R
```diff
@@ -307,6 +307,26 @@ test_that("Syntax errors in Rmd or qmd don't choke lintr", {
...
+    "# New section",
+    "",
+    "```{r unindented_chunk}",
+    '  "improperly indented"',
+    "```"
+  ))
+
+  parsed_lines <- get_source_expressions(tmp)$lines
+  expect_identical(parsed_lines[4L], '"properly indented"', ignore_attr = "names")
+  expect_identical(parsed_lines[10L], '  "improperly indented"', ignore_attr = "names")
```
AshesITR, line 327: Can we also test that we don't destroy code?
````
  ```{r weird indent}
"no whitespace before"
  ```
````

Runs through `rmarkdown::render()` just fine.

Comment (AshesITR) [reply tests/testthat/test-get_source_expressions.R:327]: To be clear, I don't mind this creating (or not creating) spurious lints.
It just shouldn't create parse errors or the like.

File: R/extract.R (hunk 1 above)
Comment (MichaelChirico) [reply R/extract.R:28]: just avoiding an overwrite if it can be avoided (which is true in vast majority of cases)

File: R/extract.R (hunk 2 above)
Comment (MichaelChirico) [reply R/extract.R:83]: this is extracted from the knitr pattern, so knitr won't recognize anything else

File: tests/testthat/test-get_source_expressions.R (hunk 3 above)
Comment (MichaelChirico) [reply tests/testthat/test-get_source_expressions.R:327]: ah i really wish `rmarkdown::render()` would error there :)

Comment (MichaelChirico) [reply tests/testthat/test-get_source_expressions.R:327]: I'm tempted to punt on this. It's easy enough to do something like `gsub("^[ \t>]{1,N}", "", line)` where `N` is the indent column number, but then what about

       ```{r weirder_indent}
    "at column 1"
     "at column 2"
       ```

Ideally we'd lint (I think `indentation_linter()`?) on the second line, so actually we need to find the first column for actual code in each chunk and outdent that much (harder than current setup)

Comment (MichaelChirico) [reply tests/testthat/test-get_source_expressions.R:327]: actually in writing this comment, maybe I answered my own question :)

File: NEWS.md [hunk 4]
```diff
@@ -109,6 +109,8 @@
 
 * `missing_argument_linter()` allows missing arguments in `quote()` calls (#1889, @IndrajeetPatil). 
 
+* `get_source_expressions()` outdents code extracted from indented chunks, which helps avoid spurious lints related to whitespace (#1945, @MichaelChirico).
```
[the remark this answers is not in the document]
MichaelChirico: Great! Also added a note about using the leftmost column in the weird staggered case.

tests/testthat/test-get_source_expressions.R
MichaelChirico: Still a bit messy, but not horrible; done.

R/extract.R
MichaelChirico: OK, wrote it to be explicitly based on the knitr pattern to make that a bit clearer.

AshesITR (approved):

PR #1955, opened 2023-04-07
Rename no_tab_linter to whitespace_linter

### LINKED ISSUE #1954: Rename no_tabs_linter() to whitespace_linter()
Inspired by https://github.com/r-devel/rdevguide/issues/133 -- I know base R has a lot of places where `\t` is used, so adding flexbility to the linter will help.

The new suggested name is also more stylistically similar to other linters in the suite.

This issue is just about renaming the linter, filing now to get in before the 3.1.0 "deadline" alongside the other renamings. Future issues will address extending flexibility. Not doing so here because it's not trivial -- setting `\t` as the desired whitespace probably just refers to _initial_ indentation, with other spaces in the file still just ` `. The current linter doesn't support this flexibility (it just looks for `\t` anywhere), so it's more of an investment to get there.

MichaelChirico:
Closes #1954


AshesITR (approved):
Not sure whether we should explicitly state future planned development in the docs.

MichaelChirico:
> Not sure whether we should explicitly state future planned development in the docs.

Do you mean in NEWS? In `?whitespace_linter`, we have "Currently, only supports linting in the presence of tabs." which hints at something being done later but gives no specifics/commitments. I think that's OK; submitting for now but feel free to follow-up

PR #1958, opened 2023-04-08
exclude deprecated linters from overall list & tag-specific lists

### LINKED ISSUE #1956: Drop deprecated linters from individual tag pages
Currently, we surface linters that may be deprecated on the tag descriptions' pages, e.g.

https://github.com/r-lib/lintr/blob/fb294ccca4267c5b6530c7be886ebcdd34afec41/man/style_linters.Rd#L45

This is not ideal since users may attempt to use these without realizing they're deprecated if it's referenced on the tag page without context.

We could either annotate `(deprecated)` on the tag page, or simply omit these from the tag page.

I would opt for the latter.

There are two options:

 1. Amend the code generating the pages to skip if `deprecated` is among the tags
 2. Retain only the `deprecated` tag for such linters

We may still want to keep the tags recorded in inst/lintr/linters.csv for reference (it's still correct metadata about the linter), hence I lean towards option 1 on the assumption it's only a line or two of difference to implement it.

MichaelChirico:
Closes #1956


MichaelChirico:
Fixing the tests will be a bit of a pain. Hold off for now.

MichaelChirico:
ready for review now that tests are passing

R/linter_tags.R
```diff
@@ -180,10 +183,11 @@ rd_linters <- function(tag_name) {
 #'
 #' @noRd
 rd_taglist <- function() {
-  linters <- available_linters(exclude_tags = NULL)
+  linters <- available_linters()
```
AshesITR, line 186: This looks a bit hacky.
Maybe instead overwrite the `tags` column for deprecated linters?

tests/testthat/test-linter_tags.R
```diff
@@ -112,27 +112,35 @@ test_that("lintr help files are up to date", {
...
   lintr_db$package <- NULL
   lintr_db$tags <- lapply(lintr_db$tags, sort)
 
-  expect_true(exists("linters", envir = help_env), info = "?linters exist")
-  # objects in help_env are class Rd, see ?as.character.Rd (part of 'tools')
-  linter_help_text <- paste(as.character(help_env$linters), collapse = "")
+  expect_true(exists("linters", envir = help_env), info = "?linters exists")
+
+  # NB: objects in help_env are class Rd, see ?as.character.Rd (part of 'tools')
+  rd_as_string <- function(rd) paste(as.character(rd), collapse = "")
```
AshesITR, line 130: Nice helper 😃

MichaelChirico: the whole test is still a bit of a mess to understand from scratch, this is a small improvement :)

R/linter_tags.R
MichaelChirico: I'll take another pass once #1961 is done

R/linter_tags.R
```diff
@@ -160,7 +160,10 @@ rd_tags <- function(linter_name) {
 #'
 #' @noRd
 rd_linters <- function(tag_name) {
-  linters <- available_linters(tags = tag_name, exclude_tags = NULL)
+  # don't show any deprecated linters on a tag's page
+  #   (except, of course, on the 'deprecated' tag's page)
+  exclude_tags <- if (tag_name == "deprecated") NULL else "deprecated"
```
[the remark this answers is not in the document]
MichaelChirico: Yep, done

R/linter_tags.R
MichaelChirico: Good idea. It's still pretty hacky, but at least easier to follow. Ditto I hope for the tests.

AshesITR (approved):

PR #1960, opened 2023-04-10
handle multiline case for function_left_parentheses_linter

### LINKED ISSUE #1953: Error in rep.int(character, length) : invalid 'times' value
```R
> install.packages("lintr")
[...]
> lintr::lint_dir("R")
...................................................
```

Is there any way to make it tell me at least what file it errored out in? Or which line in the lintr code? Anything? I know this bug report is supremely unhelpful, but there is nothing else I've got.

MichaelChirico:
Closes #1953

Open to suggestions for avoiding the horrible `range_end_xpath` but I wasn't seeing anything...

### DIFF
--- a/R/function_left_parentheses_linter.R
+++ b/R/function_left_parentheses_linter.R
@@ -61,7 +61,7 @@ function_left_parentheses_linter <- function() { # nolint: object_length.
     bad_line_lints <- xml_nodes_to_lints(
       bad_line_exprs,
       source_expression = source_expression,
-      lint_message = "Left parenthesis should be on the same line as the function's symbol.",
+      lint_message = "Left parenthesis should be on the same line as the function's symbol."
     )
 
     bad_col_exprs <- xml2::xml_find_all(xml, bad_col_xpath)
--- a/R/function_left_parentheses_linter.R
+++ b/R/function_left_parentheses_linter.R
@@ -64,8 +64,10 @@ function_left_parentheses_linter <- function() { # nolint: object_length.
       # this is a janky XPath 1.0 implementation for if(same line, "before (", "range[1]")
       #   by converting if(COND, x, y) to 1[COND] * x + (1 - 1[COND]) * y for 1[.] an indicator function
       range_end_xpath = "
-        number(./@line1  = ./following-sibling::OP-LEFT-PAREN/@line1) * number(./following-sibling::OP-LEFT-PAREN/@col1 - 1) +
-        number(./@line1 != ./following-sibling::OP-LEFT-PAREN/@line1) * number(./@col2 + 1)
+        number(./@line1  = ./following-sibling::OP-LEFT-PAREN/@line1) *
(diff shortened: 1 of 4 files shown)


R/function_left_parentheses_linter.R
```diff
@@ -60,7 +60,15 @@ function_left_parentheses_linter <- function() { # nolint: object_length.
       source_expression = source_expression,
       lint_message = "Remove spaces before the left parenthesis in a function call.",
       range_start_xpath = "number(./@col2 + 1)", # start after function / fun
-      range_end_xpath = "number(./following-sibling::OP-LEFT-PAREN/@col1 - 1)" # end before (
+      # if '(' is on the same line, end before '(', otherwise, just set range[2]=range[1].
+      # this is a janky XPath 1.0 implementation for if(same line, "before (", "range[1]")
+      #   by converting if(COND, x, y) to 1[COND] * x + (1 - 1[COND]) * y for 1[.] an indicator function
+      range_end_xpath = "
```
MichaelChirico, line 66: alternatively, we might change the lint message to something like "Move the left parenthesis to the same line as the start of the function call."

R/function_left_parentheses_linter.R
```diff
@@ -41,11 +41,11 @@
 #' - [spaces_left_parentheses_linter()]
 #' @export
 function_left_parentheses_linter <- function() { # nolint: object_length.
-  xpath <- "
-  //FUNCTION[@col2 != following-sibling::OP-LEFT-PAREN/@col1 - 1]
-  |
-  //SYMBOL_FUNCTION_CALL/parent::expr[@col2 != following-sibling::OP-LEFT-PAREN/@col1 - 1]
-  "
+  pos_xpath <- xp_or(
```
AshesITR, line 44: The variable name seems odd to me. Maybe add `bad` somewhere?

R/function_left_parentheses_linter.R
AshesITR: Are we vectorized on `range_end_xpath`?
This sounds like we should do a branch for each of the two lints (wrong col on same line or wrong line).

AshesITR (commented):
Quick question: Is it syntactically valid to have more than one line of difference?

```r
x <- function

() {
  42L
}
```

R/function_left_parentheses_linter.R
MichaelChirico: We should be, but it makes sense to add a test as such.

> branch for each of the two lints

yep, that's what i was getting at w/ changing the lint message.

MichaelChirico:
> Is it syntactically valid to have more than one line of difference?

yes, that parses

R/function_left_parentheses_linter.R
MichaelChirico: Went that route now

R/function_left_parentheses_linter.R
MichaelChirico: Should be clearer in the new version.

AshesITR (approved):

PR #1961, opened 2023-04-10
prioritize tags over exclude_tags in available_linters

### LINKED ISSUE #1959: available_linters(tags = "deprecated") should "just work"
I find the current behavior a bit strange:

```r
available_linters(tags = "deprecated")
# [1] linter  package tags   
# <0 rows> (or 0-length row.names)
```

Almost surely, that's not what the user wanted. The problem is `exclude_tags = "deprecated"` by default, so we need to also specify `exclude_tags = NULL`:

```r
nrow(available_linters(tags = "deprecated", exclude_tags = NULL))
# [1] 7
```

I see two options:

 1. Special case `"deprecated"` with an `if (missing(exclude_tags) && identical(tags, "deprecated")) exclude_tags <- NULL`
 2. Always run `exclude_tags <- setdiff(exclude_tags, tags)`, i.e., `exclude_tags` takes priority over `tags` if there's any collision

Any thoughts?

MichaelChirico:
Closes #1959

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -115,6 +115,8 @@
 
 * `get_source_expressions()` correctly extracts indented code chunks from R Markdown documents, which helps avoid spurious lints related to whitespace (#1945, @MichaelChirico). The convention taken is that, within each chunk, all code is anchored relative to the leftmost non-whitespace column.
 
+* `available_linters()` gives priority to `tags` over `exclude_tags` in the case of overlap. In particular, this means that `available_linters(tags = "deprecated")` will work to return deprecated linters without needing to specify `exclude_tags` (#1959, @MichaelChirico).
+
 ### New linters
 
 * `matrix_apply_linter()` recommends use of dedicated `rowSums()`, `colSums()`, `colMeans()`, `rowMeans()` over `apply(., MARGIN, sum)` or `apply(., MARGIN, mean)`. The recommended alternative is much more efficient and more readable (#1869, @Bisaloo).
--- a/R/linter_tags.R
+++ b/R/linter_tags.R
@@ -7,7 +7,8 @@
 #' returned. If `tags` is `NULL`, all linters will be returned. See `available_tags("lintr")` to find out what
 #' tags are already used by lintr.
 #' @param exclude_tags Tags to exclude from the results. Linters with at least one matching tag will not be returned.
-#' If `except_tags` is `NULL`, no linters will be excluded.
+#' If `except_tags` is `NULL`, no linters will be excluded. Note that `tags` takes priority, meaning that any
+#' tag found in both `tags` and `exclude_tags` will be included, not excluded.
 #'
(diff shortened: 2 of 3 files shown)


AshesITR (commented):
LGTM, feel free to replace the failing test by another that produces empty output.
Maybe `tags = "default"` with `exclude_tags = c("style", ...)`?

NEWS.md
```diff
@@ -1,11 +1,12 @@
 # lintr (development version)
 
-## Deprecations
+## Deprecations & Breaking Changes
 
 * `single_quotes_linter()` is deprecated in favor of the more generalizable `quotes_linter()` (#1729, @MichaelChirico).
 * `unneeded_concatentation_linter()` is deprecated in favor of `unnecessary_concatenation_linter()` for naming consistency (#1707, @IndrajeetPatil).
 * `consecutive_stopifnot_linter()` is deprecated in favor of the more general (see below) `consecutive_assertion_linter()` (#1604, @MichaelChirico).
 * `no_tab_linter()` is deprecated in favor of `whitespace_linter()` for naming consistency and future generalization (#1954, @MichaelChirico).
+* `available_linters()` prioritizes `tags` over `exclude_tags` in the case of overlap, i.e., tags listed in both arguments are included, not excluded. We don't expect many people to be affected by this, and the old behavior was not made explicit in the documentation, but make note of it here since it required changing a test in lintr's own suite where `linters_with_tags()` implicitly assumed this behavior.
```
MichaelChirico, line 9: I hesitated a bit about even mentioning this, but I do think tests should count as "documented behavior", so if we change a test, that's worth noting in the release notes.

To be clear, I would be willing to make such a "breaking change" even on a minor release since I think it's such an edge case, but I think it's good practice to make note of whenever we have to retrofit our own tests.

Open to suggestions here.

tests/testthat/test-with.R
```diff
@@ -44,7 +44,7 @@ test_that("all default linters are tagged default", {
   skip_on_covr()
 
   expect_identical(linters_with_tags("default"), linters_with_defaults())
-  expect_length(linters_with_tags("default", exclude_tags = "default"), 0L)
+  expect_length(linters_with_tags(NULL, exclude_tags = available_tags()), 0L)
```
MichaelChirico, line 47: It might be nice to support `linters_with_tags(exclude_tags = ...)` without specifying `tags = NULL`.

That would require setting a default of `tags = NULL` (or hacking around with `missing()`), but I hesitated since this would allow `linters_with_tags()` (i.e., no arguments) to work, which doesn't quite read as well as `linters_with_tags(NULL)`.

We could allow both behaviors ([1] require at least one argument specified and [2] allow only `exclude_tags` to be specified), again if we hack around with `missing()`. I held off on that for now since I'm recently more loath to fidget with `missing()`.

PR #1962, opened 2023-04-11
Update compare_branches not to fail when one branch errors in some cases

MichaelChirico:
Erroring on `main` is a good way to suss out new logic errors, but erroring on the old branch is mostly an annoyance.


.dev/compare_branches.R
```diff
@@ -311,15 +313,23 @@ lint_one_package <- function(package, linters, out_dir, check_deps) {
   #   ignore them because they're innocuous.
   # also ignore lintr's deprecations to make including all linters easier
   suppressMessages(withCallingHandlers(
-    lints <- as.data.frame(lint_dir(package, linters = linters, parse_settings = FALSE)),
+    {
+      # Don't fail out if one branch exhibits errors (e.g. if the new version fixes a known error)
+      lints <- tryCatch(lint_dir(package, linters = linters, parse_settings = FALSE), error = identity)
```
AshesITR, line 318: You are already within a `withCallingHandlers()` block so you could just add an error handler?

AshesITR (approved):

PR #1989, opened 2023-06-29
More helpful failure message for object_usage_linter()

### LINKED ISSUE #1985: New warning in 'bupaR' for object_usage_linter during glue() evaluation
Haven't confirmed this is a regression yet, but we observe this warning on `main` that is not present on `v3.0.2`:

```
Evaluating glue expression while testing for local variable usage failed: Expecting '}'
```

MichaelChirico:
Closes #1985


AshesITR:
Can you add a test?
Also, the issue should be referenced in `NEWS.md`.

AshesITR (approved):

PR #1994, opened 2023-07-01
more informative error message if a lintr config setting is broken

### LINKED ISSUE #886: Detect broken configuration
A lot of new users seem to stumble upon the quirkiness of the dcf format we use for `.lintr`.
I see three major pitfalls:

1. Indentation of multiline code is not intuitive

```
# Wrong
linters: with_defaults(
    line_length_linter = line_length_linter(120)
)

# Right
linters: with_defaults(
    line_length_linter = line_length_linter(120)
  )
```

2. Trailing newline is mandatory, but not reported as an error

```
# Wrong
linters: with_defaults(
    line_length_linter = line_length_linter(120)
  )
(EOF)

# Right
linters: with_defaults(
    line_length_linter = line_length_linter(120)
  )

(EOF)
```

3. The proper location of the `.lintr` config files sometimes causes confusion.

It would be nice of us to catch these and report a warning (or maybe even an error?) if the config file is malformed.

Regarding the location issue, maybe reporting the config file used (if any) would help diagnosing the problem more quickly.
I see a few options here as to when to report it:

1. Only if some option (`linter.verbose`, or `lintr.debug`?) is set.
2. Only if some argument (`verbose`, or `debug`?) is TRUE.
3. Always if running interactively.
4. Always at the start of a top-level call (lint_dir, lint_package or lint if the latter isn't called from within lint_*).

WDYT?
I think the most user-friendly version would be to use `getOption("lintr.???", default = interactive())` to decide whether to print this diagnostic info at the start of a top-level call.
If no config file is found, the info could be

  Using the default linter configuration. See vignette("using_lintr", package = "lintr") for customization options.

or something like that.

AshesITR:
fixes #886

### DIFF
--- a/tests/testthat/test-settings.R
+++ b/tests/testthat/test-settings.R
@@ -67,6 +67,13 @@ test_that("it errors if the config file does not end in a newline", {
   expect_error(lintr:::read_settings("foo"), "Malformed config file")
 })
 
+test_that("it gives informative erros if the config file contains errors", {
+  f <- withr::local_tempfile()
+  cat("linters: linters_with_defaults(\n   closed_curly_linter = NULL, \n )\n", file = f)
+  withr::local_options(list(lintr.linter_file = f))
+  expect_error(lintr:::read_settings("foo"), "Malformed config setting 'linters'")
+})
+
 test_that("rot utility works as intended", {
   expect_identical(lintr:::rot(letters), c(letters[14L:26L], LETTERS[1L:13L]))
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-settings.R
```diff
@@ -67,6 +67,13 @@ test_that("it errors if the config file does not end in a newline", {
   expect_error(lintr:::read_settings("foo"), "Malformed config file")
 })
 
+test_that("it gives informative erros if the config file contains errors", {
+  f <- withr::local_tempfile()
```
MichaelChirico, line 71: use the lines= argument for local_tempfile()?

MichaelChirico (approved):

Issue #1924, opened 2023-02-28
Document "symbols" naming style in object_name_linter 


hedsnz:
The [documentation](https://github.com/r-lib/lintr/blob/main/R/object_name_linter.R) for `object_name_linter` describes:

> ... The default naming styles are "snake_case" and "symbols".

After that, there is no explanation of what "symbols" means, and there is no example that uses `styles = "symbols"`. Could this documentation please be added?

Thanks

luma-sb:
I was recently asked by my team what the `"symbols"` style for `object_name_linter`  linter did - so did some digging and hit same issue as you @hedsnz . Appears this was introduced to resolve issue #615 which provides helpful examples to demonstrate applicable cases.

AshesITR:
Not sure where to document the regexes, but I would describe `"symbols"` as function names containing only non-alphanumeric characters, e.g., `` `%+%` `` from ggplot2 or `` `%>%` `` from magrittr.

PR #2001, opened 2023-07-05
document "symbols" in object_name_linter (#1924)

hedsnz:
Resolves https://github.com/r-lib/lintr/issues/1924 by adding a short description of the "symbols" style in the `object_name_linter` documentation.

### DIFF
--- a/R/object_name_linter.R
+++ b/R/object_name_linter.R
@@ -45,7 +45,9 @@ object_name_xpath <- local({
 #'
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
-#'   match at least one of these styles.
+#'   match at least one of these styles. The `"symbols"` style refers to
+#'   names containing only non-alphanumeric characters, e.g., `%+%` from
+#'   ggplot2 or `%>%` from magrittr.
 #' @param regexes A (possibly named) character vector specifying a custom naming convention.
 #'   If named, the names will be used in the lint message. Otherwise, the regexes enclosed by `/` will be used in the
 #'   lint message.
--- a/R/object_name_linter.R
+++ b/R/object_name_linter.R
@@ -46,8 +46,10 @@ object_name_xpath <- local({
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
 #'   match at least one of these styles. The `"symbols"` style refers to
-#'   names containing only non-alphanumeric characters, e.g., `%+%` from
-#'   ggplot2 or `%>%` from magrittr.
+#'   names containing *only* non-alphanumeric characters; e.g., `%+%` from
+#'   ggplot2 or `%>%` from magrittr will not generate lint markers, whereas
+#'   `%in%` (containing both alphanumeric *and* non-alphanumeric characters)
+#'   will.
 #' @param regexes A (possibly named) character vector specifying a custom naming convention.
 #'   If named, the names will be used in the lint message. Otherwise, the regexes enclosed by `/` will be used in the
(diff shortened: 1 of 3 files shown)


R/object_name_linter.R
```diff
@@ -45,7 +45,9 @@ object_name_xpath <- local({
 #'
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
-#'   match at least one of these styles.
+#'   match at least one of these styles. The `"symbols"` style refers to
+#'   names containing only non-alphanumeric characters, e.g., `%+%` from
```
MichaelChirico, line 49: maybe `*only*` to emphasize that e.g. `%in%` is not covered?

hedsnz: Sure -- I'm not totally sure what the appropriate grammar is, but how does this sound? https://github.com/r-lib/lintr/pull/2001/commits/521f248e92db4650908dc8a952bf6009d90e56b0

R/object_name_linter.R
```diff
@@ -45,7 +45,11 @@ object_name_xpath <- local({
 #'
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
-#'   match at least one of these styles.
+#'   match at least one of these styles. The `"symbols"` style refers to
+#'   names containing *only* non-alphanumeric characters; e.g., `%+%` from
+#'   ggplot2 or `%>%` from magrittr will not generate lint markers, whereas
+#'   `%in%` (containing both alphanumeric *and* non-alphanumeric characters)
```
AshesITR, line 51: Actually this is considered to be the same as `"in"` for the purpose of `object_name_linter()`, so it will match any lowercase style.
`` lubridate::`%m+%` `` is a better example I think.

hedsnz: Fair enough -- happy to make that change, but I've just found that `%>%` will not generate a lint regardless of whether `"symbols"` is included; is that expected?

```{r}
# R/test.R
library(magrittr)
iris %>%
  nrow()
```

```{r}
# Does not generate a lint
lint("R/test.R", linters = list(object_name_linter(styles = "snake_case")))
```

AshesITR: The linter only lints object names that you define. Your example doesn't contain any assignments that could be linted.

hedsnz: Ah right, okay. How does it look now?

R/object_name_linter.R
```diff
@@ -45,7 +45,12 @@ object_name_xpath <- local({
 #'
 #' @param styles A subset of
 #'   \Sexpr[stage=render, results=rd]{lintr:::regexes_rd}. A name should
-#'   match at least one of these styles.
+#'   match at least one of these styles. The `"symbols"` style refers to
+#'   names containing *only* non-alphanumeric characters; e.g., `%+%` from
+#'   ggplot2 or `%>%` from magrittr will not generate lint markers, whereas
```
MichaelChirico, line 50: I think "would" is probably grammatically more appropriate, as in "`%>%` from magrittr _would_ not generate lint markers". This helps convey that usage of other packages' objects won't throw lints regardless of style, only objects "you" define. That is, _using_ `%m+%` won't throw a lint, but _defining_ it will.

MichaelChirico:
Thanks for your patience -- a long review for a seemingly minor change :)

But it's important to get user-facing docs as nice as possible, IMO.

Please also credit yourself in the NEWS! I think this would be most appropriate in the `Notes` section.

hedsnz:
I'm very much in agreement :)

Thanks!

MichaelChirico (approved):

PR #2024, opened 2023-07-27
Use file-level expressions and avoid slow loop for whitespace_linter() for up to 50% speedup

MichaelChirico:
I was testing out why our lint suite ran so slow on a particular long file & found somewhat surprisingly that `whitespace_linter()` was the slowest of ~100 linters.

We can test this on R's `src/library/tools/R/QC.R`:

```r
exp <- get_source_expressions("QC.R")
l <- whitespace_linter()
system.time(replicate(100, lapply(exp$expressions, l)))

## BEFORE
#    user  system elapsed 
# 398.371   1.090 400.302 

## AFTER
#    user  system elapsed 
#   1.153   0.004   1.163 
```

End-to-end timing including `get_source_expressions()`:

```r
system.time(replicate(10, lint("QC.R", whitespace_linter())))

## BEFORE
#    user  system elapsed 
#  66.333   1.945  68.471 

## AFTER
#    user  system elapsed 
#  31.435   0.819  32.291 
```

So an end-to-end (user-facing) 50% improvement, inclusive of parsing time.

So while this approach is not as cache-friendly, of course checking initial whitespace in a file should not be expensive, so I don't think we lose much by switching to a file-level search.

### DIFF
--- a/.lintr_new
+++ b/.lintr_new
@@ -2,7 +2,7 @@ linters: linters_with_defaults(
     any_duplicated_linter(),
     any_is_na_linter(),
     backport_linter("oldrel-4", except = c("R_user_dir", "str2lang", "str2expression", "deparse1")),
-    consecutive_stopifnot_linter(),
+    consecutive_assertion_linter(),
     expect_comparison_linter(),
     expect_length_linter(),
     expect_named_linter(),
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 * `inner_combine_linter()` no longer throws on length-1 calls to `c()` like `c(exp(2))` or `c(log(3))` (#2017, @MichaelChirico). Such usage is discouraged by `unnecessary_concatenation_linter()`, but `inner_combine_linter()` _per se_ does not apply.
 
+## Changes to defaults
+
+* `assignment_linter()` lints the {magrittr} assignment pipe `%<>%` (#2008, @MichaelChirico). This can be deactivated by setting the new argument `allow_pipe_assign` to `TRUE`.
+
 # lintr 3.1.0
 
 ## Deprecations & Breaking Changes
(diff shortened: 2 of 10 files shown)


MichaelChirico:
Hmm, I think the error fixed by b5793a7efcf0333c83a7d83387eb5320fe80832f also reduced the efficiency improvement. I only saw a 3% improvement just now.

jimhester:
Can you update the timings after the full change. I doubt they will change much, but still good to be sure

MichaelChirico:
I'm looking more closely at what the bottleneck is, and `re_matches()` seems to be a big piece of it

c.f.

```r
source_expressions <- get_source_expressions("QC.R")$expressions
full_expression <- source_expressions[[length(source_expressions)]]

regex <- "^(?:\\s)*(?:\t)+"

system.time(re_matches(full_expression[["file_lines"]], regex, locations = TRUE, global = TRUE))
#    user  system elapsed 
#   1.956   0.016   1.982 
system.time(grepl(regex, full_expression[["file_lines"]]))
#    user  system elapsed 
#    0.01    0.00    0.01 
```

MichaelChirico:
I think the issue is `global=TRUE()`. That regex doesn't need `global=TRUE` -- we can use `regexpr()` and avoid `gregexpr()`.

```r
system.time(re_matches(full_expression[["file_lines"]], regex, locations = TRUE, global = FALSE))
#    user  system elapsed 
#   0.002   0.000   0.002 
```

`global=TRUE` is slow because it has to do a slow iteration over a long list of `gregexpr()` outputs, few of which even have regex matches.

MichaelChirico:
Bad news: my simple 5-character PR has ballooned substantially

Good news: this PR now marks a 400x (!!) speed improvement for QC.R

(timings added to the top)

AshesITR (approved):

PR #2025, opened 2023-07-27
Use node-level | for efficiency in infix_spaces_linter

MichaelChirico:
Similar to #2024, I found `infix_spaces_linter()` to be among the slowest, and apparently owing to the `//*` expression.

It's really slow! Even with 22 entries in `//NODE1[...] | //NODE2[...] | ... | //NODE22[...]`, I find this approach about 3x as fast in the core loop:

```r
exp <- get_source_expressions("QC.R")
ll <- lintr::infix_spaces_linter()
system.time(replicate(100, lapply(exp$expressions, ll)))

## BEFORE
#    user  system elapsed 
# 148.841   0.800 157.128

## AFTER
#    user  system elapsed 
#  48.209   0.286  51.664
```

### DIFF
--- a/R/infix_spaces_linter.R
+++ b/R/infix_spaces_linter.R
@@ -144,10 +144,11 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces 
 
   # NB: preceding-sibling::* and not preceding-sibling::expr because
   #   of the foo(a=1) case, where the tree is <SYMBOL_SUB><EQ_SUB><expr>
-  # NB: position() > 1 for the unary case, e.g. x[-1]
+  # NB: parent::*[count(expr) + count(OP-LEFT-PAREN) > 1] for the unary case, e.g. x[-1]
+  #  OP-LEFT-PAREN for EQ_SUB case with missing argument like alist(a =)
   # NB: the last not() disables lints inside box::use() declarations
   xpath <- paste(collapse = "|", glue::glue("//{infix_tokens}[
-    position() > 1
+    parent::*[count(expr) + count(SYMBOL_SUB) > 1]
     and (
       (
         @line1 = preceding-sibling::*[1]/@line2
--- a/R/infix_spaces_linter.R
+++ b/R/infix_spaces_linter.R
@@ -146,9 +146,8 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces 
   #   of the foo(a=1) case, where the tree is <SYMBOL_SUB><EQ_SUB><expr>
   # NB: position() > 1 for the unary case, e.g. x[-1]
   # NB: the last not() disables lints inside box::use() declarations
-  xpath <- glue::glue("//*[
-    ({xp_or(paste0('self::', infix_tokens))})
-    and position() > 1
+  xpath <- paste(collapse = "|", glue::glue("//{infix_tokens}[
+    position() > 1
     and (
       (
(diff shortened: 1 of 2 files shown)


MichaelChirico:
I started a Wiki page here:

https://github.com/r-lib/lintr/wiki/Writing-robust,-performant-linters

My first thought was to add notes like this or an efficiency/"advanced" section to the custom_linters vignette, but I think a Wiki may be preferable here for reasons noted there.

AshesITR:
The `position()` condition breaks your approach unfortunately.
Maybe we can avoid a top-level `//*` by constructing an XPath that has `/*[position() > 1]` as an element?

MichaelChirico:
how about //OP[parent::expr[count(expr) >2]]

untested but hopefully it's clear what I mean -- test against unary operator on parent instead of self

AshesITR:
Worth a try.

MichaelChirico:
Not quite so simple (h/t our extensive test suite!), but the basic idea works. The complications for future reference were:

 1. `parent::expr` is not always true, we see `<expr_or_assign_or_help>` for some `EQ_ASSIGN` cases
 2. `count(expr) > 1` is not enough for the `EQ_SUB` (assignment of function parameters) case, where a missing argument means there's no `<expr>` on the RHS, while the LHS doesn't get an `<expr>` because it's `SYMBOL_SUB` there.

R/infix_spaces_linter.R
```diff
@@ -144,11 +144,11 @@ infix_spaces_linter <- function(exclude_operators = NULL, allow_multiple_spaces
...
   #   of the foo(a=1) case, where the tree is <SYMBOL_SUB><EQ_SUB><expr>
-  # NB: position() > 1 for the unary case, e.g. x[-1]
+  # NB: parent::*[count(expr) + count(OP-LEFT-PAREN) > 1] for the unary case, e.g. x[-1]
+  #  OP-LEFT-PAREN for EQ_SUB case with missing argument like alist(a =)
   # NB: the last not() disables lints inside box::use() declarations
-  xpath <- glue::glue("//*[
-    ({xp_or(paste0('self::', infix_tokens))})
-    and position() > 1
+  xpath <- paste(collapse = "|", glue::glue("//{infix_tokens}[
+    parent::*[count(expr) + count(SYMBOL_SUB) > 1]
```
AshesITR, line 151: is `count(expr | SYMBOL_SUB)` a thing?

MichaelChirico: Nice! I still don't have a good sense of when `|` can be used like this. It passed tests locally, so let's go for it!

AshesITR:
Do we have a test for something like `-1-1` to ensure the XPath refactoring scopes correctly and only finds one lint in the example?

tests/testthat/test-infix_spaces_linter.R
```diff
@@ -185,3 +185,7 @@ test_that("native pipe is supported", {
   expect_lint("a |> foo()", NULL, linter)
   expect_lint("a|>foo()", rex::rex("Put spaces around all infix operators."), linter)
 })
+
+test_that("mixed unary & binary operators aren't mis-lint", {
+  expect_lint("-1-1", rex::rex("Put spaces around all infix operators."), infix_spaces_linter())
```
AshesITR, line 190: I think it's worth specifying expected lint metadata for this one, i.e., column names.

AshesITR (commented):
Nice work.

AshesITR (approved):

PR #2027, opened 2023-07-28
Add library call linter

nicholas-masel:
I've got a use case where I need all `library()` calls to be at the top of the script so the search path is consistent across the entire script.

I will incorporate this into the linting for the [logrx package](https://pharmaverse.github.io/logrx/articles/options.html#log.rx.lint),  but thought others might want this as well, so better in lintr than buried in a logging package.  I hope you all agree.


nicholas-masel:
I'm not sure why the lint is returning multiple and causing the tests to fail.  When I run this on my side it returns one and everything is running cleanly. 

``` r
library(lintr)

lint(
  text = c("library(dplyr)", "print('test')", "library(tidyr)"),
  linters = library_call_linter()
)
#> <text>:3:1: warning: [library_call_linter] Move all library calls to the top of the script.
#> library(tidyr)
#> ^~~~~~~
```

<sup>Created on 2023-07-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

AshesITR:
AFAIK the `last()` predicate refers to the last element that satisfies a condition within each node.
That means it probably doesn't do anything in the XPath, because all the library calls are contained inside their own expr element.

nicholas-masel:
> AFAIK the `last()` predicate refers to the last element that satisfies a condition within each node. That means it probably doesn't do anything in the XPath, because all the library calls are contained inside their own expr element.

@AshesITR Thanks!  I'll try [(//element[@name='D'])[last()]](https://stackoverflow.com/a/1459262) as this post suggests to see if that gives me the last node, rather than the last element within each node.

I'm still unclear why it returns the expected result on my end and fails when running through ci/cd here.  It makes it so the only way I can test is to push updates and check the actions, which is probably destroying your email inbox with fail messages. Sorry about this.

nicholas-masel:
> > probably destroying your email inbox with fail messages
> 
> FWIW I don't get those. only the code coverage bot comments when CI passes.

Good to know, thanks!  Now I don't feel so bad hacking through it this way :)

MichaelChirico:
> I'm still unclear why it returns the expected result on my end and fails when running through ci/cd here

could be an R version thing? there are subtle differences in the XML structure between versions sometimes. To help, you could add `writeLines(as.character(XML))` to facilitate print debugging.

nicholas-masel:
> writeLines(as.character(XML))

Great idea!  I'm not clear how would I retrieve this from the action?  Can I access this somehow after the action is complete?

nicholas-masel:
> > writeLines(as.character(XML))
> 
> Great idea! I'm not clear how would I retrieve this from the action? Can I access this somehow after the action is complete?

@MichaelChirico nvm!  This defaults to standard out!

R/library_call_linter.R
```diff
@@ -0,0 +1,60 @@
+#' Library call linter
+#'
+#' Force library calls to all be at the top of the script.
+#'
+#' @examples
+#'  # will produce lints
+#'  lint(
+#'    text = c("library(dplyr)", "print('test')", "library(tidyr)"),
```
AshesITR, line 8: We usually use multiline text instead of character vectors. We should stay consistent within the docs.
Can you refactor to 

```
text = "
  library(dplyr)
  print('test')
  library(tidyr)
"
```

or similar?

R/library_call_linter.R
```diff
@@ -0,0 +1,60 @@
...
+#'    linters = library_call_linter()
+#'  )
+#'
+#' @evalRd rd_tags("library_call_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+library_call_linter <- function() {
+
+  xpath <- "
+  (//SYMBOL_FUNCTION_CALL[text() = 'library'])[last()]
```
AshesITR, line 34: This can be indented by 2 spaces without changing the logic. That should improve readability a little.

R/library_call_linter.R
```diff
@@ -0,0 +1,60 @@
...
+  //preceding::expr
```
AshesITR, line 35: The // throws away the scope by restarting from the root element. Are you sure you don't want a single /?

tests/testthat/test-library_call_linter.R
```diff
@@ -0,0 +1,51 @@
...
+    library_call_linter()
+  )
+
+  expect_lint(c("print('test')", "# library(dplyr)"),
+    NULL,
+    library_call_linter()
+  )
+})
+
+test_that("library_call_linter warns on disallowed usages", {
```
AshesITR, line 24: Maybe add a test that also checks the location metadata?

R/library_call_linter.R
nicholas-masel: Sure thing.  I'll cover this is my next commit!

R/library_call_linter.R
nicholas-masel: Sure thing.  I'll cover this is my next commit!

R/library_call_linter.R
nicholas-masel: Thanks, yep, that was not my intention. I definitely misunderstood the XPATH documentation here.  I'll make the update to / here and for line 36 and 37.

tests/testthat/test-library_call_linter.R
```diff
@@ -0,0 +1,51 @@
+test_that("library_call_linter skips allowed usages", {
+
+  expect_lint(c("library(dplyr)", "print('test')"),
```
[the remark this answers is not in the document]
nicholas-masel: Will do!  Let me know if it still isn't what you expected and I can adjust again.

tests/testthat/test-library_call_linter.R
nicholas-masel: I thought about it and should have done it to start with 😄.  I'll add it for those that return more than one lint.

AshesITR (approved):

AshesITR:
Good work.
Don't forget to mention yourself in a NEWS bullet.

inst/lintr/linters.csv
```diff
@@ -42,6 +42,7 @@ infix_spaces_linter,style readability default configurable
 inner_combine_linter,efficiency consistency readability
 is_numeric_linter,readability best_practices consistency
 lengths_linter,efficiency readability best_practices
+library_call_linter,style best_practices
```
MichaelChirico, line 45: this counts as `readability` to me too

MichaelChirico (commented):
You might want to add an exception for `library()` calls inside functions -- e.g. for libraries calling `library()` on behalf of the user. The current logic seems tailored to scripts, but the linter will end up being used in packages too.

It would also be easy to consider `require()` the same as `library()` for this linter.

inst/lintr/linters.csv
nicholas-masel: This one is updated.

nicholas-masel:
> You might want to add an exception for `library()` calls inside functions -- e.g. for libraries calling `library()` on behalf of the user. The current logic seems tailored to scripts, but the linter will end up being used in packages too.
> 
> It would also be easy to consider `require()` the same as `library()` for this linter.

@MichaelChirico Would you be open to completing this PR as is and adding this as a new issue?  I'm happy to continue to work on it to add these features, but it would allow me to get this functionality in so I can have another package use it for it's upcoming release.

MichaelChirico:
SGTM, please help to file follow-up issues

MichaelChirico (approved):

PR #2030, opened 2023-08-01
keyword_quote_linter for unnecessary quoting

MichaelChirico:
Continuing on with #884.

There are a few customizations/options we could add, LMK if we should do those here or in follow-up issues.


MichaelChirico:
Draft: review #2031 first

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,129 @@
...
+    xml <- source_expression$xml_parsed_content
+
+    call_arg_expr <- xml_find_all(xml, call_arg_xpath)
+
+    invalid_call_quoting <- is_valid_r_name(get_r_string(call_arg_expr))
+
+    call_arg_lints <- xml_nodes_to_lints(
+      call_arg_expr[invalid_call_quoting],
+      source_expression = source_expression,
+      lint_message = "Only quote named arguments to functions if necessary, i.e., the name is not a valid R symbol (see ?make.names).",
```
AshesITR, line 66: > ,i.e., _if_ the name ...

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,129 @@
...
+      type = "warning"
+    )
+
+    assignment_expr <- xml_find_all(xml, assignment_xpath)
+
+    invalid_assignment_quoting <- is_valid_r_name(get_r_string(assignment_expr), no_quote = TRUE)
+
+    assignment_lints <- xml_nodes_to_lints(
```
AshesITR, line 74: I think it's worth splitting the cases here

> Use ` for quoting names, not %s.

And the other message for unnecessary quoting can be same as above.

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,129 @@
...
+    extraction_expr <- xml_find_all(xml, extraction_xpath)
+
+    invalid_extraction_quoting <-
+      is_valid_r_name(get_r_string(extraction_expr), no_quote = TRUE)
+
+    extraction_expr <- extraction_expr[invalid_extraction_quoting]
+    extractor <- xml_find_chr(extraction_expr, "string(preceding-sibling::*[1])")
+    gen_extractor <- ifelse(extractor == "$", "[[", "slot()")
+
+    extraction_lints <- xml_nodes_to_lints(
```
AshesITR, line 93: Same here.

> Use ` for quoted names or [[ to extract by string.

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,129 @@
...
+#' See [make.names()] for the description of syntactically valid names in R.
+#'   we could also replace this with `make.names(x) == x`
+#' @noRd
+is_valid_r_name <- function(x, no_quote = FALSE) {
+  if (no_quote) {
+    bad_quote <- !startsWith(x, "`")
+  } else {
+    bad_quote <- FALSE
+  }
+  is_valid_symbol <- grepl("^([a-zA-Z][a-zA-Z0-9._]*|[.]|[.][a-zA-Z._][a-zA-Z0-9._]*)$", x)
```
AshesITR, line 127: This regex can be shortened to

`^([a-zA-Z]|[.]$|[.][a-zA-Z._])[a-zA-Z0-9._]*$`

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,129 @@
...
+  return(bad_quote | (is_valid_symbol & !(x %in% r_reserved_words)))
```
AshesITR, line 128: Unnecessary return call.

R/keyword_quote_linter.R
MichaelChirico: that nested `$` is messy... let's just use `make.names()` to offshore the logic.

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,162 @@
...
+      call_arg_expr[invalid_call_quoting],
+      source_expression = source_expression,
+      lint_message = paste("Only quote named arguments to functions if necessary,", clarification),
+      type = "warning"
+    )
+
+    assignment_expr <- xml_find_all(xml, assignment_xpath)
+
+    invalid_assignment_quoting <- is_valid_r_name(get_r_string(assignment_expr))
+    assignment_to_string <- xml_name(xml_find_first(assignment_expr, "*")) == "STR_CONST"
```
AshesITR, line 111: Use `xml_children()[[1L]]` instead of an XPath?

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,162 @@
...
+    "starts-with(text(), '\"')",
+    "starts-with(text(), '`')",
+    'starts-with(text(), "\'")'
+  )
+  # SYMBOL_SUB for backticks, STR_CONST for quoted names
+  call_arg_xpath <- glue("
+  //SYMBOL_FUNCTION_CALL
+    /parent::expr
+    /parent::expr
+    /*[(self::SYMBOL_SUB or self::STR_CONST) and {quote_cond}]
```
AshesITR, line 69: I suspect this will also catch STR_CONST passed as arguments.

Maybe `/EQ_SUB/preceding-sibling::*[{quote_cond}]` instead?

Tests should also cover calls without named arguments and with string values as arguments to make sure this works as intended.

MichaelChirico: Added tests to be sure / prevent regression. FWIW, this series has been tested against 10K+ files in our code base already, so common cases are not likely to slip through.

Recall that for formals, the LHS of `EQ_SUB` doesn't get its own `<expr>`, but the RHS does, so we'll see `c("A" = "a")` look like

```xml
<expr>
  <STR_CONST>"A"</STR_CONST>
  <EQ_SUB>=</EQ_SUB>
  <expr>
    <STR_CONST>"a"</STR_CONST>
  </expr>
</expr>
```

R/keyword_quote_linter.R
MichaelChirico: Nice, much cleaner

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,162 @@
...
+  ")
+
+  # also exclude $ or @, which are handled below
+  assignment_xpath <- "
+  (//EQ_ASSIGN | //LEFT_ASSIGN[text() != ':='])
```
AshesITR, line 74: Is it worth supporting RIGHT_assign, too?

R/keyword_quote_linter.R
```diff
@@ -0,0 +1,162 @@
...
+      call_arg_expr[invalid_call_quoting],
+      source_expression = source_expression,
+      lint_message = paste("Only quote named arguments to functions if necessary,", clarification),
+      type = "warning"
+    )
+
+    assignment_expr <- xml_find_all(xml, assignment_xpath)
+
+    invalid_assignment_quoting <- is_valid_r_name(get_r_string(assignment_expr))
+    assignment_to_string <- xml_name(xml2::xml_children(assignment_expr)) == "STR_CONST"
```
AshesITR, line 111: Worth a comment that we know `xml_children()` returns exactly one child because all `assignment_expr` results are `expr` nodes containing exactly one of SYMBOL or STR_CONST?

R/keyword_quote_linter.R
MichaelChirico: Sure, though it's not as simple as just adding `//RIGHT_ASSIGN` since we need to change the sibling direction. Just a few lines.

R/keyword_quote_linter.R
MichaelChirico: Just changed to `xml_child()` instead, it is by default a wrapper to `xml_children(.)[[1]]`

AshesITR:
I was stumped and debugged.
Turns out the actual problem is `xml_child()` is not vectorized for nodesets:
image

The correct code should be

```r
assignment_to_string <- vapply(assignment_expr, function(x) xml_name(xml2::xml_child(x)) == "STR_CONST", logical(1L))
```

MichaelChirico:
Thanks for checking. Are you suggesting we shouldn't use `xml_children()`? Your workaround looks a lot clunkier given that the code already works as intended with `xml_children()`

AshesITR:
I'm suggesting to fix the comment.
It's actually not `<= 1` but `= 1` children guaranteed by the XPath, making `xml_children()` correct.
The empty case (0 elements) occurs when there are no matches for the XPath, where `xml_children()` also behaves correctly by returning `xml_missing()`.

MichaelChirico:
oh, but that's what I meant -- 1 if there's a match, 0 if not, so `<= 1`, and it works in both cases. Let me reword for clarity.

AshesITR:
The part about xml_child is wrong. It's simply not vectorized over nodesets so it would return the first child of the first match instead of the first child for each element in the resultset.

MichaelChirico:
Ohh I get you now. even better.

NEWS.md
```diff
@@ -25,6 +25,7 @@
 
 ### New linters
 
+* `keyword_quote_linter()` for finding unnecessary or discouraged quoting of symbols in assignment, function arguments, or extraction (part of #884, @MichaelChirico). Quoting is unnecessary when the target is a valid R name, e.g. `c("a" = 1)` can be `c(a = 1)`. The same goes to assignment (`"a" <- 1`) and extraction (`x$"a"`). Where quoting is necessary, the linter encourages doing so with backticks (e.g. `` x$`a b` `` instead of `x$"a"`).
```
AshesITR, line 28: The last code snippet should be `x$"a b"`.

man/length_levels_linter.Rd
```diff
@@ -8,6 +8,20 @@ length_levels_linter()
 }
 \description{
 \code{length(levels(x))} is the same as \code{nlevels(x)}, but harder to read.
+}
+\examples{
```
AshesITR, line 12: I wonder how that slipped...

MichaelChirico: I must have forgotten to `roxygenize()` after adding the examples in the other PR...

AshesITR (approved):

PR #2031, opened 2023-08-01
Add ubiquitously-called internal functions to NAMESPACE

MichaelChirico:
Split off from #2030 for clarity.

These calls are used so much it's odd we wouldn't put them in the NAMESPACE. After that, it's redundant IMO to always namespace-qualify them.


AshesITR:
There are some stray diffs in .Rd files here.

R/indentation_linter.R
```diff
@@ -162,10 +162,10 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
     "number(",
     paste(
       c(
-        glue::glue("self::{paren_tokens_left}/following-sibling::{paren_tokens_right}/preceding-sibling::*[1]/@line2"),
-        glue::glue("self::*[{xp_and(paste0('not(self::', paren_tokens_left, ')'))}]
+        glue("self::{paren_tokens_left}/following-sibling::{paren_tokens_right}/preceding-sibling::*[1]/@line2"),
+        glue("self::*[{xp_and(paste0('not(self::', paren_tokens_left, ')'))}]
                       /following-sibling::SYMBOL_FUNCTION_CALL/parent::expr/following-sibling::expr[1]/@line2"),
```
AshesITR, line 167: XPath indentation is off here now.

R/indentation_linter.R
```diff
@@ -178,7 +178,7 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
   global_nodes <- function(nodes) paste0("//", nodes, collapse = "|")
   xp_indent_changes <- paste(
     c(
-      glue::glue("//{paren_tokens_left}[not(@line1 = following-sibling::expr[
+      glue("//{paren_tokens_left}[not(@line1 = following-sibling::expr[
                     @line2 > @line1 and
```
AshesITR, line 182: XPath indentation

R/paste_linter.R
```diff
@@ -144,22 +144,22 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE) {
         source_expression = source_expression,
         lint_message = paste(
           'toString(.) is more expressive than paste(., collapse = ", ").',
-          "Note also glue::glue_collapse() and and::and()",
```
AshesITR, line 147: I wouldn't change this lint message.

R/lintr-package.R
```diff
@@ -8,11 +8,11 @@
 "_PACKAGE"
 
 ## lintr namespace: start
-#' @importFrom glue glue
+#' @importFrom glue glue glue_collapse glue_data
 #' @importFrom rex rex regex re_matches re_substitutes character_class
 #' @importFrom stats na.omit
 #' @importFrom utils capture.output head getParseData relist
-#' @importFrom xml2 xml_find_all as_list
+#' @importFrom xml2 xml_attr xml_attrs xml_find_all xml_find_chr xml_find_num xml_find_first xml_text as_list
```
AshesITR, line 15: `xml_attrs()` is only used twice in a single line. Not worth importing IMO.
Similiarly, `glue_data()` only has one usage.

MichaelChirico: I think there's no cost, but might keep on namespace-qualifying.

OTOH, I think the main advantage of doing so is readability, I'm not sure any contributor can get too tripped up guessing where `xml*` and `glue*` functions come from internally (`xmlparsedata::xml_parse_data()` is a good counterexample where we should definitely qualify).

Anyway I don't feel strongly; added back the qualification.

AshesITR: Can you remove the import as well?

AshesITR:
I've checked all changes. LGTM except for the unnecessary namespace imports.

R/lintr-package.R
MichaelChirico: I would leave it in. I think it's better for dependencies to be incurred at build time than run time, where appropriate. I don't think the current NAMESPACE is exhaustive in that regard, but this is a good start.

AshesITR: Then I'd rather not ns-qualify their usages, it seems counter-intuitive to have an unused explicit import.

MichaelChirico: OK... I get that it feels redundant. In this case, I think of `::` as being an emphasis for readability. "Yes, we imported it, but just in case you forgot, here's where it came from". Seems appropriate for "uncommon" or infrequent calls.

Anyway, I'll just remove the imports, since I do think the readability aspect is more important in this case.

AshesITR (approved):

PR #2039, opened 2023-08-04
also skip slot extractions akin to dollar

MichaelChirico:
Noticed this during #2038.

Figured it would affect other linters too, so I `grep`'d through for `OP-DOLLAR` to determine which others to include.


NEWS.md
```diff
@@ -8,6 +8,14 @@
 ## New and improved features
 
 * `library_call_linter()` can detect if all library calls are not at the top of your script (#2027, @nicholas-masel).
+* Several linters avoiding false positives in `$` extractions get the same exceptions for `@` extractions, e.g. `S4@T` will no longer throw a `T_and_F_symbol_linter()` hit (@MichaelChirico).
```
AshesITR, line 11: Add PR number?

R/for_loop_index_linter.R
```diff
@@ -34,10 +34,7 @@ for_loop_index_linter <- function() {
   //forcond
     /SYMBOL[text() =
       following-sibling::expr
-        //SYMBOL[not(
-          preceding-sibling::OP-DOLLAR
-          or parent::expr[preceding-sibling::OP-LEFT-BRACKET]
-        )]
+        //SYMBOL[not(parent::expr[OP-DOLLAR or OP-AT or preceding-sibling::OP-LEFT-BRACKET])]
```
AshesITR, line 37: The `//` seems off to me. Shouldn't this be a single slash?

R/object_name_linter.R
```diff
@@ -10,7 +10,7 @@ object_name_xpath <- local({
   #   is not possible for strings, though we do still have to
   #   be aware of cases like 'a$"b" <- 1'.
   xp_assignment_target_fmt <- paste0(
-    "not(preceding-sibling::OP-DOLLAR)",
+    "not(parent::expr[OP-DOLLAR or OP-AT])",
```
AshesITR, line 13: I would have expected this to change behavior on `badName$a <- 3`. But the tests have us covered 😃

tests/testthat/test-T_and_F_symbol_linter.R
```diff
@@ -26,6 +26,9 @@ test_that("T_and_F_symbol_linter blocks disallowed usages", {
     linter
   )
 
+  expect_lint("DF$bool <- T", msg_true, linter)
```
AshesITR, line 29: also test with `@`?

R/for_loop_index_linter.R
MichaelChirico: I think the intent is "don't overwrite any symbol in the parent env":

https://github.com/r-lib/lintr/blob/bb7669afa4b90843e8358397e60d02e2fc512953/tests/testthat/test-for_loop_index_linter.R#L19

Maybe this isn't the best approach but it seems to work OK. Of course there'll be false positives from NSE column names... Let's leave as is until an issue is filed?

R/object_name_linter.R
MichaelChirico: yep! different `<expr>`. I think the parser is being "smart" about not dropping the RHS of `$`/`@` into a child `<expr>` because only `STR_CONST`/`SYMBOL` are parseable. But this subverts our usual expectation about where to find `SYMBOL` nodes, hence clauses like this.

R/for_loop_index_linter.R
AshesITR: I'm expecting this to trigger a false positive for the expression

```r
foo <- function() { # force single expression
  for (i in 1:10) {
    42L
  }
  i <- 7L
}
```

I think instead using `/descendant::expr` will restrict (correctly) to the RHS of the forcond.

Not sure though, I'm on my phone.
We can leave this in and file a separate issue if my suspicion is correct since this PR has a different scope and the old code had a // as well.

MichaelChirico: no lint for that, just checked. i'll add it as a regression test. I think `//` is the same as `.//` inside `[`. I will use `.//` to make it clearer.

MichaelChirico: oh wait, read more closely. Note that it's

```xpath
following-sibling::expr
  //SYMBOL
```

i.e. `following-sibling::expr//SYMBOL`

AshesITR (approved):

PR #2046, opened 2023-08-04
Consistent logic around magrittr pipes

### LINKED ISSUE #2008: `unnecessary_concatenation_linter` wrongly complains about usage with magrittr's assignment pipe operator `%<>%`
Consider the following reprex:

``` r
tmp_file <- tempfile(fileext = "R")
writeLines(con = tmp_file,
           text = "
library(magrittr)
x <- 1:2
x %<>% c(3)
x
")

lintr::lint(filename = tmp_file,
            linters = list(lintr::unnecessary_concatenation_linter()))
#> /tmp/RtmpyaBMnL/filee0674c13fd78R:4:8: style: [unnecessary_concatenation_linter] Unneeded concatenation of a constant. Remove the "c" call.
#> x %<>% c(3)
#>        ^~~~
```

<sup>Created on 2023-07-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

`x %<>% c(3)` above is the [magrittr way](https://magrittr.tidyverse.org/reference/compound.html) to write `x <- c(x, 3)`, so lintr should ideally allow `c()` with a single constant *if* it is preceded by the `%<>%` operator.

MichaelChirico:
Closes #2008


R/pipe_call_linter.R
```diff
@@ -33,12 +34,16 @@ pipe_call_linter <- function() {
...
       source_expression = source_expression,
-      lint_message = "Use explicit calls in magrittr pipes, i.e., `a %>% foo` should be `a %>% foo()`.",
+      lint_message =
+        sprintf("Use explicit calls in magrittr pipes, i.e., `a %1$s foo` should be `a %1$s foo()`.", pipe),
       type = "warning"
     )
   })
 }
+
+magrittr_pipes <- c("%>%", "%!>%", "%T>%", "%$%", "%<>%")
```
AshesITR, line 49: Since it's re-used in other linters, this should probably go in a central file.

tests/testthat/test-pipe_continuation_linter.R
```diff
@@ -188,3 +188,21 @@ local({
     code_string = valid_code
   )
 })
+
+local({
+  linter <- pipe_continuation_linter()
+  pipes <- c("%>%", "%!>%", "%T>%", "%$%", "%<>%")
+  cases <- expand.grid(pipe1 = pipes, pipe2 = pipes, stringsAsFactors = FALSE)
+  cases <- within(cases, {
+    .test_name <- sprintf("(%s, %s)", pipe1, pipe2)
```
AshesITR, line 197: `pipe2 = %<>%` makes no sense IINM.

tests/testthat/test-unnecessary_concatenation_linter.R
```diff
@@ -45,18 +45,26 @@ test_that("unnecessary_concatenation_linter blocks disallowed usages", {
   )
 })
 
-test_that("Correctly handles concatenation within magrittr pipes", {
+local({
+  pipes <- c("%>%", "%!>%", "%T>%", "%<>%")
```
AshesITR, line 49: Does the backpipe support strings as symbols?

`a <- 1; "a" %<>% c("b")` does what?

tests/testthat/test-unnecessary_placeholder_linter.R
```diff
@@ -14,8 +14,8 @@ patrick::with_parameters_test_that(
     # . used inside a scope --> ok
     expect_lint(sprintf("x %s { foo(arg = .) }", pipe), NULL, linter)
   },
-  .test_name = c("forward", "assignment", "tee"),
-  pipe = c("%>%", "%<>%", "%T>%")
+  .test_name = c("forward", "assignment", "tee", "greedy", "extraction"),
```
AshesITR, line 17: Test names could be made consistent with the other tests.

tests/testthat/test-yoda_test_linter.R
```diff
@@ -37,10 +37,15 @@ test_that("yoda_test_linter ignores strings in $ expressions", {
 })
 
 # if we only inspect the first argument & ignore context, get false positives
-test_that("yoda_test_linter ignores usage in pipelines", {
-  expect_lint("foo() %>% expect_identical(2)", NULL, yoda_test_linter())
-  skip_if_not_r_version("4.1.0")
-  expect_lint("bar() |> expect_equal('a')", NULL, yoda_test_linter())
+local({
+  pipes <- c("%>%", "%!>%", "%T>%", "%<>%", if (getRversion() >= "4.1.0") "|>")
```
AshesITR, line 41: backpipe makes no sense here.

tests/testthat/test-pipe_continuation_linter.R
MichaelChirico: IIRC I've seen it used (possibly by mistake) before we cleaned up `%<>%` usage a while back. On balance, can't hurt.

R/pipe_call_linter.R
MichaelChirico: I was looking around for such a file, seems we don't have one yet. Would apply to other internal constants as well, e.g.

https://github.com/r-lib/lintr/blob/bb7669afa4b90843e8358397e60d02e2fc512953/R/expect_type_linter.R#L76

https://github.com/r-lib/lintr/blob/bb7669afa4b90843e8358397e60d02e2fc512953/R/object_name_linter.R#L1

Follow-up, or I'll make that sweep in this PR?

tests/testthat/test-unnecessary_concatenation_linter.R
MichaelChirico: Quite surprisingly it does the same thing as symbol `a` would do!

```r
a <- 1; "a" %<>% c("b")
a
# [1] "a" "b"
```

R/pipe_call_linter.R
AshesITR: Follow up for the other constants to keep the diff smaller?

tests/testthat/test-pipe_continuation_linter.R
AshesITR: Is that even functional code?

tests/testthat/test-yoda_test_linter.R
MichaelChirico: pretty baroque for sure... is there a cost to keeping it in?

tests/testthat/test-unnecessary_concatenation_linter.R
AshesITR: Wow that's surprising behavior to say the least 🙈
I wouldv'e at least expected `a <- c("1", "b")`...

R/pipe_call_linter.R
MichaelChirico: #2048

tests/testthat/test-unnecessary_concatenation_linter.R
MichaelChirico: You're right, it's even more surprising than my first understanding!

I guess it gets evaluated as

```r
"a" <- c("a", "b")
```

which fits the pattern of

```r
<expr> %<>% foo(y)
# equivalent to
<expr> <- foo(<expr>, y)
```

A great example of why `%<>%` should be avoided...

tests/testthat/test-yoda_test_linter.R
AshesITR: I'm just not sure it's worth testing such a piece of code that couldn't be functional even in theory.
That said, it's probably slight overkill to test all pipes anyway.

tests/testthat/test-unnecessary_concatenation_linter.R
AshesITR: Maybe a feature for a pipe linter to warn against using `%<>%` with non-SYMBOL LHS?

MichaelChirico: For now I'm content with `%<>%` being covered by `assignment_linter()` to dissuade using it at all. Seems the simplest option.

tests/testthat/test-pipe_continuation_linter.R
MichaelChirico: No, it works:

```r
x <- 1L
x %<>% c(2L) %<>% c(3L)
x
# [1] 1 2 3
```

Can't say I understand the mechanism for it to do so!

tests/testthat/test-yoda_test_linter.R
MichaelChirico: I see what you mean more on second look, at first I convinced myself we could come up with an example that works.

But such an example would have to be like

```r
x <- "a"
"a" %<>% expect_equal(x)
expect_equal(a, "a")
```

Very peculiar.

tests/testthat/test-unnecessary_placeholder_linter.R
MichaelChirico: Shared logic in helpers.R now, PTAL

tests/testthat/test-yoda_test_linter.R
MichaelChirico: Removed `%<>%` from consideration for `yoda_test_linter()`

tests/testthat/test-pipe_continuation_linter.R
AshesITR: Seems like the second `%<>%` is silently treated as `%>%`.

I remember looking at magrittrs implementation at some time in the past and it was one generic implementation for all the pipes which deparsed the symbols to make sense of the full pipeline at once...

Nevermind then.

NAMESPACE
```diff
@@ -157,3 +157,5 @@ importFrom(utils,relist)
 importFrom(utils,tail)
 importFrom(xml2,as_list)
 importFrom(xml2,xml_find_all)
+importFrom(xml2,xml_find_first)
```
AshesITR, line 160: how did this end up in the diff?

tests/testthat/test-unnecessary_placeholder_linter.R
AshesITR: looks good

tests/testthat/test-unnecessary_placeholder_linter.R
```diff
@@ -34,6 +34,6 @@ patrick::with_parameters_test_that(
       unnecessary_placeholder_linter()
     )
   },
-  .test_name = c("forward", "assignment", "tee"),
-  pipe = c("%>%", "%<>%", "%T>%")
+  .test_name = names(pipes(exclude = "|>")),
```
AshesITR, line 37: Any reason this test doesn't use the `local({ pipes <- ... })` pattern?

NAMESPACE
MichaelChirico: intentional, note the corresponding diff in lintr-package.R.

Just to avoid merge conflicts / possibly forgetting to remove the :: on merge later, since they're used in the diff here.

tests/testthat/test-unnecessary_placeholder_linter.R
MichaelChirico: I do prefer avoiding it, but with the new change it's probably better to switch

tests/testthat/test-unnecessary_placeholder_linter.R
```diff
@@ -14,26 +14,29 @@ patrick::with_parameters_test_that(
     # . used inside a scope --> ok
     expect_lint(sprintf("x %s { foo(arg = .) }", pipe), NULL, linter)
   },
-  .test_name = c("forward", "assignment", "tee"),
-  pipe = c("%>%", "%<>%", "%T>%")
+  .test_name = c("forward", "assignment", "tee", "greedy", "extraction"),
+  pipe = c("%>%", "%<>%", "%T>%", "%!>%", "%$%")
```
AshesITR, line 18: This one should use `pipes()` as well.

tests/testthat/test-unnecessary_placeholder_linter.R
```diff
@@ -1,8 +1,9 @@
+linter <- unnecessary_placeholder_linter()
```
AshesITR, line 1: Defining `linter` at the top level lets it leak out of the test file, potentially causing confusing error messages down the line if another test forgets to define the linter.

I think we should refrain from that.

MichaelChirico: I don't think that's so -- my understanding is each file is run in its own environment.

I tested this like so:

```r
writeLines(con  = "test-a.R", '
a = 1

test_that("a = 1", {
  expect_equal(a, 1)
})')

writeLines(con = "test-b.R", '
stopifnot(!exists("a"))

b = 3
test_that("b = 3", {
  expect_equal(b, 3)
})')

testthat::test_dir(".")
✔ | F W S  OK | Context
✔ |         1 | a [0.1s]                                                                                                                                                                                                                                                                  
✔ |         1 | b                                                                                                                                                                                                                                                                         

══ Results ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 0.2 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2 ]
```

AshesITR: How are we including `helper.R` to make it survive? 😅

MichaelChirico: That's done here AIUI

https://github.com/r-lib/testthat/blob/4c6bf287f59d5e48470d267f5792cd0eeb4af319/R/test-files.R#L210

```r
writeLines("c = 1", "helper.R")

writeLines(con  = "test-a.R", '
stopifnot(exists("c"))
a = 1

test_that("a = 1", {
  expect_equal(a, 1)
})
')

writeLines(con = "test-b.R", '
stopifnot(exists("c"))
stopifnot(!exists("a"))

b = 3
test_that("b = 3", {
  expect_equal(b, 3)
})
')

testthat::test_dir(".")
# ✔ | F W S  OK | Context
# ✔ |         1 | a [0.2s]                                                                                                                                                                                                    
# ✔ |         1 | b                                                                                                                                                                                                           

# ══ Results ═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
# Duration: 0.3 s

# [ FAIL 0 | WARN 0 | SKIP 0 | PASS 2 ]
```

AshesITR: Ah, interesting. Thanks for the enlightenment.

MichaelChirico: Related -- I had asked if {withr} would support a `local*` function to make this clearer:

https://github.com/r-lib/withr/issues/205

Seems they recommend `rlang::local_bindings()` to do that instead -- happy to make that change if you think it'll make it clearer rather than relying on this testthat behavior which is not well-known / particularly well-documented.

(will merge for now)

PR #2049, opened 2023-08-07
Support for piping into `sprintf()` in `sprintf_linter()`

### LINKED ISSUE #1943: sprintf_linter is not pipe-aware
```r
letters |>
  sprintf(fmt = "%s")
```

will throw a lint "too few arguments". Ditto for using `%>%`.

We can probably re-use the code from `unnecessary_concatentation_linter()`.

MichaelChirico:
Closes #1943


MichaelChirico:
It would be slightly easier/safer to run `eval()` on the expression including the pipe (i.e. instead of `sprintf(fmt = '%s')`, `x %>% sprintf(fmt = '%s')`), but this would induce an implicit dependency on {magrittr} which AFAICT we don't have as of now.

We _do_ have an implicit _suggested_ dependency on {magrittr} via {testthat}, so we could try something like

"if {magrittr} available, run `eval()`, else try the injection approach" but I'm not sure it's worth maintaining that complexity.

tests/testthat/test-sprintf_linter.R
```diff
@@ -91,3 +91,10 @@ test_that("edge cases are detected correctly", {
     linter
   )
 })
+
+test_that("piping into sprintf works", {
+  linter <- sprintf_linter()
+
+  expect_lint("x %>% sprintf(fmt = '%s')", NULL, linter)
+  expect_lint("x %>% sprintf(fmt = '%s%s')", "too few arguments", linter)
```
AshesITR, line 99: Maybe also add a test for the format being piped?

`"%s" %>% sprintf("%s%s")` shouldn't lint for example.

R/sprintf_linter.R
```diff
@@ -59,57 +118,3 @@ sprintf_linter <- function() {
...
-#' Try running a static sprintf() call to determine whether it will produce warnings or errors due to format
-#' misspecification
-#'
-#' @param xml An XML node representing a `sprintf()` call (i.e. the `<expr>` node containing the call)
-#'
-#' @return A string, either `NA_character_` or the text of generated errors and warnings from the `sprintf()` call when
-#' replacing all dynamic components by 0, which is compatible with all format specifiers.
-#'
-#' @noRd
-capture_sprintf_warning <- function(xml) {
```
AshesITR, line 105: Moving the functions inside cleans up the namespace, but why remove the roxygen markup?

AshesITR:
> It would be slightly easier/safer to run `eval()` on the expression including the pipe (i.e. instead of `sprintf(fmt = '%s')`, `x %>% sprintf(fmt = '%s')`), but this would induce an implicit dependency on {magrittr} which AFAICT we don't have as of now.
> 
> 
> 
> We _do_ have an implicit _suggested_ dependency on {magrittr} via {testthat}, so we could try something like
> 
> 
> 
> "if {magrittr} available, run `eval()`, else try the injection approach" but I'm not sure it's worth maintaining that complexity.

I don't think the complexity of manually piping is that bad. OTOH it seems a valid assumption that magrittr is available when we are presented with code that contains a magrittr pipe.

R/sprintf_linter.R
MichaelChirico: it will get included in the parent function's roxygen, iinm. that's how R6 documentation works... let me confirm

tests/testthat/test-sprintf_linter.R
MichaelChirico: Not sure we'll catch the second case, but nor do we at HEAD... seems unlikely to come up often, and not easy to evaluate nested piping statically.

I'd push forward with the change as is for now.

MichaelChirico:
> OTOH it seems a valid assumption that magrittr is available when we are presented with code that contains a magrittr pipe.

For package authors, I'd say yes, for "repo maintainers" less so. E.g. for us, we have a binary that takes a single source file as input and runs throughout the whole code base, so it doesn't have much state (e.g. we have `object_usage_linter()` disabled for this reason). IINM we'd have to add `magrittr` as a dependency to that binary.

R/sprintf_linter.R
MichaelChirico: Yep, if I restored the `#'` above `zap_extra_args()`, we get this roxygen2 warning from `parse_file("R/sprintf_linter.R")`:

```
Warning message:
[sprintf_linter.R:28] @export must be a single line, not 3
ℹ The first line is "Zap sprintf() call to contain only constants" 
```

because the `#' @export` line terminating the `sprintf_linter` assignment is being extended to include the beginning of the internal function's header.

MichaelChirico:
> I don't think the complexity of manually piping is that bad.

I agree, but I worry about edge cases (e.g. we don't do anything about placeholders yet), `eval()` would remove responsibility from us getting it all right.

tests/testthat/test-sprintf_linter.R
MichaelChirico: > `x |> sprintf("a") # should be ignored`

this was an error, fixed now

AshesITR: No need to catch it but does it at least not cause errors?

MichaelChirico: It's treated the same as `sprintf(x, "a")`, which we skip b/c it can't be analyzed statically:

https://github.com/r-lib/lintr/blob/2e141e95ac6c58751f43ecf58ab75f4fe465641c/tests/testthat/test-sprintf_linter.R#L105

MichaelChirico: For `"%%sb" |> sprintf("%s") |> sprintf("a")`, it doesn't throw an error (it lints once, on the middle `sprintf()`), but I hesitate to put a test down as it sort-of commits us to whatever behavior is there (which is probably wrong)

AshesITR:
But zapping constants to 0L will not work with the eval approach if we don't.first decide whether the zapped piece of code will end up as the fmt argument or not?

tests/testthat/test-sprintf_linter.R
MichaelChirico: Added an `expect_no_error()` test for the latter case.

AshesITR: > For `"%%sb" |> sprintf("%s") |> sprintf("a")`, it doesn't throw an error (it lints once, on the middle `sprintf()`), but I hesitate to put a test down as it sort-of commits us to whatever behavior is there (which is probably wrong)

That is actually correct linting for this example. What do you feel is wrong?

tests/testthat/test-sprintf_linter.R
```diff
@@ -91,3 +91,18 @@ test_that("edge cases are detected correctly", {
...
+  linter <- sprintf_linter()
+
+  expect_lint("x %>% sprintf(fmt = '%s')", NULL, linter)
+  # no fmt= specified -> this is just 'sprintf("%s", "%s%s")', which won't lint
+  expect_lint('"%s" %>% sprintf("%s%s")', NULL, linter)
+  expect_lint("x %>% sprintf(fmt = '%s%s')", "too few arguments", linter)
+
+  skip_if_not_r_version("4.1.0")
+  # Cannot evaluate statically --> skip
+  expect_lint('x |> sprintf("a")', NULL, linter)
```
AshesITR, line 105: This one is so important to test we should run it with `%>%` to avoid accidental skips IMO.
I'd also test all the basic cases with the native pipe as well just in case the XML structure differs at any time in the future.

tests/testthat/test-sprintf_linter.R
MichaelChirico: Hmm I'm actually quite surprised it seems to be working for some simple nested cases:

```r
lint('x |> sprintf(fmt = "%s") |> sprintf(fmt = "%s%s")\n', sprintf_linter())
# <text>:1:29: warning: [sprintf_linter] too few arguments
# x |> sprintf(fmt = "%s") |> sprintf(fmt = "%s%s")
#                             ^~~~~~~~~~~~~~~~~~~~~
lint('x |> sprintf(fmt = "%s%s") |> sprintf(fmt = "%s")\n', sprintf_linter())
# <text>:1:6: warning: [sprintf_linter] too few arguments
# x |> sprintf(fmt = "%s%s") |> sprintf(fmt = "%s")
#      ^~~~~~~~~~~~~~~~~~~~~
```

Are we confident enough to commit these tests to the suite?

AshesITR: Sounds good to me.
It's not that surprising to me if you recall that the output of sprintf is just another zappable argument for the next sprintf call in line unless it yields the fmt argument.

tests/testthat/test-sprintf_linter.R
MichaelChirico: parameterized the test

tests/testthat/test-sprintf_linter.R
```diff
@@ -91,3 +91,42 @@ test_that("edge cases are detected correctly", {
...
+    {
+      expect_lint(paste("x", pipe, "sprintf(fmt = '%s')"), NULL, linter)
+      # no fmt= specified -> this is just 'sprintf("%s", "%s%s")', which won't lint
+      expect_lint(paste('"%s"', pipe, 'sprintf("%s%s")'), NULL, linter)
+      expect_lint(paste("x", pipe, "sprintf(fmt = '%s%s')"), unused_fmt_msg, linter)
+
+      # Cannot evaluate statically --> skip
+      expect_lint(paste("x", pipe, 'sprintf("a")'), NULL, linter)
+      # Nested pipes
+      pipe_width <- nchar(pipe)
```
AshesITR, line 111: unused?

AshesITR (commented):
Basically LGTM

tests/testthat/test-sprintf_linter.R
MichaelChirico: Ah right, vestigial from defining `column_number` with that. I switched to the `nchar()` approach for readability, hopefully it's less arbitrary.

PR #2051, opened 2023-08-07
length_levels_linter() for length(levels(x)) --> nlevels(x)

MichaelChirico:
Part of #884


AshesITR:
There are also `data.table::uniqueN()` and `dplyr::n_distinct()` for the similar `length(unique(x))`. WDYT about optionally linting this here?

MichaelChirico:
I would put that as a separate linter.

The closer analogue is `nested_ifelse_linter()`, whose primary recommended alternatives are `data.table::fcase()` and `dplyr::case_when()`.

But in that case, there _are_ ways to avoid the nested `ifelse()` using base functions, so the user doesn't _strictly_ need any new packages to fix the issue.

For `length(unique(.))`, there's no improvement in {base} (AFAIK).

Doesn't mean we shouldn't offer such a linter (since it'll be optional for users to turn on anyway), just by way of explaining why we didn't write that so far (it's certainly crossed my mind).

AshesITR (approved):

PR #2056, opened 2023-08-08
cleanup internal namespace, use shared_constants.R for shared objects

### LINKED ISSUE #2048: Keep shared internal constants in a central location
I was looking around for such a file, seems we don't have one yet. Would apply to other internal constants as well, e.g.

https://github.com/r-lib/lintr/blob/bb7669afa4b90843e8358397e60d02e2fc512953/R/expect_type_linter.R#L76

https://github.com/r-lib/lintr/blob/bb7669afa4b90843e8358397e60d02e2fc512953/R/object_name_linter.R#L1

Follow-up, or I'll make that sweep in this PR?

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/2046#discussion_r1285082348_

---

As usual the only trouble is naming (for the file). 'aaa.R' and 'constants.R' are the natural choices, any other thoughts/suggestions?

MichaelChirico:
Closes #2048

Approach is to keep _any_ object in a `*linter.R` file besides the linter itself in that linter's own environment, unless that object is already used in >1 source file.

There are exceptions in two files:

https://github.com/r-lib/lintr/blob/f8bb91e7f0fd749f9a28523006a4cedd369c2688/R/backport_linter.R#L123-L170

https://github.com/r-lib/lintr/blob/f8bb91e7f0fd749f9a28523006a4cedd369c2688/R/object_name_linter.R#L214-L247

For backports.R, I did so on "vibes", for object_name_linter.R, those objects are needed outside the linter environment. We could work around this, e.g. by doing `get("style_regex", environment(object_name_linter()))`, but that seems rather abstruse.

I'm also not committed to this particular approach, and would be fine scaling back (e.g. to something as simple as "objects in *linter.R are fine, but shared objects should be moved to the shared file").


MichaelChirico:
`cyclocomp_linter()` on `object_usage_linter()` will take a bit of lift.

Options:

 1. Push forward trying to simplify the complexity of the linter
 2. Restore the helper functions to the lintr namespace to avoid burdening the linter with their complexity

AshesITR:
From a readablility standpoint I feel like we are deteriorating by separating the function signature from its body for the linter factories.

Keeping the dependencies of all files a DAG where the *linter files are sinks seems sound to me.
This would imply moving all shared objects and functions to "util" files while still allowing more complex helper functions used only for a single linter to live in that linters respective file outside of the function factory.

There is also a (probably negligible) performance benefit to only constructing the functions once in lintrs namespace instead of at runtime.

MichaelChirico:
> There is also a (probably negligible) performance benefit to only constructing the functions once in lintrs namespace instead of at runtime.

I was thinking the opposite... by avoiding populating the namespace with names that aren't used (except if running `all_linters()` is common), we improve efficiency by only loading functions lazily.

Anyway I did do a quick `R CMD INSTALL` and `loadNamespace()` comparison and whatever difference there is is indeed negligible (I just saw noise in the timings).

MichaelChirico:
> From a readablility standpoint I feel like we are deteriorating by separating the function signature from its body for the linter factories.

OK, makes sense & agreed. I restored the biggest changes, but kept a few that felt inconsistent with other files.

I also retained some of the refactoring that was done to reduce cyclomatic complexity in the meantime, particularly for `object_usage_linter()`, which I think is a fair amount cleaner now.

R/indentation_linter.R
```diff
@@ -313,15 +309,12 @@ indentation_linter <- function(indent = 2L, hanging_indent_style = c("tidy", "al
...
-  } else if (change_type == "hanging") {
-    hanging_indent
-  } else if (change_type == "double") {
-    current_indent + 2L * indent
-  } else {
+  switch(change_type,
+    suppress = current_indent,
+    hanging = hanging_indent,
+    double = current_indent + 2L * indent,
     current_indent + indent
```
AshesITR, line 316: That's `block =` then.

R/inner_combine_linter.R
AshesITR, line 1: revert this maybe?
No strong feeling since the linter has no arg, just makes the diff larger to review...

R/keyword_quote_linter.R
```diff
@@ -52,6 +52,11 @@
 #   //expr[expr[SYMBOL_FUNCTION_CALL]]/SYMBOL_SUB[starts-with(text(), '`')]
 #   //expr[expr[SYMBOL_FUNCTION_CALL]]/STR_CONST[{is_quoted(text())}]
 keyword_quote_linter <- function() {
+  # Check if a string could be assigned as an R variable.
+  #
+  # See [make.names()] for the description of syntactically valid names in R.
+  is_valid_r_name <- function(x) make.names(x) == x
```
AshesITR, line 58: This is always used with `x = get_r_string(...)`. Worth pulling it inside the function?

R/namespace_linter.R
AshesITR, line 1: Can you separate this? Harder to review if mixed with the other less material changes.

R/inner_combine_linter.R
MichaelChirico: Yea... I think it's a bit off an anomaly compared to other linters' logic. Maybe better to mark it for internal cleanup.

MichaelChirico: #2062

R/namespace_linter.R
MichaelChirico: Done: #2063

R/shared_constants.R
```diff
@@ -0,0 +1,218 @@
...
+    consume_to <- attr(current_match, "match.length")
+    token_content <- substr(static_regex, token_start, consume_to)
+    fixed_string <- paste0(fixed_string, get_token_replacement(token_content, token_type))
+    static_regex <- substr(static_regex, start = consume_to + 1L, stop = nchar(static_regex))
+    current_match <- regexpr(rx_first_static_token, static_regex, perl = TRUE)
+  }
+  paste0(fixed_string, static_regex)
+}
+
+get_token_replacement <- function(token_content, token_type) {
```
AshesITR, line 85: This function could use some `@noRd` roxygen to document what it does.

R/indentation_linter.R
AshesITR: Can you add this? I think it improves clarity to specify all cases.

AshesITR (commented):
Thanks, that diff is more manageable. Almost GTG.

R/indentation_linter.R
MichaelChirico: Oh, I think I added it to the wrong branch maybe... done now

AshesITR (approved):

PR #2063, opened 2023-08-09
cleanup internal namespace, use shared_constants.R for shared objects

MichaelChirico:
Follow-up to #2056 since it pushed beyond a simple copy-paste refactor

Not sure why the diff looks so strange. Marking as draft until the first PR is merged.


MichaelChirico:
Weird, seems we can't edit the target branch of this PR. I'll open a new one.

PR #2066, opened 2023-08-10
include line length in lint message

### LINKED ISSUE #2057: line_length_linter() message should mention the current line length
This can be helpful to tell at a glance how wide the line is, e.g. in contexts where we don't have our handy IDE ruler visible, we can often guess how to split up the line based on if it's just a few or many characters too long.

I'd also get some mileage out of it while linting our own package -- when running `lint_package(linters = all_linters())`, the default of 80 characters applies --> many false positives, we could filter those out if we could subset on the message being >= 120 instead. This is not the proper way to handle this problem, but would be handy in interactive situations.

MichaelChirico:
Closes #2057


NEWS.md
```diff
@@ -23,6 +23,7 @@
   + `unreachable_code_linter()`
   + `yoda_test_linter()`
 * `sprintf_linter()` is pipe-aware, so that `x %>% sprintf(fmt = "%s")` no longer lints (#1943, @MichaelChirico).
+* `line_length_linter()` helpful includes the line length in the lint message (#2057, @MichaelChirico).
```
AshesITR, line 26: grammar

AshesITR (approved):

PR #2068, opened 2023-08-10
Some refactoring to clarify logic in complicated linters

### LINKED ISSUE #2069: False positive in object_usage_linter() for infix operators in glue strings
```r
lint(linters = object_usage_linter(), "
foo <- function() {
  `%++%` <- `+`
  glue('{x %++% y}')
}")
# <text>:3:3: warning: [object_usage_linter] local variable '%++%' assigned but may not be used
#   `%++%` <- `+`
#   ^~~~~~
```

The problem is we only look for `SYMBOL` or `SYMBOL_FUNCTION_CALL` in the parse tree of the glue expression:

https://github.com/r-lib/lintr/blob/2813b6dcd29edb06676481df9cf879fd2c11e76c/R/object_usage_linter.R#L225

#2068 already fixed this by using `all.vars()` instead.

MichaelChirico:
Closes #2069 

Follow-up #2063, split off the refactoring pieces from the pure copy-paste pieces.


R/namespace_linter.R
```diff
@@ -123,74 +123,55 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
 build_ns_get_int_lints <- function(packages, symbols, symbol_nodes, namespaces, source_expression) {
-  lints <- list()
-
   ## Case 2: foo does not exist in pkg:::foo
-
-  non_symbols <- !vapply(
-    seq_along(symbols),
-    function(ii) symbols[[ii]] %in% ls(namespaces[[ii]], all.names = TRUE),
-    logical(1L)
+  non_symbols <- !mapply(
```
AshesITR: Isn't `vapply()` faster and safer to use because of type checks?

R/namespace_linter.R
```diff
@@ -123,74 +123,55 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
   })
 }
 
-get_all_exports <- function(namespace) {
-  c(getNamespaceExports(namespace), names(.getNamespaceInfo(namespace, "lazydata")))
+exported_symbols <- function(ns) c(getNamespaceExports(ns), names(.getNamespaceInfo(ns, "lazydata")))
+is_exported <- function(symbols, namespaces) {
+  mapply(
```
AshesITR, line 134: Isn't `vapply()` better?

R/object_usage_linter.R
```diff
@@ -55,105 +52,110 @@ object_usage_linter <- function(interpret_glue = TRUE, skip_with = TRUE) {
     | descendant::LEFT_ASSIGN[text() = ':=']
   ")
 
+  lint_assignment <- function(fun_assignment, source_expression, declared_globals, env) {
```
AshesITR, line 55: Why move this inside?

MichaelChirico:
Maybe makes sense to further split off the `namespace_linter()` edits from here given that this is now a bugfix for `object_usage_linter()`. LMK if you agree.

R/object_usage_linter.R
```diff
@@ -166,63 +168,46 @@ extract_glued_symbols <- function(expr) {
...
-
-  if (length(glue_calls) == 0L) {
-    return(character())
-  }
+  glue_call_xpath <- "
+    descendant::SYMBOL_FUNCTION_CALL[text() = 'glue']
+      /parent::expr
+      /parent::expr[
+        not(SYMBOL_SUB[text() = '.envir' or text() = '.transform'])
+        and not(expr[position() > 1 and not(STR_CONST)])
```
AshesITR, line 176: Why `position() > 1`?

AshesITR:
I wouldn't mind either way, but `object_usage_linter()` would need a regression test for the fixed bug.

R/object_usage_linter.R
MichaelChirico: `position()=1` is the `SYMBOL_FUNCTION_CALL` `<expr>`

AshesITR: Ah now I see you moved one expr further up in comparison to the old XPath.

R/namespace_linter.R
MichaelChirico: yes, though safety is less of a concern for internal functions since we always control the input.

basically it's a readability-efficiency tradeoff, carrying around the `[[ii]]` everywhere vs. the cleaner code with `mapply()`.

R/object_usage_linter.R
MichaelChirico: vestigial, forgot to undo that part

R/object_usage_linter.R
MichaelChirico: added a comment

R/namespace_linter.R
MichaelChirico: changed back to `vapply()` now that all 3 loops are covered by one helper, makes the readability tax minimal.

AshesITR (approved):

R/object_usage_linter.R
MichaelChirico: right -- saves some `following-sibling::` space, especially since the end anchor of the XPath is that `expr` anyway.

PR #2071, opened 2023-08-10
if_not_else_linter() for "double negation" in if statements

MichaelChirico:
Part of #884.

Internally, we use the default `exceptions=`, but I'd be open to setting that to `character()` by default, or extending the current list.


R/if_not_else_linter.R
```diff
@@ -0,0 +1,108 @@
...
+#'   linters = if_not_else_linter()
+#' )
+#'
+#' lint(
+#'   text = "ifelse(is_treatment, y, x)",
+#'   linters = if_not_else_linter()
+#' )
+#'
+#' lint(
+#'   text = "if (!is.null(x)) x else 2",
```
IndrajeetPatil, line 47: I think this example still needs to be changed?

MichaelChirico: This example is meant to show the default behavior that `!is.null()` passes without linting

AshesITR (approved):

Issue #1108, opened 2022-04-27
Further extensions to paste_linter


MichaelChirico:
Branched off from #998 -- creating this issue to cover the remaining `paste`-related lints mentioned there that aren't making the 3.0.0 cut

 - [x] #884 `PasteFilePathLinter`: `paste(sep = "/")` ➡️ `file.path()`
 - [x] #884 `PasteStrrepLinter`: `paste(rep(), collapse="")` ➡️ `strrep()` #1652
 - [ ] `paste0(x, collapse = )` ➡️ `paste(x, collapse = )`

olivroy:
I just tripped on 

```r
paste0(x, collapse = )
```

Would love to see this implemented!

MichaelChirico:
Could you elaborate what you mean by "tripped on"? The behavior of `paste(x, collapse=)` should be identical to that of `paste0(x, collapse=)`, the advantage here is only for consistency/possibly readability, not anything to do with correctness.

olivroy:
Oh, actually, I tripped on `paste0(sep =)`. `sep` is ignored by `paste0()` which yielded bad results. Sorry for the confusion!

MichaelChirico:
Your use case is already covered by `paste_linter()`:

```
lintr::lint(text='paste0(a, sep="b")', linters=lintr::paste_linter())
# <text>:1:1: warning: [paste_linter] sep= is not a formal argument to paste0(); did you mean to use paste(), or collapse=?
# paste0(a, sep="b")
# ^~~~~~~~~~~~~~~~~~
```

PR #2074, opened 2023-08-10
file path detection in paste_linter

MichaelChirico:
Part of #884 and #1108. We wrote this as a separate `PasteFilePathLinter()`, but I believe it fits well as an extension of `paste_linter()`.

### DIFF
--- a/R/paste_linter.R
+++ b/R/paste_linter.R
@@ -191,7 +191,7 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE, allow
   ")
 
   empty_paste_note <-
-    'Note well that paste() converts empty inputs to "", whereas file.path() leaves it empty.'
+    'Note that paste() converts empty inputs to "", whereas file.path() leaves it empty.'
 
   Linter(function(source_expression) {
     if (!is_lint_level(source_expression, "expression")) {
--- a/R/paste_linter.R
+++ b/R/paste_linter.R
@@ -253,7 +253,8 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE, allow
         source_expression = source_expression,
         lint_message = paste(
           'Construct file paths with file.path(...) instead of paste(..., sep = "/").',
-          'If you are using paste(sep = "/") to construct a date, consider using format() or lubridate helpers instead.',
+          'If you are using paste(sep = "/") to construct a date,',
+          "consider using format() or lubridate helpers instead.",
           empty_paste_note
         ),
         type = "warning"
--- a/R/paste_linter.R
+++ b/R/paste_linter.R
@@ -141,40 +141,32 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE, allow
   ")
 
   # Type II: paste0(x, "/", y, "/", z)
-  #   more specifically, paste0 need not be replaced by file.path when
-  #   (1) there are consecutive SYMBOL|expr (like paste0(x, y))
-  #   (2) a STR_CONST precedes a SYMBOL|expr, but doesn't end with /
-  #   (3) a STR_CONST follows a SYMBOL|expr, but doesn't start with /
-  #   (4) there are consecutive STR_CONSTs but neither
-  #       the first ends with / nor the second starts with /
-  #   (5) there is only one argument
-  #   (6) there is any NUM_CONST
-  #   (7) collapse is used
-  #   (8) / is the first character in an initial STR_CONST or the
-  #       last character in a terminal STR_CONST, where file.path
-  #       replacement would somewhat awkwardly look like
-  #       file.path("", ...) and/or file.path(..., "")
-  #   (9) a STR_CONST beginning or ending with >1 slashes (like "//x")
-  # use count() to exclude paste0(x)
-  # don't use starts-with() -- (1) there's no ends-with() in XPath 1.0 and
-  #   (2) we need to cater for both ' and " quotations which ^ complexity
-  # NB: XPath's substring() signature is (text, initial_index, n_characters)
-  # an expression matching any of the not(...) conditions is _not_ a path
-  paste0_file_path_xpath <- glue("
+  paste0_file_path_xpath <- xp_strip_comments(glue("
   //SYMBOL_FUNCTION_CALL[text() = 'paste0']
     /parent::expr
     /parent::expr[
+      (: exclude paste0(x) :)
       count(expr) > 2
+      (: An expression matching _any_ of these conditions is _not_ a file path :)
       and not(
+        (: Any numeric input :)
         expr/NUM_CONST
+        (: A call using collapse= :)
         or SYMBOL_SUB[text() = 'collapse']
+        (: First input is '/', meaning file.path() would need to start with '' :)
         or expr[2][{slash_str}]
+        (: Last input is '/', meaning file.path() would need to end with '' :)
         or expr[last()][{slash_str}]
+        (: String starting or ending with multiple / :)
+        (: TODO(#2075): run this logic on the actual R string :)
         or expr/STR_CONST[
+          (: NB: this is (text, initial_index, n_characters) :)
           substring(text(), 2, 2) = '//'
           or substring(text(), string-length(text()) - 2, 2) = '//'
         ]
+        (: Consecutive non-strings like paste0(x, y) :)
         or expr[({non_str}) and following-sibling::expr[1][{non_str}]]
+        (: A string not ending with /, followed by non-string or string not starting with / :)
         or expr[
           {str_not_end_with_slash}
           and following-sibling::expr[1][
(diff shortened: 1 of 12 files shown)


R/paste_linter.R
```diff
@@ -111,6 +123,76 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE) {
...
+  slash_str <- sprintf("STR_CONST[%s]", xp_text_in_table(c("'/'", '"/"')))
+  str_not_start_with_slash <-
+    "STR_CONST[not(substring(text(), 2, 1) = '/')]"
+  str_not_end_with_slash <-
+    "STR_CONST[not(substring(text(), string-length(text()) - 1, 1) = '/')]"
+  non_str <- "SYMBOL or expr"
+
+  # Type I: paste(..., sep = "/")
+  paste_file_path_xpath <- glue("
+  //SYMBOL_FUNCTION_CALL[text() = 'paste']
```
AshesITR, line 135: This could also catch `paste0(..., sep = "/")` - or is it already linted for any non-missing `sep=`?
The latter would be even better, although it would cause two "chained" lints for users refactoring `paste0("a", "b", sep = "/") = paste("a", "b", sep = "/") = file.path("a", "b")`

R/paste_linter.R
```diff
@@ -111,6 +123,76 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE) {
...
+    /parent::expr
+    /parent::expr[
+      SYMBOL_SUB[text() = 'sep']/following-sibling::expr[1][{slash_str}]
+      and not(SYMBOL_SUB[text() = 'collapse'])
+    ]
+  ")
+
+  # Type II: paste0(x, "/", y, "/", z)
```
AshesITR, line 143: Due to the size of these comments, maybe you should interlace them with the XPath code.
NB `glue("a", "b") == "ab"` allowing for easy commenting - especially since the whitespace in the string isn't syntactic for XPath.

R/paste_linter.R
```diff
@@ -111,6 +123,76 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE) {
...
+        or expr[
+          {str_not_start_with_slash}
+          and preceding-sibling::expr[1][{non_str}]
+        ]
+      )
+    ]
+  ")
+
+  empty_paste_note <-
+    'Note well that paste() converts empty inputs to "", whereas file.path() leaves it empty.'
```
AshesITR, line 194: "Note _well_ that" broke my reading flow. Just a nit, but consider using just "Note that"?

R/paste_linter.R
```diff
@@ -111,6 +123,76 @@ paste_linter <- function(allow_empty_sep = FALSE, allow_to_string = FALSE) {
...
+    /parent::expr
+    /parent::expr[
+      count(expr) > 2
+      and not(
+        expr/NUM_CONST
+        or SYMBOL_SUB[text() = 'collapse']
+        or expr[2][{slash_str}]
+        or expr[last()][{slash_str}]
+        or expr/STR_CONST[
+          substring(text(), 2, 2) = '//'
```
AshesITR, line 174: How robust is this with regards to raw string literals?

tests/testthat/test-paste_linter.R
```diff
@@ -113,3 +113,95 @@ test_that("paste_linter blocks simple disallowed usages", {
...
+  expect_lint("paste0(foo(x), '/abc/', 'def/', bar(y))", lint_msg, linter)
+})
+
+test_that("paste_linter skips initial/terminal '/' and repeated '/' for paths", {
+  linter <- paste_linter()
+
+  expect_lint("paste0('/', x)", NULL, linter)
+  expect_lint("paste0(x, '/')", NULL, linter)
+  expect_lint("paste0(x, '//hey/', y)", NULL, linter)
+  expect_lint("paste0(x, '/hey//', y)", NULL, linter)
```
AshesITR, line 180: Seeing this test I wonder if we can force a false positive with `paste0(a, "/not//a/path/", b)`?

R/paste_linter.R
MichaelChirico: I've been thinking about a wrapper that strips XPath "comments" for this, WDYT?

```r
xpath <- strip_comments("
  //path
    # we need this because...
    /expr
")
```

R/paste_linter.R
MichaelChirico: that's just me translating NB :)

R/paste_linter.R
MichaelChirico: Not at all! I'll file a follow-up?

R/paste_linter.R
MichaelChirico: in fact I would recommend using XPath2.0-style comments, in case we eventually have a back-end that supports XPath2.0:

https://stackoverflow.com/a/22648576/3576984

AshesITR: IDE syntax highlighting and performance is likely better with R comments.
I'lm have to try and see if language injections for XPath highlight correctly in my IDE.
If we do put them in the string then definitely yes - use XPath 2.0 comment syntax.

R/paste_linter.R
AshesITR: It's probably fine, will just cause false negatives in the case they are used, right?

R/paste_linter.R
MichaelChirico: performance-wise, I'd expect a trivial difference (`strip_comments()` should just be a `gsub()` run in the linter factory).

will the difference for syntax highlighting be a big gain? as long as the comments are always on their own line, at least.

in favor of doing it within one big string literal b/c I find all the `"` and `,` for the approach of building the XPath line-by-line a bit distracting.

R/paste_linter.R
MichaelChirico: Right... though I think we've usually tried to be consistent about running logic on the evaluated string itself (`get_r_string()`), rather than assuming patterns on the `STR_CONST`

MichaelChirico: #2075

tests/testthat/test-paste_linter.R
MichaelChirico: whether that's a path is a bit ambiguous (it's a valid path on many OS), and would also disallow it being used to construct URLs... maybe that's another option to add? Since I know some (maybe many) authors don't like building URLs with `file.path()`, we could disable the linter if `//` is present? Or should we just do that by default

R/paste_linter.R
MichaelChirico: PTAL

AshesITR: Looks good.

tests/testthat/test-paste_linter.R
AshesITR: A default-TRUE argument to ignore URLs sounds good.

R/paste_linter.R
MichaelChirico: :sweat_smile:aesthetically it makes the code look a lot happier :upside_down_face: 🙂

R/xp_utils.R
```diff
@@ -65,3 +65,17 @@ xp_find_location <- function(xml, xpath) {
...
+#' Strip XPath 2.0-style comments from an XPath
+#'
+#' xml2 uses XPath 1.0, which has no support for comments. But comments are
+#'   useful in a codebase with as many XPaths as we maintain, so we fudge our
+#'   way to XPath 2.0-ish support by writing this simple function to remove comments.
+#'
+#' @noRd
+xpath_comment_re <- rex::rex(
+  "(:",
+  zero_or_more(not(":)")),
```
AshesITR, line 78: What actual regex does this compile to?
Also XPath is not regular - comments are only allowed in "ignorable whitespace".
This approximation should be noted here:

`/STR_CONST[text() = '"(: not a comment :)"']` 

Would produce a wrong XPath.

MichaelChirico: That's right -- but without a full XPath engine, we won't really be able to prevent that. And it seems unlikely to show up in anything we do in {lintr}. I'll make a note in the documentation.

> What actual regex does this compile to?

```regex
\(:(?:(?:(?!:\)).)*)*:\)
```

When I wrote it by hand I did `[(][:](?:[^:]|[:][^)])*[:][)]`, the simple `rex()` version is probably a little less efficient (by always checking two behind the current character), but easy to read.

AshesITR: Thanks, both regexes seem acceptable.

tests/testthat/test-paste_linter.R
MichaelChirico: In the spirit of importing something closer to the original linter we used internally first, marked this piece for follow-up: #2082. It's part of the release milestone to ensure this is the default behavior before the next CRAN version.

AshesITR (approved):

PR #2077, opened 2023-08-10
extend conjunct_test_linter for dplyr::filter()

### LINKED ISSUE #2073: filter_and: New linter, or part of `conjunct_test_linter()`?
One linter for #884 looks for `dplyr::filter(x, A && B)` and recommends `dplyr::filter(x, A, B)`.

Should this be a new linter (we call it `filter_and_condition_linter()`, but I think `conjunct_filter_linter()` would fit the {lintr} style better), or be subsumed as part of the existing `conjunct_test_linter()`?

MichaelChirico:
Closes #2073, part of #884


MichaelChirico:
Marking the `stats::filter()` issue for follow-up: #2078

R/conjunct_test_linter.R
```diff
@@ -72,24 +86,36 @@ conjunct_test_linter <- function(allow_named_stopifnot = TRUE) {
...
+    operator <- xml_find_chr(test_expr, "string(expr/*[self::AND2 or self::OR2])")
     replacement_fmt <- ifelse(
       matched_fun %in% c("expect_true", "expect_false"),
       "write multiple expectations like %1$s(A) and %1$s(B)",
       "write multiple conditions like %s(A, B)."
     )
     lint_message <- paste(
       sprintf("Instead of %s(A %s B),", matched_fun, operator),
-      sprintf(replacement_fmt, matched_fun),
+      sprintf(as.character(replacement_fmt), matched_fun),
```
AshesITR, line 100: This could use an explanation. I.e. what is `ifelse(logical(0L), "a", "b")`

R/conjunct_test_linter.R
```diff
@@ -44,26 +47,37 @@
 conjunct_test_linter <- function(allow_named_stopifnot = TRUE) {
```
AshesITR, line 47: `assume_dpylr_loaded = FALSE` or a dynamic check if the global `filter()` is from dplyr to remove the NS prefix condition in the XPath?
I think both approaches would greatly increase the amount of true positives.

MichaelChirico: Let's assume `filter()` is `dplyr::filter()` unless another namespace is provided for this initial PR, and fine-tune in follow-up

AshesITR: All right

AshesITR (dismissed):

AshesITR (approved):

PR #2079, opened 2023-08-11
Remove '::' from more frequent calls from {rex}, {utils}

MichaelChirico:
Follow-up to #2031


R/unused_import_linter.R
```diff
@@ -44,7 +44,7 @@
 unused_import_linter <- function(allow_ns_usage = FALSE, except_packages = c("bit64", "data.table", "tidyverse")) {
   # Get dataset names lazy-loaded by imported packages
   get_datasets <- function(pkg) {
```
AshesITR, line 46: namespace_linter has `namespace_symbols()` already, taking lazydata into account.
We can probably refactor this and not use `utils::data()` at all.

R/linter_tags.R
```diff
@@ -70,7 +70,7 @@ available_linters <- function(packages = "lintr", tags = NULL, exclude_tags = "d
   if (!file.exists(csv_file)) {
     return(empty_linters())
   }
-  available <- utils::read.csv(csv_file, encoding = "UTF-8", as.is = TRUE)
+  available <- read.csv(csv_file, encoding = "UTF-8", as.is = TRUE)
```
AshesITR, line 73: AFAICT this is the only usage of `read.csv()`. If that is the case, I wouldn't import it.

MichaelChirico: I went back and forth a few times on this, deciding to just go ahead and import it since it's a default package & "everyone" knows where `read.csv()` comes from, so may as well import it. Ditto for `data()` and `packageVersion()`. I would either import all or none of those 3.

AshesITR: Re `data()` see my comment - I wouldn't import but instead get rid of the call if it's possible to reuse `namespace_linter()` logic.

Re `packageVersion()` I don't mind because it's used multiple times.

I found myself having to look up which package these "base" functions come from, so the ns-prefixed version helps when deciding if a base Imports: is necessary. But that's not a big deal I think.

All in all maybe not import from `utils`?

R/unused_import_linter.R
MichaelChirico: Since it's besides the point here, filed as #2080

AshesITR (approved):

PR #2086, opened 2023-08-14
withr is Suggests

MichaelChirico:


AshesITR (approved):

AshesITR (approved):
LGTM. Is it possible for a new linter to detect this? Something like unused import linter, but the other way around.
Für Suggests it would have to check that the function has a matching requireNamespace() test, maybe?

MichaelChirico:
> LGTM. Is it possible for a new linter to detect this? Something like unused import linter, but the other way around. Für Suggests it would have to check that the function has a matching requireNamespace() test, maybe?

I'm not sure how easy this is in general. Started #2087 to centralize discussion.

PR #2089, opened 2023-08-16
Add new 'pkg_testthat' tag to testthat linters

MichaelChirico:
`package_development` is a bit broader, I think it's good to have a tag more specifically designating these linters as related to {testthat}.


AshesITR:
How would the tag apply to linters that are extended beyond expectation functions? Would they keep the tag if they work with unit testing frameworks?
`conjunct_test_linter()`, for example.

Also: I'd keep the tag name general in case we improve support for other testing packages in the future. Maybe `package_tests` or `testing`?

R/linter_tag_docs.R
```diff
@@ -99,3 +99,11 @@ NULL
 #' @evalRd rd_linters("executing")
 #' @seealso [linters] for a complete list of linters available in lintr.
 NULL
+
+#' Testthat linters
+#' @name testthat_linters
+#' @description
+#' Linters encouraging best practices within testthat suites.
```
AshesITR, line 106: This could use some external references to testthat docs, preferably about best practices.

MichaelChirico:
> I'd keep the tag name general in case we improve support for other testing packages in the future.

What I had in mind though is `linters_with_tags("testthat")` for users running testthat. It wouldn't be useful to include RUnit tests, in that case.

Maybe we should add two tags here: `package_testing` and `testthat`. The former is between `package_development` and `testthat` in generality, WDYT?

> How would the tag apply to linters that are extended beyond expectation functions? Would they keep the tag if they work with unit testing frameworks?

`yoda_test_linter()` is an example here, it can easily extend to include e.g. `3 == x`, so it becomes more general than just `testthat`. I think the tag system is to say "this tag applies to this linter" so it's better to keep it even if the linter scope increases.

R/linter_tag_docs.R
MichaelChirico: I just linked to the testthat site, I don't know of any "best practices" docs they produce. Even if so I worry about introducing link rot.

AshesITR: SGTM. Maybe also reference the R Packages section?

https://r-pkgs.org/testing-basics.html

AshesITR:
> > I'd keep the tag name general in case we improve support for other testing packages in the future.
> 
> 
> 
> What I had in mind though is `linters_with_tags("testthat")` for users running testthat. It wouldn't be useful to include RUnit tests, in that case.
> 
> 
> 
> Maybe we should add two tags here: `package_testing` and `testthat`. The former is between `package_development` and `testthat` in generality, WDYT?

Sounds okay. Maybe prefix the tag with `pkg_` in case we add e.g. `pkg_dplyr` or `pkg_data.table` in the future?
Thinking more about `package_testing`, it would only be useful as an except= argument I think and that makes it hard to decide what should be done about linters supporting more than just testing (or do so in the future) - so maybe it's not a good idea at all.

> > How would the tag apply to linters that are extended beyond expectation functions? Would they keep the tag if they work with unit testing frameworks?
> 
> 
> 
> `yoda_test_linter()` is an example here, it can easily extend to include e.g. `3 == x`, so it becomes more general than just `testthat`. I think the tag system is to say "this tag applies to this linter" so it's better to keep it even if the linter scope increases.

Then you're missing `conjunct_test_linter()` here.

AshesITR (approved):

PR #2092, opened 2023-08-21
new option to skip URLs in paste_linter for file paths

### LINKED ISSUE #2082: paste_linter() should ignore URL-like paths for file.path() recommendation by default
Follow-up to #2074.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2074#discussion_r1290899615_

Not all authors will like using `file.path()` to build URLs, so `paste0()` may be preferable in such cases. The proposed solution is to skip any `paste0()` call with a double-slash `//`, and use an argument for authors who may want to lint such paths anyway.

MichaelChirico:
Closes #2082


AshesITR:
The argument name is very long.
WDYT about `allow_file_path_url` or changing `allow_file_path = c("never", "double_slash", "always")`?

MichaelChirico:
> TRUE and FALSE being treated as "always" and "never" respectively?

Since this hasn't been part of a release yet, I'll drop the option to have `TRUE`/`FALSE`, I'd rather keep it simple.

AshesITR (approved):

PR #2093, opened 2023-08-22
require & library both linted in library_call_linter()

### LINKED ISSUE #2045: pkgdown shows lintr version 3.0.2 instead of 3.1.0
image

MichaelChirico:
Closes #2045 

Maybe a renaming is in order? `attach_up_front_linter()`? `preamble_attach_linter()`? `package_preamble_linter()`?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,7 +31,7 @@
 
 ### New linters
 
-* `library_call_linter()` can detect if all library calls are not at the top of your script (#2027, @nicholas-masel).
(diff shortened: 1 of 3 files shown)


AshesITR:
The present name seems okay to me.
Still, throwing `dependency_preamble_linter()` into the mix if you want to rename.

This could later be extended to also require `source()`d scripts at the top.

MichaelChirico:
Similar for if we add {box} support... let's cross that bridge when we get to it & we can go through a deprecation cycle if needed.

AshesITR (approved):

PR #2094, opened 2023-08-23
fixed_regex_linter is pipe-aware

### LINKED ISSUE #1811: fixed_regex_linter() throws false positives for piped-in `stringr` calls
```r
x %>%
  str_replace(fixed("abc"), "cde")
```

Since our logic currently uses the argument number and doesn't account for whether the call is piped-in, it sees `cde` as being "fixed" and hence throws a lint.

MichaelChirico:
Closes #1811


AshesITR (commented):
LGTM, maybe add a test for `|>`?

MichaelChirico:
Good catch. Basically the same amount of code to factorialize the test & cover all pipes, so did that while we're at it.

AshesITR (approved):

PR #2095, opened 2023-08-23
detect missing() usage in function_argument_linter

### LINKED ISSUE #1546: function_argument_linter() could look for missing() handling and mention it in the lint message

MichaelChirico:
Closes #1546


AshesITR:
Can we add a test case without `{}` in the function body?

```r
function(x = 2, y) if (missing(y)) x else y
```

Just to make sure the XPath works everywhere.
The rest looks good.

NEWS.md
```diff
@@ -30,6 +30,8 @@
 * `conjunct_test_linter()` also lints usage like `dplyr::filter(x, A & B)` in favor of using `dplyr::filter(x, A, B)` (part of #884, @MichaelChirico).
 * `sort_linter()` checks for code like `x == sort(x)` which is better served by using the function `is.unsorted()` (part of #884, @MichaelChirico).
 * `paste_linter()` gains detection for file paths that are better constructed with `file.path()`, e.g. `paste0(dir, "/", file)` would be better as `file.path(dir, file)` (part of #884, @MichaelChirico).
+* `fixed_regex_linter()` gains an option `allow_unescaped` (default `FALSE`) to toggle linting regexes not requiring any escapes or character classes (#1689, @MichaelChirico). Thus `fixed_regex_linter(allow_unescaped = TRUE)` would lint on `grepl("[$]", x)` but not on `grepl("a", x)` since the latter does not use any regex special characters.
```
AshesITR, line 33: Merge messed up?

AshesITR (dismissed):

NEWS.md
```diff
@@ -30,6 +30,7 @@
 * `conjunct_test_linter()` also lints usage like `dplyr::filter(x, A & B)` in favor of using `dplyr::filter(x, A, B)` (part of #884, @MichaelChirico).
 * `sort_linter()` checks for code like `x == sort(x)` which is better served by using the function `is.unsorted()` (part of #884, @MichaelChirico).
 * `paste_linter()` gains detection for file paths that are better constructed with `file.path()`, e.g. `paste0(dir, "/", file)` would be better as `file.path(dir, file)` (part of #884, @MichaelChirico).
+* `function_argument_linter()` detects usage of `missing()` for the linted argument (#1546, @MichaelChirico). The simplest fix for `function_argument_linter()` lints is typically to set that argument to `NULL` by default, in which case it's usually preferable to update function logic checking `missing()` to check `is.null()` instead.
```
MichaelChirico, line 33: lol this zombie NEWS item refuses to die

MichaelChirico:
Oh, whoops. Should open my eyes wider. I mis-read and thought it was the `fixed_regex_linter()` bullet back again. Restored.

AshesITR (approved):

PR #2099, opened 2023-08-29
`length(x):1` now mentions `rev()` in the lint message

### LINKED ISSUE #1542: seq_linter() gives the wrong message for n:1 lints

MichaelChirico:
Closes #1542


AshesITR (dismissed):

tests/testthat/test-seq_linter.R
```diff
@@ -158,3 +158,12 @@ test_that("Message vectorization works for multiple lints", {
     seq_linter()
   )
 })
+
+test_that("Message recommends rev() correctly", {
+  linter <- seq_linter()
+
+  expect_lint(".N:1", rex::rex("Use rev(seq_len(.N))"), linter)
```
AshesITR, line 165: BTW if this is data.table code, it should be `rev(.I)`.
Not correct in general, though, so the current message is fine I think.

AshesITR (approved):

tests/testthat/test-seq_linter.R
MichaelChirico: Good point. It also applies in the `1:.N` case, though, so I'll file it as a separate issue.

PR #2107, opened 2023-09-05
Add `repeat_linter` for infinite loops 

### LINKED ISSUE #2106: Detection of while(TRUE) and while(FALSE)
t would be handy to introduce a new linter that detects `while(TRUE)` constructs and suggests adopting `repeat` for clearer infinite loop indication. In the same linter, a feature could be developed to identify `while(FALSE)` constructs, signaling unreachable code or potentially as an additional feature of the existing `unreachable_code_linter`.

MEO265:
PR resolves #2106


R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
+#' Repeat linter
+#'
+#' Check that `while` is not used for infinite loops.
```
MichaelChirico, line 3: wording suggestion here. this could read as "any infinite while loop is detected" which is broader than the actual linter. maybe "simple infinite loops", or just mention while(TRUE) here

R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
...
+#'   linters = repeat_linter()
+#' )
+#'
+#'
+#' @evalRd rd_tags("repeat_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+repeat_linter <- function() {
+  xpath <- "
+  //WHILE[following-sibling::*[1][self::OP-LEFT-PAREN] and
```
MichaelChirico, line 25: first condition looks redundant to me. is it possible to write a while loop without '('? if you're worried about 'while' being used in other contexts, while is a reserved word so it will show up as a SYMBOL IINM (worth adding a test for this to be sure)

R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
...
+  following-sibling::*[2][self::expr[NUM_CONST[text()='TRUE'] and count(*)=1]]
+  and following-sibling::*[3][self::OP-RIGHT-PAREN]]"
```
MichaelChirico, line 27: this is also redundant. if you're worried about more complicated expressions in the condition, your test for NUM_CONST is implicitly handling that -- TRUE would show up at further nesting of <expr> in that case.

R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
...
+#' )
+#'
+#'
+#' @evalRd rd_tags("repeat_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+repeat_linter <- function() {
+  xpath <- "
+  //WHILE[following-sibling::*[1][self::OP-LEFT-PAREN] and
+  following-sibling::*[2][self::expr[NUM_CONST[text()='TRUE'] and count(*)=1]]
```
MichaelChirico, line 26: I would simplify this as following-sibling::expr/NUM_CONST[...]

the node structure of while() loops is pretty simple in terms of what can parse, so no need to be overly defensive IMO.

R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
...
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+    xml <- source_expression$xml_parsed_content
+
+
+    lints <- xml_nodes_to_lints(
+      xml_find_all(xml, xpath),
+      source_expression = source_expression,
+      lint_message = "'while (TRUE)' is not recommended for infinite loops. Use 'repeat' instead.",
```
MichaelChirico, line 39: We prefer the actionable bit first (there's an issue about doing so consistently in the tracker somewhere...). here that would be "Use repeat instead of while(TRUE) for infinite loops"

tests/testthat/test-repeat_linter.R
MichaelChirico, line 1: more suggested test cases:

```r
# while as a function call
`while`(TRUE)
# TRUE is the expression
while (j < 5) TRUE
# TRUE in a more complicated condition
while (TRUE && j < 5) { ... }
# sensitivity to '{'
while (TRUE) NULL
while (j < 5) NULL
```

MichaelChirico (commented):
The code looks good and works, thanks! My comments are for polish/completeness.

MEO265:
> The code looks good and works, thanks! My comments are for polish/completeness.

Understood. I'm always happy about comments that help to improve.

tests/testthat/test-repeat_linter.R
AshesITR: Also should test for lint locations, especially with multiline code and multiple lints in one expression.

```r
{
  while (TRUE) {
  }
  while (TRUE) {
  }
}
```

R/repeat_linter.R
MEO265: > while is a reserved word so it will show up as a SYMBOL IINM (worth adding a test for this to be sure)
When I try to create a test for it, the parser always flags an error. I would skip the tests if you don't have a suggestion

MichaelChirico: This works: ``parse(text = "`while`(TRUE, NULL)")``, but actually resolves as equivalent to `while (TRUE) NULL`.

We can actually get into really funky territory here:

```r
`while` <- identity
while(TRUE)
# [1] TRUE
while (FALSE) { }
# Error in while (FALSE) { : unused argument ({
# })
```

I'm OK to skip tests.

R/repeat_linter.R
MEO265: I implemented it as suggested but had to test that an expression existed at all

R/repeat_linter.R
MEO265: I implemented it as suggested but still had to test that an expression existed at all

R/repeat_linter.R
```diff
@@ -0,0 +1,46 @@
...
+#'   linters = repeat_linter()
+#' )
+#'
+#'
+#' @evalRd rd_tags("repeat_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+repeat_linter <- function() {
+  xpath <- "
+  //*[(self::WHILE or self::expr[SYMBOL_FUNCTION_CALL[.='`while`']])
```
MichaelChirico, line 25: `//*` is really troublesome for XPath -- they are evaluated greedily and so this linter will be really expensive on files with a lot of expressions. See https://github.com/r-lib/lintr/wiki/Writing-robust,-performant-linters.

I really think we can ignore the `` `while` `` case unless there's a specific request for it -- that's a really unusual way to write a `while` loop. Otherwise, we would usually write this as `//WHILE[...] | //SYMBOL_FUNCTION_CALL[text() = '`while`']/parent::expr[...]`.

(TIL you can use `. = '...'` to check the text on a node btw! Nice find. I still like `text() = '...'` for expressiveness, `. = '...'` is a little bit harder to reason about)

MEO265:
> @MEO265 FYI, I edited the PR description. "Solves" is not the ideal keyword since it won't auto-close the issue -- see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue.

Thanks for the info.

R/repeat_linter.R
MEO265: I will ignore the `` `while` `` case. 
I misunderstood you I thought you would have wished for this case by mentioning it as a possible test  [here](https://github.com/r-lib/lintr/pull/2107#discussion_r1316290069)

MEO265:
Can someone tell me why pipeline "check-link-rot" fails and whether I can influence it at all?

MichaelChirico:
> Can someone tell me why pipeline "check-link-rot" fails and whether I can influence it at all?

ignore it, it's been happening to other builds as well, that link appears fragile.

MichaelChirico (approved):

MEO265:
@MichaelChirico I looked at your changes and will try to do the same next time

MichaelChirico (approved):

Issue #2108, opened 2023-09-05
`conjunct_test_linter`: Allow to opt out of linting `dplyr::filter()`


salim-b:
https://github.com/r-lib/lintr/pull/2077 added support to lint logical conjunctions inside `dplyr::filter()` calls. Quote:

> Relatedly, `dplyr::filter(DF, A & B)` is the same as `dplyr::filter(DF, A, B)`, but the latter will be more readable / easier to format for long conditions.

I respectfully disagree with the "latter will be more readable" part and would like to use `conjunct_test_linter` *without* linting `dplyr::filter()` calls. Would you be willing to introduce a flag like `allow_inside_dplyr_filter` to opt out of this behaviour?

MichaelChirico:
I think in that sentence the emphasis is on "for long conditions". We often see code like:

```r
DF |>
  filter(
    is_treatment &
      (expenditure - cost > 0) &
      group == "x"
  )
```

In this case it's not very controversial that using `,` is better:

```r
DF |>
  filter(
    is_treatment,
    expenditure - cost > 0,
    group == "x"
  )
```

It's more of a toss-up for simple cases, like `A & B` vs `A, B`.

Anyway, happy to accept a PR. I would simply call such an argument `allow_filter`.

salim-b:
> It's more of a toss-up for simple cases, like `A & B` vs `A, B`.

Yeah, but I also think `&` is clearer than `,` in differentiating logical conjunctions from logical disjunctions (` | `).

> Anyway, happy to accept a PR. I would simply call such an argument `allow_filter`.

Cool, see https://github.com/r-lib/lintr/pull/2110.

PR #2110, opened 2023-09-05
Allow to opt out of linting `filter()` in `conjunct_test_linter`

salim-b:
Fixes https://github.com/r-lib/lintr/issues/2108.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,7 @@
 * `seq_linter()` recommends `rev()` in the lint message for lints like `nrow(x):1` (#1542, @MichaelChirico).
 * New `xp_call_name()` helper to facilitate writing custom linters (#2023, @MichaelChirico). This helper converts a matched XPath to the R function to which it corresponds. This is useful for including the "offending" function in the lint's message.
 * `function_argument_linter()` detects usage of `missing()` for the linted argument (#1546, @MichaelChirico). The simplest fix for `function_argument_linter()` lints is typically to set that argument to `NULL` by default, in which case it's usually preferable to update function logic checking `missing()` to check `is.null()` instead.
+* Allow to opt out of linting `filter()` expressions in `conjunct_test_linter` via new `allow_filter` argument which defaults to `FALSE` (#2110, @salim-b)
 
 ### New linters
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,13 +27,12 @@
   + `yoda_test_linter()`
 * `sprintf_linter()` is pipe-aware, so that `x %>% sprintf(fmt = "%s")` no longer lints (#1943, @MichaelChirico).
 * `line_length_linter()` helpfully includes the line length in the lint message (#2057, @MichaelChirico).
-* `conjunct_test_linter()` also lints usage like `dplyr::filter(x, A & B)` in favor of using `dplyr::filter(x, A, B)` (part of #884, @MichaelChirico).
+* `conjunct_test_linter()` also lints usage like `dplyr::filter(x, A & B)` in favor of using `dplyr::filter(x, A, B)` unless `allow_filter = TRUE` (part of #884, @MichaelChirico; #2110, @salim-b).
 * `sort_linter()` checks for code like `x == sort(x)` which is better served by using the function `is.unsorted()` (part of #884, @MichaelChirico).
 * `paste_linter()` gains detection for file paths that are better constructed with `file.path()`, e.g. `paste0(dir, "/", file)` would be better as `file.path(dir, file)` (part of #884, @MichaelChirico).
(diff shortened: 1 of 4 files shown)


R/conjunct_test_linter.R
```diff
@@ -32,6 +33,12 @@
 #'   linters = conjunct_test_linter(allow_named_stopifnot = FALSE)
 #' )
 #'
+#' lint(
+#'   text = "dplyr::filter(mtcars,
+#'                         mpg > 20 & vs == 0)",
```
MichaelChirico, line 38: please use tidyverse style in examples

R/conjunct_test_linter.R
```diff
@@ -7,13 +7,14 @@
 #'
 #' Similar reasoning applies to `&&` usage inside [stopifnot()] and `assertthat::assert_that()` calls.
 #'
-#' Relatedly, `dplyr::filter(DF, A & B)` is the same as `dplyr::filter(DF, A, B)`, but the
-#'   latter will be more readable / easier to format for long conditions. Note that this linter
-#'   assumes usages of `filter()` are `dplyr::filter()`; if you're using another function named `filter()`,
-#'   e.g. [stats::filter()], please namespace-qualify it to avoid false positives.
+#' Relatedly, `dplyr::filter(DF, A & B)` is the same as `dplyr::filter(DF, A, B)`, but the latter will be more readable
```
MichaelChirico, line 10: looks like a lot of spurious diff here, any particular reason to widen the lines?

MichaelChirico:
Thanks!

Please add

1. regression tests
2. extend the NEWS item to mention this option & cite yourself

R/conjunct_test_linter.R
MichaelChirico: Yes that's the upper limit... generally better to keep things narrower.

In this case, it's more of a distraction when reviewing to check line-by-line if there's any diff when the only "real" change is your added sentence. So, if you were writing documentation from scratch / a new paragraph, a whichever width you align to is fine. But here, adhering to the "prevaling" width makes review simpler by focusing the diff on what's actually changed.

(two paragraphs about a very minor point, alas)

salim-b: You should use a more powerful diff viewer/algorithm then :wink: I can very much recommend the Rust-based [diffr](https://github.com/mookid/diffr), which implements Myers' longest common subsequence algorithm.

MichaelChirico: Tell it to the GitHub UI team :)

R/conjunct_test_linter.R
salim-b: Ok, see https://github.com/r-lib/lintr/pull/2110/commits/f5c599b19a9b03d6d6ed3e71f53dc95894b76069

NEWS.md
```diff
@@ -33,6 +33,7 @@
 * `seq_linter()` recommends `rev()` in the lint message for lints like `nrow(x):1` (#1542, @MichaelChirico).
 * New `xp_call_name()` helper to facilitate writing custom linters (#2023, @MichaelChirico). This helper converts a matched XPath to the R function to which it corresponds. This is useful for including the "offending" function in the lint's message.
 * `function_argument_linter()` detects usage of `missing()` for the linted argument (#1546, @MichaelChirico). The simplest fix for `function_argument_linter()` lints is typically to set that argument to `NULL` by default, in which case it's usually preferable to update function logic checking `missing()` to check `is.null()` instead.
+* Allow to opt out of linting `filter()` expressions in `conjunct_test_linter` via new `allow_filter` argument which defaults to `FALSE` (#2110, @salim-b)
```
MichaelChirico, line 36: sorry, could you merge it with the existing item on L30? since the inclusion of filter() has not been released yet, I think it makes more sense to have only one item.

salim-b: Ok, see https://github.com/r-lib/lintr/pull/2110/commits/3638b4adba954cadfdf9b3d6267d80fa620628f3

R/conjunct_test_linter.R
```diff
@@ -32,6 +33,14 @@
 #'   linters = conjunct_test_linter(allow_named_stopifnot = FALSE)
 #' )
 #'
+#' lint(
+#'   text = "dplyr::filter(
```
MichaelChirico, line 37: sorry for the toil -- would you mind putting this into an inline expression? it'll be more consistent with examples from other linters. Typically when examples run to >1 line we write the examples  little differently (search for writeLines()), but I don't think it's necessary in this case.

salim-b: Like this? https://github.com/r-lib/lintr/pull/2110/commits/88734f31e1e7cc081916d999ce40a047e2c04937

MichaelChirico (approved):

PR #2112, opened 2023-09-07
lint x %in% NA in equals_na_linter

### LINKED ISSUE #2088: throw equals_na_linter() for x %in% NA
_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2084#discussion_r1296315374_

Originally suggested for inclusion in `scalar_in_linter()`, but agreed it makes more sense to come from `equals_na_linter()`.

MichaelChirico:
Closes #2088

In writing this I had three follow-up questions

 1. Should we customize the lint message for `%in%`/`==`/`!=` (yes)
 2. Should we lint `x %in% c(NA, y)` to encourage `is.na(x) | x %in% y` (I think no)
 3. Should we lint `NA %in% x` (should be `anyNA(x)`; I am less sure, maybe it could go under `any_is_na_linter()`?)


man/conjunct_test_linter.Rd
```diff
@@ -44,10 +44,7 @@ lint(
 )
 
 lint(
-  text = "dplyr::filter(
-    mtcars,
-    mpg > 20 & vs == 0
-  )",
+  text = "dplyr::filter(mtcars, mpg > 20 & vs == 0)",
```
MichaelChirico, line 47: didn't `/document` the previous PR before merging

tests/testthat/test-equals_na_linter.R
```diff
@@ -56,3 +56,7 @@ test_that("equals_na_linter blocks disallowed usages in edge cases", {
   # correct line number for multiline code
   expect_lint("x ==\nNA", list(line_number = 1L, column_number = 1L, ranges = list(c(1L, 4L))), linter)
 })
+
+test_that("x %in% NA is caught", {
+  expect_lint("x %in% NA", "Use is.na for comparisons to NA", equals_na_linter())
```
AshesITR, line 61: Why not add this to the existing patrick case?

AshesITR (commented):
Re 3.: I Agree it should be linted in the `any_is_na_linter()`

AshesITR:
Also should test that `NA %in% x` doesn't lint here.

tests/testthat/test-equals_na_linter.R
MichaelChirico: Mainly b/c i was too preoccupied thinking about the 3 things mentioned in the description :)

MichaelChirico:
Follow-ups:

#2113
#2114

MichaelChirico:
> Also should test that NA %in% x doesn't lint here.

Good call -- the logic was broken (`==`/`!=` throw lint for NA on LHS or LHS of the operator, not so for `%in%`). Done now

AshesITR (approved):

PR #2116, opened 2023-09-07
catch raw strings in file path inputs

### LINKED ISSUE #2075: paste_linter() for file paths should be robust to R"()" strings
_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2074#discussion_r1290609418_

Currently, the logic operates on `STR_CONST`, we could generalize it to use `get_r_string()` instead.

MichaelChirico:
Closes #2075

### DIFF
--- a/R/paste_linter.R
+++ b/R/paste_linter.R
@@ -104,7 +104,7 @@ paste_linter <- function(allow_empty_sep = FALSE,
                          allow_file_path = c("double_slash", "always", "never")) {
   allow_file_path <- match.arg(allow_file_path)
 
-  sep_xpath <- "
+  paste_sep_xpath <- "
   //SYMBOL_FUNCTION_CALL[text() = 'paste']
     /parent::expr
     /following-sibling::SYMBOL_SUB[text() = 'sep' and following-sibling::expr[1][STR_CONST]]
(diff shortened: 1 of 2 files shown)


R/paste_linter.R
```diff
@@ -137,32 +137,36 @@ paste_linter <- function(allow_empty_sep = FALSE,
...
-    /parent::expr
-    /parent::expr[
-      SYMBOL_SUB[text() = 'sep']/following-sibling::expr[1][{slash_str}]
-      and not(SYMBOL_SUB[text() = 'collapse'])
-    ]
-  ")
-
   if (allow_file_path %in% c("always", "double_slash")) {
-    slash_cond <- "contains(text(), '//')"
+    check_double_slash <- function(str) any(grepl("//", str))
```
AshesITR, line 141: `fixed = TRUE`

tests/testthat/test-paste_linter.R
```diff
@@ -198,7 +200,24 @@ test_that("multiple path lints are generated correctly", {
       rex::rex('paste(..., sep = "/")'),
       rex::rex('paste0(x, "/", y, "/", z)')
     ),
-    paste_linter()
+    linter
+  )
+
+  # check vectorization of multiple paste0 file paths
+  expect_lint(
+    trim_some('{
```
AshesITR, line 208: Intersperse with some lints and check on locations?

MichaelChirico: Added one. This is mainly about checking the `vapply()` loop, so just not erroring was good enough for me, but checking a lint can't hurt.

AshesITR (dismissed):

AshesITR:
What's wrong with startsWith() / endsWith()?
Seems more expressive to me and should be faster...

MichaelChirico:
> What's wrong with startsWith() / endsWith()?

Indeed that's why I started with them... but then I noticed we have `which(!(startsWith(...)))` which is 3 calls vs. directly `grep()`. And then the regex is trivial enough, and the variable names are very expressive too. So I switched.

MichaelChirico:
I'm surprised how slow `grep()` is!

```
library(microbenchmark)

l = sample(letters, 1e7, TRUE)
idx <- sample(1e7, 1e5)
l[idx] <- paste0("/", idx)
microbenchmark(grep("^[^/]", l), which(!startsWith(l, "/")))
# Unit: milliseconds
#                        expr       min        lq     mean    median       uq      max neval cld
#            grep("^[^/]", l) 651.28783 663.49967 674.2574 669.93581 677.6424 769.6030   100   b
#  which(!startsWith(l, "/"))  71.68315  80.89737 102.8347  87.80899 100.1424 279.1671   100  a 
```

Hmm switching back

AshesITR (approved):

PR #2119, opened 2023-09-08
Bump checkout action to the latest version

IndrajeetPatil:


MichaelChirico:
It would be nice if there was a way to only have to update this once and have it cascade to all the actions in our repo, is that possible?

IndrajeetPatil:
Not that I am aware of. But, TBH, we shouldn't be doing this manually anyway. And, once https://github.com/r-lib/lintr/pull/2120 is merged, we won't need to. This process will get automated.

MichaelChirico:
> Not that I am aware of. But, TBH, we shouldn't be doing this manually anyway. And, once #2120 is merged, we won't need to. This process will get automated.

Thanks! I read this PR before #2120 🙈. As long as it's automated it will be fine -- I was more worried about "missing" one action by mistake if done manually.

MichaelChirico (approved):

MichaelChirico:
Looks like #2121 will be ready to merge first & it's a wee bit nicer to have this generated automatically. Closing, but thanks!

PR #2120, opened 2023-09-08
Add dependabot workflow

IndrajeetPatil:
To avoid manually doing upkeep like in https://github.com/r-lib/lintr/pull/2119.

This Dependabot configuration is designed to keep GitHub Actions dependencies up-to-date
by running weekly checks and creating pull requests to update them as needed.

This ensures that the GitHub Actions workflows remain compatible with the latest versions
of their dependencies, potentially improving the stability and security of automation processes.

For more, see:
https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
+  # Keep dependencies for GitHub Actions up-to-date
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,8 @@
+version: 2
```
MichaelChirico, line 1: how often will we have to update this version instead 😂

MichaelChirico:
How does this work exactly? It would help to add a link to documentation in the PR description

.github/dependabot.yaml
IndrajeetPatil: Haha. No free lunch 🙈 

It's actually quite stable, and I haven't needed to change it in a couple of years, I think.

MichaelChirico (approved):

Issue #2105, opened 2023-09-04
Enhancement of `unreachable_code_linter` to Detect Unreachable Code in Loops


MEO265:
The `unreachable_code_linter` should also mark code inside if, else, for, and while loops as unreachable, especially when a return or stop function is called.

The following code currently generates only one annotation:

```r
x <- function() { 
  if(TRUE) { 
    stop('This is a test') 
    5+5 
  } 
  return(5L)
  4+4 
} 
```

And this one generates none:
```r
x <- function() { 
  if(TRUE) { 
    stop('This is a test') 
    5+5 
  } 
  4+4 
} 
```

MichaelChirico:
I think this is a duplicate of #1428, though I don't think if(TRUE) constitutes _unreachable_ code. I think it's more relevant for unnecessary _nesting_ (which we have a linter for, though not in {lintr}) yet IIRC

MEO265:
In loops, attention should also be paid to "next" and "break". 

@MichaelChirico I'm not sure if that's a part of what you might have meant in your TODO as well.
https://github.com/r-lib/lintr/blob/c8f00746501e041d2516da58ccca3e054cf305f1/tests/testthat/test-unreachable_code_linter.R#L175-L180

PR #2123, opened 2023-09-08
unreachable_code_linter() lints on code in if (FALSE), while (FALSE)

### LINKED ISSUE #1428: Lint if (FALSE) / while (FALSE) snippets through commented_code_linter?
Such code is _effectively_ commented code, right?

MEO265:
Resolves #1428. 

I held back from linting comments with the new features since, unlike the previous linting scenario, I don't see any issues with it and didn't find any guidelines on this in the tidyverse style guide. But I'm open to tweaking this if needed. Maybe we could even add a flag for this option?

Empty loops are intentionally ignored.

When I get a moment, I'm also planning to tackle issue #2105.

I noticed that the previous examples each contained an unnecessary line in my opinion, but I didn't want to just change it. Am I missing something why `writeLines(code_lines)` is necessary?


MEO265:
I have seen the lines; I will correct them.

R/unreachable_code_linter.R
```diff
@@ -51,16 +100,37 @@ unreachable_code_linter <- function() {
...
+      lints_return_stop[!is_nolint_end_comment],
       source_expression = source_expression,
       lint_message = "Code and comments coming after a top-level return() or stop() should be removed.",
       type = "warning"
     )
+
+    lints_if_while <- xml_nodes_to_lints(
+      lints_if_while,
+      source_expression = source_expression,
+      lint_message = "Code inside a conditional loop with a deterministically false condition is not reachable.",
```
AshesITR, line 123: s/is not reachable/should be removed

R/unreachable_code_linter.R
```diff
@@ -14,6 +15,18 @@
...
+#'
+#' code_lines <- "f <- while (FALSE) {\n 2 + 2\n}"
+#' lint(
+#'   text = code_lines,
+#'   linters = unreachable_code_linter()
+#' )
+#'
 #' # okay
 #' code_lines <- "f <- function() {\n  return(1 + 1)\n}"
 #' writeLines(code_lines)
```
AshesITR, line 32: Feel free to refactor these. They stem from a time before `text=` existed.

AshesITR (commented):
Good work. Don't forget to add a NEWS bullet.

R/unreachable_code_linter.R
MichaelChirico: that's not it -- the idea is these examples run by example(). it is hard to parse what the code actually looks like when there's escapes and newlines. for max readability, writeLines() will show you the exact code being linted.

MichaelChirico: Here's a good example:

https://lintr.r-lib.org/reference/brace_linter.html

`writeLines()` helps the user grok the spacing & newlines without having to build a file mentally

MichaelChirico:
> I held back from linting comments

Makes sense to me... even though some such comments should likely be removed, some are OK, and we won't have a way to distinguish them that I can think of.

NEWS.md
```diff
@@ -35,6 +35,7 @@
 * New `xp_call_name()` helper to facilitate writing custom linters (#2023, @MichaelChirico). This helper converts a matched XPath to the R function to which it corresponds. This is useful for including the "offending" function in the lint's message.
 * `function_argument_linter()` detects usage of `missing()` for the linted argument (#1546, @MichaelChirico). The simplest fix for `function_argument_linter()` lints is typically to set that argument to `NULL` by default, in which case it's usually preferable to update function logic checking `missing()` to check `is.null()` instead.
 * `equals_na_linter()` checks for `x %in% NA`, which is a more convoluted form of `is.na(x)` (#2088, @MichaelChirico).
+* `unreachable_code_linter()` checks for code inside a conditional loop with a deterministically false condition (#1428, @ME0265).
```
MichaelChirico, line 38: I suspect many users won't know what a "conditional loop with a deterministically false condition" means, prompting with an example will make this more accessible. How about "checks for code inside `if (FALSE)` and other conditional loops with deterministically false conditions"?

R/unreachable_code_linter.R
```diff
@@ -1,6 +1,7 @@
 #' Block unreachable code and comments following return statements
 #'
-#' Code after a top-level [return()] or [stop()] can't be reached; typically
+#' Code after a top-level [return()] or [stop()]
+#'   or in deterministically false conditional loops can't be reached; typically
```
MichaelChirico, line 4: same concern about "deterministically false conditional loops", how about "deterministically false conditional loops like `if (FALSE)`"?

R/unreachable_code_linter.R
```diff
@@ -43,6 +72,30 @@ unreachable_code_linter <- function() {
     ]
     /following-sibling::*[1]
   "
+  xpath_if_while <- "
+  //WHILE[following-sibling::expr[1]/NUM_CONST[text() = 'FALSE']]/following-sibling::expr[2]
```
MichaelChirico, line 76: `(//WHILE | //IF)[ ... ]` should work

R/unreachable_code_linter.R
```diff
@@ -43,6 +72,30 @@ unreachable_code_linter <- function() {
...
+      lints,
+      FUN = function(x) {
+        if (xml2::xml_name(xml2::xml_child(x)) == "OP-LEFT-BRACE") {
+          xml2::xml_find_first(x, "expr")
+        } else {
+          x
+        }
+      }
+    )
+    lints <- lints[lapply(lints, FUN = xml2::xml_length) != 0L]
```
MichaelChirico, line 97: shouldn't this be `vapply(lints, xml2::xml_length, integer(1L))`?

R/unreachable_code_linter.R
```diff
@@ -43,6 +72,30 @@ unreachable_code_linter <- function() {
...
+  xpath_else <- "
+  //IF[following-sibling::expr[1]/NUM_CONST[text() = 'TRUE']]
+  /following-sibling::ELSE/following-sibling::expr[1]
+  "
+
+  handle_inline_conditions <- function(lints) {
+    lints <- lapply(
+      lints,
+      FUN = function(x) {
+        if (xml2::xml_name(xml2::xml_child(x)) == "OP-LEFT-BRACE") {
```
MichaelChirico, line 90: `xml_name` and `xml_find_first` are in our namespace, no need to qualify:

https://github.com/r-lib/lintr/blob/c8f00746501e041d2516da58ccca3e054cf305f1/NAMESPACE#L170-L173

R/unreachable_code_linter.R
```diff
@@ -51,16 +104,37 @@ unreachable_code_linter <- function() {
 
     xml <- source_expression$xml_parsed_content
 
-    bad_expr <- xml_find_all(xml, xpath)
+    lints_return_stop <- xml_find_all(xml, xpath_return_stop)
```
MichaelChirico, line 107: nit: elsewhere in the codebase, the flow of logic is rearranged when there's >1 XPath

Here you've done

(1) run all `xml_find_all()`
(2) run all `xml_nodes_to_lints()`

We usually do

(1) Run `xml_find_all()` and `xml_nodes_to_lints()` for XPath 1
(2) ...
(n) Run `xml_find_all()` and `xml_nodes_to_lints()` for XPath n

`package_hooks_linter()` may be a good example.

The result is the same of course, but for consistency I would use the latter pattern.

R/unreachable_code_linter.R
```diff
@@ -51,16 +104,37 @@ unreachable_code_linter <- function() {
...
+      lints_return_stop[!is_nolint_end_comment],
       source_expression = source_expression,
       lint_message = "Code and comments coming after a top-level return() or stop() should be removed.",
       type = "warning"
     )
+
+    lints_if_while <- xml_nodes_to_lints(
+      lints_if_while,
+      source_expression = source_expression,
+      lint_message = "Code inside a conditional loop with a deterministically false condition.",
```
MichaelChirico, line 127: ... should be removed, for consistency with the existing message

tests/testthat/test-unreachable_code_linter.R
```diff
@@ -148,6 +148,151 @@ test_that("unreachable_code_linter ignores terminal nolint end comments", {
...
+  ")
+
+  expect_lint(
+    lines,
+    list(
+      line_number = 4L,
+      message = msg
+    ),
+    unreachable_code_linter()
+  )
```
MichaelChirico, line 293: Suggest a test covering multiple branches, such as

```r
{
  if (FALSE) {
    code
  }
  while (FALSE) {
    code
  }
  if (TRUE) {
  } else {
    code
  }
  stop(".")
  code
}
```

MichaelChirico: There's also no test of `handle_inline_conditions()` where `{` is not used

MichaelChirico:
> I held back from linting comments

OTOH, what about cases like this?

```r
if (FALSE) {
  # some_code()
  # that_i_commented_out()
  # that_should_be_removed()
}
```

MichaelChirico:
Overall, looks good! Thanks for taking this on!

R/unreachable_code_linter.R
```diff
@@ -43,6 +72,30 @@ unreachable_code_linter <- function() {
...
   "
+  xpath_if_while <- "
+  //WHILE[following-sibling::expr[1]/NUM_CONST[text() = 'FALSE']]/following-sibling::expr[2]
+  |
+  //IF[following-sibling::expr[1]/NUM_CONST[text() = 'FALSE']]/following-sibling::expr[2]
+  "
+
+  xpath_else <- "
+  //IF[following-sibling::expr[1]/NUM_CONST[text() = 'TRUE']]
+  /following-sibling::ELSE/following-sibling::expr[1]
```
MichaelChirico, line 83:
```suggestion
    /following-sibling::ELSE/following-sibling::expr[1]
```

MEO265:
> > I held back from linting comments
> 
> OTOH, what about cases like this?
> 
> ```r
> if (FALSE) {
>   # some_code()
>   # that_i_commented_out()
>   # that_should_be_removed()
> }
> ```

I think linter should have independent separate tasks and that would be the `commented_code_linter`'s job

NEWS.md
```diff
@@ -40,6 +40,7 @@
 * `seq_linter()` recommends `rev()` in the lint message for lints like `nrow(x):1` (#1542, @MichaelChirico).
 * `function_argument_linter()` detects usage of `missing()` for the linted argument (#1546, @MichaelChirico). The simplest fix for `function_argument_linter()` lints is typically to set that argument to `NULL` by default, in which case it's usually preferable to update function logic checking `missing()` to check `is.null()` instead.
 * `equals_na_linter()` checks for `x %in% NA`, which is a more convoluted form of `is.na(x)` (#2088, @MichaelChirico).
+* `unreachable_code_linter()` checks for code inside `if (FALSE)` and other conditional loops with deterministically false conditions (#1428, @ME0265).
```
MichaelChirico, line 43: merge this with the bullet on the same linter below, see e.g line 17

tests/testthat/test-unreachable_code_linter.R
```diff
@@ -197,6 +197,229 @@ test_that("unreachable_code_linter ignores terminal nolint end comments", {
...
+    }
+  ")
+
+  expect_lint(
+    lines,
+    list(
+      line_number = 3L,
+      message = msg
+    ),
+    unreachable_code_linter()
```
MichaelChirico, line 219: `linter`, then everything should fit on one line

MichaelChirico (approved):

PR #2124, opened 2023-09-08
length_test_linter() for length(x == 0)

### LINKED ISSUE #1991: Linter for length(x == y) where almost certainly length(which(x == y)) or length(x) == y was intended
This is a mistake that I find myself making all the time:

```r
if (length(x == 0)) ...
```

It makes itself into a really subtle bug and many wasted minutes.

I can't think of a case where `length(x == y)` is preferable to `length(x)` or `length(y)`, unless there's some weird coercion going on and that should be made explicit anyway.

Therefore it seems like a good candidate for a new linter.

~I don't really know what to name such a linter... I also wonder if maybe it's good for an umbrella linter like `common_mistakes_linter()`, though I can't really think of other good candidate rules for that off the top of my head.~ see comment below

MichaelChirico:
Closes #1991.

`length_equals_linter()` generalized to `length_test_linter()` to incorporate other cases where `length()` is applied to a logical vector, e.g. `length(x > 0)`.


tests/testthat/test-length_test_linter.R
```diff
@@ -0,0 +1,30 @@
...
+
+local({
+  ops <- c(lt = "<", lte = "<=", gt = ">", gte = ">=", eq = "==", neq = "!=")
+  linter <- length_test_linter()
+  lint_msg_stub <- rex::rex("Checking the length of a logical vector is likely a mistake. Did you mean ")
+
+  patrick::with_parameters_test_that(
+    "all logical operators detected",
+    {
+      expect_lint(paste("length(x", op, "y)"), rex::rex("`length(x) ", op, " y`?"), linter)
```
IndrajeetPatil, line 25: Nit: Separate the actual and expected outputs to separate lines? 
All those commas make it difficult to figure out where the first argument ends and the second begins, at least for me.

```suggestion
      expect_lint(
        paste("length(x", op, "y)"), rex::rex("`length(x) ", op, " y`?"), 
        linter
      )
```

R/length_test_linter.R
```diff
@@ -0,0 +1,55 @@
+#' Check for a common mistake where length is applied in the wrong place
+#'
+#' Usage like `length(x == 0)` is a mistake. If checking the length of `x`,
+#'   `length(x) == 0` was intended. The length of the logical vector `x == 0`
+#'   will be the same as the length of `x`, so it would be more direct to use
+#'   `length(x)` if that value was really intended. Even in more complicated
+#'   cases like `length(x == y)` where `x` and `y` may have different lengths
+#'   and this expression relies on recycling rules, it is far preferable to
+#'   make the intent clearer, e.g. `max(length(x), length(y))`.
```
IndrajeetPatil, line 9: I found this paragraph difficult to read, and I was trying to slightly rephrase it. This is what I came up with. What do you think?

> It's almost always a mistake to use `length(x == 0)`. If you intended to check the length of `x` and see if it's equal to zero, then you should use `length(x) == 0`. Using `x == 0` as a logical vector will give you the same length as `x`, so it's more direct to use `length(x)` if that's what you intended. In more complex cases like `length(x == y)`, where `x` and `y` may have different lengths and rely on recycling rules, it's better to make your intent clearer. For example, you could use `max(length(x), length(y))`.

tests/testthat/test-length_test_linter.R
MichaelChirico: Split to 3 lines instead

R/length_test_linter.R
MichaelChirico: I just axed most of this. Those cases are quite abstruse -- the point is `length(<obj> <comparator> <obj>)` is always a mistake.

tests/testthat/test-length_test_linter.R
```diff
@@ -0,0 +1,32 @@
...
+  expect_lint("length(DF[key == val, cols])", NULL, linter)
+})
+
+test_that("blocks simple disallowed usages", {
+  linter <- length_test_linter()
+  lint_msg_stub <- rex::rex("Checking the length of a logical vector is likely a mistake. Did you mean ")
+
+  expect_lint("length(x == 0)", rex::rex(lint_msg_stub, "`length(x) == 0`?"), linter)
+  expect_lint("length(x == y)", rex::rex(lint_msg_stub, "`length(x) == y`?"), linter)
+  expect_lint("length(x + y == 2)", rex::rex(lint_msg_stub, "`length(x+y) == 2`?"), linter)
```
IndrajeetPatil, line 14: What's up with the spaces around the infix operator disappearing in the lint message?

R/length_test_linter.R
IndrajeetPatil: Much better. Thanks!

tests/testthat/test-length_test_linter.R
MichaelChirico: that's xml_text() combining across nodes. it's not ideal but more trouble than it's worth to fix imo.

open to other ways of suggesting an alternative that avoids this issue but I was stuck.

AshesITR: How about fetching the code from line data?

MichaelChirico: Not a bad idea, but what about

```r
length(
  foo(
    x, y, z
  ) ==
  bar(
    a, b, c
  )
)
````

I think the current code is pretty simple & handles this case. I would revisit only upon user request as I think it's not very common anyway.

Looking at the linked fixes I filed using this PR, one would have used bad spacing in the lint message:

https://github.com/r-devel/r-svn/pull/144/files#diff-470e0ffa86df95f53eafdae7529ba5fdf7f2f029afef998c998784fbd894db87R507

MichaelChirico: See also https://github.com/r-lib/xmlparsedata/issues/34. I'm not sure a general solution is currently possible.

AshesITR (approved):

PR #2129, opened 2023-09-09
More robust `unreachable_code_linter`

### LINKED ISSUE #2127: More robust `unreachable_code_linter`
As soon as there is a comment after the `return` or `stop`, it is correctly not linted, but neither are the following lines.

Will produce lints:
```r
foo  <- function() {
  return(3)
  5 + 2
}
```

False negative: 
```r
foo  <- function() {
  return(3)  # A comment
  5 + 2
}
```
The same applies to something that @MichaelChirico  has already marked as a todo in the code, but I haven't found an issue about it:
https://github.com/r-lib/lintr/blob/c8f00746501e041d2516da58ccca3e054cf305f1/tests/testthat/test-unreachable_code_linter.R#L181
However, not only will the following expression in the same line not be marked, but also those in subsequent lines will not be marked.

False negative: 
```r
foo  <- function() {
  return(3); x + 1
  5 + 2
}
```

MEO265:
Fixes #2127

In my opinion, statement `(position() != last() - 1 or not(following-sibling::OP-RIGHT-BRACE))`
 is not necessary (At least with the new addition) and made no difference to the tests. If I miss something, let me know.


tests/testthat/test-unreachable_code_linter.R
```diff
@@ -78,6 +78,69 @@ test_that("unreachable_code_linter finds unreachable comments", {
...
+  lines <- trim_some("
+    foo <- function(x) {
+      return(
+        y^2
+      ); 3 + 1
+    }
+  ")
+  expect_lint(
+    lines,
+    rex::rex("Code and comments coming after a top-level return() or stop()"),
```
MichaelChirico, line 91: nit: store these

```r
linter <- unreachable_code_linter()
lint_msg <- rex::rex(...)
```

MichaelChirico:
Great find & thanks for fixing! Only the small change please.

MichaelChirico:
@MEO265 FYI we reserve the "bugs" section if the NEWS for things like "running the linter threw an unexpected error or warning". tightened logic (reduced false positives or false negatives) goes under "new/improved".

MichaelChirico:
(re-reading both items currently there don't belong either! sorry)

MichaelChirico (approved):

PR #2132, opened 2023-09-11
xml2lang retains some whitespace

### LINKED ISSUE #2131: sprintf_linter() fails on complex input
Found in r-devel:

https://github.com/r-devel/r-svn/blob/5da93fa94cd61d510acff8e8975e429a68c839d8/src/library/base/R/all.equal.R#L228-L229

```
lintr::lint("src/library/base/R/all.equal.R", lintr::sprintf_linter())
# Error: Linter 'sprintf_linter' failed in /usr/local/google/home/chiricom/svn/r-devel/src/library/base/R/all.equal.R: <text>:1:65: unexpected symbol
# 1: sprintf("Class definitions are not identical%s",if(isTRUE(ae))""elsepaste
```

MichaelChirico:
Closes #2131

The key for a reprex is that removing whitespace causes a parse failure, so `if (A) x else y` won't work because `if (A) xelsey` is actually fine, and apparently `if (A) "" else ""` is also fine. `if (A) "" else y` is the shortest I could see that causes the parse failure when ` ` is stripped.


NEWS.md
```diff
@@ -6,8 +6,7 @@
 
 ## Bug fixes
 
-* `inner_combine_linter()` no longer throws on length-1 calls to `c()` like `c(exp(2))` or `c(log(3))` (#2017, @MichaelChirico). Such usage is discouraged by `unnecessary_concatenation_linter()`, but `inner_combine_linter()` _per se_ does not apply.
```
AshesITR, line 9: Why move these bugfixes?

MichaelChirico: they're not really bugfixes (which for us is like the issue at hand, where running lint() throws an error) but lint precision/recall fixes. those have gone under new/improved for us historically (and in the ongoing release)

AshesITR (approved):

PR #2133, opened 2023-09-11
improve handling of progress in lint_dir, expose option show_progress

### LINKED ISSUE #972: Allow controlling verbosity of `lintr` during linting
what about `verbose = interactive()` (or `progress=`) as an exposed argument instead? I have wondered how to suppress this output myself sometimes.

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/967#discussion_r831292928_

MichaelChirico:
Closes #972


AshesITR (approved):

AshesITR:
A nice feature extension would be to somehow show the current file in such a way that generated warnings() can be attributed.

MichaelChirico:
> A nice feature extension would be to somehow show the current file in such a way that generated warnings() can be attributed.

agree it's useful... not aware of a way to do so with {utils} progress bars. eventually I think we could allow the user to supply their own progress bar like {purrr} does. for now keeping things simple -- warnings/errors ideally don't happen often (they should usually be caught as lints), so I'm happy for the quick win for now.

PR #2137, opened 2023-09-11
simplify implicit_assignment_linter logic, and catch false negatives

### LINKED ISSUE #2136: implicit_assignment_linter() misses assignments in later arguments of calls
```r
lintr::lint("foo(a, x <- y)\n", lintr::implicit_assignment_linter())
```

MichaelChirico:
Closes #2136 

Note that this addresses the other comments in that issue:

 1. One test is converted to a true positive lint because the operator precedence demands it IMO
 2. One test is dropped since it should be covered by the `except=` argument if so intended

Note also that the lint metadata has changed because we now land on the `<expr>` for the assignment instead of just the assignment operator itself -- I think this is a welcome improvement.


tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -317,11 +304,27 @@ test_that("implicit_assignment_linter blocks disallowed usages in function calls
...
+
+  expect_lint(
+    trim_some("
+      map(
+        .x = 1:4,
+        .f = ~ .x + 1 -> x
+      )"),
+    lint_message,
+    linter
+  )
```
IndrajeetPatil, line 321: Can you please also add a test with a version of this that does _not_ lint?

MichaelChirico: see here: https://github.com/r-lib/lintr/issues/2138#issuecomment-1714362558

in this PR we could do f. = ~(.x + 1 -> x)

but the last PR invalidates that I'm favor of requiring braces

IndrajeetPatil: Okay, then let's leave that for the later PRs.

IndrajeetPatil (approved):

PR #2140, opened 2023-09-11
allow_lazy argument for implicit_assignment_linter to skip lazy assignment

### LINKED ISSUE #2016: New argument for implicit_assignment_linter allowing lazy assignment
Related but I'm not sure it's the same as #2015.

Another type of violation to `implicit_assignment_linter()` that I find a bit awkward to refactor looks like:

```r
if (rare_condition1 && rare_condition2 && foo(key <- val)) {
  bar(key)
}
```

Lazy assignment means it's rare that `key` is actually assigned, so this refactor is wasteful:

```r
key <- val
if (rare_condition1 && rare_condition2 && foo(key)) {
  bar(key)
}
```

The version that seems preferred gets close to an `unnecessary_nested_if()` violation:

```r
if (rare_condition1 && rare_condition2) {
  key <- val
  if (foo(key)) {
    bar(key)
  }
}
```

This increases the indentation vs. the original version; this problem gets worse (possibly combinatorially) if there's more than one such assignment in the same `if()` condition.

IINM this is a subset of #2015, but still could be used in different situations. Again there's no such carveout in the style guide, so this usage should be disallowed by default.

MichaelChirico:
Closes #2016.


NEWS.md
```diff
@@ -47,6 +47,7 @@
 * `implicit_assignment_linter()`
   + finds assignments in call arguments besides the first one (#2136, @MichaelChirico).
   + finds assignments in parenthetical expressions like `if (A && (B <- foo(A))) { }` (#2138, @MichaelChirico).
+  * gains an argument `allow_lazy` (default `FALSE`) that allows optionally skipping lazy assignments like `A && (B <- foo(A))` (#2016, @MichaelChirico).
```
AshesITR, line 50: Bullet consistency

AshesITR (commented):
Not sure if lazy assignment in an `||` expression can be useful. LGTM, though.

MichaelChirico:
> Not sure if lazy assignment in an `||` expression can be useful. LGTM, though.

I found at least 50 cases using it in whatever repos I have cloned locally, here's a neat example:

https://github.com/r-devel/r-svn/blob/b50422e6f5ecaa0fa6a297bbd32d7a1e24b4ac35/src/library/tools/R/build.R#L438

IndrajeetPatil:
Should we deal with the two `backport_linter()` lints seen here? E.g.

```
file=R/lint.R,line=42,col=14,[backport_linter] ...names (R 4.1.0) is not available for dependency R >= 3.5.3. Use the `except` argument of `backport_linter()` to configure available backports.
```

MichaelChirico:
> Should we deal with the two `backport_linter()` lints seen here? E.g.
> 
> ```
> file=R/lint.R,line=42,col=14,[backport_linter] ...names (R 4.1.0) is not available for dependency R >= 3.5.3. Use the `except` argument of `backport_linter()` to configure available backports.
> ```

yes -- I'm not sure why they're happening though:

https://github.com/r-lib/lintr/blob/a0ab99a967567d0d2cbe776d570e66baa9d1806c/.lintr_new#L4

MichaelChirico:
> > Should we deal with the two `backport_linter()` lints seen here? E.g.
> > ```
> > file=R/lint.R,line=42,col=14,[backport_linter] ...names (R 4.1.0) is not available for dependency R >= 3.5.3. Use the `except` argument of `backport_linter()` to configure available backports.
> > ```
> 
> yes -- I'm not sure why they're happening though:
> 
> https://github.com/r-lib/lintr/blob/a0ab99a967567d0d2cbe776d570e66baa9d1806c/.lintr_new#L4

nvm, it's probably #2164

IndrajeetPatil (approved):

Issue #1347, opened 2022-06-02
unreachable_code_linter hits #nolint messages


MichaelChirico:
```
function() {
  do_something
  # nolint start: one_linter.
  edge_case_stuff()
  # nolint stop: one_linter.
}
```

will throw the lint on `# nolint stop`. this seems less than ideal but I'm not sure if it can be easily fixed.

AshesITR:
Maybe never lint `COMMENT` nodes?

MichaelChirico:
Actually comments were one of the things I definitely wanted to lint. Often see things like

```
function() {
  do_something()
  # print(x)
}
```

i.e., vestigial comments that should probably be removed (or at least moved away from the very end of the function)

AshesITR:
Shouldn't that be a separate lint message at least? Trailing comments aren't exactly unreachable code.

Also, just to clarify, these lints are excluded by `exclude()`, right?

MichaelChirico:
Oh, good point. Edited the issue -- they're not excluded because they're specific to a different linter.

AshesITR:
I'm having trouble reproducing this:

```r
expect_lint(
  trim_some("
    foo <- function() {
      do_something
      # nolint start: one_linter.
      a = 42
      # nolint end
    }
  "),
  NULL,
  list(unreachable_code_linter(), one_linter = assignment_linter())
)
```

doesn't fail on main.

PR #2141, opened 2023-09-11
Unreachable loops

### LINKED ISSUE #2105: Enhancement of `unreachable_code_linter` to Detect Unreachable Code in Loops

MEO265:
Closes #2105 

It would also be possible to put the two cases `stop`/`return` and `next`/`break` together. Then there would be no separate message and more than the sensible situations would be tested, e.g. `next` in functions, but it would be faster.

I would be grateful for a hint as to why it is only necessary to filter out the end of "nolint" blocks but not the beginning


MEO265:
I realized I forgot `for` loops, after that it becomes a normal PR

MichaelChirico:
> I would be grateful for a hint as to why it is only necessary to filter out the end of "nolint" blocks but not the beginning

See the original #1347. The idea is that we don't want to lint as "unreachable comment" some code that only exists to terminate a `# nolint` region.

In principle a `# nolint start` region could begin there, too, but in practice I've never seen this. Plain `# nolint` are also ignored since they should only come after actual expressions -- the expression itself will get the `unreachable_codee_linter()` hit, so no need to double that up by linting the comment as well.

R/unreachable_code_linter.R
```diff
@@ -57,14 +57,22 @@
...
     /expr[expr[1][not(OP-DOLLAR or OP-AT) and SYMBOL_FUNCTION_CALL[text() = 'return' or text() = 'stop']]]
     /following-sibling::*[
       not(self::OP-RIGHT-BRACE or self::OP-SEMICOLON)
       and (not(self::COMMENT) or @line2 > preceding-sibling::*[1]/@line2)
     ][1]
   "
+  xpath_next_break <- "
+  ((//REPEAT | //ELSE | //FOR)/following-sibling::expr | (//IF | //WHILE)/following-sibling::expr[2])
+    /expr[NEXT or BREAK]
+    /following-sibling::*[
```
MichaelChirico, line 70: looks like there's a lot shared with the previous XPath -- pull out the shared parts to variables for easier / less repetitive reading?

R/unreachable_code_linter.R
```diff
@@ -104,7 +112,19 @@ unreachable_code_linter <- function() {
...
       expr_return_stop[!is_nolint_end_comment],
       source_expression = source_expression,
-      lint_message = "Code and comments coming after a top-level return() or stop() should be removed.",
+      lint_message = "Code and comments coming after a return() or stop() should be removed.",
+      type = "warning"
+    )
+
+    expr_next_break <- xml_find_all(xml, xpath_next_break)
+
+    is_nolint_end_comment <- xml2::xml_name(expr_next_break) == "COMMENT" &
```
MichaelChirico, line 121: I would add a helper like drop_nolint_end_comment to filter the initial match since it's re-used. that way it's easier to keep the logic in sync & modular.

tests/testthat/test-unreachable_code_linter.R
```diff
@@ -7,6 +7,196 @@ test_that("unreachable_code_linter works in simple function", {
   expect_lint(lines, NULL, unreachable_code_linter())
 })
 
+test_that("unreachable_code_linter works in sub expressions", {
+  linter <- unreachable_code_linter()
+  msg <- rex::rex("Code and comments coming after a return() or stop()")
+
+  lines <- trim_some("
```
MichaelChirico, line 14: long tests like this are good for robustness, but best to start with very succinct tests of only one aspect of the lint logic.

MichaelChirico: it may also help to either annotate the test with a comment on what all is being tested (it looks like "various types of if/else cases"), or to split into different test_that() and record that in the test's name

tests/testthat/test-unreachable_code_linter.R
```diff
@@ -7,6 +7,196 @@ test_that("unreachable_code_linter works in simple function", {
...
+})
+
+test_that("unreachable_code_linter works with next and break in sub expressions", {
+  linter <- unreachable_code_linter()
+  msg <- rex::rex("Code and comments coming after a `next` or `break`")
+
+  lines <- trim_some("
+    foo <- function(bar) {
+      if (bar) {
+        next
```
MichaelChirico, line 114: next doesn't do anything inside an if statement I thought?

```
if (TRUE) {
  next
  print(1)
}
if (TRUE) {
  break
  print(2)
}
```

will give warnings but execute the print IIRC?

MEO265: If there is a `while` or `for` loop around it, the following code will never be executed.  De Linter shouldn't worry about correct use.

I thought that if there is no such loop around it, an error will be thrown.  But I'll test it again.

MEO265: An error is generated, I just tried it with the code you posted.

MichaelChirico: > De Linter shouldn't worry about correct use.

It can be useful to know "this code can't work" when working in an IDE that displays lints while coding. But you're right it is not particularly material to the lint logic here. Feel free to ignore if accounting for this will complicate the linter logic substantially.

MichaelChirico:
Hi @MEO265, we're closing in on sealing up a release (lintr 3.1.1). If you'd like this PR included, please take a chance to wrap up. If you'd like it included and won't have time to work on it in the next few weeks, let us know and we can try taking over from here. Thanks!

MEO265:
> Hi @MEO265, we're closing in on sealing up a release (lintr 3.1.1). If you'd like this PR included, please take a chance to wrap up. If you'd like it included and won't have time to work on it in the next few weeks, let us know and we can try taking over from here. Thanks!

Unfortunately, I have absolutely no idea whether I will find time this week or next.  I would fix everything the following Monday at the latest. 
But feel free to change things yourself, especially if it doesn't fit with the release otherwise.

MichaelChirico:
Hi @MEO265, friendly ping :)

MichaelChirico:
Fixed the more trivial requests. I'll let the long tests slide for now, @MEO265 a follow-up improving the tests would be appreciated if you find the time! Thanks for the PR.

MichaelChirico (approved):

Issue #2135, opened 2023-09-11
lint_dir gets files= argument? Or lint accepts vector of files?


MichaelChirico:
Sometimes, we want to lint a vector of files that's not neatly captured by `lint_dir()`.

Currently, there's three ways to do so:

 1. Use `lint_dir()` with non-default `patterns=`. This is not very flexible -- `patterns=` is passed to `dir(patterns=)`, which only operates on `basename()`.
 1. manually loop over the vector with `lint()`. This sacrifices several things handled by `lint_dir()`, like a progress meter, possible parallelization (#485), `flatten_lints()` merging results, etc.
 2. Use a `.lintr` config to set exclusions. This works well for simple packages/projects, but is not fully general and very clunky for interactive use cases.

One use case I have in mind is linting the R sources. `lint_dir()` includes lots of files that we'd typically not want to lint, e.g. test scripts for the R parser that intentionally use bad R syntax, and the concatenated source `all.R` files. It's very hard to get `lint_dir()` to ignore those `all.R` files especially, which is very inefficient, so the best way to lint the R sources currently is to filter `list.files()` by hand, then `lintr:::flatten_lints(lapply(filtered_files, lint, ...))`.

I can see two options for improving this situation:

 1. `lint_dir()` gets an argument specifying a vector of files, e.g. `files=`, user can specify either `path=` or `files=` and the rest, including `lint_package()` is unchanged.
 2. `lint()` accepts a vector of file names in `filename=`.

I can see arguments for both; I'm currently leaning towards option (1) because the code change will be very small & will make it easier to plug in to parallelization in the future. The main advantage of option (2) is not growing the argument surface of any functions.

Filing issue here for feedback/discussion.

AshesITR:
Slightly leaning to `lint()`. Vectorizing `lint()` should keep a stable output type IINM, so has no apparent downsides to me.

Progress handling and flattening should be refactored and `lint_package()` should probalby also be rewritten to use `lint()` directly.

While we're at it, we should split `lint.R` into bite sized chunks. It's become quite long.

New calls could be architected like this:

`lint_{dir,package} -> lint() -> lint_impl()`

Where `lint_impl()` is the unvectorized `lint()` and not exported.

`lint_{dir,package}()` would do file enumeration and delegate the rest to `lint()`.

MichaelChirico:
NB: This might also simplify the `lint_changed_files` workflow:

[...]

PR #2142, opened 2023-09-11
rename `lint_dir()` test file for consistency

MichaelChirico:
Related: #2135


AshesITR:
Pull `lint_dir()` into `R/lint_dir.R`?

MichaelChirico:
> Pull `lint_dir()` into `R/lint_dir.R`?

That'll probably happen... my thinking is we should nail down exactly how #2135 will happen before doing that. This change can happen regardless since (1) `git mv` shows cleanly in the git history (whereas moving a code chunk is a bit harder to trace) (2) it's consistent with how `lint_package()` is tested in `test-lint_package.R` (https://github.com/r-lib/lintr/blob/a59c6c06229c6e9c71d910fe31d379a554255bea/tests/testthat/test-lint_package.R).

AshesITR (approved):

PR #2143, opened 2023-09-11
interpret glue in unused_import_linter, too

### LINKED ISSUE #2042: False positive [unused_import_linter]
It seems like when using a function inside a glue() call, this lintr doesn't recognise a function is being used.
An example
```
library(praise)

glue::glue(
  "Hi Team,
          
          {praise()}  

          All the best"
)
```

Which is flagged as
`[unused_import_linter] Package 'praise' is attached but never used.`

MichaelChirico:
Closes #2042


R/shared_constants.R
```diff
@@ -228,6 +228,68 @@ strip_names <- function(x) {
   x
 }
 
+extract_glued_symbols <- function(expr, interpret_glue) {
```
AshesITR, line 231: Add roxygen?
Maybe add tests so we can later expand more glue() like functions?

R/shared_constants.R
```diff
@@ -228,6 +228,68 @@ strip_names <- function(x) {
...
+    parsed_call <- tryCatch(xml2lang(glue_call), error = unexpected_error, warning = unexpected_error)
+    parsed_call[[".envir"]] <- glued_symbols
+    parsed_call[[".transformer"]] <- symbol_extractor
+    # #1459: syntax errors in glue'd code are ignored with warning, rather than crashing lint
+    tryCatch(eval(parsed_call), error = parse_failure_warning)
+  }
+  names(glued_symbols)
+}
+
+unexpected_error <- function(cond) {
```
AshesITR, line 270: These functions should be either local to `extract_glued_symbols()` or named more specific.

R/shared_constants.R
MichaelChirico: > Maybe add tests so we can later expand more glue() like functions?

Not sure I follow, there are already a lot of tests in `object_usage_linter()`:

https://github.com/r-lib/lintr/blob/a59c6c06229c6e9c71d910fe31d379a554255bea/tests/testthat/test-object_usage_linter.R#L345-L538

R/shared_constants.R
MichaelChirico: Done. Pulled them out to avoid overhead of declaring several functions each time `extract_glued_symbols` is invoked.

R/shared_constants.R
AshesITR: My thinking was that since this now got moved, we might more explicitly nail down the interface. Fine without tests, too. But roxygen should be there I think (with `@noRd`)

R/shared_constants.R
```diff
@@ -228,6 +228,78 @@ strip_names <- function(x) {
...
 
+#' Pull out symbols used in glue strings under the current sub-tree
+#'
+#' Required by any linter (e.g. [object_usage_linter()] / [unused_imports_linter()])
+#'   that lints based on whether certain symbols are present, to ensure any
+#'   symbols only used inside glue strings are also visible to the linter.
+#'
+#' @param expr An XML AST
+#' @param interpret_glue Logical, if `FALSE` return nothing.
+#' @noRd
```
AshesITR, line 239: `@return A character vector of extracted symbols`?

AshesITR (dismissed):

AshesITR (approved):

PR #2146, opened 2023-09-12
test lint() in test-lint

MichaelChirico:
Companion to #2145.


MichaelChirico:
>   File names must not differ just by case to be usable on all R platforms.

I see... that's somewhat annoying.

IndrajeetPatil (approved):

PR #2151, opened 2023-09-12
update lint_dir to lint .qmd files by default; fixes r-lib/lintr#2150

### LINKED ISSUE #2150: lint_dir() does not lint .qmd files by default
The documentation for `lint_dir()` claims that .qmd files are included in the default argument for 'pattern', but I have discovered that this is not the case. The default argument for patter is a `rex()`-generated regex that matches '.Rmd' and '.rmd', but neither '.qmd' nor '.Qmd'.

I would make a pull request but I find rex pretty confusing, and the default argument is already quite long.

Here is `lint_dir()` as documented. It's easy to see that there's no allowance made for qmd files:

```r
lint_dir(
  path = ".",
  ...,
  relative_path = TRUE,
  exclusions = list("renv", "packrat"),
  pattern = rex::rex(".", one_of("Rr"), or("", "html", "md", "nw", "rst", "tex", "txt"),
    end),
  parse_settings = TRUE
)
```

And here is a worked example using the same regex:

``` r
lint_dir_regex <- rex::rex(".", one_of("Rr"), or("", "html", "md", "nw", "rst", "tex", "txt"), end)

# .rmd works:
stringr::str_detect(".rmd", lint_dir_regex)                                                      
#> [1] TRUE

# .qmd does not:
stringr::str_detect(".qmd", lint_dir_regex)
#> [1] FALSE

```

<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dave-lovell:
Closes #2150 

Updates the default of `lint_dir()`s 'pattern' argument, so that '.Qmd'/'.qmd' files are linted by default, as described in the function documentation.


dave-lovell:
Okay I do not understand git. Not sure why this is failing when all I've done is slightly adjust a default argument

dave-lovell:
> Okay I do not understand git. Not sure why this is failing when all I've done is slightly adjust a default argument

Oh okay I've just edited a single file in-browser like a lemon, but that means the documentation is now out of sync with the actual function formals.

MichaelChirico:
@dave-lovell try clicking through the ❌ and see if you can't understand what it's telling you.

You've only updated the R file, but the documentation needs to be updated as well (that's what gets displayed here & it has to be in sync w the actual code: https://lintr.r-lib.org/reference/lint.html).

The safest thing is to run roxygenize().

MichaelChirico:
> > Okay I do not understand git. Not sure why this is failing when all I've done is slightly adjust a default argument
> 
> Oh okay I've just edited a single file in-browser like a lemon, but that means the documentation is now out of sync with the actual function formals.

yep exactly... if you don't have access to a machine that can run roxygen, let me know & I'll try to do it for you.

MichaelChirico:
ideally, well also add a test to ensure this works & continues to do so. let me know if you'd like to have a go at that -- you could imitate other code in tests/testthat/test-lint_dir.R

R/lint.R
```diff
@@ -128,7 +128,8 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
 lint_dir <- function(path = ".", ...,
                      relative_path = TRUE,
                      exclusions = list("renv", "packrat"),
-                     pattern = rex(".", one_of("Rr"), or("", "html", "md", "nw", "rst", "tex", "txt"), end),
+                     pattern = rex(".", or(group(one_of("Rr"), or("", "html", "md", "nw", "rst", "tex", "txt")),
```
AshesITR, line 131: No need for a capturing `group()`, `list()` should work just fine IINM.

vignettes/test.Qmd
```diff
@@ -0,0 +1,52 @@
+---
+title: "Test"
+author: "dave-lovell"
+format: html
+editor: visual
+---
```
IndrajeetPatil, line 6: Adding this vignette is unnecessary. 

For testing purposes, you can probably use- `tests/testthat/knitr_formats/test.qmd`.

dave-lovell: This just seemed like the easiest implementation since the tests in `test-lint_dir.R` already use `test_path("dummy_packages", "package", "vignettes")` for the "lint all files in a directory" test.

dave-lovell: Oh wait are you saying it would be simpler to copy the file you've mentioned into this directory? Got it.

R/lint.R
dave-lovell: I've just tested this: using `list()` instead of `group()' doesn't work - no files are linted in tests.

vignettes/test.Qmd
dave-lovell: I've implemented this in f96b8b0

dave-lovell:
Okay I think this is ready:

✅ Function defaults updated
✅ Roxygenised
✅ Relevant tests updated 
✅ All checks complete
✅ All tests passing

Thanks a lot everyone, it's been a rewarding educational journey.

R/lint.R
MichaelChirico, line 131: @AshesITR WDYT about making this case-insensitive ourselves by prepending `(?i)` to the `rex()` output?

AFAICT there's not a way to do that within {rex} itself: https://github.com/r-lib/rex/issues/85

That would make this a lot simpler: `rex(".", or("r", "rmd", "qmd", "rnw", "rhtml", "rrst", "rtex", "rtxt"), end)` then later `dir(pattern = paste0("(?i)", pattern))`.

Alternatively we can add a helper for this to hide the ugly/distracting details from the user: `pattern = r_files_pattern()`.

MichaelChirico: @AshesITR I was also skeptical, but note that `rex:::group()` is a _non-capturing_ group:

```r
rex(group(or("a", "bc")))
# (?:(?:a|bc))
```

MichaelChirico:
Thanks @dave-lovell, the change & tests LGTM.

R/lint.R
dave-lovell: This seems smart. I was wondering about a helper when I found myself writing that quite unwieldy line. Would it have been smart to raise that question in the issue before creating this RP? I'm still trying to get a handle on GH etiquette/best practice

MichaelChirico: Thanks for asking!

I'm not sure there's a universal rule. Certainly it can't hurt to ask "Hey, it seems a helper would be more useful here, WDYT?" esp. along with a proposed implementation. Whether to go ahead and do so without suggestion is a bit more dicey.

Ultimately we as reviewers/maintainers will have the best understanding on if it's the right choice for the codebase -- if you're new to the codebase, it's possible there's something missing from the bigger picture, or there's just a style preference why things are done a certain way.

AshesITR: Thanks for double checking and sorry for the confusion.

MichaelChirico:
@dave-lovell Don't forget to add a NEWS bullet and cite yourself!

R/lint.R
MichaelChirico: @AshesITR gentle ping on the question above. I lean towards a helper (and exporting it), it will also be useful for users customizing lint patterns after #2135.

AshesITR: Silently adding `(?i)` to used-supplied egexes seems like asking for trouble. For our default regex I think the only important thing is that the list of file extensions is easily visible to the user.

What about creating a helper `rex_file_exts("R", ..., ignore_case = TRUE)`?

AshesITR: Note that filesystems are case-sensitive on all platforms except windows.

AshesITR: I'm also fine with ditching rex() here until the issue is resolved and instead using a string regex.

MichaelChirico: My memory is MacOS is also case-insensitive, which a quick google suggests is true (at least by default): https://discussions.apple.com/thread/251191099... anyway you're right. What I had in mind is silently applying that to the default argument, not user-supplied arguments, which we could do with `missing()` but better to avoid that.

`rex_file_exts()` looks reasonable enough.

Let's merge this as-is for now & handle as a follow-up? If so I'll write the NEWS.

MichaelChirico: > I'm also fine with ditching rex() here until the issue is resolved and instead using a string regex.

one concern there is `(?i)` is not particularly common (?), so it might be (?) confusing to expose that in the default value.

AshesITR: > > I'm also fine with ditching rex() here until the issue is resolved and instead using a string regex.
> 
> 
> 
> one concern there is `(?i)` is not particularly common (?), so it might be (?) confusing to expose that in the default value.

If that were the case, we'd to a (humble) job towards educating our users about it 😊
I expect R programmers to be able to google for `regex (?i)` if they don't understand and want to know.

MichaelChirico: I was worried `(?i)` wouldn't google well but I am wrong :)

MichaelChirico (approved):

PR #2152, opened 2023-09-12
fix internal lints

MichaelChirico:
Uncovered by #2137

`brace_linter()` and `implicit_assignment_linter()` together create a lot of whitespace in comparison 👀


R/expect_lint.R
```diff
@@ -80,11 +80,12 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
 
   local({
     itr <- 0L
+    this_env <- environment()
```
AshesITR, line 83: Maybe call this `outer_env` everywhere for consistency? Also saves having to come up with new names all the time.

The inner code will be more clear too, I feel.
`outer_env$itr <- outer_env$itr + 1L`

R/zzz.R
```diff
@@ -289,13 +289,17 @@ settings <- NULL
...
+  #   that requries such calls to pass asNamespace("lintr") explicitly instead of
+  #   re-using the getNamespace() outcome as we do here.
   for (base_fun in c("str2lang", "str2expression")) {
     if (!exists(base_fun, base_ns)) {
       assign(base_fun, get(base_fun, backports_ns), lintr_ns)
     }
   }
 
-  default_settings <<- list(
+  assign("default_settings", envir = lintr_ns, list(
```
AshesITR, line 302: Use `assignInMyNamespace()`?

vignettes/lintr.Rmd
```diff
@@ -9,6 +9,7 @@ vignette: >
 ---
 
 ```{r, include = FALSE}
+library(lintr)
```
AshesITR, line 12: Is it standard practice in vignettes to silently load the package?

R/zzz.R
MichaelChirico: see the comment... I tried that but it threw an R CMD check warning (it looks for potentially risky calls to assignInNamespace(), flagging anything that doesn't use asNamespace("lintr") exactly). we could match that, but we already use assign() like this for backports, so I continued to do so.

AshesITR: There is assignIn*My*Namespace for that particular use case, no?

MichaelChirico: > There is assignIn_My_Namespace for that particular use case, no?

Ah, sorry, read on phone. Trying now.

Side note/related: https://bugs.r-project.org/show_bug.cgi?id=18595

vignettes/lintr.Rmd
MichaelChirico: Hmm, not super common:

https://github.com/search?q=org%3Acran%20path%3Avignettes%20%2Fopts_chunk%5B%24%5Dset%2F&type=code

I'd call this a false positive of `library_call_linter()` which should probably exempt antecedent `opts_chunk$set()` calls (though our `library(lintr)` probably _should_ appear higher in the vignette). I'll skip this change for now.

R/expect_lint.R
```diff
@@ -80,23 +80,24 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
 
   local({
     itr <- 0L
```
MichaelChirico, line 82: bah... `object_usage_linter()` false positives... let's ditch the `<<-` updates for now. Our `<<-` usage is pretty safe overall, where the initial assignment of the `<<-` target is always _just_ before the scoped function using `<<-`.

tests/testthat/test-methods.R
```diff
@@ -144,7 +144,11 @@ test_that("split.lint works as intended", {
 test_that("within.list is dispatched", {
   l <- lint(text = "a=1\nb=2", linters = infix_spaces_linter())
   expect_silent({
-    l <- lapply(l, within, line_number <- line_number + 1L)
+    l <- lapply(l, function(li) {
```
MichaelChirico, line 147: Seems there's a fundamental conflict among linters here. The current code violates `unnecessary_lambda_linter()`...

I'll proceed by `# nolint`ing the original code. It's much more concise anyway.

(actually, since we haven't turned on `implicit_assignment_linter()`, I'll just revert this change for now)

PR #2154, opened 2023-09-12
Add a test for library_call_linter

MichaelChirico:
This passes current tests, though I think there's a fair amount of work to do in light of that.


AshesITR:
```r
fun()
library(moreFun)
oops()
```

should fail IINM.

MichaelChirico:
> ```r
> fun()
> library(moreFun)
> oops()
> ```
> 
> should fail IINM.

Thanks, I've added this as a test case instead.

AshesITR (approved):

PR #2155, opened 2023-09-13
add more emphasis about executing linters

### LINKED ISSUE #1137: docs: package to be linted should be installed
Linting a package that is not installed may generate false positives for (at least) `object_usage_linter`, see e.g. https://github.com/r-lib/actions/issues/557

1. Why is that exactly? Is #91 still relevant here?
2. (Where) should this be documented? 

Re 2: I took https://github.com/r-lib/lintr/blob/fff78086a1f70f42f8994f2f37054075f3ff3f14/README.md?plain=1#L217-L218 to mean that the package to be linted must be installed, but I guess it just means that lintr needs to be installed (which is self-evident, though, I suppose)? Anyway, my first thought was to document it at the beginning of the the https://github.com/r-lib/lintr/blob/fff78086a1f70f42f8994f2f37054075f3ff3f14/README.md?plain=1#L192 
section, because it is also valid for GHA. 

But the "package should be installed" applies locally as well, so maybe it should be in the `lint_package()` docs? Maybe something like  
> Note: To avoid false positives, the package to be linted needs to be installed.

MichaelChirico:
Closes #1137

Not sure if there's anywhere else we should make this note... the [README](https://lintr.r-lib.org/)? [Editor setup](https://lintr.r-lib.org/articles/editors.html)? [Get started](https://lintr.r-lib.org/articles/lintr.html)? I gave those a quick once-over but didn't see an easy fit.


AshesITR:
Getting started sounds like a good place to put the hint, too.
Maybe where `lint_package()` is mentioned?

IndrajeetPatil (approved):

PR #2164, opened 2023-09-14
Add backports exception to .lintr too

MichaelChirico:


MichaelChirico:
I forget why we keep separate `.lintr` and `.lintr_new`, @IndrajeetPatil is that a relic of when the package was "lintier"?

IndrajeetPatil (approved):
LGTM

IndrajeetPatil:
> I forget why we keep separate `.lintr` and `.lintr_new`, @IndrajeetPatil is that a relic of when the package was "lintier"?

No, I think we use `.lintr` for `lint` workflow, while `.lintr_new` for `lint-changed-files` workflow.

I think we should be considering using `.lintr` for both workflows to avoid such confusion and duplication.

AshesITR:
We aimed once to converge to a single `.lintr` but decided to keep `.lintr_new` more strict for performance reasons IIRC.

MichaelChirico:
> We aimed once to converge to a single `.lintr` but decided to keep `.lintr_new` more strict for performance reasons IIRC.

I see... doesn't appear to be a concern anymore, the `lint` workflow has always finished before the `R CMD check` tests IME:

https://github.com/r-lib/lintr/actions/workflows/lint.yaml?page=5&query=is%3Acompleted

Issue #2015, opened 2023-07-24
New argument for implicit_assignment_linter to skip "scoped" usages


MichaelChirico:
I sometimes find the following type of usage to be appropriate for "implicit" assignment:

```
if (any(idx <- x > 0)) {
  stop("Found positive entries: [", toString(which(idx)), "]")
}
```

The object `idx` is not used outside the branch, so I think it can be distracting to define it in the "main" body of the function. I believe @IndrajeetPatil has a cite on this being good practice in some contexts too.

It would be nice to allow this sort of implicit assignment, defined by the object of assignment is (1) inside `if` and (2) only referenced inside that `if`'s parent `<expr>`.

This argument (let's say `allow_scoped_if_assignment`) would be `FALSE` by default since the style guide does not consider this exception.

MichaelChirico:
One question is whether `foo(x <- 1)` should be linted if `x` is not referenced anywhere else in the code. I think yes (esp. since this is a common mistake where `=` was intended), but it will make implementation a bit harder.

PR #2165, opened 2023-09-14
Remove separate .lintr_new

MichaelChirico:
Is there a reason to keep both workflows now @IndrajeetPatil ? Or is lint-changed-files sufficient?

Or maybe we should drop lint-changed-files and make the regular lint workflow error instead? My thinking is the latter will catch new lints that pop up in unchanged files as we improve linter logic.


IndrajeetPatil:
> Or maybe we should drop lint-changed-files and make the regular lint workflow error instead? My thinking is the latter will catch new lints that pop up in unchanged files as we improve linter logic.

Yes, I would also personally prefer this option. 

E.g., in this very PR, `lint` is flagging three [new lints](https://github.com/r-lib/lintr/actions/runs/6188594718/job/16800944012?pr=2165) that we will discover only when the respective files change.

MichaelChirico:
Some of those are fixed in #2152. I left this implicit assignment on purpose:

https://github.com/r-lib/lintr/blob/306a078d0da9f035e131ff62d9f4a432f20850ee/R/tree_utils.R#L9

I would mark it as `# nolint` pending #2015.

MichaelChirico:
Marked as draft until #2152 is merged.

tests/testthat/test-methods.R
```diff
@@ -144,7 +144,7 @@ test_that("split.lint works as intended", {
 test_that("within.list is dispatched", {
   l <- lint(text = "a=1\nb=2", linters = infix_spaces_linter())
   expect_silent({
-    l <- lapply(l, within, line_number <- line_number + 1L)
+    l <- lapply(l, within, line_number <- line_number + 1L) # nolint: implicit_assignment_linter. Workaround is pretty horrid.
```
IndrajeetPatil, line 147: This change is producing `line_length_linter()` lint.

IndrajeetPatil (approved):

PR #2170, opened 2023-09-14
allow_preamble in library_call_linter() to skip setup code

### LINKED ISSUE #2161: `library_call_linter()` should relax for Rmd files
Hmm, not super common:

https://github.com/search?q=org%3Acran%20path%3Avignettes%20%2Fopts_chunk%5B%24%5Dset%2F&type=code

I'd call this a false positive of `library_call_linter()` which should probably exempt antecedent `opts_chunk$set()` calls (though our `library(lintr)` probably _should_ appear higher in the vignette). I'll skip this change for now.

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/2152#discussion_r1323530530_

Maybe an easy workaround would be to add `require_first=TRUE` and make the linter optionally more lax in the sense that it only lints if `library()` calls are interrupted by other code. This would force grouping all library calls together without forcing this group all the way to the top.

MichaelChirico:
Closes #2161

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,7 +58,7 @@
 
 ### New linters
 
-* `library_call_linter()` can detect if all library/require calls are not at the top of your script (#2027, #2043, and #2163, @nicholas-masel and @MichaelChirico).
+* `library_call_linter()` can detect if all library/require calls are not at the top of your script (#2027, #2043, #2163, and #2170, @nicholas-masel and @MichaelChirico).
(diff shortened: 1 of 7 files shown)


AshesITR:
❤️ our test suite.

Code LGTM, just add the configurable tag.

AshesITR (approved):

AshesITR (approved):

PR #2171, opened 2023-09-14
allow_scoped in implicit_assignment_linter() for scope-limited assignments

### LINKED ISSUE #2015: New argument for implicit_assignment_linter to skip "scoped" usages

MichaelChirico:
Closes #2015.


MichaelChirico:
Ran this on my local github directory.

Compared to `allow_scoped = FALSE` (4316 lints), `allow_scoped = TRUE` has 3172 lints (1144=27% of lints removed).

Here is a sample of 5 that were eliminated:

 - https://github.com/r-devel/r-svn/blob/5359b1d3a586bb56e70727abbd68da6dba3af13a/src/library/tools/R/QC.R#L5597-L5608 (also skipped under `allow_lazy = TRUE`)
 - https://github.com/r-devel/r-svn/blob/5359b1d3a586bb56e70727abbd68da6dba3af13a/src/library/grDevices/R/device.R#L456
 - https://github.com/r-devel/r-svn/blob/5359b1d3a586bb56e70727abbd68da6dba3af13a/src/library/tools/R/QC.R#L8164-L8167
 - https://github.com/ycroissant/plm/blob/541e160683dc1e08d71b0a193969212db216bdf1/inst/tests/test_order_between_fixef_ranef.R#L180 **This is a false negative...**
 - https://github.com/lme4/lme4/blob/b5b6ad96a05394795b045102664035712b1f9f82/tests/refit.R#L176-L177 **This is a false negative...**

R/implicit_assignment_linter.R
```diff
@@ -75,6 +92,14 @@ implicit_assignment_linter <- function(except = c("bquote", "expression", "expr"
   if (allow_lazy) {
     xpath <- paste0(xpath, "[not(ancestor::expr/preceding-sibling::*[self::AND2 or self::OR2])]")
   }
+  if (allow_scoped) {
+    # force 2nd preceding to ensure we're in the loop condition, not the loop expression
+    in_branch_cond <- "ancestor::expr[preceding-sibling::*[2][self::IF or self::WHILE]]"
+    xpath <- paste0(
+      xpath,
+      glue("[not({in_branch_cond}) or expr[1]/SYMBOL = {in_branch_cond}/parent::expr/following::SYMBOL]")
```
AshesITR, line 100: What does the second or cond do?

MichaelChirico: `expr[1]/SYMBOL = {in_branch_cond}/parent::expr/following::SYMBOL` this one?

It's looking for the symbol on the LHS of `<-` matching a symbol anywhere later in the file. That's very conservative (e.g. it'll match that symbol in NSE and in different environments like other functions), but seems like a reasonable/simple enough start. I'll add a comment.

tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -383,3 +383,87 @@ test_that("allow_lazy lets lazy assignments through", {
...
+      if (any(idx <- foo()) && BB) {
+        stop('Invalid foo() output: ', toString(idx))
+      }
+    "),
+    NULL,
+    linter
+  )
+  expect_lint(
+    trim_some("
+      if (any(idx <- foo()) && BB) {
```
AshesITR, line 461: Add a truly lazy test? (assignment on RHS of AND2)

AshesITR (approved):

PR #2176, opened 2023-09-15
Add some more clarity to settings docs

MichaelChirico:
I think there's still more we can do, but this is a nice quick improvement


AshesITR (approved):
Good idea - especially aliasing `?.lintr`!

Issue #2134, opened 2023-09-11
Improve error message for misspelled arguments?


IndrajeetPatil:
I am reporting a minimal version of the function call here, but it took me a long time to figure out that linting was failing for a very simple reason: I had misspelled an argument. It took so long because the error message just spit out a big block of code, instead of an informative error message. 

Should we improve error message here?

```r
library(lintr)
lint_package(litners = all_linters())
```

## Actual error message

```r
Error in FUN(X[[i]], ...) : 
  unused argument (litners = list(function (source_expression) 
{
    lapply(ids_with_token(source_expression, "STR_CONST"), function(id) {
        token <- with_id(source_expression, id)
        path <- get_r_string(token$text)
        if (path_function(path)) {
            start <- token[["col1"]] + 1
            end <- token[["col2"]] - 1
            Lint(filename = source_expression[["filename"]], line_number = token[["line1"]], column_number = start, type = "warning", message = message, line = source_expression[["lines"]][[as.character(token[["line1"]])]], ranges = list(c(start, end)))
        }
    })
}, function (source_expression) 
{
    if (!is_lint_level(source_expression, "expression")) {
        return(list())
    }
    xml <- source_expression$xml_parsed_content
    any_duplicated_expr <- xml_find_all(xml, any_duplicated_xpath)
    any_duplicated_lints <- xml_nodes_to_lints(any_duplicated_expr, source_expression = source_expression, lint_message = "anyDuplicated(x, 
```

## Expected error message

```r
Error: There is no argument called `litners`. Did you misspell argument name?
```

MichaelChirico:
More context: `exclude()` and then `parse_exclusions()` is picking up `...` so that's where the error's coming from.

PR #2177, opened 2023-09-15
allow config to be an .R file (RFC)

### LINKED ISSUE #1210: Writing R code in a DCF is unappealing
Since you don't get syntax highlighting, help, autocomplete, ...

What do you think about supporting some additional way of defining custom linters that uses a file with a `.R` extension?

MichaelChirico:
Closes #1210

Here's a minimal change to allow `getOption("lintr.linter_file")` to be an R script. My thinking is to include this in the release as an "experimental" feature & decide by 3.2.0 the future of config files for {lintr}. Some notes:

 - There was some concern in the issue about "what if multiple configs are found", I'm not sure how to map that to the code, because everything is governed by the `lintr.linter_file` option. That points us to which config to use.
 - We use the file extension to decide how to parse the file
 - The change in `get_setting()` is a bit awkward. It might be preferable to `parse()` the DCF keys up-front instead.

Still needs tests/NEWS, but wanted to open discussion first.

### DIFF
--- a/R/settings.R
+++ b/R/settings.R
@@ -22,18 +22,7 @@ read_settings <- function(filename) {
     default_settings[["encoding"]] <- default_encoding
   }
 
-  if (!is.null(config_file)) {
-    malformed <- function(e) {
-      stop("Malformed config file, ensure it ends in a newline\n  ", conditionMessage(e), call. = FALSE)
-    }
-    config <- tryCatch(
-      read.dcf(config_file, all = TRUE),
-      warning = malformed,
-      error = malformed
-    )
-  } else {
-    config <- NULL
-  }
+  config <- read_config_file(config_file)
 
   for (setting in names(default_settings)) {
     value <- get_setting(setting, config, default_settings)
@@ -50,18 +39,45 @@ read_settings <- function(filename) {
   }
 }
 
+read_config_file <- function(config_file) {
+  if (is.null(config_file)) {
+    return(NULL)
+  }
+
+  if (endsWith(config_file, ".R")) {
+    config <- new.env()
+    read <- function(file) {
+      sys.source(file, config)
+      config
+    }
+    malformed <- function(e) {
+      stop("Malformed config file, ensure it is valid R syntax\n  ", conditionMessage(e), call. = FALSE)
+    }
+  } else {
+    read <- function(file) read.dcf(file, all = TRUE)
+    malformed <- function(e) {
+      stop("Malformed config file, ensure it ends in a newline\n  ", conditionMessage(e), call. = FALSE)
+    }
+  }
+  tryCatch(read(config_file), warning = malformed, error = malformed)
+}
+
 get_setting <- function(setting, config, defaults) {
   option <- getOption(paste(sep = ".", "lintr", setting))
   if (!is.null(option)) {
     option
   } else if (!is.null(config[[setting]])) {
-    malformed <- function(e) {
-      stop("Malformed config setting '", setting, "'\n  ", conditionMessage(e), call. = FALSE)
+    if (is.environment(config)) {
+      config[[setting]]
+    } else {
+      malformed <- function(e) {
+        stop("Malformed config setting '", setting, "'\n  ", conditionMessage(e), call. = FALSE)
+      }
+      tryCatch(
+        eval(parse(text = config[[setting]])),
+        error = malformed
+      )
     }
-    tryCatch(
-      eval(parse(text = config[[setting]])),
-      error = malformed
-    )
   } else {
     defaults[[setting]]
   }


R/settings.R
```diff
@@ -50,18 +39,45 @@ read_settings <- function(filename) {
...
+    config <- new.env()
+    read <- function(file) {
+      sys.source(file, config)
+      config
+    }
+    malformed <- function(e) {
+      stop("Malformed config file, ensure it is valid R syntax\n  ", conditionMessage(e), call. = FALSE)
+    }
+  } else {
+    read <- function(file) read.dcf(file, all = TRUE)
```
AshesITR, line 57: This is where the `get_setting()` akwardness should go IMO

R/settings.R
```diff
@@ -50,18 +39,45 @@ read_settings <- function(filename) {
...
 
+read_config_file <- function(config_file) {
+  if (is.null(config_file)) {
+    return(NULL)
+  }
+
+  if (endsWith(config_file, ".R")) {
+    config <- new.env()
+    read <- function(file) {
+      sys.source(file, config)
```
AshesITR, line 50: How safe is this with respect to file encodings?

R/settings.R
MichaelChirico: Yep, that's what I was thinking. Will do.

R/settings.R
MichaelChirico: Not sure... but `read.dcf()` would have the same issue, right?

MichaelChirico: it's using `readLines()`, so in principle we could add encoding to the a `conn` object passed in...

https://github.com/r-devel/r-svn/blob/d12e38f0c510e3315265e7ada26e3e33370278ba/src/library/base/R/source.R#L269

Similarly, `read.dcf()` accepts a connection, could do the same thing.

AshesITR: Maybe worth assuming the file is UTF-8?
But fine as a follow up if the current behavior is locale-dependent as well.

MichaelChirico: #2181

MichaelChirico:
Ah that's annoying. IIUC we can't even write `sys.source(keep.parse.data = FALSE)` in the source because `codetools()` will complain about "**possibly** unused argument". I can only think of an ugly workaround...

MichaelChirico:
(alternatively, we can just ignore the `R CMD check` note on R3.5.0, since CRAN is not running any checks so old anymore)

R/settings.R
```diff
@@ -50,18 +39,45 @@ read_settings <- function(filename) {
...
+    read <- function(file) {
+      dcf_values <- read.dcf(file, all = TRUE)
+      out <- lapply(names(dcf_values), function(setting) {
+        tryCatch(
+          eval(parse(text = dcf_values[[setting]], keep.source = FALSE)),
+          error = function(e) stop("Malformed config setting '", setting, "'\n  ", conditionMessage(e), call. = FALSE)
+        )
+      })
+      names(out) <- names(dcf_values)
+      out
```
AshesITR, line 66: `list2env()` for type stability?

AshesITR:
Since R code can to pretty wild things, I'd suggest to add schema validation of the settings.

MichaelChirico:
> Since R code can to pretty wild things, I'd suggest to add schema validation of the settings.

Technically also affects DCF too, right? Since we just `eval(parse(.))` the code there. Agree it's a bit more urgent now that we accept plain R code. I would push to follow-up to nail down what level of validation should be done as a separate issue (e.g. is `isTRUE(all.equal(sappy(config, typeof) == sapply(default_settings, typeof)))` sufficient?)

One thing discussed in the issue that I ended up omitting is any check that users are "doing too much" in their config. I think it's fine since we write to a "clean" environment, and only copy over the "correct" keys (instead of, say, copying over every key found in the new environment).

R/zzz.R
```diff
@@ -277,6 +277,18 @@ default_undesirable_operators <- all_undesirable_operators[names(all_undesirable
 default_settings <- NULL
 
 settings <- NULL
+# TODO(R>=3.6.0): Just use sys.source() directly. Note that we can't
```
IndrajeetPatil, line 280: Can't we already bump the minimum needed R version to `3.6.0`?

`{styler}` already [has](https://cran.r-project.org/web/packages/styler/index.html), and this is also not violating tidyverse/r-lib's [stated policy](https://www.tidyverse.org/blog/2019/04/r-version-support/).

Screenshot 2023-09-18 at 22 44 40

MichaelChirico: I still don't see any "core" tidyverse packages using that -- I'd wait for them to make the bump first:

https://github.com/tidyverse/dplyr/blob/73742713bc61d4ef079e72b806143b23cb7c4871/DESCRIPTION#L22-L23

AshesITR:
I would actually err on the side of safety and suggest consciously using `rm()` or `local()` if temporary variables are needed.

Re validation:

`linters` should be a list of linters (we already have that validation somewhere).
Exclusions should also be checked "manually".
The rest seems okay to check against types of `default_settings`.

But most importantly, no unused variable should be defined. And yes, the validation should be executed after loading the config from either source.

MichaelChirico:
> I would actually err on the side of safety

Could you elaborate the risk/unsafe practice you have in mind?

AshesITR:
Mainly worried about misspellings, like in #2134.

e.g. `exclude` vs. `exclusions` etc.

MichaelChirico:
Thanks. Makes sense; I guess that applies equally to DCF configs though. So that type of checking can be part of the overall validation. I still think we can allow unused variables in the script & manage to have a friendly failure interface along the lines of "Expected config with schema [...], got config with schema [...]. Complete config as parsed: [...]".

AshesITR:
Yeah I agree validation should run after reading the config, regardless of the branch.

MichaelChirico:
OK, I filed #2195 as follow-up to be included in this release & run config validation. Since it also applies to DCF, I think it's out of scope for this PR in particular.

R/settings.R
```diff
@@ -49,18 +38,41 @@ read_settings <- function(filename) {
   }
 }
 
+read_config_file <- function(config_file) {
+  if (is.null(config_file)) {
+    return(NULL)
+  }
+
+  config <- new.env()
+  if (endsWith(config_file, ".R")) {
```
AshesITR, line 47: This check should be upgraded later down the line if we stick with R config files.

R/settings.R
```diff
@@ -49,18 +38,41 @@ read_settings <- function(filename) {
...
+    load_config <- function(file) sys_source(file, config)
```
AshesITR, line 48: Does the code need to import `lintr` like this or can it use, e.g., `linters <- all_linters()`?

R/zzz.R
AshesITR: Note the core packages drop testing for the old versions IINM. They don't explicitly update the `Depends` field.

AshesITR (commented):
Implementation seems fine for experimental use. Still needs a lot of documentation hinting at the experimental nature of the feature.

R/settings.R
MichaelChirico: Not stress-tested, but I did copy over `modify_defaults()` & that worked, so I suppose `all_linters()` will as well. Will keep in mind when writing tests.

MichaelChirico:
> Note the core packages drop testing for the old versions IINM. They don't explicitly update the Depends field.

Hmm, I do see {dplyr} updated their `Depends` 7 months ago. Maybe there's just a delay. https://github.com/tidyverse/dplyr/commit/1b4adf75ef043c553f25b4dc6d3536ed9fc87f1f

AshesITR:
See https://github.com/tidyverse/purrr/issues/1045

IndrajeetPatil:
Oops. Sorry, @MichaelChirico, for updating the PR; I thought it had already been approved.

R/settings.R
```diff
@@ -49,18 +38,41 @@ read_settings <- function(filename) {
...
     option
   } else if (!is.null(config[[setting]])) {
-    malformed <- function(e) {
-      stop("Malformed config setting '", setting, "'\n  ", conditionMessage(e), call. = FALSE)
-    }
-    tryCatch(
-      eval(parse(text = config[[setting]])),
-      error = malformed
-    )
+    config[[setting]]
```
AshesITR, line 75: We could use `%||%` here for a little less cyclocomp.

MichaelChirico: way cleaner, nice.

R/zzz.R
MichaelChirico: Let's defer to next release, otherwise I might prefer making this release 3.2.0.

MichaelChirico:
cc issue author @hadley who might be interested in the implementation here.

R/utils.R
```diff
@@ -1,5 +1,5 @@
 `%||%` <- function(x, y) {
-  if (is.null(x) || length(x) <= 0L || is.na(x[[1L]])) {
+  if (is.null(x) || length(x) <= 0L || (is.atomic(x[[1L]]) && is.na(x[[1L]]))) {
```
AshesITR, line 2: While you're at it, `length(x) == 0L` seems more clear.

MichaelChirico:
TODO: `find_config()` only looks for the `.lintr` file. So `lint_package('tests/testthat/dummy_packags/...')` will search upwards to `lintr/.lintr`, ignoring the `_lintr.R` file in the package directory. Need a way to get the defaults to recognize "this package has an R config". Needs some more thought.

AshesITR:
Why don't we name the R config file `.lintr.R` by default?

R/settings.R
MichaelChirico: `linters_with_defaults()` is included in the tests, LMK if you think we need to test more thoroughly.

AshesITR:
Can we use `callr` in our tests to ensure a fresh test environment without lintr attached?

Diff seems dirty, but there's a merge conflict wrt main. It'll probably look fine after a manual merge.

NEWS.md
```diff
@@ -17,6 +17,7 @@
...
 * `lint()`, `lint_dir()`, and `lint_package()` fail more gracefully when the user mis-spells an argument name (#2134, @MichaelChirico).
 * Quarto files (.qmd) are included by `lint_dir()` by default (#2150, @dave-lovell).
+* There is experimental support for writing config in plain R scripts (as opposed to DCF files; #1210, @MichaelChirico). The script is run in a new environment and variables matching settings (`?default_settings`) are copied over. In particular, this removes the need to write R code in a DCF-friendly way, and allows normal R syntax highlighting in the saved file. We may eventually deprecate the DCF approach in favor of this one; user feedback is welcome on strong preferences for either approach, or for a different approach like YAML. Generally you should be able to convert your existing `.lintr` file to an equivalent R config by replacing the `:` key-value separators with assignments (`<-`).
```
AshesITR, line 20: This should include the default file name for the new config, `.lintr.R`

R/settings_utils.R
```diff
@@ -61,44 +68,41 @@ find_config <- function(filename) {
...
-
-  while (!has_config(path, config)) {
-    gh <- file.path(path, ".github", "linters")
-    if (has_config(gh, config)) {
-      return(file.path(gh, config))
+find_local_config <- function(path, config_file) {
+  repeat {
+    guesses_in_dir <- c(
+      file.path(path, config_file),
+      file.path(path, paste0(config_file, ".R")),
```
AshesITR, line 93: Swapping the order allows for smoother testing (no need to delete the `.lintr` file for testing and a rollback can be achieved by deleting `.lintr.R`)

MichaelChirico: not sure I follow you here, sorry

MichaelChirico:
> Can we use `callr` in our tests to ensure a fresh test environment without lintr attached?
> 
> Diff seems dirty, but there's a merge conflict wrt main. It'll probably look fine after a manual merge.

I was wondering if we should run sys.source() in a subprocess to avoid any side effects e.g. config runs library().

AshesITR:
Worth a try. The process could implicitly run `library(lintr)` before sourcing.

MichaelChirico:
> Worth a try. The process could implicitly run `library(lintr)` before sourcing.

right... one issue is line numbers will be off in error messages if we just paste that into the script. is messing with an .Rprofile the best workaround? but we should be running the subprocess as --vanilla?

AshesITR:
`--vanilla` can mess up local configurations with nonstandard libraries etc.

Either create a temp script that does the setup and source, or, e.g. with `parallel`, use `clusterEvalQ()` before sourcing to set up the environment.

MichaelChirico:
> --vanilla can mess up local configurations with nonstandard libraries etc.

My thinking is the script should be very plain & running into issues with `--vanilla` is a code smell. We can record where {lintr} itself is found and use a smarter `library()` call in the subprocess. Other approaches seem like overkill, to the point where I'm wanting to kill the experiment already 😅

hadley:
IMO running this file in a subprocess is overkill, and will have a substantial performance cost for relatively little gain.

MichaelChirico:
@AshesITR WDYT about sticking with simple `sys.source()` in the same session for now? Any demons that come from users running too much code with side effects in their config is "their own fault" & we are a "front end" package after all.

AshesITR:
I'm fine with that. My original point was to run or tests in a separate process to make sure they don't depend on lintr being attached.

MichaelChirico:
> I'm fine with that. My original point was to run our tests in a separate process to make sure they don't depend on lintr being attached.

OK, added a test, let me know if that's what you had in mind.

tests/testthat/test-lint_package.R
```diff
@@ -225,3 +225,17 @@ test_that("package using .lintr.R config lints correctly", {
...
+test_that("lintr need not be attached for .lintr.R configs to use lintr functions", {
+  exprs <- paste(
+    'options(lintr.linter_file = "lintr_test_config")',
+    sprintf('lints <- lintr::lint_package("%s")', test_path("dummy_packages", "RConfig")),
+    # simplify output to be read from stdout
+    'cat(paste(as.data.frame(lints)$linter, collapse = "|"), "\n", sep = "")',
+    sep = "; "
+  )
+  expect_identical(
+    system2("Rscript", c("-e", shQuote(exprs)), stdout = TRUE),
```
AshesITR: This should be `file.path(R.home("bin"), "Rscript")` with a branch for `"Rscript.exe"` on windows.
See `callr:::setup_rscript_binary_and_args` to see how callr does it.

R/settings_utils.R
AshesITR: `first_exists()` returns the first of the files that exist.
If a `.lintr` file exists, it will be returned even if `.lintr.R` exists.

I imagine noone will have a `.lintr.R` lying around.
Now if you want to test the feature, you need to rename your `.lintr` to some other file name; otherwise `find_config()` will still pick up the DCF file.

tests/testthat/test-lint_package.R
MichaelChirico: Comment on a stale commit, sorry for the confusion. Latest already does that.

R/settings_utils.R
MichaelChirico: OK, changed the precedence

AshesITR:
Implementation looks good.
Should be put in a more prominent position in NEWS imo.
Technically, it is a breaking change. Or we could add a Section "Experimental features"

MichaelChirico:
I want to take a pass at reorganizing the NEWS, let's merge for now

Issue #1832, opened 2022-12-10
New linter to suggest parentheses to clarify boolean expressions?


IndrajeetPatil:
Parentheses are cheap, and they can radically improve readability of complex conditional expressions. 
It would be a great if we have a linter that suggests doing so in apt contexts.

```r
# actual
if (a <= b && c == d || !okay) {
  ...
}

# suggested
if (((a <= b) && (c == d)) || (!okay)) { # `|| !okay) {` is also fine here
  ...
}
```

AshesITR:
Maybe more generally mixing of different binary operators without parentheses?
We should allow common cases like `%>%`, `|>`, `%+%` by default, but generally discourage relying on operator precedence.
This would also lint e.g. `1:5 + 1`

IndrajeetPatil:
> There's a lot of overlap with https://github.com/r-lib/lintr/issues/1566 -- is it the same?

I was just going to comment that. I think these two can be collapsed to one (e.g.) `implicit_precedence_linter()`?

> "parentheses are cheap" rings pretty false

Sorry for the confusion. I meant performance-wise; i.e., one shouldn't have to worry that using them profusely is going to slow down the program.

> Keeping track of matching parentheses is notoriously difficult 

Fair enough, but it is nowhere near as difficult as creating your own parentheses mentally by taking precedence into account and then matching these invisible parentheses. Also, thankfully, many IDEs (VS Code, RStudio, etc.) support rainbow parentheses, which does make this task slightly easier.

--- 

Just to clarify: I think the lint message should only mention that the user can benefit from using more parentheses for the expression in question. It shouldn't suggest how many or where these should be placed, because this will be hard to decide statically.

Issue #1566, opened 2022-09-27
Proposed linter for mixing logical operators without grouping


MichaelChirico:
Consider:

```r
x | y & z
# or
x & y | z
```

I think this is a common pitfall for users not well-versed in logical operator precedence rules.

At least "`&&` evaluates before `||`" appears consistent across languages from what I've seen; this Q&A hints that this order was chosen "to reduce the number of parentheses" (would love a citation on that), while also highlighting an interesting language (Fortress) which declines to set arbitrary operator precedences and instead enforces parentheses in any uncommon combinations:

https://softwareengineering.stackexchange.com/q/391263/221381

The above would lint unless the user adds parentheses:

```r
x | (y & z)
(x & y) | z
```

would be the logical equivalents. This would also apply to `&&` / `||`.

Related: in the Google suite, we have a linter for `!(x == 2)` --> `x != 2`. To me, the precedence of `!` is clear enough not to need extra parentheses when combined with `&&`/`||`, so I didn't include it here.

IndrajeetPatil:
Love it! Would indeed be very useful, for the naive and the advanced users alike.

> Related: in the Google suite, we have a linter for `!(x == 2)` --> `x != 2`. To me, the precedence of `!` is clear enough not to need extra parentheses when combined with `&&`/`||`, so I didn't include it here.

The one place I have seen people really struggle to make sense of precedence is `!` combined with `%in%`. Here, `!(x %in% y)` is definitely easier to reason about than `!x %in% y`.

MichaelChirico:
i personally like the `!x %in% tbl` approach (especially in if() clauses where the added parens tend to clutter) but agree it would be a good linter to offer!

MichaelChirico:
Related issue coming up in #1656:

---

Currently, the linter marks `x %>% ... %>% length(.) > 0`

This _technically_ works as intended because it's parsed as `` `>`(x %>% ... %>% length(), 0)`` but I think the code is not quite behaving as intended.

`x %>% ... %>% (length(.) > 0)` is probably closer to what was intended. So for now we mark the code as a lint, but that also doesn't feel quite right (we're using this linter as a signal of a different code quality issue).

---

So perhaps this proposed linter could be something like `operator_precedence_linter()` that encourages grouping with `()` whenever operators across different "levels" of precedence are mixed without clearly denoting intended evaluation order with `()`?

PR #2185, opened 2023-09-18
Clarify precedence using parentheses

IndrajeetPatil:
This code always makes me do a double-take.


MichaelChirico:
Related: #1832, #1566

MichaelChirico (approved):

PR #2187, opened 2023-09-18
Use linter variable in `expect_type_linter()` tests

IndrajeetPatil:


tests/testthat/test-expect_type_linter.R
```diff
@@ -1,48 +1,52 @@
...
+  expect_lint("expect_equal(typeof(x), t, expected.label = 'type')", NULL, linter)
+  expect_lint("expect_true(is.double(x), info = 'x should be double')", NULL, linter)
 })
 
 test_that("expect_type_linter blocks simple disallowed usages", {
+  linter <- expect_type_linter()
+
   expect_lint(
     "expect_equal(typeof(x), 'double')",
     rex::rex("expect_type(x, t) is better than expect_equal(typeof(x), t)"),
```
MichaelChirico, line 27: this is repeated too, right?

IndrajeetPatil: Yes, but there is also `rex::rex("expect_type(x, t) is better than expect_true(is.<t>(x))")` in the same unit, so I didn't create a new variable for lint message.

MichaelChirico: OK, we can skip for now

MichaelChirico (approved):

PR #2193, opened 2023-09-19
unnecessary_lambda_linter() finds explicit return cases

### LINKED ISSUE #1567: Unnecessary lambda linter misses multi-line cases
Follow-up to #1541 to catch more cases

MichaelChirico:
Closes #1567


tests/testthat/test-unnecessary_lambda_linter.R
```diff
@@ -6,6 +6,7 @@ test_that("unnecessary_lambda_linter skips allowed usages", {
   # the first argument may be ... or have a cumbersome name, so an anonymous
   #   function may be preferable (e.g. this is often the case for grep() calls)
```
IndrajeetPatil, line 7: I wonder if this comment should actually be part of the description for this function. As a user, I am likely to wonder about what's special about the example with `grep()` that it doesn't produce a lint.

IndrajeetPatil (approved):

IndrajeetPatil (approved):

PR #2201, opened 2023-09-20
Extract XML-related utilities into a separate file

IndrajeetPatil:
Since the `utils.R` is getting quite large.

Similar to `xp_utils.R`.


MichaelChirico:
I would let this PR simmer a while. There's some cost to understanding git history when moving code between files like this. If there will only ever be two functions in the new utils file, I'm not sure it's worth it.

It might be more fruitful to group some of the other utils together, or to try removing some of them.

AshesITR (approved):

PR #2202, opened 2023-09-20
Return early if there's nothing to lint

MichaelChirico:
Accidentally linted a directory where everything was excluded, leading to an error from `txtProgressBar()`:

```
Error in txtProgressBar(max = length(files), style = 3L) : 
  must have 'max' > 'min'
```

Better to just skip running anything if there's nothing to lint.


IndrajeetPatil:
Just need to update NEWS here. Othewise, LGTM.

MichaelChirico:
Nothing to update. This is really a regression after #2133. So it's a dev-only improvement.

AshesITR (approved):

PR #2205, opened 2023-09-21
Simplify find config

MichaelChirico:
Feeds into #2177

I found the structure of the `while()` loop confusing, it took me quite some time to figure out it's checking _both_ `${PATH}/.lintr` _and_ `${PATH}/.github/linters/.lintr` on each pass.

I hope this rewrite makes what's going on much clearer. It will also make it trivial to extend for #2177 to also check for `.lintr.R`.

### DIFF
--- a/R/settings_utils.R
+++ b/R/settings_utils.R
@@ -78,16 +78,6 @@ find_config <- function(filename) {
   NULL
 }
 
-<<<<<<< Updated upstream
-find_config2 <- function(path) {
-  config <- basename(getOption("lintr.linter_file"))
-  path <- normalizePath(path, mustWork = FALSE)
-
-  while (!has_config(path, config)) {
-    gh <- file.path(path, ".github", "linters")
-    if (has_config(gh, config)) {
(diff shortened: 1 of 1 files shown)


R/settings_utils.R
```diff
@@ -34,14 +34,21 @@ is_root <- function(path) {
...
+is_directory <- function(filename) isTRUE(file.info(filename)$isdir)
+
+#' Return the first of a vector of files that exists.
+#'
+#' Avoid running 'expensive' [file.exists()] for the full vector,
+#'   since typically the first entries will lead to early exit.
+#' TODO(#2204): check if the implementation should be simpler
+#' @noRd
+first_exists <- function(files) {
+  for (file in files) {
```
AshesITR, line 46: This can be avoided because `file.exists()` is vectorized.
Not sure if faster, we should check:

```r
file_exists <- file.exists(files)
if (any(file_exists)) {
  files[which.max(file_exists)]
} else {
  NULL
}
```

AshesITR (approved):

PR #2207, opened 2023-09-21
Minor refactoring with XML node helpers

IndrajeetPatil:
Well, not truly a refactoring because I am changing one error message.

### DIFF
--- a/R/xml_nodes_to_lints.R
+++ b/R/xml_nodes_to_lints.R
@@ -52,7 +52,7 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
     )
     class(lints) <- "lints"
     return(lints)
-  } else if (!inherits(xml, "xml_node")) {
+  } else if (!is_node(xml)) {
     stop(
       "Expected an xml_nodeset, a list of xml_nodes or an xml_node, got an object of class(es): ",
       toString(class(xml))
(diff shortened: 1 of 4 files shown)


R/xml_nodes_to_lints.R
```diff
@@ -96,7 +96,9 @@ xml_nodes_to_lints <- function(xml, source_expression, lint_message,
   )
 }
 
+is_node <- function(xml) inherits(xml, "xml_node")
+is_nodeset <- function(xml) inherits(xml, "xml_nodeset")
 is_nodeset_like <- function(xml) {
-  inherits(xml, "xml_nodeset") ||
-    (is.list(xml) && all(vapply(xml, inherits, logical(1L), what = "xml_node")))
+  is_nodeset(xml) ||
+    (is.list(xml) && all(vapply(xml, is_node, logical(1L))))
```
IndrajeetPatil, line 103: These helpers can probably move to the XML utils file (#2201).

R/xp_utils.R
```diff
@@ -78,10 +78,17 @@ xp_or <- function(...) paren_wrap(..., sep = "or")
...
 xp_call_name <- function(expr, depth = 1L, condition = NULL) {
   stopifnot(
-    inherits(expr, c("xml_node", "xml_nodeset")),
     is.numeric(depth), depth >= 0L,
     is.null(condition) || is.character(condition)
   )
+  is_valid_expr <- is_node(expr) || is_nodeset(expr)
+  if (!is_valid_expr) {
+    stop(
+      "Expected an xml_nodeset, a list of xml_nodes or an xml_node, got an object of class(es): ",
```
AshesITR, line 87: The error message is inconsistent with the test (re list of nodes)

AshesITR (commented):
Not sure it's worth it to factor out `inherits(x, "xml_node")` into its own helper given it has so few call sites.
The error message improvement seems good, though.

Maybe revert to `inherits()` and keep the improved error message?

IndrajeetPatil:
> Not sure it's worth it to factor out `inherits(x, "xml_node")` into its own helper given it has so few call sites. The error message improvement seems good, though.

But it has five different call sites, and everywhere the same `inherits()` call is repeated. I think this deserves a separate function.

AshesITR (approved):

PR #2209, opened 2023-09-21
Use `Map()` instead of `mapply()`

IndrajeetPatil:
Since `mapply()` is part of `default_undesirable_functions`.

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -83,7 +83,6 @@ get_chunk_positions <- function(pattern, lines) {
   # use 'ws_re' to make clear that we're matching knitr's definition of initial whitespace.
   ws_re <- sub("```.*", "", pattern$chunk.begin)
   extract_indent <- function(start, end) {
-    lines_subset <- lines[start:end]
     matches <- gregexpr(ws_re, lines[start:end], perl = TRUE)
     lengths <- vapply(matches, attr, integer(1L), "match.length")
     min(lengths)
--- a/R/extract.R
+++ b/R/extract.R
@@ -87,7 +87,8 @@ get_chunk_positions <- function(pattern, lines) {
     lengths <- vapply(matches, attr, integer(1L), "match.length")
     min(lengths)
   }
-  indents <- Map(extract_indent, starts, ends)
+  # min() guarantees length(indents) == length(starts)
+  indents <- unlist(Map(extract_indent, starts, ends))
   list(starts = starts, ends = ends, indents = indents)
 }
 
--- a/R/extract.R
+++ b/R/extract.R
@@ -82,10 +82,13 @@ get_chunk_positions <- function(pattern, lines) {
   #   set the initial column to the leftmost one within each chunk (including the start+end gates). See tests.
   # use 'ws_re' to make clear that we're matching knitr's definition of initial whitespace.
   ws_re <- sub("```.*", "", pattern$chunk.begin)
-  indents <- mapply(
-    function(start, end) min(vapply(gregexpr(ws_re, lines[start:end], perl = TRUE), attr, integer(1L), "match.length")),
-    starts, ends
-  )
+  extract_indent <- function(start, end) {
+    lines_subset <- lines[start:end]
+    matches <- gregexpr(ws_re, lines_subset, perl = TRUE)
+    lengths <- vapply(matches, attr, integer(1L), "match.length")
+    min(lengths)
+  }
+  indents <- Map(extract_indent, starts, ends)
   list(starts = starts, ends = ends, indents = indents)
 }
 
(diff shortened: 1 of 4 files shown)


IndrajeetPatil:
Why doesn't the lint GHA workflow fail on the `main` if there is an undesirable function lint present in the codebase?!

MichaelChirico:
because undesirable_function_linter is not a default linter. with good reason IMO, these uses of mapply() are perfectly fine when I last checked.

MichaelChirico:
yes, I wouldn't make either change. mapply() with specific SIMPLIFY=FALSE doesn't have the return type instability risk. and the other case is much less readable, with Map, and that code has been really stress tested with the type of R code we throw at it.

IndrajeetPatil:
I personally find the `Map()` versions to be much more readable than `mapply()`. Plus, I also like having one consistent way throughout the codebase when it comes to solving similar problems (which is `Map()` in other contexts, but `mapply()` in these two places).

But you raise a good point about not touching code that has been stress tested in the wild. I am not sure under what circumstances these two functions would produce different results (I'd have thought our comprehensive test suite would pick up on it), but I guess it's better to be cautious anyway.

MichaelChirico:
> I am not sure under what circumstances these two functions would produce different results

It's like `sapply()` -- `mapply()` attempts `simplify2array()`. So your version is returning a `list()` where `mapply()` is returning a vector. I'm surprised it passed tests, but even if it does, the result will be less efficient since it can't take advantage of fast subsetting.

`mapply()` is fine here because `min()` is type-stable (always producing length-1 real, even for empty input). And IIRC this helper is only invoked when chunks are detected, so empty input is not really possible anyway.

R/extract.R
```diff
@@ -82,10 +82,13 @@ get_chunk_positions <- function(pattern, lines) {
   #   set the initial column to the leftmost one within each chunk (including the start+end gates). See tests.
   # use 'ws_re' to make clear that we're matching knitr's definition of initial whitespace.
   ws_re <- sub("```.*", "", pattern$chunk.begin)
-  indents <- mapply(
-    function(start, end) min(vapply(gregexpr(ws_re, lines[start:end], perl = TRUE), attr, integer(1L), "match.length")),
-    starts, ends
-  )
+  extract_indent <- function(start, end) {
+    lines_subset <- lines[start:end]
+    matches <- gregexpr(ws_re, lines_subset, perl = TRUE)
```
MichaelChirico, line 87:
```suggestion
    matches <- gregexpr(ws_re, lines[start:end], perl = TRUE)
```

R/extract.R
```diff
@@ -82,10 +82,13 @@ get_chunk_positions <- function(pattern, lines) {
   #   set the initial column to the leftmost one within each chunk (including the start+end gates). See tests.
   # use 'ws_re' to make clear that we're matching knitr's definition of initial whitespace.
   ws_re <- sub("```.*", "", pattern$chunk.begin)
-  indents <- mapply(
-    function(start, end) min(vapply(gregexpr(ws_re, lines[start:end], perl = TRUE), attr, integer(1L), "match.length")),
-    starts, ends
-  )
+  extract_indent <- function(start, end) {
+    lines_subset <- lines[start:end]
```
MichaelChirico, line 86:
```suggestion
```

MichaelChirico:
> I personally find the Map() versions to be much more readable than mapply().

I was reading on mobile, actually it looks OK on desktop. Happy to merge if you'd like & agree w the suggestions (if so please add `mapply()` to our `.lintr` as undesirable).

IndrajeetPatil:
Thanks, @MichaelChirico, for the suggestions. I have accepted the suggested changes and also added `mapply()` to `.lintr`.

R/extract.R
```diff
@@ -82,10 +82,13 @@ get_chunk_positions <- function(pattern, lines) {
   #   set the initial column to the leftmost one within each chunk (including the start+end gates). See tests.
   # use 'ws_re' to make clear that we're matching knitr's definition of initial whitespace.
   ws_re <- sub("```.*", "", pattern$chunk.begin)
-  indents <- mapply(
-    function(start, end) min(vapply(gregexpr(ws_re, lines[start:end], perl = TRUE), attr, integer(1L), "match.length")),
-    starts, ends
-  )
+  extract_indent <- function(start, end) {
+    matches <- gregexpr(ws_re, lines[start:end], perl = TRUE)
```
MichaelChirico, line 86: This is probably a follow-up, but shouldn't we be using `regexpr()` here, not `gregexpr()`?

MichaelChirico (approved):

R/extract.R
IndrajeetPatil: Tracking in https://github.com/r-lib/lintr/issues/2211

MichaelChirico (approved):

PR #2213, opened 2023-09-21
Ship SARIF template as a JSON file in external data folder

IndrajeetPatil:
Storing the template in a `.json` file makes it easier to read, esp. because of proper indentation, and also to format them in case of modification.

Inlining the JSON as a string in the source code makes both of these difficult.

WDYT?


MichaelChirico:
cc @shaopeng-gh who authored SARIF support.

looks ok to me. is there any prior art on what templated json should look like?

inst/extdata/sarif-template.json
```diff
@@ -0,0 +1,58 @@
...
+      ],
+      "columnKind": "utf16CodeUnits",
+      "originalUriBaseIds": {
+        "ROOTPATH": {
+          "uri": "file:///C:/repos/repototest/"
+        }
+      }
+    }
+  ]
+}
```
AshesITR, line 58: Missing terminal newline

IndrajeetPatil: Thanks. Fixed. 

VS code formatter wouldn't let me add it. So had to do it from RStudio.

AshesITR (approved):
This should be safe regardless of feedback since all we do is move the constant to another file.

PR #2219, opened 2023-09-25
Update tests for methods

IndrajeetPatil:
- ~~Use snapshot testing for `print.lint`~~ (not going to work outside RStudio)
- Harmonize with our preference for inlining lint calls in tests


tests/testthat/test-methods.R
```diff
@@ -107,36 +101,35 @@ test_that("print.lint works for inline data, even in RStudio", {
 
   l <- lint("x = 1\n")
 
-  # Make sure lints print to console.
-  # The full output is:
-  #
-  # > <text>:1:3: style: Use <-, not =, for assignment.
-  # > x = 1
-  # >   ^
```
IndrajeetPatil, line 115: This is exactly what snapshot testing is for. We are rarely going to check this output manually.

tests/testthat/test-methods.R
```diff
@@ -152,16 +145,14 @@ test_that("within.list is dispatched", {
 
 test_that("as_tibble.list is _not_ dispatched directly", {
   skip_if_not_installed("tibble")
-  expect_identical(
-    nrow(tibble::as_tibble(lint(text = "a = 1", linters = assignment_linter()))),
```
IndrajeetPatil, line 156: Dagwood sandwich making it difficult to parse this test. The same for the test below.

tests/testthat/test-methods.R
```diff
@@ -107,36 +101,35 @@ test_that("print.lint works for inline data, even in RStudio", {
...
+      print(l)
+    })
   )
 })
 
 test_that("print.lints works", {
   withr::local_options(lintr.rstudio_source_markers = FALSE)
-  tmp <- withr::local_tempfile()
 
-  expect_true(file.create(tmp))
```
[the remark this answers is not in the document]
MichaelChirico: Yes, but if file creation fails, we'd like to stop early for clarity. Maybe `stopifnot(file.create(tmp))` instead?

tests/testthat/test-methods.R
```diff
@@ -152,16 +146,14 @@ test_that("within.list is dispatched", {
 
 test_that("as_tibble.list is _not_ dispatched directly", {
   skip_if_not_installed("tibble")
-  expect_identical(
-    nrow(tibble::as_tibble(lint(text = "a = 1", linters = assignment_linter()))),
-    1L
-  )
+
+  l <- lint(text = "a = 1", linters = assignment_linter())
```
MichaelChirico, line 150: Better to unpack this, indeed. But let's avoid the `l` shorthand in committed code. `lint_output` or just `lints`?

tests/testthat/test-methods.R
```diff
@@ -70,21 +69,16 @@ test_that("as.data.frame.lints", {
 })
 
 test_that("summary.lints() works (no lints)", {
-  no_lints <- lint(
-    "x <- 1\n",
-    linters = assignment_linter()
-  )
+  no_lints <- lint("x <- 1\n", linters = assignment_linter())
```
MichaelChirico, line 72: This `lint(text = "x = 1", linters = assignment_linter())` is repeated a lot. Should we:

 - Have a helper in helpers.R: `run_assignment_linter()` (perhaps with argument `should_lint = TRUE/FALSE`?
 - Just cache the output within this test file: `example_with_lint <- lint(...); example_without_lint <- lint(...)`
 - Neither?

tests/testthat/test-methods.R
```diff
@@ -128,15 +122,15 @@ test_that("print.lint works for inline data, even in RStudio", {
 
 test_that("print.lints works", {
   withr::local_options(lintr.rstudio_source_markers = FALSE)
-  tmp <- withr::local_tempfile()
 
-  expect_true(file.create(tmp))
-  expect_invisible(print(lint(tmp)))
+  withr::with_tempfile("tmp", code = {
```
AshesITR, line 126: The `local_tempfile()` approach seems preferrable to me because it explicitly sets `tmp`.

tests/testthat/test-methods.R
IndrajeetPatil: I would prefer to do neither. I learned from you the concept of keeping tests [DAMP](https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/), and I think that's relevant here. 

IMO, as they stand, the tests are more readable than they would be with this additional abstraction introduced.

tests/testthat/test-methods.R
IndrajeetPatil: I have reverted this change.

IndrajeetPatil:
Btw, we have a few tests that check for existence of a file or its absence, and I am wondering if we should introduce custom expectations for these? Or you think it's not worth it?

```r
expect_file_exists <- function(...) {
  expect_true(file.exists(...))
}

expect_file_doesnt_exist <- function(...) {
  expect_false(file.exists(...))
}
```

So, e.g.

```r
expect_true(file.exists("lintr_results.sarif")) # before
expect_file_exists("lintr_results.sarif") # after
```

AshesITR:
A custom expectation only seems worth it if the failure message is more helpful. Otherwise, it's just an additional layer of indirection hampering readability.

IndrajeetPatil:
> A custom expectation only seems worth it if the failure message is more helpful. Otherwise, it's just an additional layer of indirection hampering readability.

Good point! I think in that case this can be something that can be implemented upstream.

At any rate, the current PR is ready for another round of review. I don't think there is any pending comment anymore.

MichaelChirico:
> A custom expectation only seems worth it if the failure message is more helpful. Otherwise, it's just an additional layer of indirection hampering readability.

agree... I'd rather not have to memorize another helper for this (I have enough trouble remembering skip_if_not_r_version)

PR #2225, opened 2023-10-03
Validate config files

### LINKED ISSUE #2195: Validate configs
I would actually err on the side of safety and suggest consciously using `rm()` or `local()` if temporary variables are needed.

Re validation:

`linters` should be a list of linters (we already have that validation somewhere).
Exclusions should also be checked "manually".
The rest seems okay to check against types of `default_settings`.

But most importantly, no unused variable should be defined. And yes, the validation should be executed after loading the config from either source.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/issues/2177#issuecomment-1724488803_

There's also more discussion in the PR about validation.

MichaelChirico:
Closes #2195.


MichaelChirico:
@AshesITR PTAL at the updated description of valid `exclusions`. AFAICT this tested usage was not described till now:

https://github.com/r-lib/lintr/blob/9c49c6ec5efb4632616f0973747d2cc578ed5655/tests/testthat/test-exclusions.R#L88

MichaelChirico:
This is another undocumented usage of `exclusions`:

https://github.com/r-lib/lintr/blob/9c49c6ec5efb4632616f0973747d2cc578ed5655/tests/testthat/dummy_packages/clean/lintr_test_config#L1

Should we break that, or extend the description of a valid `exclusions` to cover this?

MichaelChirico:
That form's used here too:

https://github.com/r-lib/lintr/blob/9c49c6ec5efb4632616f0973747d2cc578ed5655/tests/testthat/test-lint_dir.R#L88

AshesITR:
> This is another undocumented usage of `exclusions`:
> 
> 
> 
> https://github.com/r-lib/lintr/blob/9c49c6ec5efb4632616f0973747d2cc578ed5655/tests/testthat/dummy_packages/clean/lintr_test_config#L1
> 
> 
> 
> Should we break that, or extend the description of a valid `exclusions` to cover this?

I wouldn't want to break character vectors of file names to exclude.

R/settings.R
```diff
@@ -77,6 +78,72 @@ read_config_file <- function(config_file) {
   config
 }
 
+validate_config_file <- function(config, config_file, defaults) {
+  matched <- names(config) %in% names(defaults)
+  if (!all(matched)) {
+    warning("Found non-setting objects in config '", config_file, "': ", toString(names(config)[!matched]))
```
AshesITR, line 84: "Found unused settings ..."?

R/settings.R
```diff
@@ -77,6 +78,72 @@ read_config_file <- function(config_file) {
...
+  }
+
+  check_types <- intersect(names(config), names(defaults))
```
AshesITR, line 87: Wouldn't this test throw on character exclusions?

It seems more expressive to explicitly list the keys and what they should be though. That would also allow for regex validation for the nolint configs.

AshesITR (changes requested):

R/settings.R
```diff
@@ -77,6 +78,72 @@ read_config_file <- function(config_file) {
...
+    if (observed_type != expected_type) {
+      stop("Setting '", setting, "' should be of type '", expected_type, "', not '", observed_type, "'.")
+    }
+  }
+
+  validate_linters(config)
+  validate_exclusions(config)
+}
+
+validate_linters <- function(config) {
```
AshesITR, line 100: Don't we have a more advanced implementation of this already? e.g. auto-converting factories to linters by calling them.

R/settings.R
MichaelChirico: It increases the complexity a lot for not much benefit IMO.

R/settings.R
MichaelChirico: That's a deprecated feature though, right? So ideally we'd remove the code to do that.

MichaelChirico: I searched for `"linter"` in our sources to find other related functionality, it turned up `is_linter()` as the `inherits(., "linter")` wrapper (which is useful here), but that's all I saw (possible I scanned past something useful).

R/settings.R
AshesITR: I'd argue it makes the actual validations much more clear. Currently, you have to go look up the definition of `default_settings` to figure out what types are expected.

NB regex validation could be very simple to implement. Not sure how nice the error messages are if you provide an invalid regex to `exclude` and friends

AshesITR: Okay, fair enough.
In that case, I would disable the check for exclusions because it can vary in types. I see no need to introduce a breaking change.

MichaelChirico: > I would disable the check for exclusions because it can vary in types

Yes, the current state is flexible enough -- tests pass without needing to adapt any old tests.

Currently, I only added tests of errors -- do you think it's worth adding tests of valid configs in this PR as well?

AshesITR: Are they? What if `exclusions` is given as a character?

MichaelChirico: Added some explicit tests for clarity, PTAL

R/settings.R
AshesITR: `validate_linter_object()`, or `define_linters()` seem useful.

In particular `linters: object_name_linter()` appears to be supported by the CLI usage (it's very handy indeed). Do we (currently) support providing a linter via settings instead of a list of linters?

AshesITR: `validate_linter_object()`, or `define_linters()` seem useful.

In particular `linters: object_name_linter()` appears to be supported by the CLI usage (it's very handy indeed). Do we (currently) support providing a linter via settings instead of a list of linters?

AshesITR: `validate_linter_object()`, or `define_linters()` seem useful.

In particular `linters: object_name_linter()` appears to be supported by the CLI usage (it's very handy indeed). Do we (currently) support providing a linter via settings instead of a list of linters?

AshesITR: `validate_linter_object()`, or `define_linters()` seem useful.

In particular `linters: object_name_linter()` appears to be supported by the CLI usage (it's very handy indeed). Do we (currently) support providing a linter via settings instead of a list of linters?

MichaelChirico: > In particular linters: object_name_linter() appears to be supported by the CLI usage

Is it? I tested at v3.1.0, current `main`, and this PR, all give an error for that:

```r
withr::local_dir(withr::local_tempdir())
writeLines("linters: infix_spaces_linter()", ".lintr")
writeLines("a=1", "aaa.R")
lint_dir()

# v3.1.0
# .Error in x[missing] : object of type 'closure' is not subsettable

# main @ 62858ba5675dbd4c86942bf1175dd605b97c4781
# Error in x[[1L]] : object of type 'closure' is not subsettable

# this branch @ 62858ba5675dbd4c86942bf1175dd605b97c4781
# Setting 'linters' should be of type 'list', not 'closure'.
```

AshesITR: With CLI usage I meant the `linters=` argument for `lint()` and friends.
Since the answer is no, we don't support this shorthand for the config, I agree with your implementation 😊

MichaelChirico: oh, got it. feel free to file as a follow-up FR, it seems reasonable enough & not too hard to support.

AshesITR: #2230

R/settings.R
```diff
@@ -77,6 +78,73 @@ read_config_file <- function(config_file) {
...
 
+validate_config_file <- function(config, config_file, defaults) {
+  matched <- names(config) %in% names(defaults)
+  if (!all(matched)) {
+    warning("Found unused settings in config '", config_file, "': ", toString(names(config)[!matched]))
+  }
+
+  # exclude can flexibly accept list() or character()
+  check_types <- setdiff(intersect(names(config), names(defaults)), "exclusions")
+  for (setting in check_types) {
```
AshesITR, line 89: Could also check the lengths for string-type settings.

Or does this work?
`exclude: c("#nolint", "#mynolint")`

MichaelChirico: No, it's passed to `rex::re_matches()`:

https://github.com/r-lib/lintr/blob/b49ee9dac3eddb9285648c9e2b5f15e85659ee3a/R/exclude.R#L140

Which AFAICT only accepts scalar. I suppose we could accept that (just collapse with `|`) but that seems like asking for trouble, user can do that themselves if needed.

AshesITR: So we should validate for length 1.

MichaelChirico: Refactored the approach, I think it balances complexity with extensibility pretty well. Basically we class the types of options there are and run the validator on each of those. Now there are 5 types of options: (1) character string, (2) regex (=character string+), (3) TRUE/FALSE, (4) linters (5) exclusions.

NEWS.md
```diff
@@ -5,6 +5,7 @@
...
+  + We also validate config files up-front make it clearer when invalid configs are present (#2195, @MichaelChirico). There is a warning for "invalid" settings, i.e., settings not part of `?default_settings`. We think this is more likely to affect users declaring settings in R, since any variable defined in the config that's not a setting must be removed.
```
AshesITR, line 8: ... *to* make it clearer

R/settings.R
```diff
@@ -77,6 +78,97 @@ read_config_file <- function(config_file) {
...
+  validate_character_string(config, c("encoding", "cache_directory", "comment_token"))
+  validate_true_false(config, c("comment_bot", "error_on_lint"))
+  validate_linters(config$linters)
+  validate_exclusions(config$exclusions)
+}
+
+is_character_string <- function(x) is.character(x) && length(x) == 1L && !is.na(x)
+is_valid_regex <- function(str) !inherits(tryCatch(grepl(str, ""), condition = identity), "condition")
+is_single_regex <- function(x) is_character_string(x) && is_valid_regex(x)
+is_true_false <- function(x) is.logical(x) && length(x) == 1L && !is.na(x)
```
AshesITR, line 99: Can we already use `isTRUE(x) || isFALSE(x)`?

tests/testthat/test-lint_package.R
```diff
@@ -217,7 +217,11 @@ test_that("package using .lintr.R config lints correctly", {
...
   # config produces unused variables
   withr::local_options(lintr.linter_file = "lintr_test_config_extraneous")
-  expect_length(lint_package(r_config_pkg), 2L)
+  expect_warning(
+    expect_length(lint_package(r_config_pkg), 2L),
+    "Found unused settings in config",
+    fixed = TRUE
+  )
 
   # DCF is preferred if multiple matched configs
```
AshesITR, line 226: Comment needs updating?

tests/testthat/test-settings.R
```diff
@@ -152,3 +152,94 @@ test_that("it has a smart default for encodings", {
...
+  expect_warning(lint_dir(), "Found unused settings in config", fixed = TRUE)
+
+  writeLines("encoding: FALSE", .lintr)
+  expect_error(lint_dir(), "Setting 'encoding' should be a character string, not 'FALSE'", fixed = TRUE)
+
+  writeLines("encoding: NA_character_", .lintr)
+  expect_error(lint_dir(), "Setting 'encoding' should be a character string, not 'NA'", fixed = TRUE)
+
+  writeLines("encoding: letters", .lintr)
+  expect_error(lint_dir(), "Setting 'encoding' should be a character string, not '[a-z, ]+'")
```
AshesITR, line 175: What's up with that expectation? `rex()` clash?

R/exclude.R
```diff
@@ -10,18 +10,24 @@
...
-#'   "2. line range in the source file. default: `# nolint start`, `# nolint end`. `# nolint start` accepts linter",
+#'   "2. Line range in the source file. default: `# nolint start`, `# nolint end`. `# nolint start` accepts linter",
 #'   "   lists in the same form as `# nolint`.",
-#'   "3. exclusions parameter, a named list of files with named lists of linters and lines to exclude them on, a named",
-#'   "   list of the files and lines to exclude, or just the filenames if you want to exclude the entire file, or the",
-#'   "   directory names if you want to exclude all files in a directory."
+#'   "3. Exclusions parameter, a list with named and/or unnamed entries. If present, the name is a path relative to",
+#'   "   the config. Moreover, elements have the following characteristics:",
+#'   "   1. Unnamed elements specify filenames or directories.",
+#'   "   2. Named elements are numeric or a list. If present, the name is a linter.",
```
AshesITR, line 25: Isn't the name a filename or a directory?

R/exclude.R
```diff
@@ -10,18 +10,24 @@
...
+#'   "      1. Numeric or unnamed list elements are line numbers to be excluded.",
```
AshesITR, line 26: `Inf` should be documented as a sentinel to mean all lines in the file / folder.

R/settings.R
MichaelChirico: Even so we could backport, but note that it's redundant:

```
isTRUE <- function(x) is.logical(x) && length(x) == 1L && !is.na(x) && x
isFALSE <- function(x) is.logical(x) && length(x) == 1L && !is.na(x) && !x
```

This version avoids double-testing 3/4 conditions

tests/testthat/test-settings.R
MichaelChirico: Hmm I think `letters` is obscuring this. Better to use a simple literal instead for clarity.

R/exclude.R
MichaelChirico: Agree it's confusing. But "the name is a path relative to the config" above refers to those names, this one refers to the more nested names. Taking another pass.

R/exclude.R
MichaelChirico: Wasn't sure whether to document it, but I guess this is the easiest way to exclude certain linters from whole files.

MichaelChirico: Documented and added tests

R/exclude.R
MichaelChirico: PTAL

R/settings.R
AshesITR: Huh? Isn't it `identical(x, TRUE)` and `identical(x FALSE)` respectively?

MichaelChirico: (note the comment above about why they abandoned that implementation)

AshesITR (approved):
Looks good for now. ?exclude still needs some work (examples, maybe?), but I think that can be a follow-up.

Thanks for all the work and patience!

MichaelChirico:
> Thanks for all the work and patience!

I'm really happy w the progress! thanks for the thorough review & insistence!

PR #2226, opened 2023-10-03
Create condition_call_linter

### LINKED ISSUE #2222: Enforce use of `call. = FALSE` in `stop()` and `warning()`
This is part of the official tidyverse guidelines: https://design.tidyverse.org/err-call.html.

Some people might prefer to display the call but we should at least ensure that use of `call.` is consistent throughout the package. So an argument like `display_call` defaulting to `FALSE` might be helpful.

Before I start working on this, I wanted to double check there is not already an issue or work in progress for this that I missed.

Bisaloo:
Fix #2222 

I'm opening this PR so we have something tangible to discuss but I'm happy to take any feedback and do any necessary changes.


Bisaloo:
As far as I can tell, tests fail because there are snapshot tests to control the addition of new defaults. It's indeed probably safer to not include this linter as default at least for now. Please let me know what you think.

AshesITR:
Since it's part of the style guide I am in favor of adding it to the defaults.
The default linter test code would need to be expanded to create a lint for this linter.

Regarding the design, I'm not sure we should lint any `call.=` calls under the assumption that the author has put thought into the value of `call.`.
Therefor I'd only lint `stop()` without a `call.` argument by default. The lint message could still suggest `call. = FALSE`.
LMK what you think about that.

MichaelChirico:
> Since it's part of the style guide

Is it? The 'design' book is not the style guide... the latter AFAICT is mum on the topic. Would help to know if we intend to enforce both the style guide and the design guide by default (what does {styler} do here @lorenzwalthert?)

R/condition_call_linter.R
```diff
@@ -0,0 +1,98 @@
...
+#'   linters = condition_call_linter()
+#' )
+#'
+#' lint(
+#'   text = "stop('test', call. = TRUE)",
+#'   linters = condition_call_linter(display_call = TRUE)
+#' )
+#'
+#' @evalRd rd_tags("condition_call_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
```
AshesITR, line 42: Add a reference link to design.tidyverse.org?

AshesITR:
> > Since it's part of the style guide
> 
> 
> 
> Is it? The 'design' book is not the style guide... the latter AFAICT is mum on the topic. Would help to know if we intend to enforce both the style guide and the design guide by default (what does {styler} do here @lorenzwalthert?)

Good catch.

lorenzwalthert:
{styler} only implements the style guide as per the docs.

Bisaloo:
> Regarding the design, I'm not sure we should lint any call.= calls under the assumption that the author has put thought into the value of call..
Therefor I'd only lint stop() without a call. argument by default. The lint message could still suggest call. = FALSE.
LMK what you think about that.

That's a good question :thinking: 

I could still imagine that it's useful to ensure consistency across different contributors (or different projects using the same lintr settings). The code complexity added by linting `call.` seems low enough that it's worth offering this flexibility.

I don't feel super strongly on this though as I personally prefer `call. = FALSE` but I have heard diverging opinions on the matter. Happy to follow your recommendation.

AshesITR:
Yes to offering flexibility.

What do you think about these options?

```r
call. = NA # force providing call.=, ignore its value
call. = TRUE # lint call. = FALSE, ignore missing call.=
call. = FALSE # force call. = FALSE
```

Default could be `FALSE` as by the design guide.

Bisaloo:
Any thoughts about the argument name? I have used `display_call` since I wasn't really inspired. I see you used `call.`, should I update it to this?

AshesITR:
I'm not sure, `use_call` comes to mind but that implies to me that `TRUE` does what `NA` is supposed to do and `FALSE` requires `call.` to be missing so that's not ideal, really.

@MichaelChirico, @IndrajeetPatil any thoughts?

AshesITR:
Other linters tend to have some `*style` argument. Maybe that's also a valid idea.

Bisaloo:
BTW, I went back to the style guide history and I confirm this used to be part of the official style guide but is now obsolete since rlang is now recommended instead of base conditions:

https://github.com/tidyverse/style/commit/78bc2e2b12db78d627d5cc6d40c659b19fff5858

I'm not completely sure what makes sense in this case.

AshesITR:
Let's leave it as non-default for the moment.

Bisaloo:
Hi all, any action required from my end to move forward with this PR?

AshesITR:
Hi, you could address my comment (adding a link to the location in tidy design principles which states what this linter checks).
There are also a few lints that need fixing and an entry in `NEWS.md` giving you credit should be added.

R/condition_call_linter.R
Bisaloo: I missed the location of this comment. The link is now provided here: https://github.com/Bisaloo/lintr/blob/16a9b3f021bf960fca19e91568a0cecc75f2764e/R/condition_call_linter.R#L3-L5. Please let me know if you don't think it's appropriate.

Bisaloo:
How should I go about applying this linter to lintr itself? I think you usually do a separate PR. Is that right?

IndrajeetPatil:
> How should I go about applying this linter to lintr itself? I think you usually do a separate PR. Is that right?

No, we actually fix the lints in lintr in the PR itself. This exercise acts as a good proof of concept 😺

R/condition_call_linter.R
```diff
@@ -0,0 +1,116 @@
...
+    frag <- xp_or(call_xpath, no_call_xpath)
+  }
+
+  xpath <- glue::glue("
+    //SYMBOL_FUNCTION_CALL[text() = 'stop' or text() = 'warning']
+    /parent::expr[{frag}]
+    /parent::expr
+  ")
+
+  Linter(function(source_expression) {
```
IndrajeetPatil, line 85: You will also need to update this PR to use this new attribute introduced in https://github.com/r-lib/lintr/pull/2352

Bisaloo:
> > > How should I go about applying this linter to lintr itself? I think you usually do a separate PR. Is that right?
> > 
> > 
> > No, we actually fix the lints in lintr in the PR itself. This exercise acts as a good proof of concept 😺
> 
> I would say 'it depends', if there are only a few hits on {lintr}, go ahead and include fixes, but if there are dozens of things to change it will be better in a separate/focused PR.
> 
> Anyway, maybe you are asking about whether the new linter will be part of our .lintr config. As of #2328 all linters are applied to {lintr} right away by default.

I've made the change in https://github.com/r-lib/lintr/pull/2226/commits/0b2e34d5feed4d036f8f5c8e1b0b396c25a6620f and https://github.com/r-lib/lintr/pull/2226/commits/427b258e0e61a1937dcc079a19ea312905bcc0af but I can easily cherry-pick them in a different branch and remove them from here if you prefer.

MichaelChirico:
> > > > How should I go about applying this linter to lintr itself? I think you usually do a separate PR. Is that right?
> > > 
> > > 
> > > No, we actually fix the lints in lintr in the PR itself. This exercise acts as a good proof of concept 😺
> > 
> > 
> > I would say 'it depends', if there are only a few hits on {lintr}, go ahead and include fixes, but if there are dozens of things to change it will be better in a separate/focused PR.
> > Anyway, maybe you are asking about whether the new linter will be part of our .lintr config. As of #2328 all linters are applied to {lintr} right away by default.
> 
> I've made the change in [0b2e34d](https://github.com/r-lib/lintr/commit/0b2e34d5feed4d036f8f5c8e1b0b396c25a6620f) and [427b258](https://github.com/r-lib/lintr/commit/427b258e0e61a1937dcc079a19ea312905bcc0af) but I can easily cherry-pick them in a different branch and remove them from here if you prefer.

Yes, please do. The diff is quite large now.

IndrajeetPatil:
Sorry, @Bisaloo, for the incorrect advice. Didn't realize there would be so many new lints found in lintr for the current linter.

Bisaloo:
> Yes, please do. The diff is quite large now.

There is now a separate PR for this: https://github.com/r-lib/lintr/pull/2377. I'll rebase this PR on `main` once it's merged.

> Sorry, @Bisaloo, for the incorrect advice. Didn't realize there would be so many new lints found in lintr for the current linter.

No worries, I understand there can be some variability depending on the specifics of each case and it wasn't much work to move these commits to a separate branch.

Bisaloo:
This PR has now been rebased on `main`.

R/condition_call_linter.R
AshesITR: We usually put it in `@seealso`.

AshesITR (commented):
LGTM. I'd move the link to `@seealso` for consistency among the help pages, though.

AshesITR (approved):

PR #2227, opened 2023-10-04
avoid structure()

MichaelChirico:
`structure()` is IME always slower (admittedly on very small absolute scale) than the alternatives, with little tradeoff of readability (if anything I think using setters is more canonical -> more readable). E.g.

```r
library(microbenchmark)

l <- list()

f1 <- function(l) structure(l, class = c("a", "b"), name = "nm")
f2 <- function(l) {
  class(l) <- c("a", "b")
  attr(l, "name") <- "nm"
  l
}

microbenchmark(times = 1e6, f1(l), f2(l))
# Unit: microseconds
#   expr   min    lq     mean median    uq      max neval cld
#  f1(l) 5.432 5.746 7.478789  5.895 6.280 74773.89 1e+06   b
#  f2(l) 1.469 1.616 2.235726  1.681 1.795 69402.46 1e+06  a 

identical(f1(l), f2(l))
# [1] TRUE
```

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -15,10 +15,9 @@
 }
 
 flatten_lints <- function(x) {
-  structure(
-    flatten_list(x, class = "lint"),
-    class = "lints"
-  )
+  x <- flatten_list(x, class = "lint")
+  class(x) <- "lints"
+  x
 }
 
 # any function using unlist or c was dropping the classnames,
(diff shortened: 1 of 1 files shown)


AshesITR (approved):

klmr:
This feels very subjective. For what it’s worth, I find the immutable logic using `structure()` clearer and thus preferable to the reassignment with replacement functions.

I don’t object to providing the lint (after all, many lints are subjective and they are configurable) but I am surprised that this one is being made part of the _default_ list of undesirable functions, since it does not share the objectively objectionable characteristics of the other functions in that list (such as causing global side effects) — instead, it is an entirely subjective preference as far as I can tell.

PR #2228, opened 2023-10-04
set structure() as an undesirable function by default

MichaelChirico:
Follow-up to #2227


R/zzz.R
```diff
@@ -194,12 +196,14 @@ default_undesirable_functions <- all_undesirable_functions[names(all_undesirable
...
   "source",
+  "structure",
   "Sys.setenv",
   "Sys.setlocale",
   "Sys.unsetenv",
   "trace",
   "undebug",
-  "untrace"
+  "untrace",
+  NULL
```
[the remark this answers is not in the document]
MichaelChirico: it will shrink any future diff from appending to the end of the list (instead of 2 lines diff, one of which is just adding a trailing comma, the diff is focused on the one line added). Similar motivation in Python (c.f. `[1,2,]` and `(1,2,)`) and Go for allowing trailing `,` in constructors (in fact it's _required_ in Go: https://dave.cheney.net/2014/10/04/that-trailing-comma).

Also here:

https://github.com/r-lib/lintr/blob/8d420b5a035645116be2e1ad216ebea5cabe4ce5/R/shared_constants.R#L159

Basically accomplishing `rlang::list2()` for `c()`

AshesITR:
LGTM holding off merging until NEWS is up to speed so we can't forget the bullet.

.lintr
```diff
@@ -31,7 +31,7 @@ linters: linters_with_defaults(
     sort_linter(),
     sprintf_linter(),
     strings_as_factors_linter(),
-    undesirable_function_linter(c(Sys.setenv = NA_character_, mapply = NA_character_)),
+    undesirable_function_linter(c(Sys.setenv = NA_character_, mapply = NA_character_, structure = NA_character_)),
```
AshesITR, line 34: Didn't we clean this up?

MichaelChirico: Not sure I follow...

AshesITR: NVM I had thought we were using all default undesirables.

R/zzz.R
```diff
@@ -171,7 +171,9 @@ all_undesirable_functions <- modify_defaults(
   untrace = paste(
     "remove this likely leftover from debugging.",
     "It is only useful for interactive debugging with trace()"
-  )
+  ),
+  structure =
```
AshesITR, line 175: Is it possible to do a trailing comma trick here, too?

MichaelChirico: not trivially -- rlang::list2() inside modify_defaults() would be the right approach to let someone else handle the implementation, or we'd have to hack it together ourselves looking for missing() at the end of ...

AshesITR: Okay, never mind then.

AshesITR (approved):

PR #2239, opened 2023-10-13
Check if vigilant workflow is working as expected

IndrajeetPatil:


IndrajeetPatil:
Ignore the lint workflow failure, that's deliberate.

The relevant thing to note is that this workflow doesn't fail even if there is a [warning](https://github.com/r-lib/lintr/actions/runs/6512077944/job/17689024105?pr=2239) in tests:

```r
✔ | F W  S  OK | Context

⠏ |          0 | testthat                                                       [ FAIL 0 | WARN 1 | SKIP 0 | PASS 5149 ]

[ FAIL 0 | WARN 1 | SKIP 0 | PASS 5149 ]
```

Issue #1807, opened 2022-12-07
`brace_linter()` could allow skipping braces in function factories


AshesITR:
``` r
lintr::lint(text = "function() function(x) {\n  x\n}")
#> <text>:1:1: style: [brace_linter] Any function spanning multiple lines should use curly braces.
#> function() function(x) {
#> ^~~~~~~~~~~~~~~~~~~~~~~~
```

<sup>Created on 2022-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This could optionally be allowed in addition to
```r
# always allowed
function() {
  function(x) {
    x
  }
}

# could optionally be allowed
function() function(x) {
  x
}
```

MichaelChirico:
WDYT about making it the same option as allowing e.g.

```
function(x) AWrapper(
  y,
  x
)
```

?

AshesITR:
Wouldn't that disable that particular lint completely?

MichaelChirico:
No, there's still

```r
function(x)
  AWrapper(y, x)

function(x)
  AWrapper(
    y,
    x
  )
```

AshesITR:
Ah okay. Any ideas for an option name? `allow_simple_multiline = FALSE` is what comes to my mind, but I'm not really happy with it.

IndrajeetPatil:
But omitting braces here reduces readability, since the reader now needs to mentally add those braces back in. 

```r
function() function(x) {
  x
}
```

Why should we make an exception for such an omission?  That is, why is it so important here to leave out braces?

AshesITR:
Because you need two additional lines and an additional level of indent.
Not arguing that this should be default, but if you define a bunch of function factories, you end up with a lot of essentially empty lines and reduced available line width for the actual implementations.

MichaelChirico:
Agree it can be useful, and warrants an argument, but also that it should not be the default choice.

IndrajeetPatil:
Okay, thanks for the explaining. Makes sense.

As for the parameter name, how about `curly_braces_required` (default: `TRUE`)?

MichaelChirico:
is `strict=TRUE` too glib?

AshesITR:
`strict` sounds okay, while not very specific. `curly_braces_required` would suggest that the other examples mentioned by @MichaelChirico don't lint, which they always should.

PR #2240, opened 2023-10-15
Add param `function_bodies` to `brace_linter()`

salim-b:
Fixes https://github.com/r-lib/lintr/issues/1807.


AshesITR:
This seems too aggressive at removing lints.
The following are meant to lint even with the option enabled:

```r
function(x)
  AWrapper(y, x)

function(x)
  AWrapper(
    y,
    x
  )
```

See https://github.com/r-lib/lintr/issues/1807#issuecomment-1340515302

So omitting braces is only okay if the next (and only) token that makes up the function body starts on the same line, e.g.

```r
function(x) AWrapper(
  y,
  x
)
```

salim-b:
> So omitting braces is only okay if the next (and only) token that makes up the function body starts on the same line

I see, sorry didn't look at the mentioned issue closely enough then.

Personally, I don't care too much about these special variations, so I'd be happy with having the `require_multi_line` param as it is now. And I'm not familiar enough with XPath to implement it the way you outline, I'm afraid. So I hope it's okay I let this up to you guys :slightly_smiling_face:

MichaelChirico:
what's supported by this PR is not style guide-compliant, but that's ok, as long as the default is compliant. Agree that this does not close the referenced issue (though related -- it's even more lax). maybe we can have this parameter take 3 values instead:

function_braces = c("always", "multi_call", "never")

(the naming of the middle option is as usual tricky)

if you're interested, give it a try to implement the middle option -- I suspect it's doable with some effort for you.

Otherwise, feel free to implement the design above with a stop() for "not yet implemented" on the middle option & we'll handle it in follow up (before the subsequent release).

salim-b:
@MichaelChirico Thanks for the encouragement,  I think I got it. IMHO `"multi_line"` is a better name for the middle option than `"not_inline"` (although I'm no native speaker; but "inline function" seems to evoke [different concepts](https://en.wikipedia.org/wiki/Inline_function)).

Open question: When using `function_braces = c("always", "multi_line", "never")` in the `brace_linter` definition (with `function_braces <- match.arg(function_braces)` afterwards), the **default changes**. I wouldn't care personally, but the current tests do, especially https://github.com/r-lib/lintr/blob/4a8b9316e0349804a87eff3cb25152d196535265/tests/testthat/test-expect_lint.R#L63

Is it fine to change the default (and adapt the dummy package content accrodingly)? Or should I set `function_braces = "multi_line"`?

AshesITR:
The default behavior should not change, because it aligns to the style guide.

Also, IINM, `allow_single_line` becomes deprecated by this PR and should be replaced by `function_braces = "multi_line"` with a warning.

```r
function(x) x

function(x) wrapper(
  x,
  y
)

function(x)
  wrapper(
    x,
    y
  )
```

salim-b:
> The default behavior should not change, because it aligns to the style guide.

Ok. Still bears the question which `brace_linter()` signature you prefer, i.e. option 2 or 3 [above](https://github.com/r-lib/lintr/pull/2240#issuecomment-1767404769).

> Also, IINM, `allow_single_line` becomes deprecated by this PR and should be replaced by `function_braces = "multi_line"` with a warning.

I don't think so. `function(x) { x + 4 }` will be linted by `brace_linter(allow_single_line = FALSE, function_braces = "multi_line")`, but not by `brace_linter(allow_single_line = TRUE, function_braces = "multi_line")`. This is the way it should be, no?

> Also we have four levels I think

I think we can go with three options only, so `brace_linter(function_braces = "multi_line")` will lint both of the following:

```r
function(x)
  x + 4
  
function(x) x +
  4
```

AshesITR:
> > The default behavior should not change, because it aligns to the style guide.
> 
> Ok. Still bears the question which `brace_linter()` signature you prefer, i.e. option 2 or 3 [above](https://github.com/r-lib/lintr/pull/2240#issuecomment-1767404769).

Option 3 seems to be viable.

> > Also, IINM, `allow_single_line` becomes deprecated by this PR and should be replaced by `function_braces = "multi_line"` with a warning.
> 
> I don't think so. `function(x) { x + 4 }` will be linted by `brace_linter(allow_single_line = FALSE, function_braces = "multi_line")`, but not by `brace_linter(allow_single_line = TRUE, function_braces = "multi_line")`. This is the way it should be, no?

🤦 sorry, I was confused.
You are right.

> > Also we have four levels I think
> 
> I think we can go with three options only, so `brace_linter(function_braces = "multi_line")` will lint both of the following:
> 
> ```r
> function(x)
>   x + 4
>   
> function(x) x +
>   4
> ```

I think the former should lint, but the latter isn't easily distinguishable in the AST from
```r
function() wrapper(
  x,
  y
)
```
so it shouldn't lint (at least for some value of `function_braces`).

salim-b:
> so it shouldn't lint (at least for some value of function_braces).

It won't lint for `brace_linter(function_braces = "never")`. Is that ok for you?

AshesITR:
I would like to have an option that distinguishes between function definitions where the body starts on the same line, like
```r
function(x) x +
  4
```
, and those where the body does _not_ start on the same line, like
```r
function(x)
  x + 4
```

salim-b:
> I would like to have an option that distinguishes between function definitions where the body starts on the same line, like (...), and those where the body does not start on the same line, like (...)

Ok. I'm afraid I can't help with that. But you could extend the `function_braces` param with a forth option `not_start_on_same_line`[^n] in the future. I've updated this PR with the three-option solution.

[^n]: Hopefully with a better name 😄

AshesITR:
Thanks for the effort, I'll try and take a look some time this week to see if I can implement the fourth option.
Name suggestions welcome 😆

AshesITR:
@salim-b PTAL, I named it `"not_inline"` for now.
Please refrain from force-pushing since I've cloned your branch now and it would cause divergent histories.
We squash all commits when a PR is merged into main, so don't worry about creating too many commits.

AshesITR:
Also, `brace_linter(function_braces = "always")` effectively prohibits inline function definitions. Not sure whether we should

1. ignore this fact
2. warn the user about using `function_braces = "always"` without `allow_single_line = TRUE`
3. implicitly set `allow_single_line = TRUE` if `function_braces = "always"`

What do you think, @MichaelChirico?

AshesITR:
We should also discuss what to do about multi-line function headers.

```r
# these all trigger "always", "multi_line" and "not_inline" atm
function(
    a,
    b
) a + b

function(
    a,
    b
) a + 
  b

function(
    a,
    b
)
  a + b

# this forces allow_single_line = TRUE to be set
function(
    a,
    b
) { a + b }

# this never lints
function(
    a,
    b
) {
  a + b
}
```

AshesITR:
> I'm not sure whether I understand correctly... but didn't you mean to write **`brace_linter(function_braces = "never")`** (instead of `"always"`) effectively prohibits inline function definitions?

I meant `"always"`, see these tests:

https://github.com/r-lib/lintr/pull/2240/files#diff-42f1e9787771f9204786c9d84ece7bc678f89e7cbd6b3d826f9a2e62ba90da5eR325
https://github.com/r-lib/lintr/pull/2240/files#diff-42f1e9787771f9204786c9d84ece7bc678f89e7cbd6b3d826f9a2e62ba90da5eR336

None of these two ways to define an inline function, with or without braces, is acceptable for `brace_linter(allow_single_line = FALSE, function_braces = "always")`.
```r
function(x) x
function(x) { x }
```

MichaelChirico:
FYI: I'm holding off on review until after release. thanks for your patience.

salim-b:
> Added Option 4.: Carve out `allow_single_line = FALSE` to allow `function(x) { x }` if and only if `function_braces = "always"`.

I think I'd still prefer option 1. But you and @MichaelChirico are certainly more qualified to decide on that^^

AshesITR:
Yes and we pin down the behavior to make sure it works as intended.

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
 #'  - Closing curly braces in `if` conditions are on the same line as the corresponding `else`.
 #'  - Either both or neither branch in `if`/`else` use curly braces, i.e., either both branches use `{...}` or neither
 #'    does.
-#'  - Functions spanning multiple lines use curly braces.
+#'  - Function bodies are wrapped in curly braces.
 #'
-#' @param allow_single_line if `TRUE`, allow an open and closed curly pair on the same line.
+#' @param allow_single_line If `TRUE`, allow an open and closed curly pair on the same line.
```
MichaelChirico, line 13: what's the source of the diff on this line?

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
...
+#' @param function_braces Character specifying whether to require function bodies to be wrapped in curly braces.
```
MichaelChirico, line 14:
```suggestion
#' @param function_braces Character string specifying whether to require function bodies to be wrapped in curly braces.
```

(NB: also need to `roxygenize()`)

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
...
+#'   `"always"` requires braces for all function definitions, including inline functions.
+#'   `"not_inline"` requires braces when a function body does not start on the same line as its header.
+#'   `"multi_line"` requires braces when a function definition spans multiple lines.
```
MichaelChirico, line 17:
```suggestion
#'   `"multi_line"`, the default, requires braces when a function definition spans multiple lines.
```

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
...
 #'  - Either both or neither branch in `if`/`else` use curly braces, i.e., either both branches use `{...}` or neither
 #'    does.
-#'  - Functions spanning multiple lines use curly braces.
+#'  - Function bodies are wrapped in curly braces.
 #'
-#' @param allow_single_line if `TRUE`, allow an open and closed curly pair on the same line.
+#' @param allow_single_line If `TRUE`, allow an open and closed curly pair on the same line.
+#' @param function_braces Character specifying whether to require function bodies to be wrapped in curly braces.
+#'   `"always"` requires braces for all function definitions, including inline functions.
+#'   `"not_inline"` requires braces when a function body does not start on the same line as its header.
```
MichaelChirico, line 16: I think "header" is not the right term here, "signature" seems better

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
 #'  - Closing curly braces in `if` conditions are on the same line as the corresponding `else`.
 #'  - Either both or neither branch in `if`/`else` use curly braces, i.e., either both branches use `{...}` or neither
 #'    does.
-#'  - Functions spanning multiple lines use curly braces.
+#'  - Function bodies are wrapped in curly braces.
 #'
-#' @param allow_single_line if `TRUE`, allow an open and closed curly pair on the same line.
+#' @param allow_single_line If `TRUE`, allow an open and closed curly pair on the same line.
+#' @param function_braces Character specifying whether to require function bodies to be wrapped in curly braces.
+#'   `"always"` requires braces for all function definitions, including inline functions.
```
MichaelChirico, line 15: This is written in the roxygen markup as a list, but it won't be rendered in the Rd as such. Use a bulleted list (`*`), maybe?

R/brace_linter.R
MichaelChirico, line 14: Sorry for the late comment on this, I am wondering if `function_braces` is too redundant (cf `brace_linter(function_braces  = ...)`, WDYT about `function_bodies`? cc @AshesITR

R/brace_linter.R
salim-b: Since it's a sentence, I changed the first letter (If) to uppercase.

R/brace_linter.R
salim-b: See https://github.com/r-lib/lintr/pull/2240/commits/229993aebeb98047bbf96ae3faf937d4dccd5da7

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
...
+#'   `"not_inline"` requires braces when a function body does not start on the same line as its header.
+#'   `"multi_line"` requires braces when a function definition spans multiple lines.
+#'   `"never"` never requires braces in function bodies.
```
MichaelChirico: we don't have to include it in this PR, but there is one more style we might cater to in the future, namely, use braces _if and only if_ they're syntactically required.

Consider

```r
A <- function() { NULL }
B <- function() { NULL; NULL }
C <- function() NULL
D <- function() {
  NULL
}
E <- function()
  NULL
F <- function() {
  NULL
  NULL
}
G <- function() invisible(
)
H <- function() { invisible(
) }
```

And the corresponding lint table ( :heavy_check_mark: :arrow_right: throw a lint):

| `function_braces=` | A | B | C | D | E | F | G | H |
|---------------------------- | -- | -- | -- | -- | -- | -- | -- | -- |
| always |  |  | :heavy_check_mark:   | | :heavy_check_mark: | | :heavy_check_mark: | |
| not_inline | | | | | :heavy_check_mark: | | | |
| multi_line | | | | | :heavy_check_mark: | | :heavy_check_mark: | |
| never | | | | | | | | |
| _strict_ | :heavy_check_mark:  | | | :heavy_check_mark:  | | | | :heavy_check_mark: |

(PS we might consider adding any of the above as test cases if they're not yet represented)

R/brace_linter.R
MichaelChirico: Thanks, missed that.

R/brace_linter.R
MichaelChirico: Mentioning this here to possibly align on the naming for such a style. If we can't come up with a name that's natural for the current `function_braces` (or `function_bodies`), we might need to think of a different parameter name to accommodate. Hope that makes sense.

R/brace_linter.R
```diff
@@ -124,7 +132,25 @@ brace_linter <- function(allow_single_line = FALSE) {
...
+
+    xp_function_brace <- glue(
+      "(//FUNCTION | //OP-LAMBDA)/parent::expr[{xp_cond_function_brace} and not(expr[OP-LEFT-BRACE])]"
+    )
+
+    msg_function_brace <- switch(
+      function_braces,
+      always = "Any function body should be wrapped in curly braces.",
+      multi_line = "Any function body spanning multiple lines should be wrapped in curly braces.",
+      not_inline = "Any function body starting on a new line should be wrapped in curly braces."
```
MichaelChirico, line 151:
```suggestion
      always = "Wrap function bodies in curly braces.",
      multi_line = "Wrap multi-line function bodies in curly braces.",
      not_inline = "Wrap function bodies starting on a new line in curly braces."
```

(will require updates to tests)

R/brace_linter.R
salim-b: I think this makes sense and renamed it.

salim-b (commented):
https://github.com/r-lib/lintr/pull/2240/commits/5eaae61e1e3f023d3db4e09e71f8d0efb98165b8 broke tests, thus I reverted it.

R/brace_linter.R
salim-b: > use braces *if and only if they're syntactically required*.

What exactly do you mean by "syntactically required"? I'm confused by your examples above, `A`, `D` and `H` are fine without braces (same result)...

salim-b:
@MichaelChirico This PR has been sitting here for almost three months without any further changes. I'd love to get it merged.

R/brace_linter.R
MichaelChirico: > What exactly do you mean by "syntactically required"? 

i.e., "without braces, the code either wouldn't parse, or would parse incorrectly". In other words "function uses braces" if and only if "function has more than one expression" (I think that should be equivalent to `length(body(f)) > 1`)

> A, D and H are fine without braces (same result)...

Precisely why they would throw a lint (under "strict" mode, or maybe "minimalist" mode is more evocative) -- that style ditches `{` whenever possible. In A/D/H, ditching `{` doesn't impact the validity of the code --> lint

NEWS.md
```diff
@@ -55,6 +55,7 @@
 * `todo_comment_linter()` has a new argument `except_regex` for setting _valid_ TODO comments, e.g. for forcing TODO comments to be linked to GitHub issues like `TODO(#154)` (#2047, @MichaelChirico).
 * `vector_logic_linter()` is extended to recognize incorrect usage of scalar operators `&&` and `||` inside subsetting expressions like `dplyr::filter(x, A && B)` (#2166, @MichaelChirico).
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
+* `brace_linter()`' has a new argument `function_bodies` (default `"multi_line"`) which controls whether to require function bodies to be wrapped in curly braces, with the options `"always"`, `"multi_line"` (only require curly braces when a function body spans over multiple lines), `"not_inline"` (only require curly braces when a function body starts on a new line) and `"never"` (#1807, #2240, @salim-b).
```
MichaelChirico, line 58:
```suggestion
* `brace_linter()`' has a new argument `function_bodies` (default `"multi_line"`) which controls whether to require function bodies to be wrapped in curly braces, with the options `"always"`, `"multi_line"` (only require curly braces when a function body spans multiple lines), `"not_inline"` (only require curly braces when a function body starts on a new line) and `"never"` (#1807, #2240, @salim-b).
```

R/brace_linter.R
```diff
@@ -8,9 +8,14 @@
 #'  - Closing curly braces in `if` conditions are on the same line as the corresponding `else`.
 #'  - Either both or neither branch in `if`/`else` use curly braces, i.e., either both branches use `{...}` or neither
 #'    does.
-#'  - Functions spanning multiple lines use curly braces.
+#'  - Function bodies are wrapped in curly braces.
 #'
-#' @param allow_single_line if `TRUE`, allow an open and closed curly pair on the same line.
+#' @param allow_single_line If `TRUE`, allow an open and closed curly pair on the same line.
+#' @param function_bodies Whether to require function bodies to be wrapped in curly braces. One of
```
MichaelChirico, line 14: "Whether" kind of implies a binary choice

```suggestion
#' @param function_bodies Character string governing when function bodies without curly braces should generate lints. One of
```

R/brace_linter.R
```diff
@@ -124,7 +132,25 @@ brace_linter <- function(allow_single_line = FALSE) {
...
+  if (function_bodies != "never") {
+    xp_cond_function_brace <- switch(
+      function_bodies,
+      always = "true",
+      multi_line = "@line1 != @line2",
+      not_inline = "@line1 != expr/@line1"
+    )
+
+    xp_function_brace <- glue(
+      "(//FUNCTION | //OP-LAMBDA)/parent::expr[{xp_cond_function_brace} and not(expr[OP-LEFT-BRACE])]"
```
MichaelChirico, line 144: save one character+level of nesting

```suggestion
      "(//FUNCTION | //OP-LAMBDA)/parent::expr[{xp_cond_function_brace} and not(expr/OP-LEFT-BRACE)]"
```

tests/testthat/test-brace_linter.R
MichaelChirico, line 1: Suggested new test case:

```r
foo <- function(x) ({
  x + 1
})
```

(I think it should lint, this is an unusual way to declare a function that should be discouraged in favor of plain `{`.)

R/brace_linter.R
MichaelChirico: Also, as @AshesITR points out, I think we should be careful here to specify "function body" for all the rules, since a function _definition_ includes the parameters and those can take up many lines even if the function itself doesn't (not that we expect to see that except in strange cases).

tests/testthat/test-brace_linter.R
MichaelChirico: Also, something like:

```r
foo <- function(x) { x +
  4 }
```

Is slightly different from the current set of cases & worth including as a regression test I think.

R/brace_linter.R
salim-b: Addressed in https://github.com/r-lib/lintr/pull/2240/commits/01bb196ed89797d089b5d19a96152a878617c57e. I've opted for a shorter description without the param type ("character string") since it should become clear from the second sentence (and we do `match.arg(function_bodies)` anyways).

salim-b:
@MichaelChirico I've [reverted](https://github.com/r-lib/lintr/pull/2240/commits/e21c4e797a14fd27a6c996e50b8bf7fe24ad3529) your [XPath "readability improvement"](https://github.com/r-lib/lintr/pull/2240/commits/7620e3791ce570a5d0fca3958ba8f8c55d257952) (the [2nd time](https://github.com/r-lib/lintr/pull/2240#pullrequestreview-1932966374)!) since it's apparently wrong (run the tests!).

tests/testthat/test-brace_linter.R
salim-b: I've added two additional test sets for these cases: https://github.com/r-lib/lintr/pull/2240/commits/689f8036fe1e62aab5e1ea13dff06de09eaa0143 and https://github.com/r-lib/lintr/pull/2240/commits/153adef6a8e69d071978e5c4a98ad104cf95a6fe. I'm not sure whether the latter is what you expect, though.

MichaelChirico:
> @MichaelChirico I've [reverted](https://github.com/r-lib/lintr/pull/2240/commits/e21c4e797a14fd27a6c996e50b8bf7fe24ad3529) your [XPath "readability improvement"](https://github.com/r-lib/lintr/pull/2240/commits/7620e3791ce570a5d0fca3958ba8f8c55d257952) (the [2nd time](https://github.com/r-lib/lintr/pull/2240#pullrequestreview-1932966374)!) since it's apparently wrong (run the tests!).

apologies 😅

MichaelChirico (approved):
Looks great, thanks for your persistence!

katrinabrock:
❤️  Thanks again!

PR #2254, opened 2023-11-04
more considered handling of conditions in config parsing

### LINKED ISSUE #2253: Linter warnings bubble up as config formatting errors with confusing message
See the following reprex package: https://github.com/Bisaloo/testlintrconfigerror/:

```dcf
linters: linters_with_defaults(
    backport_linter("2.10.0")
  )
```

The config file is correctly formatted but the `backport_linter()` setting will generate a warning:

https://github.com/r-lib/lintr/blob/1493c5e0bafc2483bea58a4f12404c68cf483e59/R/backport_linter.R#L116-L119

This warning is totally unrelated with the config formatting but will bubble up when config is read and produce a red herring message:

```r
lintr::lint_package()
```

> Error: Malformed config file, ensure it ends in a newline
>  Malformed config file, ensure it ends in a newline
>  It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.

I looked again and again for the missing newline and couldn't figure out what was wrong until I realized the error came from the warning added at the end of the error message.

MichaelChirico:
Closes #2253.

The output in @Bisaloo's case now looks like this:

```r
withr::with_dir("/tmp/testlintrconfigerror", lint_package())
# Warning message:
# Warning encountered while loading config:
#   Warning from config setting 'linters':
#     It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.
```

Open to fine-tuning it a bit but feels like we're in the right direction here.


AshesITR:
Can we get the call that produced the warning to appear?

MichaelChirico:
> Can we get the call that produced the warning to appear?

Do you just want to print the string that caused the warning (in this case, it would be `"linters_with_defaults(\n    backport_linter(\"2.10.0\")\n  )"`)? Or to isolate it to `backport_linter(\"2.10.0\")`? The latter would require a lot more code than just `eval(.)`. My only concern the former is printing an arbitrarily formatted string can be quite long (e.g. our own `linters` entry is 39 lines long), output may look scary / hard to understand in that case.

MichaelChirico:
Here's what the call stack looks like from within the relevant warning handler:

```
[[17]]
eval(parsed_setting)

[[18]]
eval(parsed_setting)

[[19]]
linters_with_defaults(backport_linter("2.10.0"))

[[20]]
dots <- list(...)

[[21]]
r_version <- normalize_r_version(r_version)

[[22]]
warning("It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.")

[[23]]
.signalSimpleWarning("It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.", 
    base::quote(normalize_r_version(r_version)))

[[24]]
withRestarts({
    .Internal(.signalCondition(simpleWarning(msg, call), msg, 
        call))
    .Internal(.dfltWarn(msg, call))
}, muffleWarning = function() NULL)

[[25]]
withOneRestart(expr, restarts[[1L]])

[[26]]
doWithOneRestart(return(expr), restart)

[[27]]
(function(w) {
            browser()
            warning("Warning from config setting '", setting, "':\n    ", conditionMessage(w))
            invokeRestart("muffleWarning")
          })(list(message = "It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.", 
    call = normalize_r_version(r_version)))
```

`w$call` (what would usually be shown under `call.=TRUE`) is `normalize_r_version(r_version)`.

Would including `w$call` be an improvement in this case?

AshesITR:
`normalize_r_version()` seems worth including.
How does dplyr & friends handle the dots?
Maybe we can improve `with_defaults()`, too.

```r
test <- dplyr::mutate(iris, okay = Sepal.Length + Sepal.Width, warny = warning("warn"))
```

Provides a very nice warning:
```
Warning: There was 1 warning in `dplyr::mutate()`.
i in argument: `warny = warning("warn")`.
Caused by warning:
! warn
```

MichaelChirico:
IINM that would require hooking into the {rlang} error+stack system (which I don't know anything about). Not sure it's worth it for this use case.

MichaelChirico:
Current output from the test case in the issue:

```
withr::with_dir("/tmp/testlintrconfigerror", lint_package())
Warning message:
Warning encountered while loading config:
  Warning from config setting 'linters' at call 'normalize_r_version(r_version)':
    It is not recommended to depend on an R version older than 3.0.0. Resetting 'r_version' to 3.0.0.
```

MichaelChirico:
> Warning: There was 1 warning in `dplyr::mutate()`.
> i in argument: `warny = warning("warn")`.
> Caused by warning:
> ! warn
> ```

BTW, `mutate()` already evaluates its arguments one-by-one (necessary to make `mutate(DF, a = 1, b = a +1)` possible), which makes it more natural to include a warning message naming the culprit argument.

AshesITR:
Fine for now. "at call..." sounds akward to me, though.

AshesITR:
Base R does "Error in <call>: <message>".
WDYT about "Warning from config setting 'linters' in 'normalize_r_version(r_version)'..."?

MichaelChirico:
Sure, that works

NEWS.md
```diff
@@ -5,7 +5,7 @@
...
-  + We also validate config files up-front make it clearer when invalid configs are present (#2195, @MichaelChirico). There is a warning for "invalid" settings, i.e., settings not part of `?default_settings`. We think this is more likely to affect users declaring settings in R, since any variable defined in the config that's not a setting must be removed to make it clearer which variables are settings vs. ancillary.
+  + We also validate config files up-front make it clearer when invalid configs are present (#2195, @MichaelChirico). There is a warning for "invalid" settings, i.e., settings not part of `?default_settings`. We think this is more likely to affect users declaring settings in R, since any variable defined in the config that's not a setting must be removed to make it clearer which variables are settings vs. ancillary. Thanks also to @Bisaloo for early testing to improve the failure interface for invalid configs (#2253).
```
AshesITR, line 8: rm "also"

AshesITR:
LGTM except for the nit in NEWS.

NEWS.md
MichaelChirico: i actually think it's fine

AshesITR: I like the more concise style. But feel free to merge anyway

AshesITR (dismissed):

NEWS.md
IndrajeetPatil: I guess this NEWS item should now move to the next version?

MichaelChirico: Yes, waiting on the CRAN release commit to land here first.

MichaelChirico: (moot in the end since this didn't make it to 3.1.1)

AshesITR (approved):

PR #2271, opened 2023-11-10
Linter for explicit/implicit returns

### LINKED ISSUE #1100: New `return_linter` to enforce consistent return style
Tidyverse expects implicit returns, so those should be the default.

Basically check that the last expr in all function definitions is a) never a return() or b) always a return().

Signature could maybe be `return_linter(use_implicit_returns = TRUE)`?

The google linters introduced a bunch of these (`grep return\(lapply\( R/*.R`)

MEO265:
Closes #1100.

I have designed the implementation in such a way that as far as possible no false positive errors occur. This leads to a relatively large number of false negatives, but I would like to accept that.
If anyone can construct false positive cases, I would be grateful if you would write them in the comments.

Inline functions have been ignored so far, this is to a certain extent intentional, as I personally am a supporter of explicit returns, but find them unnecessary in inline functions. On the other hand, this case would have become much more complex.
If someone ever wants to implement it, I think an `lint_inline` parameter would be nice.

FYI: 
There are 10 explicit `returns` (that the linter finds) in the `linter` package and although I prefer them, they should probably be removed for consistency.


AshesITR:
The warnings stem from exclusions of linters that you removed by providing `linters=`.
To avoid them, add `return_linter()` to our `.lintr` instead.

MichaelChirico:
Hmm, we do already have an `explicit_return_linter()` [our default conflicts with the Tidyverse guide & we require all returns to be _explicit_]. Seems like we've got some duplicate effort here. OTOH yours already goes further by trying to support both styles (implicit + explicit)

We have tested ours pretty extensively (no false positives on 10,000+ files) -- here's all the tests we've written:

https://gist.github.com/MichaelChirico/9164672a4e762627c09ad6b521e54964

Please ensure these tests pass under `use_implicit_returns = FALSE`

Here's our XPath:

https://gist.github.com/MichaelChirico/084bc163c01ad0debba8b1b945b75009

The major issue I'm aware of with our implementation is it doesn't handle arbitrary nesting in terminal `if`/`else` branches -- we only look at top-level `if`/`else` (i.e. depth=1)

MichaelChirico:
> Inline functions have been ignored so far

I agree we shouldn't require `return()` on inline functions -- that's how we implemented it as well.

MichaelChirico:
> There are 10 explicit returns (that the linter finds) in the linter package and although I prefer them, they should probably be removed for consistency.

Most likely, these are from copy-pasting our internal code where the styles are opposite. Yes, we should remove them here for consistency (and because we're meant to enforce the tidyverse guide)

MEO265:
> We have tested ours pretty extensively (no false positives on 10,000+ files) -- here's all the tests we've written:
> 
> https://gist.github.com/MichaelChirico/9164672a4e762627c09ad6b521e54964
> 
> Please ensure these tests pass under `use_implicit_returns = FALSE`

Just by looking at it I can tell that my linter would produce lots of false positives. Many that one should definitely pay attention to, e.g. `invokeRestart`, `tryInvokeRestart`, `UseMethod`, etc. and others that I would say are actually marked correctly, namely  `message`, `stopifnot` and maybe even `warning`.

Maybe in addition to a fixed number of basic functions that are taken into account in the code, there should be the option to accept more functions (from other packages).

> Here's our XPath:
> 
> https://gist.github.com/MichaelChirico/084bc163c01ad0debba8b1b945b75009
> 
> The major issue I'm aware of with our implementation is it doesn't handle arbitrary nesting in terminal `if`/`else` branches -- we only look at top-level `if`/`else` (i.e. depth=1).

Thanks for access to the code. I see that you have already paid a lot more attention. What is the best way to proceed? I can try to bring the two together and make sure everything passes, although that might take a while (which doesn't bother me), but if you want to take over that would be ok too.

> See #884 for linters we've already written that are just a matter of porting over.

Is there a way to see the code of all linters still to be transferred or to get access to certain ones? If so, I would like to try to support whenever my time allows.

MEO265:
> The warnings stem from exclusions of linters that you removed by providing `linters=`. To avoid them, add `return_linter()` to our `.lintr` instead.

Thanks for the answer, what surprised me is that not a single warning came when I called up my linter without argument

MichaelChirico:
> Is there a way to see the code of all linters still to be transferred or to get access to certain ones?

Not easily... I have to strip some internal-only pieces from the code before uploading them here. If you'd like to see specific linters I am happy to do so one a one-by-one basis.

MichaelChirico:
> when it comes to controls like for and while, my code is also a little stricter.

I think I'd be open to simplifying things a bit -- always lint on a terminal `for`/`while`/`repeat`, requiring the function to terminate with `return(invisible())` instead (or perhaps a `stop()` like you mentioned in case the function is meant to return early). I think I didn't realize at the time we initially implemented this that those control flows always return `NULL` implicitly:

```r
foo <- function() {
  for (ii in 1:10) {
    ii
  }
}
x <- foo()
dput(x)
# NULL
```

So under explicit return style, IMO it's always preferable to make that behavior more explicit:

```r
foo <- function() {
  for (ii in 1:10) {
    ii
  }
  return(invisible())
}
```

MichaelChirico:
> Maybe in addition to a fixed number of basic functions that are taken into account in the code, there should be the option to accept more functions (from other packages).

Yes, that's what I would do... one issue is the list is already a bit long, the signature will get pretty messy if the default list is huge. So I would propose the parameter being _extra_ functions to skip.

Something like

```r
# User has no control here; all of these come from base
default_allowed_functions <- c(
  # Normal calls
  "return", "stop", "warning", "message", "stopifnot", "q", "quit",
  "invokeRestart", "tryInvokeRestart",

  # Functions related to S3 methods
  "UseMethod", "NextMethod",

  # Functions related to S4 methods
  "standardGeneric", "callNextMethod",

  # Functions related to C interfaces
  ".C", ".Call", ".External", ".Fortran"
)

# User supplies these; internally we would use this
extra_allowed_functions <- c(
  # Normal calls from non-default libraries
  "LOG", "abort",

  # tests in the RUnit framework are functions ending with a call to one
  #   of the below. would rather users just use a different framework
  #   (e.g. testthat or tinytest), but already 250+ BUILD files depend
  #   on RUnit, so just cater to that. confirmed the efficiency impact
  #   of including these is minimal.
  # RUnit tests look like 'TestInCamelCase <- function()'
  #   NB: check for starts-with(text(), 'Test') below is not sufficient, e.g.
  #   in cases of a "driver" test function taking arguments and the main unit
  #   test iterating over those.
  "checkEquals", "checkEqualsNumeric", "checkException", "checkIdentical",
  "checkStop", "checkTrue", "checkWarnings",
)
```

AshesITR:
I'm assuming `"LOG"` is internal to google?
In that case, I'd remove it from the official defaults.

Agree we should silently add all base R exceptions and maybe rlang (abort)?
Or keep `"abort"` in the signature to hint at the usage, dropping RUnit support out of the box?

AshesITR:
Actually, why allow `warning()` as the last call? It quietly returns the message as a string.

MichaelChirico:
> Actually, why allow `warning()` as the last call? It quietly returns the message as a string.

Hmm, not sure we noticed that at the time. I do find it very strange that behavior differs from `message()` (which returns `NULL`).

Anyway, for our case I would still skip `return()`. I think requiring it for `warning()` would make `tryCatch()`/`withCallingHandlers()` usage overly verbose, e.g.

```r
tryCatch(
  foo(),
  error = \(e) {
    cat(Sys.time(), file="log")
    cat(e$message, file="log", append=TRUE)
    warning(e)
  }
)
```

Treating all signalling functions the same makes more sense to me.

That said, for {lintr}'s defaults, we can consider something different.
```

**Comment by AshesITR** [comment]
Shouldn't your example lint on `foo()`, or except `tryCatch()` and friends?

File: R/return_linter.R [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+#'
+#' @evalRd rd_tags("return_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#'
+#' @export
+return_linter <- function(use_implicit_returns = TRUE) {
+
+  if (use_implicit_returns) {
+    xpath <- "
+      (//FUNCTION | //OP-LAMBDA)[following-sibling::expr[1]/*[1][self::OP-LEFT-BRACE]]
```
MichaelChirico, line 15: we should be able to combine the repetitive `following-sibling::expr[1]`, e.g.

```
(//FUNCTION | //OP-LAMBDA)
  /following-sibling::expr[1][OP-LEFT-BRACE]
  /expr[last()][...]
```

ditto for the XPath in the other branch

AshesITR:
Maybe instead of allowing `warning()` etc. we should carve out functions passed to `withCallingHandlers()` and friends?

R/return_linter.R
MEO265: If you leave that out, wouldn't you have a problem with in-line functions? This way you can easily exclude them from the lint

MEO265:
@MichaelChirico I think it would be really useful, even outside of this PR, if we included some of the functions you use for glue, like `.XpTextInTable`, in the package here too. Would make some linter easier to read. 
Or is it already in the package under a different name and I just overlooked it?

MichaelChirico:
> @MichaelChirico I think it would be really useful, even outside of this PR, if we included some of the functions you use for glue, like `.XpTextInTable`, in the package here too. Would make some linter easier to read. Or is it already in the package under a different name and I just overlooked it?

in {lintr} that's xp_text_in_table, see xp_utils.R for similar helpers

tests/testthat/test-return_linter.R
```diff
@@ -0,0 +1,841 @@
...
+  )
+})
+
+test_that("Do not lint control statements (with return) on end of function", {
+  linter <- return_linter(use_implicit_returns = FALSE)
+
+  expect_lint(
+    trim_some("
+      function() {
+        repeat {
```
MichaelChirico, line 68: @MEO265 shouldn't this lint? Otherwise we're treating `while` and `repeat` differently right?

Currently I'm thinking the rule for explicit returns should be that `for`/`while`/`repeat` shouldn't terminate the function -- better to have an explicit `return(invisible())` come last.

tests/testthat/test-return_linter.R
```diff
@@ -0,0 +1,841 @@
...
+      }
+    "),
+    list(msg, line_number = 2L),
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      function() {
+        if(x == 2L){
```
AshesITR, line 143: This code should be reformatted to have some whitespace around ().

tests/testthat/test-return_linter.R
```diff
@@ -0,0 +1,841 @@
...
```
MichaelChirico, line 131: Here is an implicit `else` return, I will leave it out from this PR and handle such cases in #2321

tests/testthat/test-return_linter.R
MichaelChirico: Thanks... I would hold off on stylistic comments until the PR is marked for review (not draft). Some of this will be removed gradually as tests are fixed.

tests/testthat/test-return_linter.R
```diff
@@ -0,0 +1,841 @@
...
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      function(x) {
+        switch(x, a = return(1), 'b' = stop(2), '3' = return(3), stop('End'))
+      }
+    "),
+    NULL,
```
MichaelChirico, line 224: This also looks off to me -- I would expect this to lint. For explicit returns, I think it's better to force "outer" return: `return(switch(...))`.

R/return_linter.R
MichaelChirico: See updated `use_implicit_returns` XPath

MichaelChirico:
Still some work to do, but we're at least passing all existing tests, so it's worth removing from draft status.

MichaelChirico:
> Is there a way to see the code of all linters still to be transferred or to get access to certain ones? If so, I would like to try to support whenever my time allows.

@MEO265 I've now added all of the linters as PRs -- you might try looking at the follow-up issues (see recent new issues), reporting any issues you see with the new linters on your own code, or reviewing pending PRs if you want to pitch in :)

MEO265:
> Still some work to do, but we're at least passing all existing tests, so it's worth removing from draft status.

I'll try to look at it this evening (which is in about 10 hours according to my time zone).

MEO265:
@MichaelChirico I believe we have discussed this before, but nothing was decided. I would like to remove `warning`, `message`, and `stopifnot` from the allowed functions. The first two certainly return values implicitly and `stopifnot` should not be treated differently than `if() stop()`. If I understood correctly, @AshesITR  supported something similar.
If needed, we could introduce a flag like `strict` to disallow these.

On the other hand, I would like to allow `repeat`, as this function would only give an implicit return in case of using a `break`, and we could explicitly test for it if necessary.

Additionally, I would remove `LOG` and `abort`, as well as the `RUnit` functions. We could add parameters for more functions of each category in a further PR (or directly now).

Would that be okay with you?

tests/testthat/test-return_linter.R
```diff
@@ -0,0 +1,842 @@
...
+    linter
+  )
+
+  expect_lint(
+    trim_some("
+      function(x) {
+        switch(x, a = return(1), 'b' = stop(2), '3' = return(3), stop('End'))
+      }
+    "),
+    list(lint_msg, line_number = 2L),
```
MEO265, line 212: @MichaelChirico I don't think we should give a lint here. In my opinion this is a false positive, as there is no way an implicit return is given.
My first implementation handled this accordingly. Maybe we could do that again.

AshesITR:
> @MichaelChirico I believe we have discussed this before, but nothing was decided. I would like to remove `warning`, `message`, and `stopifnot` from the allowed functions. The first two certainly return values implicitly and `stopifnot` should not be treated differently than `if() stop()`. If I understood correctly, @AshesITR  supported something similar.

Yes, I support this proposal.

> If needed, we could introduce a flag like `strict` to disallow these.

I don't think this is necessary. The only way `message()` and `warning()` could actually break control flow would be inside a `withCallingHandlers()`.

> On the other hand, I would like to allow `repeat`, as this function would only give an implicit return in case of using a `break`, and we could explicitly test for it if necessary.

Linting `repeat` with `break` sounds like a good solution. I expect most `repeat` statements in the wild to have either `break` or `return()` so not linting at all seems like a bad trade-off.

> Additionally, I would remove `LOG` and `abort`, and maybe also the `RUnit` functions. We could add parameters for more functions of each category in a further PR (or directly now).

`abort()` is from rlang and breaks control flow. I'd suggest to add a whitelist parameter and default it to `"abort"` to showcase its usage.
That way, packages relying on rlang for signaling won't get false positives out of the box. No hard opinion on the default, though. A customizable list of functions to allow in addition to base R exit functions seems warranted though.

MEO265:
I created commits to address the issues. However, I was completely unsure about how to deal with `Runit` and would be happy if someone suggested an alternative.

Tomorrow I would like to implement the refined handling of `switch` and `repeat`.

>  I'd suggest to add a whitelist parameter and default it to "abort"
I read the note too late. Feel free to add it yourself

R/return_linter.R
```diff
@@ -0,0 +1,206 @@
...
+#'
+#' @param use_implicit_returns Whether to use implicit or explicit returns
+#'
+#' @param additional_allowed_func Names of additional functions that are
+#'  accepted as return if `!use_implicit_returns`
+#'
+#' @param additional_side_effect_func Names of additional functions that are
+#'  not checked for an explicit retun if `!use_implicit_returns`
+#'
+#' @param use_runit Whether to ignore Runit like functions or not
```
AshesITR, line 13: I would leave this argument out, the feature is already present by using the function whitelist arguments.

R/return_linter.R
```diff
@@ -0,0 +1,206 @@
+#' Return linter
+#'
+#' This linter checks for explicit [return()] at the end of a function
+#'
+#' @param use_implicit_returns Whether to use implicit or explicit returns
+#'
+#' @param additional_allowed_func Names of additional functions that are
```
AshesITR, line 7: Better name `return_functions`?

R/return_linter.R
```diff
@@ -0,0 +1,206 @@
...
+#'  accepted as return if `!use_implicit_returns`
+#'
+#' @param additional_side_effect_func Names of additional functions that are
```
AshesITR, line 10: Maybe call this `except`?

Doc could be:
Names of functions that do not need an explicit return.

R/return_linter.R
MEO265: In my opinion, we can also do without this parameter.  However, it cannot be replicated with the other two, because the way it was implemented, all functions that start with “Test” were ignored.  I wanted to keep that with this parameter, even though I'm not a supporter of this functionality.  I would also prefer without this parameter.

R/return_linter.R
MEO265: Maybe "additional_return_functions"?

R/return_linter.R
MEO265: I would add an “additional” to the doc, but otherwise I think it’s a good idea.

R/return_linter.R
MichaelChirico: Yea, it's hard to capture the exemption for `Test*` functions with an allowlist. It's hardly practical for users to supply every single `Test*` function in their suite, and I don't think there's much need for a pattern based argument accepting `Test*` (glob) or `^Test.*` (regex). I think `use_runit` or `allow_runit` is the best bet here.

R/return_linter.R
AshesITR: Okay. Additional to what should then also be documented, i.e. the built-in list of exceptions.

R/return_linter.R
AshesITR: It seems overly verbose to me. Maybe a good way would be to document the base return functions while keeping the argument name short.

R/return_linter.R
AshesITR: I'm not familiar with RUnit but would it be an option to directory-exclude the RUnit test files instead?

MichaelChirico: not really, no. (1) there's not really a consistent directory structure or file name pattern for RUnit tests (2) it won't work in distributed setup where per-directory settings are not possible (even if eventually supported by lintr itself).

another option is to hack it such that if (any RUnit functions exempted) then "also exempt Test* functions

MEO265: We could also add a parameter "function_exclusion_prefixes" . Then one could fully emulate the functionality without adapting the linter too much to a package.

AshesITR: What about adding rudimentary globbing support to `except=`? It could detect `*` at the end of exceptions and convert those entries to prefix-exclusions.

MichaelChirico: Something along those lines could work. OTOH I worry about this PR + discussion getting too long.

WDYT about keeping a simple allowlist for this initial PR and handling extensions as a follow-up?

MichaelChirico:
> On the other hand, I would like to allow repeat, as this function would only give an implicit return in case of using a break, and we could explicitly test for it if necessary.

I see what you mean, but it's impossible to tell statically if the `return()` we find underneath `repeat { ... }` is valid. It could easily just be an early return that's only run in exceptional cases. A `repeat { ... }` with neither `return()` nor `break` anywhere "inside" could be a `repeat_linter()` violation IMO, but that's a separate issue.

I really think we need to treat `while (...) { ... }` the same as `repeat { ... }`. `repeat { ... }` is just a convenient way of writing `while (TRUE) { ... }` so treating them differently doesn't make sense to me.

tests/testthat/test-return_linter.R
MichaelChirico: I just think using `return()` inside `switch()` is bad practice in general. In this simple case it's easy enough to read, but more practically `switch()` statements are more complicated and it will be easy to lose track of what actually gets returned. Moving `return()` to the outside of the `switch()` seems the safest approach IMO. The downside is that it's somewhat confusing to parse `return(stop('...'))`, but I think practically speaking a `stop()` is only likely to show up in the catch-all final position.

R/return_linter.R
```diff
@@ -0,0 +1,206 @@
+#' Return linter
+#'
+#' This linter checks for explicit [return()] at the end of a function
```
MichaelChirico, line 3: the wording here implies a default of enforcing explicit returns. Maybe

> This linter examines terminal expressions in functions for calls to `return()`, which are discouraged by default.

R/return_linter.R
MichaelChirico: I like `return_calls` or `return_functions`.

R/return_linter.R
```diff
@@ -0,0 +1,206 @@
...
+#'   linters = return_linter(use_implicit_returns = FALSE)
+#' )
+#'
+#'
+#' @evalRd rd_tags("return_linter")
+#' @seealso
+#'  - [linters] for a complete list of linters available in lintr.
+#'  - <https://style.tidyverse.org/functions.html?q=return#return>
+#' @export
+return_linter <- function(
```
MichaelChirico, line 53: Use tidyverse indent style

https://style.tidyverse.org/functions.html#long-lines-1

MEO265:
> > On the other hand, I would like to allow repeat, as this function would only give an implicit return in case of using a break, and we could explicitly test for it if necessary.
> 
> I see what you mean, but it's impossible to tell statically if the `return()` we find underneath `repeat { ... }` is valid. It could easily just be an early return that's only run in exceptional cases. A `repeat { ... }` with neither `return()` nor `break` anywhere "inside" could be a `repeat_linter()` violation IMO, but that's a separate issue.
> 
> I really think we need to treat `while (...) { ... }` the same as `repeat { ... }`. `repeat { ... }` is just a convenient way of writing `while (TRUE) { ... }` so treating them differently doesn't make sense to me.

I think you may have misunderstood me.  I don't want to look for a "return", but rather for a "break" and mark the "repeat" when I find one.

If there is no "break" there are only two possibilities: 
Firstly, a return or a similar function is never made.  -> no implicit return.  
Second, such a function is hit -> explicit return.
Therefore it would be a false positive for me.

tests/testthat/test-return_linter.R
MEO265: I agree that this is bad style, but I would still like to prevent as many false positives as possible and I think that's part of it, isn't it?

MichaelChirico:
> > > On the other hand, I would like to allow repeat, as this function would only give an implicit return in case of using a break, and we could explicitly test for it if necessary.
> > 
> > 
> > I see what you mean, but it's impossible to tell statically if the `return()` we find underneath `repeat { ... }` is valid. It could easily just be an early return that's only run in exceptional cases. A `repeat { ... }` with neither `return()` nor `break` anywhere "inside" could be a `repeat_linter()` violation IMO, but that's a separate issue.
> > I really think we need to treat `while (...) { ... }` the same as `repeat { ... }`. `repeat { ... }` is just a convenient way of writing `while (TRUE) { ... }` so treating them differently doesn't make sense to me.
> 
> I think you may have misunderstood me. I don't want to look for a "return", but rather for a "break" and mark the "repeat" when I find one.
> 
> If there is no "break" there are only two possibilities: Firstly, a return or a similar function is never made. -> no implicit return. Second, such a function is hit -> explicit return. Therefore it would be a false positive for me.

I think `repeat_linter()` is better suited for linting whether `repeat` loops that have neither `break` nor `return`.

```r
function(x) {
  repeat {
    if (FALSE) return()
  }
}
```

this does have a return but it is a false negative IMO to let that pass as an explicit return. actually the XPath I shared initially would indeed not lint, but I think it's the wrong behavior. In explicit return world, the repeat loops should be followed by an explicit terminal behavior.

that may be up to preference, but in that case we should treat while identically.

tests/testthat/test-return_linter.R
MichaelChirico: that makes sense, though I don't think it's a priority to handle such a case in this PR. Let's mark it as a follow-up. Have you come across such cases?

MEO265: We can do it like that.  Then all you have to do is finalize the documentation and parameter names and remove Runit.  I'll take care of it late this evening.  If you want to have it through beforehand, you can of course do it too.

R/return_linter.R
AshesITR: #2335

MichaelChirico:
- Dropped RUnit support for now, to be handled in follow-up
- Renamed all three parameters
  + `use_implicit_returns` ➡️ `return_style`. This already reads better to me, and as hinted in #2321, might get more options in the future
  + `additional_allowed_func` ➡️ `return_functions`
  + `additional_side_effect_func` ➡️ `except`

tests/testthat/test-return_linter.R
MichaelChirico: Resolving in favor of handling in follow-up: #2343

R/return_linter.R
MichaelChirico: Handled in latest batch of commits, PTAL.

MichaelChirico:
> @MichaelChirico as I see it, you've already done everything or is there still something to do for me? Thanks anyway.

yes, I tried to clear the pending tasks. good for another round of review in case there's anything pending needed for initial merge.

MEO265:
> > @MichaelChirico as I see it, you've already done everything or is there still something to do for me? Thanks anyway.
> 
> yes, I tried to clear the pending tasks. good for another round of review in case there's anything pending needed for initial merge.

Okay, I don't have any more comments from my side

AshesITR:
Nice. I'll try to review later this day.

R/return_linter.R
```diff
@@ -0,0 +1,177 @@
...
+          not(OP-DOLLAR or OP-AT)
+          and SYMBOL_FUNCTION_CALL[text() = 'return']
+        ]
+      ]
+    "
+    msg <- "Use implicit return behavior; explicit return() is not needed."
+  } else {
+    # See `?.onAttach`; these functions are all exclusively used for their
+    #   side-effects, so implicit return is generally acceptable
+    default_except <- c(".onLoad", ".onUnload", ".onAttach", ".onDetach", ".Last.lib")
```
MichaelChirico, line 77: just noticed we could probably re-use this:

https://github.com/r-lib/lintr/blob/1c36e0d03b3f39a13f1a79d173548c2f75ec9bf9/R/object_name_linter.R#L186-L194

tests/testthat/default_linter_testcode.R
```diff
@@ -20,6 +25,7 @@ f = function (x,y = 1){}
 # object_name
 # object_usage
 # open_curly
+# return
```
AshesITR: This is inaccurate. `return_linter()` doesn't lint the following function.

AshesITR:
@MichaelChirico PTAL, I've fixed what I found directly on the branch.
I think we can merge now.

AshesITR (approved):

AshesITR (approved):

tests/testthat/default_linter_testcode.R
```diff
@@ -25,7 +25,6 @@ g <- function(x) {
 # object_name
 # object_usage
 # open_curly
-# return
```
MichaelChirico: oh! in some commit I added an explicit return here. Must not have survived. I see someone (possibly me?) added a `return_linter()` test case to this file in another commit.

tests/testthat/test-return_linter.R
```diff
@@ -245,29 +245,34 @@ test_that("return_linter ignores expressions that aren't functions", {
...
 
 test_that("return_linter ignores if statements outside of functions", {
-  lines <- c(
-    "if (TRUE) {",
-    "  TRUE",
-    "} else {",
-    "  FALSE",
-    "}"
+  expect_lint(
+    trim_some("
```
MichaelChirico, line 253: thanks for cleaning these up to use `trim_some()`!

MichaelChirico:
Great work everyone and thanks for the patience! 70 comments, that may be a record :)

PR #2272, opened 2023-11-10
New comparison_negation_linter

MichaelChirico:
Part of #884

No lints on our own package


AshesITR:
Out of curiosity:
Is there some S3 magic where the negation makes sense?
Thinking set containment relations that might be implemented via Ops.

MichaelChirico:
That's absolutely something that could happen in principal, but it's definitely a code smell if so.

In fact, C++20 introduced compiler-level changes to prevent this from becoming an issue -- if you write an == method, you get != for free because the compiler rewrites x!=y to !(x==y). They also added the spaceship operator for similar reasons re: >,≥,<,≤:

https://stackoverflow.com/a/60177760/3576984

If you wrote it this way because of dispatch issues in upstream code, #nolint would be the way to go.

AshesITR (approved):
Do we want to add it to our `.lintr_new`?

MichaelChirico:
> Do we want to add it to our `.lintr_new`?

We might want to switch to something like `modify_defaults(all_linters(), ...)` for our .lintr TBH...

PR #2273, opened 2023-11-10
New stopifnot_all_linter

MichaelChirico:
Part of #884


AshesITR:
WDYT about integrating it into `conjunct_test_linter()`? It detects a similar problem with `stopifnot()`.

MichaelChirico:
> WDYT about integrating it into `conjunct_test_linter()`? It detects a similar problem with `stopifnot()`.

I lean towards "no", since it's really a particular thing for `stopifnot()` (e.g. `testthat::expect_true()` nor `assertthat::assert_that()` use this behavior).

That said, we don't have to stretch the definition of "conjunct" too much to make it work -- `all(x)` being `x[1L] && ... && x[n]`...

AshesITR:
Metadata and vectorization tests would be nice.

MichaelChirico:
> vectorization tests

I think for linters built with `make_linter_from_xpath()`, that's overkill -- better to test `make_linter_from_xpath()` directly

AshesITR:
If the linter ever gets more complex functionality in the future, these tests will ensure that it still works?

MichaelChirico:
> If the linter ever gets more complex functionality in the future, these tests will ensure that it still works?

I think the next commit would be the one to owe new tests.

OTOH, it doesn't really hurt to add new tests. And more importantly, there _is_ a new test of the XPath logic here, namely that it's not somehow over- or under-counting lints. Not sure we've come across that being an issue in the past, but still justifies the new test IMO.

PR #2275, opened 2023-11-12
New nzchar_linter

MichaelChirico:
Part of #884

Some notes:

 - We could easily make the behavior we chose to skip linting `if (x == "")` optional
 - This linter may be somewhat controversial because of the difference in handling `NA_character_`. I really find the mismatch of defaults in R confusing here. We could consider a weaker version of this linter that only lints `nchar(x, keepNA = FALSE)`
 - No lints on {lintr}


R/nzchar_linter.R
```diff
@@ -0,0 +1,103 @@
...
+nzchar_linter <- function() {
+  # use string-length to capture both "" and ''
+  # if (any(x == "")) is not treated like it's part of if(), but
+  #   any(if (x == "") y else z) _is_ treated so. this condition looks for the
+  #   expr to be inside a call that's _not_ above an IF/WHILE.
+  comparison_xpath <- "
+  //STR_CONST[string-length(text()) = 2]
+    /parent::expr
+    /parent::expr[
+      (EQ or NE or LT or GT or LE or GE)
```
AshesITR, line 23: Why lint `<, >, <=, >=`?

R/nzchar_linter.R
```diff
@@ -0,0 +1,103 @@
...
+  "
+
+  # nchar(., type="width") not strictly compatible with nzchar
+  # unsure allowNA compatible, so allow it just in case (see TODO in tests)
+  nchar_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'nchar']
+    /parent::expr
+    /parent::expr
+    /parent::expr[
+      (EQ or NE or LT or GT or LE or GE)
```
AshesITR, line 48: Of these, only `>`, `==`, `!=` and `<=` (which is `==` in disguise) make sense. `<` is always false, `>=` is always true.

R/nzchar_linter.R
```diff
@@ -0,0 +1,103 @@
...
+
+    xml <- source_expression$xml_parsed_content
+
+    comparison_expr <- xml_find_all(xml, comparison_xpath)
+    comparison_lints <- xml_nodes_to_lints(
+      comparison_expr,
+      source_expression = source_expression,
+      lint_message = paste(
+        'Instead of comparing strings to "", use nzchar().',
+        'An exception is carved out for comparing scalar strings to "" in',
```
AshesITR, line 81: I wouldn't mention the exception in the lint message.

AshesITR (commented):
Missing metadata tests.

R/nzchar_linter.R
MichaelChirico: Right... feels fine to include those anyway. `x < ""` is a code smell I guess, seems OK to lint it here. Maybe with different lint metadata?

R/nzchar_linter.R
MichaelChirico: Happens very rarely but does seem like it's done. Maybe as a little code golf thing?

https://github.com/search?q=lang%3AR+%2F%5B%5E%23%3D-%5D%3E%5Cs*%22%22%2F&type=code

AshesITR: Interesting, there are even some hits for `<=`. Thanks for checking.

R/nzchar_linter.R
AshesITR: This one is `nchar(x) < 0` // `nchar(x) >= 0`

MichaelChirico: 😅 guessed wrong from the preview. Still feels harmless to include (and I can switch to use `infix_metadata`), WDYT?

R/expect_comparison_linter.R
```diff
@@ -49,7 +49,7 @@
 #' @export
 expect_comparison_linter <- function() {
   # != doesn't have a clean replacement
-  comparator_nodes <- setdiff(as.list(infix_metadata$xml_tag[infix_metadata$comparator]), "NE")
```
AshesITR, line 52: What's with this diff?

tests/testthat/test-nzchar_linter.R
```diff
@@ -0,0 +1,74 @@
...
+})
+
+test_that("multiple lints are generated correctly", {
+  expect_lint(
+    trim_some("{
+      a == ''
+      nchar(b) != 0
+    }"),
+    list(
+      rex::rex('Instead of comparing strings to ""'),
```
AshesITR, line 69: Add line number expectations here?

R/expect_comparison_linter.R
MichaelChirico: I think it's a vestige from before `xp_or()` supported vector input (585e7cd3e0e4a3aa8f78fdf0242a908ff4cfd65e0). And `setdiff(list(), "NE")` looks weird to me.

AshesITR (dismissed):

IndrajeetPatil:
@MichaelChirico This should be ready for a merge after the merge conflict is resolved.

IndrajeetPatil (approved):

PR #2276, opened 2023-11-12
New terminal_close_linter

MichaelChirico:
Part of #884 

We might consider renaming this to just `close_linter()`, to flexibility in the future for more lints associated with use of `close()`. E.g., we might make this stricter by default and require _all_ `close()` calls to be inside `on.exit()` hooks.


AshesITR:
Is this vectorized to detect multiple (trailing) `close()` calls? Should it be?

MichaelChirico:
> Is this vectorized to detect multiple (trailing) `close()` calls? Should it be?

Not now... can't say I've encountered that either. I would reserve for a follow-up if needed.

AshesITR:
All right. Can we add a metadata test here as well? Maybe with an example containing two close statements to show which one lints?

MichaelChirico:
> All right. Can we add a metadata test here as well? Maybe with an example containing two close statements to show which one lints?

SG. Any opinion on `terminal_close_linter()` vs. `close_linter()`?

AshesITR:
Hmm not sure where this might go. I can imagine a `connection_linter()` requiring `on.exit(close(...))` immediately after any connection-opening code unless the connection is returned by the function.

My gut feeling says let's keep the name for now and see to deprecating / renaming if and when the need arises.

AshesITR (approved):

PR #2277, opened 2023-11-12
Switch to cli style

### LINKED ISSUE #2220: Use {cli}, not {crayon}
https://github.com/r-lib/lintr/blob/87b59571f1430b937f4282344a9e3712866e21da/DESCRIPTION#L39

{crayon} is superseded by {cli}:

https://github.com/r-lib/crayon/blob/f11f6e66df40513ca07b241eb986e058f91fff33/README.md?plain=1#L2-L7

MichaelChirico:
Closes #2220 

@IndrajeetPatil do you know if we need to change these options in the workflows?

https://github.com/r-lib/lintr/blob/71c7e3163ac531721accede811299ab182364f72/.github/workflows/test-package-vigilant.yaml#L30


AshesITR:
> Closes #2220 
> 
> 
> 
> @IndrajeetPatil do you know if we need to change these options in the workflows?
> 
> 
> 
> https://github.com/r-lib/lintr/blob/71c7e3163ac531721accede811299ab182364f72/.github/workflows/test-package-vigilant.yaml#L30

https://github.com/r-lib/cli/blob/main/vignettes/cli-config-internal.Rmd

Says it's deprecated. `cli.num_colors = 8L` seems to be the preferred way currently, looking at the code of `num_ansi_colors()`.
It should work without adjustment for the moment, though.

MichaelChirico:
It looks like it's really supposed to "just work" with {cli}. I would guess GHA is one use case that's gotten a lot of attention too... let's try disabling the options for now.

IndrajeetPatil:
AFAICT, the `crayon.enabled = TRUE` is relevant only for printing the workflow output in color, which is what we want. So we don't need to change anything here for the workflows.

IndrajeetPatil (approved):
LGTM, thanks!

Issue #1908, opened 2023-02-01
comment_code_linter's false positives single quotes enclose strings on roxygen comments


eitsupi:
It seems that lintr complains about the first line of the following R file.

```r
#' foo <- '
#' bar'
#' foo <- "
#' bar"
```

I found this in apache/arrow#33968

IndrajeetPatil:
Thanks. Can reproduce:

``` r
library(lintr)

lint(
  text = "#' foo <- '
    #' bar'",
  linters = commented_code_linter()
)
#> <text>:1:2: style: [commented_code_linter] Commented code should be removed.
#> #' foo <- '
#>  ^~~~~~~~~~
```

<sup>Created on 2023-02-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico:
I think this is downstream of making lintr roxygen-aware. once that's done I imagine this issue will go away on its own.

MichaelChirico:
@AshesITR WDYT about completely excluding roxygen2 comments from `commented_code_linter()`? As simple as `//COMMENT[not(starts-with(text(), "#'"))]` instead of all `//COMMENT` nodes.

PR #2279, opened 2023-11-14
New character_only_linter

MichaelChirico:
Part of #884

Here I think the main question is whether this might be subsumed into `library_call_linter()`.


tests/testthat/test-character_only_linter.R
```diff
@@ -0,0 +1,92 @@
+test_that("character_only_linter skips allowed usages", {
+  linter <- character_only_linter()
+
+  expect_lint("library(data.table)", NULL, linter)
+  expect_lint("function(pkg) library(pkg, character.only = TRUE)", NULL, linter)
+  expect_lint("function(pkgs) sapply(pkgs, require, character.only = TRUE)", NULL, linter)
```
AshesITR, line 6: What makes this one different from the test in L 81?

MichaelChirico: it's inside a `function()`. This usage is exempted by default to allow for packags like {knitr} which may load packages on the user's behalf.

AshesITR: What about a similar bootstrapping code in "userland", like

```r
pkgs <- discover_dependencies()
sapply(pkgs, library, character.only = TRUE)
```
?

Do we want to block that?
(I'm fine with that since it's non-default, just asking)

MichaelChirico: Not 100% sure what you mean by "userland" here, as in, at the top-level of a script?

Yes, that's the explicit purpose here -- discouraging usage like

```r
script_deps <- c("dplyr", "lubridate", "plm")
sapply(script_deps, library, character.only = TRUE)
```

It's much better for code analysis/tooling to have a consistent way of writing the script's dependencies, having usage like this makes it a mess to find out which script attaches which packages at scale.

Possibly we could add some parameters to fine-tune behavior here, but I'd leave that for when we get user feedback.

AshesITR: Yeah, I was thinking about top-level scripts. I agree it should be blocked.

MichaelChirico:
NB: holding off on adding examples until we decide whether this should be subsumed into `library_call_linter()`.

AshesITR:
I'm in favor of subsuming - the linter name contains `library` that way. Someone unfamiliar with `character.only` will probably wonder what `character_only_linter()` does.

MichaelChirico:
> I'm in favor of subsuming - the linter name contains `library` that way. Someone unfamiliar with `character.only` will probably wonder what `character_only_linter()` does.

OK, done. Follow-up: Should either behavior be made optional with a parameter?

R/library_call_linter.R
```diff
@@ -1,49 +1,66 @@
...
 #' lint(
-#'   text = "
-#'     library(dplyr)
-#'     print('test')
-#'   ",
+#'   text = code,
 #'   linters = library_call_linter()
 #' )
 #'
+#' code <- '# comment\nlibrary(dplyr)'
```
MichaelChirico, line 46: a very annoying false positive for `commented_code_linter()` here...

MichaelChirico: IINM this is #1908

AshesITR: Does using double quotes help?

MichaelChirico: 🤦 yes, thanks for the extra eyes. Should be using `"` there anyway. But there is an easy fix for `commented_code_linter()` for such cases, added a comment on #1908.

AshesITR:
For now, I wouldn't create an option unless we get a feature request. The intention of the two lints appears to be the same - getting a clean `library()` call list at the top of scripts.

NEWS.md
```diff
@@ -7,6 +7,7 @@
 ## New and improved features
 
 * More helpful errors for invalid configs (#2253, @MichaelChirico).
+* `library_call_linter()` is extended to encouraging all packages to be attached with `library(symbol)`, not `library("symbol", character.only = TRUE)` or "vectorized" approaches looping over package names (part of #884, @MichaelChirico).
```
IndrajeetPatil, line 10:
```suggestion
* `library_call_linter()` is extended to encourage all packages to be attached with `library(symbol)`, not `library("symbol", character.only = TRUE)` or "vectorized" approaches looping over package names (part of #884, @MichaelChirico).
```

R/library_call_linter.R
```diff
@@ -1,46 +1,63 @@
 #' Library call linter
 #'
-#' Force library calls to all be at the top of the script.
+#' This linter covers several rules related to [library()] calls:
+#'
+#'  - Enforce such calls to all be at the top of the script.
+#'  - Block usage of argument `character.only`.
```
IndrajeetPatil, line 6: Maybe also add rule about making library calls directly over vectorized calls?

MichaelChirico: Thanks. Will revisit once this is merged to `library_call_linter()`

IndrajeetPatil (dismissed):

R/library_call_linter.R
MichaelChirico: Oops, memory failed me :)

IndrajeetPatil (approved):

PR #2281, opened 2023-11-14
New which_grepl_linter

MichaelChirico:
Part of #884

No hits on {lintr} itself.


tests/testthat/test-which_grepl_linter.R
```diff
@@ -0,0 +1,27 @@
+test_that("which_grepl_linter skips allowed usages", {
+  # this _could_ be combined as p1|p2, but often it's cleaner to read this way
```
AshesITR, line 2: Actually, with conjunction, it's not even doable since matches might overlap. The comment would be correct for alternation (OR) of the `grepl()` results.

MichaelChirico: good catch

AshesITR:
LGTM. What do you think about introducing a `regex_linter()` that can support multiple lints, e.g. from `fixed_regex_linter()`?
It would allow future regex related lints to be automatically added to projects that opt into `regex_linter()`, reducing maintenance burden on `.lintr`.

MichaelChirico:
There's two avenues we've used for this, one is to have a single linter that handles a lot of things (`paste_linter()` is quintessential here), the other is using `tags`. I prefer the latter in most cases -- adding lots of parameters for turning certain lints on/off within a linter is somewhat painful for maintenance.

AshesITR:
SGTM. Should we add a tag `"regex"`, then?

R/linter_tag_docs.R
```diff
@@ -111,3 +111,11 @@ NULL
 #'  - <https://testthat.r-lib.org>
 #'  - <https://r-pkgs.org/testing-basics.html>
 NULL
+
+#' Regular expression linters
+#' @name regex_linters
+#' @description
+#' Linters that examine the usage of regular expressions in user code.
```
AshesITR, line 118: Maybe add the injunction ... and functions working with them ...?

AshesITR (approved):

AshesITR (approved):

IndrajeetPatil:
@MichaelChirico This is also ready for a merge after the merge conflict is resolved.

IndrajeetPatil (approved):

PR #2283, opened 2023-11-14
specific handling for 1:.N --> .I in seq_linter()

### LINKED ISSUE #2101: `seq_linter()` should recommend `.I` for `1:.N`
BTW if this is data.table code, it should be `rev(.I)`.
Not correct in general, though, so the current message is fine I think.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2099#discussion_r1310816164_

MichaelChirico:
Closes #2101

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 ## New and improved features
 
 * More helpful errors for invalid configs (#2253, @MichaelChirico).
+* Lint message for using `.N` from {data.table} like `1:.N` mentions the {data.table}-specific `.I` as an alternative (#2101, @MichaelChirico).
 
 ### New linters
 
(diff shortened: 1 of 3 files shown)


AshesITR:
Careful: `.I` is not `seq_len(.N)` in the presence of `by`:

``` r
library(data.table)
df <- data.table(grp = rep(c("a", "b"), each = 2L), val = rep(c(1L, 2L), times = 2L))
df[, rx := .I, by = .(grp)][rx > 1, val := 0L]
df
#>    grp val rx
#> 1:   a   1  1
#> 2:   a   0  2
#> 3:   b   0  3
#> 4:   b   0  4
df <- data.table(grp = rep(c("a", "b"), each = 2L), val = rep(c(1L, 2L), times = 2L))
df[, rx := seq_len(.N), by = .(grp)][rx > 1, val := 0L]
df
#>    grp val rx
#> 1:   a   1  1
#> 2:   a   0  2
#> 3:   b   1  1
#> 4:   b   0  2
```

<sup>Created on 2023-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #2284, opened 2023-11-14
Fully deprecate anything marked deprecated since 3.0.0

MichaelChirico:
Note that we can't use `lintr_deprecated(signal = "stop")` for the removed linters (e.g. `closed_curly_linter()`) because we test that _all_ linters (under `linters_with_tags(exclude_tags = NULL)`) can be instantiated. So we have to make a clean removal already.


tests/testthat/test-with.R
```diff
@@ -68,25 +68,16 @@ test_that("can instantiate all linters without arguments", {
...
   )
-  expect_identical(defaults, old_defaults)
 
-  # linters_with_defaults only accepts `defaults = list()` to start from blank
-  defaults <- linters_with_defaults(defaults = list(), whitespace_linter())
-  expect_warning(
-    {
-      old_defaults <- with_defaults(default = NULL, whitespace_linter())
-    },
+  expect_error(
```
AshesITR, line 77: Seems redundant to test twice for the same error.

tests/testthat/test-lint.R
```diff
@@ -151,7 +151,7 @@ test_that("exclusions work with custom linter names", {
 })
 
 test_that("compatibility warnings work", {
```
AshesITR, line 153: "warnings" isn't accurate anymore.

tests/testthat/test-with.R
MichaelChirico: I suppose in principal it could be that non-default arguments _don't_ throw the deprecation error. But no sense overanalyzing deprecated code that will soon be removed.

AshesITR:
Re removal of deprecated linters: Can't we rewrite the test somehow to allow for hard deprecations? Maybe a special tag "defunct", or logic that tests missing linters for hard-deprecation and allows them not to appear in linters.csv?
Seems a bit harsh to skip from warning to removal.

MichaelChirico:
> Maybe a special tag "defunct"

That SGTM.

R/linter_tags.R
```diff
@@ -58,7 +58,8 @@ available_linters <- function(packages = "lintr", tags = NULL, exclude_tags = "d
   }
 
   # any tags specified explicitly will not be excluded (#1959)
-  exclude_tags <- setdiff(exclude_tags, tags)
+  # never include defunct linters, which don't work / error on instantiation (#2284).
```
AshesITR, line 61: This undocumented behavior might surprise. Why not add it to the default `exclude_tags` and let users query which linters are defunct?

R/lintr-deprecated.R
```diff
@@ -5,11 +5,63 @@
 #'
 #' These functions have been deprecated from lintr.
 #'
+#' - `open_curly_linter()` and `closed_curly_linter()` check that open and closed curly braces
```
AshesITR, line 8: These descriptions no longer make sense, because the linters described are defunct.

tests/testthat/test-linter_tags.R
```diff
@@ -79,11 +79,13 @@ test_that("warnings occur only for deprecated linters", {
 
 test_that("available_linters matches the set of linters available from lintr", {
   lintr_db <- available_linters(exclude_tags = NULL)
-  linters_in_namespace <- setdiff(ls(asNamespace("lintr"), pattern = "_linter$"), "is_linter")
+  defunct_linters <- subset(read.csv(system.file("lintr", "linters.csv", package = "lintr")), tags == "defunct")$linter
```
AshesITR: `lintr_db$linter[vapply(lintr_db$tags, `%in%`, logical(1L), y = "defunct")]`
avoids re-reading the CSV.

R/linter_tags.R
MichaelChirico: That will break the workflow `available_linters(tags = NULL)`. I'll just document it instead.

AshesITR: Which one? Our own test?

AshesITR:
CRAN is down ATM 👀

R/linter_tags.R
MichaelChirico: It's really not worth spending more effort supporting a way to access defunct linters. They're defunct.

R/linter_tags.R
```diff
@@ -4,11 +4,12 @@
...
 #' @param packages A character vector of packages to search for linters.
 #' @param tags Optional character vector of tags to search. Only linters with at least one matching tag will be
-#' returned. If `tags` is `NULL`, all linters will be returned. See `available_tags("lintr")` to find out what
-#' tags are already used by lintr.
+#'   returned. If `tags` is `NULL`, all linters will be returned. See `available_tags("lintr")` to find out what
+#'   tags are already used by lintr.
 #' @param exclude_tags Tags to exclude from the results. Linters with at least one matching tag will not be returned.
-#' If `except_tags` is `NULL`, no linters will be excluded. Note that `tags` takes priority, meaning that any
-#' tag found in both `tags` and `exclude_tags` will be included, not excluded.
+#'   If `except_tags` is `NULL`, no linters will be excluded. Note that `tags` takes priority, meaning that any
```
AshesITR, line 10: Old Typo: `exclude_tags`, not `except_tags`.

R/lintr-deprecated.R
```diff
@@ -5,23 +5,12 @@
 #'
 #' These functions have been deprecated from lintr.
 #'
-#' - `open_curly_linter()` and `closed_curly_linter()` check that open and closed curly braces
-#'   are on their own line unless they follow an else, a comma, or a closing bracket.
-#'   Deprecated in favor of [brace_linter()].
+#' - `open_curly_linter()` (use [brace_linter()])
+#' - `closed_curly_linter()` (use `brace_linter`)
```
AshesITR, line 9: consistency in linking to `[brace_linter()]`?

MichaelChirico: I think it's normal practice to only link a reference doc the first time it's used. Helps avoid link overload. That's how e.g. Wikipedia does it: https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#MOS:REPEATLINK

AshesITR: Then it's still missing `()`.

MichaelChirico:
> CRAN is down ATM 👀

Noting here as well my ref from #2289 for completeness:

---

Yes, CRAN is undergoing scheduled maintenance:

https://bsky.app/profile/eddelbuettel.com/post/3keb4kxyzpj2v

AshesITR (approved):

IndrajeetPatil (approved):

PR #2286, opened 2023-11-15
New rep_len_linter

MichaelChirico:
Part of #884 

No hits on {lintr}


R/rep_len_linter.R
```diff
@@ -0,0 +1,24 @@
+#' Require usage of rep_len(x, n) over rep(x, length.out = n)
+#'
+#' `rep(x, length.out = n)` calls `rep_len(x, n)` "under the hood". The latter
+#'   is thus more direct and equally readable.
+#'
+#' @evalRd rd_tags("rep_len_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+rep_len_linter <- make_linter_from_xpath(
+  # match starts-with(text(), 'len') to get common partial matching usage.
```
AshesITR, line 10: This is not implemented, right?

R/rep_len_linter.R
```diff
@@ -0,0 +1,24 @@
+#' Require usage of rep_len(x, n) over rep(x, length.out = n)
+#'
+#' `rep(x, length.out = n)` calls `rep_len(x, n)` "under the hood". The latter
+#'   is thus more direct and equally readable.
+#'
```
AshesITR, line 5: examples are missing

R/rep_len_linter.R
MichaelChirico: Oh, good catch. I wonder what happened... maybe I didn't get any new matches from using `len`? Do you think it's worth implementing?

AshesITR: Not really. I remember we have an issue lying around that lints all partial argument matching. IINM that could even be default because the tidyverse style guide discourages it.

R/rep_len_linter.R
```diff
@@ -0,0 +1,36 @@
...
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "rep(1:3, length.out = 10)",
+#'   linters = rep_len_linter()
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "rep(1:3, each = 2L, length.out = 10L)",
```
IndrajeetPatil, line 15: Shouldn't the okay example instead showcase how to do the same using `rep_len()`?

MichaelChirico: I still don't have a clear structure in my head for what belongs under `# okay` examples. Is it examples paralleling the "bad" examples in the previous section? examples of code using the same functions but in "acceptable" ways?

it seems to get "parallelism" right in general since there's often not a 1:1 correspondence between good/bad sections.

anyway I added an example with `rep_len()`.

IndrajeetPatil: Yes, thus far we have followed the parallelism pattern where we return the same version of the code in a way that will or will not produce a lint.

AshesITR:
Vectorizatiion & metadata tests are missing. The rest looks fine.

tests/testthat/test-rep_len_linter.R
```diff
@@ -0,0 +1,54 @@
...
+      rep(x, each = 4, length.out = 50)
+      rep(x, length.out = 50)
+    }"),
+    list(
+      list(lint_msg, line_number = 3L),
+      list(lint_msg, line_number = 5L)
+    ),
+    rep_len_linter()
+  )
+})
```
AshesITR, line 54: Missing terminal newline

AshesITR (dismissed):

AshesITR (approved):

PR #2288, opened 2023-11-15
Ensure examples in linter docs

MichaelChirico:
Thanks @IndrajeetPatil for catching the mistake of `sample_int_linter()` lacking examples. I added those, and looked for other doc pages lacking `@examples` -- a few more slipped through as well.

Therefore I added a regression test to guard against this going forward.

### DIFF
--- a/R/comparison_negation_linter.R
+++ b/R/comparison_negation_linter.R
@@ -3,6 +3,34 @@
 #' `!(x == y)` is more readably expressed as `x != y`. The same is true of
 #'   other negations of simple comparisons like `!(x > y)` and `!(x <= y)`.
 #'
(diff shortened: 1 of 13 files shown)


AshesITR (approved):

IndrajeetPatil (commented):
Sorry, again, for reviewing after the PR has already been merged, but I can't do any open-source during work hours.

Thanks for adding the examples and the tests!

One thing I noticed is that we are not following the pattern we were following in all existing examples:

```r
# will produce lints
example_with_lints

# okay
the_same_example_without_lints
```

The examples here don't have the same one-to-one correspondence between with and without lint code examples.

PR #2289, opened 2023-11-16
more vestigial code from allowing R<3.6

MichaelChirico:


IndrajeetPatil (approved):
AFAICT, the failing URL check workflow is a false positive because CRAN is down.

MichaelChirico:
Yes, CRAN is undergoing scheduled maintenance:

https://bsky.app/profile/eddelbuettel.com/post/3keb4kxyzpj2v

PR #2294, opened 2023-11-16
New one_call_pipe_linter

MichaelChirico:
Part of #884

This one's slightly odd because we ended up disabling this linter for ourselves -- linting code like:

```r
x |>
  filter(a > 1)
```

Was just found to be too disruptive to normal {dplyr} workflows & we didn't invest in figuring out how to strike a balance between cases like that where it's OK and probably more readable with `|>` and other cases like `(1:10) |> sum()` where the pipeless version is clearly superior.


R/one_call_pipe_linter.R
```diff
@@ -0,0 +1,63 @@
...
+#' @export
+one_call_pipe_linter <- function() {
+  pipes_cond <- xp_text_in_table(c("%>%", "%$%", "%T>%"))
+
+  # preceding-sibling::SPECIAL: if there are ever two pipes, don't lint
+  # OP-LEFT-BRACKET/LBB: accept DT[...] %>% .[...] as a two-call pipe,
+  #   (but not DT %>% .[...])
+  # parent::expr/SPECIAL: make sure we are at the top of a pipeline
+  # count(): any call anywhere else in the AST within the pipe expression
+  # TODO(michaelchirico): Add support for native pipe |> like DT |> _[...]
```
AshesITR, line 24: Is this resolved?

R/one_call_pipe_linter.R
```diff
@@ -0,0 +1,63 @@
...
+#'   calls in the first `%>%` argument count. `rowSums(x) %>% max()` is OK
+#'   because there are two total calls (`rowSums()` and `max()`).
+#'
+#' Note also that un-"called" steps are *not* counted, since they should
+#'   be calls (see [pipe_call_linter()]).
+#'
+#' @evalRd rd_tags("one_call_pipe_linter")
+#' @seealso
+#' - [linters] for a complete list of linters available in lintr.
+#' - <https://style.tidyverse.org/pipes.html#short-pipes>
```
AshesITR, line 14: Means we should make it a default linter, no?

R/one_call_pipe_linter.R
MichaelChirico: Not yet, and I don't have a machine with a recent enough R installed to test a fix at the moment.

We can either leave for follow-up or pend this PR until I get to my personal machine.

R/one_call_pipe_linter.R
MichaelChirico: Not totally clear... internally our discussion put a lot of weight on **consider** here

> A one-step pipe can stay on one line, but unless you plan to expand it later on, you should **consider** rewriting it to a regular function call.

So there's some ambiguity in the style guide on whether to _always_ enforce this, or even if it should be removed _most_ of the time. We found the `# nolint` workaround too disruptive because there's so many false positives.

OTOH our rules are not identical to the Tidyverse rules, and we are free to continue disabling it. So {lintr}'s decision should ultimately be independent.

cc @IndrajeetPatil here.

AshesITR:
Metadata tests missing

R/one_call_pipe_linter.R
AshesITR: Okay let's keep it non-default then.

R/one_call_pipe_linter.R
```diff
@@ -0,0 +1,78 @@
...
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "rowSums(x) %>% mean()",
+#'   linters = one_call_pipe_linter()
+#' )
+#'
+#' lint(
+#'   text = "DT[src == 'a', .N, by = grp] %>% .[N > 10]",
```
AshesITR, line 30: Seems like a bad example because the pipe is unnecessary alltogether.

inst/lintr/linters.csv
```diff
@@ -61,7 +61,12 @@ numeric_leading_zero_linter,style consistency readability
 object_length_linter,style readability default configurable executing
 object_name_linter,style consistency default configurable executing
 object_usage_linter,style readability correctness default executing configurable
+<<<<<<< HEAD
```
AshesITR, line 64: Merge fail

R/one_call_pipe_linter.R
MichaelChirico: more realistically the second step would come on the next line, but such usage is not unheard of, e.g.

https://github.com/TheEconomist/big-mac-data/blob/54c2663f5237c9bdbd34c99cf1ec4f87062b6469/data-generator.R#L18

AshesITR: I would put `][` on a new line, but fine.

MichaelChirico: I've found [] chaining is not v friendly with tidy syntax (or styler at least), and mixing data.table steps with tidyverse (e.g. dplyr, tidyr, or ggplot2) steps can be more natural to use this form of chaining.

tests/testthat/test-one_call_pipe_linter.R
```diff
@@ -0,0 +1,122 @@
...
+
+test_that("multiple lints are generated correctly", {
+  expect_lint(
+    trim_some("{
+      a %>% b()
+      c %$% d()
+      e %T>% f()
+    }"),
+    list(
+      list(rex::rex("pipe %>%"), line_number = 2L),
```
MichaelChirico, line 71: @AshesITR which other metadata tests would you like?

R/one_call_pipe_linter.R
```diff
@@ -0,0 +1,78 @@
...
+#'   linters = one_call_pipe_linter()
+#' )
+#'
+#' @evalRd rd_tags("one_call_pipe_linter")
+#' @seealso
+#' - [linters] for a complete list of linters available in lintr.
+#' - <https://style.tidyverse.org/pipes.html#short-pipes>
+#' @export
+one_call_pipe_linter <- function() {
+  pipes_cond <- xp_text_in_table(c("%>%", "%$%", "%T>%"))
```
AshesITR, line 40: reuse magrittr pipes?

tests/testthat/test-one_call_pipe_linter.R
AshesITR: Maybe a multi-line case to show where the lint is placed?

AshesITR (commented):
Almost good to merge.

AshesITR (dismissed):

AshesITR (approved):

Issue #2295, opened 2023-11-17
Find redundant GHA and remove them


MichaelChirico:
Lately I've found there's a lot of lethargy in our review process because of queueing in the GHA for the repo.

My suspicion is that many of the actions we run are totally redundant -- they provide signal that is completely identical to at least one other action. That is, they _always_ share failure status with some other action. Such actions can be removed.

I will be extending this script to analyze this numerically:

https://gist.github.com/MichaelChirico/d5a1a0d864864e0d30f24e3eb87afecf

IndrajeetPatil:
In my view, we can remove:

- `R-CMD-check / ubuntu-latest (devel) zh_CN` (this workflow is actually quite flaky and often fails unpredictably)
- `test-package` (redundant if we run `R-CMD-check` with `options(warn = 2)`)
- `R-CMD-check / windows-latest (4.1)` (if this is only for checking with a newer Rtools version, isn't that also done via R-release?)

MichaelChirico:
Hmm GHA API is doing something weird:

```r
runs[, sum(conclusion == "failure"), by = name]
#                          name  V1
# 1:             check-link-rot   0
# 2:                    pkgdown   0
# 3:                       lint   0
# 4:                R-CMD-check 346
# 5:         check-all-examples   0
# 6:               test-package 173
# 7:           R-CMD-check-hard   0
# 8:              test-coverage   0
# 9: pages build and deployment   0
```

I'm fairly confident some of those other actions have failed! Especially we saw check-lint-rot fail when CRAN was down this week.

AshesITR:
The zh_CN run was added to test a non-UTF 8 locale because of some encoding related bug I think.

MichaelChirico:
Yea, I wouldn't remove `zh_CN` -- it has a clear purpose marking it distinct from the other runs. Flakiness can be addressed separately.

IndrajeetPatil:
Gabor had checked this with R-devel after discussion in https://github.com/r-lib/actions/issues/628/, and they had also mentioned not to use the non-UTF8 locale. 

If we choose to retain this, we shouldn't be running this workflow on R-devel, but maybe on R-oldrel. Otherwise, it's just going to fail sporadically.

AshesITR:
Maybe switch to `r-release`?
Dropping non UTF-8 testing seems like a bad trade. We are also working with user files that might have any encoding, so we should test that it's handled correctly.

MichaelChirico:
[...]

PR #2297, opened 2023-11-17
Check if parallel testing reduces check time in CI

IndrajeetPatil:
Related to #2295

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,6 +54,7 @@ VignetteBuilder:
     knitr
 Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
+Config/testthat/parallel: true
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3


IndrajeetPatil:
On `main` for macOS and R-release:

```r
* checking tests ...
  Running ‘testthat.R’ [67s/69s]
 [68s/69s] OK
```

In this PR for the same config:

```r
* checking tests ...
  Running ‘testthat.R’ [85s/46s]
 [85s/46s] OK
```

I don't understand why there are two timings, and which one we should care about.

The `test-package` workflows gives the exact duration, but that workflow fails here because `options(warn = 2)` seems to mess with the parallel testing.

MichaelChirico:
Part of the issue is that I think there's a maximum number of jobs that can be running at once -- speeding up individual jobs helps some, but I'd really like to reduce the number of jobs if some jobs are redundant.

IndrajeetPatil:
Agreed. But what do you think about turning on the parallel testing for `{lintr}`, in general? 
Do you see any downsides to it, apart from the failing workflow, which is likely to be a `{testthat}` bug?

MichaelChirico:
if we could have it run in serial by default and parallel according to environment variable on GHA only (not CRAN) I think that's preferable. parallelism is often a trojan horse for headaches if we don't get a huge benefit

IndrajeetPatil:
Will setting something like this in `testthat.R` work (parallel on for non-CRAN, but off for CRAN)?

```r
if (Sys.getenv("NOT_CRAN", "true") == "true") {
  Sys.setenv("TESTTHAT_PARALLEL" = "FALSE")
}
```

IndrajeetPatil:
Filed an issue upstream about the error we see in the vigilant workflow: https://github.com/r-lib/testthat/issues/1912

AshesITR:
> On `main` for macOS and R-release:
> 
> 
> 
> ```r
> 
> * checking tests ...
> 
>   Running ‘testthat.R’ [67s/69s]
> 
>  [68s/69s] OK
> 
> ```
> 
> 
> 
> In this PR for the same config:
> 
> 
> 
> ```r
> 
> * checking tests ...
> 
>   Running ‘testthat.R’ [85s/46s]
> 
>  [85s/46s] OK
> 
> ```
> 
> 
> 
> I don't understand why there are two timings, and which one we should care about.

I suspect the first is CPU time and the second is elapsed (wall) time.
So we pay 6s more computational cost for 23s time savings. Seems worth it.

IndrajeetPatil:
The reduction in test time from parallel testing has reduced significantly since the last time we checked; probably because our test suite itself has somehow got faster 👀  

- for main

```r
* checking tests ...
  Running ‘testthat.R’ [28s/29s]
 [29s/29s] OK
```

- for PR

```r
* checking tests ...
  Running ‘testthat.R’ [49s/24s]
 [49s/24s] OK
```

I am no longer sure if it is still worth it to turn on parallel testing, or if we'd wait for our test suite to get even bigger before we could reap the benefits of parallel testing.

IndrajeetPatil:
@MichaelChirico, @AshesITR WDYT? 

Should we keep this PR open for a few more months or a year, rebase once in a while, and check for time saving benefits? 
Or should we just kill it now and revisit only when we are faced with slow tests because of the growth in test suite?

AshesITR:
I'm fine with either way. It looks like most of the work is done already and rebasing from time to time isn't difficult.

What do you think is best?

IndrajeetPatil:
> What do you think is best?

I think it should be fine (and costless) to keep this pending for a while.

IndrajeetPatil:
I can reproduce this locally as well, so looks like a legit issue. 
Will have a deeper look into why this shows up only when tests are run in parallel.

```r
── Error (test-linter_tags.R:106:3): lintr help files are up to date ───────────
Error in `tools::Rd_db("lintr")`: installed help of package 'lintr' is corrupt
Backtrace:
    ▆
 1. └─tools::Rd_db("lintr") at test-linter_tags.R:106:3
 ```

PR #2299, opened 2023-11-17
New pipe_return_linter

MichaelChirico:
Part of #884

No hits on {lintr}.


R/pipe_return_linter.R
```diff
@@ -0,0 +1,40 @@
...
+  #   to lint in valid R code.
+  xpath = "
+  //SPECIAL[text() = '%>%']
+    /following-sibling::expr[expr/SYMBOL_FUNCTION_CALL[text() = 'return']]
+  ",
+  lint_message = paste(
+    "Using return() as the final step of a magrittr pipeline",
+    "is an anti-pattern. Instead, assign the output of the pipeline to",
+    "a well-named object and return that.",
+    "See go/r-primer#how-magrittr-pipes-work."
```
IndrajeetPatil, line 38: What is this linking to? Is it internal to Google?

tests/testthat/test-pipe_return_linter.R
```diff
@@ -0,0 +1,47 @@
+test_that("pipe_return_linter skips allowed usages", {
```
IndrajeetPatil, line 1: Maybe use `trim_some()` instead of passing a vector of lines of code?

R/pipe_return_linter.R
MichaelChirico: Oh, thanks!

tests/testthat/test-pipe_return_linter.R
MichaelChirico: I was editing in nano :sweat_smile: 

much easier from VScode

tests/testthat/test-pipe_return_linter.R
```diff
@@ -0,0 +1,47 @@
...
+        sapply(function(xi) {
+          return(xi ** 2)
+        })
+      return(x_squared)
+    }
+  ")
+  expect_lint(nested_return_lines, NULL, linter)
+})
+
+test_that("pipe_return_linter blocks simple disallowed usages", {
```
IndrajeetPatil, line 33: I also checked this for lambdas and it works as expected. But I don't think we need to add tests for that.

IndrajeetPatil (approved):

tests/testthat/test-pipe_return_linter.R
MichaelChirico: what do you mean by lambdas in this case? `|>`? see the comment in the R/ script -- native pipe is excluded because using return() there is a parser error

IndrajeetPatil: No, I meant with purrr's mappers :)

PR #2301, opened 2023-11-17
New nested_pipe_linter

MichaelChirico:
Part of #884

No hits on {lintr}


AshesITR:
This one I would carve out an exception at least for `select()`.

MichaelChirico:
> This one I would carve out an exception at least for `select()`. But feel free to ignore that.
> 
> Tests for metadata are missing.

Equally common I think is `DF1 |> inner_join(DF2 |> filter(...))`.

For `select()` the issue becomes cases like:

```r
DF1 |>
  inner_join(
    DF2 |>
      select(
        a, b, c, d,
        e, f, g, h, i
      )
  )
```

This was part of my motivation for the whole linter to begin with -- it becomes quite hard to tell at a glance how `|>` will parse (is it part of the `DF1 |>` pipeline... no... where does that nested pipeline begin exactly?).

For `select()`, the advantage of assigning an intermediate variable is not as clear as for `filter()`. so I see why that's a pretty good default.

But maybe we should add logic around whether the "nested" pipeline takes up >1 line?

AshesITR:
`> 1 line` looks like a good heuristic.
Maybe better than exceptions for some functions.

MichaelChirico:
> `> 1 line` looks like a good heuristic. Maybe better than exceptions for some functions.

SGTM. I am thinking to make this a parameter, `allow_inline = TRUE` by default. `FALSE` is the current logic, `TRUE` only hits when the "inner" pipeline spans >1 line

R/nested_pipe_linter.R
```diff
@@ -0,0 +1,73 @@
...
+#'
+#' @evalRd rd_tags("nested_pipe_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+nested_pipe_linter <- function(allow_inline = TRUE) {
+  multiline_and <- if (allow_inline) "@line1 != @line2 and" else ""
+  xpath <- glue("
+  (//PIPE | //SPECIAL[{ xp_text_in_table(magrittr_pipes) }])
+    /parent::expr[{multiline_and} preceding-sibling::expr/SYMBOL_FUNCTION_CALL[
+      not(text() = 'try' or text() = 'tryCatch')
```
AshesITR, line 47: Add `withCallingHandlers()` and maybe make this configurable?

MichaelChirico: Hmm I think I was mistaken about this case, `tryCatch(x %>% y(), ...)` is identical to `x %>% y() %>% tryCatch(...)`:

```r
foo <- function(x) stop("hi")
identical(
  tryCatch(1 %>% foo(), error = identity),
  1 %>% foo() %>% tryCatch(error = identity)
)
# [1] TRUE
```

So actually we excluded it as a readability preference. I'll make it configurable and include `withCallingHandlers()` in the default.

AshesITR (approved):

PR #2302, opened 2023-11-17
New unnecessary_nesting_linter

MichaelChirico:
Part of #884


R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,168 @@
...
+#' @examples
+#' # will produce lints
+#' code <- "if (A) {\n  stop('A is bad!')\n} else {\n  do_good()\n}"
+#' writeLines(code)
+#' lint(
+#'   text = code,
+#'   linters = unnecessary_nesting_linter()
+#' )
+#'
+#' code <- "tryCatch(\n  {\n    foo()\n  },\n  error = identity\n)"
```
AshesITR, line 15: Will `tryCatch({\n  foo()\n}, error = identity)` lint?

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,168 @@
...
+  #   allowing for cleanly parallel code, where breaking it would often harm readability:
+  #   if (A) {
+  #     stop()
+  #   } else {
+  #     warning()
+  #   }
+  # NB: print() is intentionally excluded since its usage is usually a mistake (?print_linter)
+  signal_calls <- c(
+    exit_calls,
+    "warning", "warn", "message", "cat", "LOG", "stopifnot"
```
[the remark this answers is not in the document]
MichaelChirico: yep... left it in here as a reminder to discuss what sort of parameterization we should build in here, WDYT?

`allow_signal_calls = c("warning", "message", "cat", "stopifnot")`?

R/unnecessary_nesting_linter.R
MichaelChirico: yes, I'll add a test:

```r
lint("tryCatch({\n  foo()\n}, error = identity)", unnecessary_nesting_linter())
# <text>:1:10: warning: [unnecessary_nesting_linter] Reduce the nesting of this statement by removing the braces {}.
# tryCatch({
#          ^
```

MichaelChirico: btw, styler will adjust that code with output:

```
tryCatch(
  {
    foo()
  },
  error = identity
)
```

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,168 @@
...
+
+    if_else_exit_expr <- xml_find_all(xml, if_else_exit_xpath)
+    # TODO(michaelchirico): customize the error message to the exit clause used
+    if_else_exit_lints <- xml_nodes_to_lints(
+      if_else_exit_expr,
+      source_expression = source_expression,
+      lint_message = paste0(
+        "Reduce the nesting of this if/else statement by unnesting the ",
+        "portion without an exit clause (i.e., ",
+        paste0(exit_calls, "()", collapse = ", "),
```
AshesITR, line 152: Can we dynamically extract the correct call?

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,168 @@
...
+        or self::WHILE
+        or self::REPEAT
+        or self::expr/SYMBOL_FUNCTION_CALL[text() = 'switch']
+        or self::expr/expr/SYMBOL_FUNCTION_CALL[text() = 'foreach']
+        or self::OP-TILDE
+        or self::LEFT_ASSIGN[text() = ':=']
+      ])
+      and not(expr/OP-LEFT-BRACE)
+      and not(preceding-sibling::OP-LEFT-BRACE)
+      and not(
```
AshesITR, line 130: Add a similar carve-out for `OP-LEFT-BRACE`?

R/unnecessary_nesting_linter.R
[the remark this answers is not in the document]
AshesITR: I'm not 100% sure yet.

R/unnecessary_nesting_linter.R
MichaelChirico: I think that's the TODO just above. Marking for follow-up as this PR is already on the big side.

MichaelChirico: #2316

R/unnecessary_nesting_linter.R
MichaelChirico: do you have an example in mind? that will help for testing

AshesITR:
```r
tryCatch({
  main()
}, error = identity, warning = identity)
```

MichaelChirico: this should not lint you mean? i.e. using either `({` or `})` will exempt the expression (regardless of what `brace_linter()` says)

AshesITR: Yes. What does brace_linter say? I don't get lints for my example.

AshesITR:
> I think the path forward is to carve out an exception when the only expression inside `{` is an assignment. Shall we mark that with a parameter to `unnecessary_nesting_linter()`?

SGTM. Should default to not linting (so `all_linters()` doesn't create a conflicting configuration).
Argument `allow_assignment = TRUE`?

R/unnecessary_nesting_linter.R
MichaelChirico: Nothing, but wondering if it should (there at least seems to be a conflict with {styler})

R/unnecessary_nesting_linter.R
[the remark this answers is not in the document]
AshesITR: So, just to understand this: These functions explicitly allow a parallel branch in `if` statements with a single control-flow breaking statement?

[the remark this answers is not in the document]
MichaelChirico: Right... they're allowed to pair with "exit" clauses, but can't cause a lint by themselves -- nested code using them can't _always_ be unnested. IOW they can remove a lint, but not cause a lint.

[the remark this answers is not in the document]
MichaelChirico: i.e. this won't lint:

```r
if (A) {
  warning('hi')
} else {
  do()
}
```

Nor will this:

```r
if (A) {
  warning('hi')
} else {
  stop('bye')
}
```

But this will:

```r
if (A) {
  stop('hi')
} else {
  do()
}
```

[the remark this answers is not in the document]
AshesITR: What about

```r
if (A) {
  stop()
} else {
  do()
  warning()
}
```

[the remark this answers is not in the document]
MichaelChirico: won't lint -- only the terminal call is examined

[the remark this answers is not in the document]
AshesITR: So
```r
if (A) {
  stop()
} else {
  warning()
  do()
}
```
Would lint?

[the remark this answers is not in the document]
MichaelChirico: yes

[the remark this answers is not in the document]
MichaelChirico: Of course, one could always disable this behavior (and always force unnesting if possibe) by passing e.g. `character()` to the argument accepting a vector of calls.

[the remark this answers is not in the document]
AshesITR: I'm just wondering how intuitive the exception logic is.

AshesITR:
I think the exceptions need some additional thought. Two options:

1. Remove exceptions for now and make a follow up issue.
2. Flesh out the interface and what should be eligible for exception.

Which do you prefer?

MichaelChirico:
> I think the exceptions need some additional thought. Two options:
> 
> 1. Remove exceptions for now and make a follow up issue.
> 2. Flesh out the interface and what should be eligible for exception.
> 
> Which do you prefer?

let's go with one, I'm for making these PRs as sleek as possible in general.

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,180 @@
+#' Block instances of excessive nesting
```
AshesITR, line 1: s/excessive/unnecessary/ ?

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,180 @@
...
+
+  Linter(function(source_expression) {
+    if (!is_lint_level(source_expression, "expression")) {
+      return(list())
+    }
+
+    xml <- source_expression$xml_parsed_content
+
+    if_else_exit_expr <- xml_find_all(xml, if_else_exit_xpath)
+    # TODO(michaelchirico): customize the error message to the exit clause used
```
AshesITR, line 157: Track this in an issue instead?

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,180 @@
...
+#'  - [linters] for a complete list of linters available in lintr.
+#' @export
+unnecessary_nesting_linter <- function(allow_assignment = TRUE) {
+  exit_calls <- c("stop", "return", "abort", "quit", "q")
+  exit_call_expr <- glue("
+    expr[SYMBOL_FUNCTION_CALL[{xp_text_in_table(exit_calls)}]]
+  ")
+  # block IF here for cases where a nested if/else is entirely within
+  #   one of the branches.
+  # TODO(michaelchirico): we could try and make the parallel exits requirement
```
AshesITR, line 70: Track this in an issue instead?

R/unnecessary_nesting_linter.R
```diff
@@ -0,0 +1,180 @@
...
+  //OP-LEFT-BRACE
+    /parent::expr[
+      count(expr) = 1
+      and not(preceding-sibling::*[
+        self::FUNCTION
+        or self::FOR
+        or self::IF
+        or self::WHILE
+        or self::REPEAT
+        or self::expr/SYMBOL_FUNCTION_CALL[text() = 'switch']
```
AshesITR, line 134: Parameterize the list of allowed functions?

R/unnecessary_nesting_linter.R
MichaelChirico: I'm actually not sure anymore what I meant here, exactly. Just deleting for now, it'll come up again if really important.

R/unnecessary_nesting_linter.R
MichaelChirico: #2329

R/unnecessary_nesting_linter.R
MichaelChirico: Have something specific in mind? Per the comment, `switch()` is included here as a _control flow_ which happens to be a function, so it's treated the same as if/for/while/repeat.

AshesITR: Those discussed in #2326 come to mind.

MichaelChirico: Let's follow up in #2334, removed the exception for `switch()` for now.

AshesITR (approved):

PR #2303, opened 2023-11-17
reduce nesting where possible

MichaelChirico:
Uncovered by #2302


inst/example/complexity.R
```diff
@@ -3,6 +3,7 @@ complexity <- function(x) {
     if (x > 10.0) {
       if (x > 20.0) {
         x <- x / 2.0
+        return(x)
```
MichaelChirico, line 6: maybe it's preferable to delete this file? Why do we keep it here... it's used in two tests, shouldn't we prefer to write out the code directly in those tests?

IndrajeetPatil: Yes, I agree. We can either directly include this code in tests or move this file somewhere to the `tests` folder and use it from there.

AshesITR: I would vote for ditching the file.
We can also lower the cyclocomp threshold in the tests to make the test code smaller.

MichaelChirico: Reminded me I forgot to push the edit I made locally

IndrajeetPatil (approved):

PR #2305, opened 2023-11-17
New consecutive_mutate_linter

MichaelChirico:
Part of #884

No hits on {lintr}


R/consecutive_mutate_linter.R
```diff
@@ -0,0 +1,76 @@
...
+#' lint(
+#'   text = code,
+#'   linters = consecutive_mutate_linter()
+#' )
+#'
+#' @evalRd rd_tags("consecutive_mutate_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+consecutive_mutate_linter <- function(invalid_backends = "dbplyr") {
+  blocked_library_xpath <- glue("
```
AshesITR, line 30: That seems quite brittle. What about ns-qualified usages of dbplyr? What about `require()`?

R/consecutive_mutate_linter.R
```diff
@@ -0,0 +1,76 @@
...
+  #   starting like mutate(DF, ...) %>% mutate(...)
+  xpath <- glue("
+  (//PIPE | //SPECIAL[{ xp_text_in_table(magrittr_pipes) }])
+    /preceding-sibling::expr[
+      expr[2][expr/SYMBOL_FUNCTION_CALL[text() = 'mutate']]
+      or expr/SYMBOL_FUNCTION_CALL[text() = 'mutate']
+    ]
+    /following-sibling::expr[
+      expr/SYMBOL_FUNCTION_CALL[text() = 'mutate']
+      and not(SYMBOL_SUB[text() = '.keep' or text() = '.by'])
```
AshesITR, line 50: Why is this condition not necessary for the first mutate call?

inst/lintr/linters.csv
```diff
@@ -14,6 +14,7 @@ comparison_negation_linter,readability consistency
 condition_message_linter,best_practices consistency
 conjunct_test_linter,package_development best_practices readability configurable pkg_testthat
 consecutive_assertion_linter,style readability consistency
+consecutive_mutate_linter,consistency readability configurable
```
AshesITR, line 17: Maybe efficiency?

R/consecutive_mutate_linter.R
```diff
@@ -0,0 +1,76 @@
...
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "x %>% mutate(a = 1) %>% mutate(b = 2)",
+#'   linters = consecutive_mutate_linter()
+#' )
+#'
+#' # okay
+#' code <- "library(dplyr)\nx %>% mutate(a = 1) %>% mutate(a = a + 1)"
```
AshesITR, line 19: I see no logic precluding this lint. Did you mean `library(dbplyr)`?

inst/lintr/linters.csv
MichaelChirico: These days I'm mostly using {dbplyr} backend so I forget about the efficiency part :)

R/consecutive_mutate_linter.R
MichaelChirico: Very brittle. Especially in packages... I don't have a great general solution here

> `require()`

We already migrated all code to use `library()` so omitted that. I'll add here.

> ns-qualified usages of dbplyr

SGTM. I'll also look for `#' @import` comments (which later can be improved by using proper roxygen2 parsing)

R/consecutive_mutate_linter.R
MichaelChirico: It's indeed necessary, nice catch!

We haven't seen much usage of this feature yet so the false positive hadn't been reported yet.

R/consecutive_mutate_linter.R
```diff
@@ -0,0 +1,96 @@
...
+#'   a variable defined in the same `mutate()` expression.
+#'
+#' @examples
+#' # will produce lints
+#' lint(
+#'   text = "x %>% mutate(a = 1) %>% mutate(b = 2)",
+#'   linters = consecutive_mutate_linter()
+#' )
+#'
+#' # okay
```
IndrajeetPatil, line 23: Shouldn't the first okay example be the non-exception case?

```r
#' lint(
#'   text = "x %>% mutate(a = 1, b = 2)",
#'   linters = consecutive_mutate_linter()
#' )
```

DESCRIPTION
```diff
@@ -81,6 +81,7 @@ Collate:
     'condition_message_linter.R'
     'conjunct_test_linter.R'
     'consecutive_assertion_linter.R'
+    'consecutive_mutate_linter.R'
```
IndrajeetPatil, line 84: Should this linter be named (e.g. `unnecessary_consecutive_calls_linter()`) in a way to anticipate further cases like this in the _tidyverse_ workflow?

E.g. I could imagine a similar linter for `dplyr::filter()`

``` r
library(dplyr, warn.conflicts = FALSE)

df <- tibble(a = 1:4, b = 4:1)

identical(
  df %>%
    filter(a > 2) %>%
    filter(b < 3),
  df %>%
    filter(a > 2 & b < 3)
)
#> [1] TRUE
```

<sup>Created on 2023-11-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Note that this is different from `conjunct_test_linter()`, which is not about successive function calls.

AshesITR: Careful:

`df %>% group_by(x) %>% filter(y == 42L) %>% filter(n() > 1L)` cannot be collapsed into a single filter call, because the first filter changes group sizes.

MichaelChirico: Yes, exactly why this case was not considered. It's impossible to tell statically whether this is happening, too.

MichaelChirico: (historical footnote: we originally had a `ConsecutiveCallLinter()` that covered `suppressMessages()`, `mutate()`, and `stopifnot()`, but each case drifted far enough apart that we gave them their own linter)

IndrajeetPatil: Hadn't thought about this. Thanks for this insight!

``` r
library(dplyr)
library(palmerpenguins)

setdiff(
  penguins %>%
    group_by(sex) %>%
    filter(island == "Dream", n() > 1L),
  penguins %>%
    group_by(sex) %>%
    filter(island == "Dream") %>%
    filter(n() > 1L)
)
#> # A tibble: 1 × 8
#> # Groups:   sex [1]
#>   species island bill_length_mm bill_depth_mm flipper_length_mm body_mass_g
#>   <fct>   <fct>           <dbl>         <dbl>             <int>       <int>
#> 1 Adelie  Dream            37.5          18.9               179        2975
#> # ℹ 2 more variables: sex <fct>, year <int>
```

<sup>Created on 2023-11-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: Basically, we either need to have `filter(x, A, B, C)` equivalent to `filter(x, A && B && C)` or `x |> filter(A) |> filter(B) |> filter(C)`; {dplyr} chooses the former.

R/consecutive_mutate_linter.R
```diff
@@ -0,0 +1,101 @@
+#' Force consecutive calls to mutate() into just one when possible
```
AshesITR, line 1: Wording could be more consistent to other linters

MichaelChirico: PTAL

AshesITR (approved):

PR #2306, opened 2023-11-17
New consecutive_suppression_linter

MichaelChirico:
Part of #884

No hits on {lintr}


AshesITR:
I'm wondering whether this would also fit into `library_call_linter()`. Alternatively, we could lint successive calls to `suppressMessages()` regardless of their code content. WDYT?

MichaelChirico:
> I'm wondering whether this would also fit into `library_call_linter()`. Alternatively, we could lint successive calls to `suppressMessages()` regardless of their code content. WDYT?

We lost some history here, but my memory is we intentionally left out `suppress{Warnings,Messages}` because of a prevalence of false positives (where `suppressWarnings()` twice in a row as done "well"). So the idea of subsuming into `library_call_linter()` SGTM.

MichaelChirico:
Marking as draft until #2279 is merged

R/library_call_linter.R
```diff
@@ -94,13 +114,13 @@ library_call_linter <- function(allow_preamble = TRUE) {
...
+  ")
 
   bad_indirect_funs <- c("do.call", "lapply", "sapply", "map", "walk")
-  call_symbol_cond <- "
-  SYMBOL[text() = 'library' or text() = 'require']
-    or STR_CONST[text() = '\"library\"' or text() = '\"require\"']
-  "
+  call_symbol_cond <- glue("
+    SYMBOL[{attach_call_cond}]
+    or STR_CONST[{ xp_text_in_table(dQuote(attach_calls, '\"')) }]
```
AshesITR, line 122: This is to catch `lapply(pkgs, "library", character.only = TRUE)`?
Is that worth it? If yes, we should catch `lapply(pkgs, 'library', character.only = TRUE)` too.

MichaelChirico: Yes, but let's mark as follow-up since it's not related to consecutive suppression but `character.only`

MichaelChirico: #2319

AshesITR:
Metadata tests and intermingled calls would be interesting to check.

MichaelChirico:
PTAL at the test in 9977822e80f3d36cb06a64105fee26eb330ac159

AshesITR:
I was thinking about

```r
suppressMessages()
suppressPackageStartupMessages()
suppressMessages()
suppressPackageStartupMessages()
suppressPackageStartupMessages()
```

To see how they interact.

MichaelChirico:
Oh, gotcha. Will add. FWIW, I've never come across this in practice. I assume any author using these consecutive calls is just doing a bunch of copy-paste, rather than thinking carefully about what needs to be suppressed in each case.

MichaelChirico:
OTOH, there is a _lot_ of R code in the world :laughing: 

https://github.com/proteinphysiologylab/FrustraEvo/blob/044963c1e8016bf957c8fc04043bd75da6e85c4e/Scripts/Generator.R#L2-L7

https://github.com/alematia/MaGplotR/blob/70c3e253635d8ba047038af19b17acc55e87f387/MaGplotR.R#L19-L25

AshesITR (approved):

PR #2307, opened 2023-11-18
New object_overwrite_linter

MichaelChirico:
Part of #884

Here's another one we didn't end up using internally, so it hasn't gotten all too much scrutiny -- it might benefit from some rethinking (e.g. is exclusion of top-level assignments and function formals really necessary? Or even worth putting behind an argument?)

Originally, we also used a fixed set of functions to check for duplicates based on the observed frequency of overwrites in our code base. Presenting that list for posterity, but I'm not sure there's a "principled" way to include it in the general package:

```r
kCommonBaseOverwrites <- unname(unlist(list(
  base = c(
    "all", "args", "beta", "c", "col", "date", "diff",
    "dir", "file", "labels", "mean", "message", "names",
    "q", "row", "scale", "summary", "t", "table", "units"
  ),
  utils = c("data", "history"),
  stats = c(
    "coef", "coefficients", "cov", "df", "dt", "fitted",
    "formula", "sd", "sigma", "start", "time", "var",
    "variable.names", "weights"
  ),
  graphics = c("lines", "par", "plot", "text", "title"),
  grDevices = "colors"
)))
```

PR #2308, opened 2023-11-18
Rename some variables that overwrite base functions

MichaelChirico:
Partial fix for some of the overwrites pointed out in #2307

### DIFF
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -35,20 +35,20 @@ exclude <- function(lints, exclusions = settings$exclusions, linter_names = NULL
     return(lints)
   }
 
-  df <- as.data.frame(lints)
+  lint_df <- as.data.frame(lints)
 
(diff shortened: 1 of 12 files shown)


AshesITR:
I can see from this diff why you disabled it internally 😅

IndrajeetPatil (approved):
Thanks! 

Irrespective of what brought about this change, I do find the majority of new names to be an improvement over their previous versions.

IndrajeetPatil:
@MichaelChirico, @AshesITR, @jimhester, @jeroen: Any ideas on why the r-universe deployments have started to remain pending? E.g. deployments from the 13th Nov are still pending. I am not too familiar with r-universe, and so I am not sure if some repo settings need to be changed for this to be fixed.

Screenshot 2023-11-18 at 11 01 33

PR #2321, opened 2023-11-20
New allow_implicit_else argument for return_linter

MichaelChirico:
Part of #884

With this, all of the internal linters we plan to "upstream" are now filed on the tracker 🎉


AshesITR:
The linter should probably be made configurable. I can see `initialize(...)` as defined within R6 classes for example.

Also, the message implies explicit returns. Should be reworded to be agnostic of the return style if possible.

MichaelChirico:
> the message implies explicit returns

Right, that's why I'm thinking it may be best to subsume into `return_linter()`. Something like `return_style = c("implicit", "allow_implicit_else", "explicit")`.

> The linter should probably be made configurable

Not seeing the connection with R6, could you elaborate?

AshesITR:
For implicit returns, the lint could also be desired because the implicit NULL might surprise.

Regarding R6:

```r
Foo <- R6::R6Class(
  "Foo",
  public = list(
    initialize = function(x) {
      self$x <- x
    },
    x = NULL
  )
)
```

Shouldn't lint.

AshesITR:
Is there functionality here that still needs to go into `return_linter()`?
I think for implicit return style it could also be a good feature to lint terminal `if` without `else`, for example.

MichaelChirico:
> I think for implicit return style it could also be a good feature to lint terminal `if` without `else`.
> 
> Suggestion: `allow_implicit_else = TRUE` (default `FALSE` would need justification from the styleguide since `return_linter()` is a default linter)

what is the lint for the implicit return case? that `else NULL` should be explicit?

MichaelChirico:
Implemented under `return_linter()` with new argument `allow_implicit_else`, PTAL.

R/return_linter.R
```diff
@@ -154,7 +172,24 @@ return_linter <- function(
...
     ")
-    msg <- "All functions must have an explicit return()."
+    return_msg <- "All functions must have an explicit return()."
+  }
+
+  if (!allow_implicit_else) {
+    # for inline functions, terminal <expr> is a sibling of <FUNCTION>, otherwise
+    #   it's a descendant of the <expr> following <FUNCTION>
+    implicit_else_xpath <- glue("
+    //FUNCTION[not(
```
AshesITR, line 182: Indentationy by one more level reads better to me.

R/shared_constants.R
```diff
@@ -309,3 +309,14 @@ purrr_mappers <- c(
   "map_raw", "map_lgl", "map_int", "map_dbl", "map_chr", "map_vec",
   "map_df", "map_dfr", "map_dfc"
 )
+
+# Shared between returns_linter and implicit_else_return_inter
+# See `?.onAttach`; these functions are all exclusively used for their
+#   side-effects, so implicit return is generally acceptable
+return_not_needed_funs <- c(
```
AshesITR, line 316: This is unused, no?
`special_funs` is the name of the constant containing package hooks.

tests/testthat/test-return_linter.R
```diff
@@ -882,3 +882,117 @@ test_that("return_linter lints `message`, `warning` and `stopifnot`", {
...
+      foo <- function(bar) {
+        if (TRUE) {
+          return(bar)
+        } else {
+          return(NULL)
+        }
+      }
+    "),
+    NULL,
+    return_linter(allow_implicit_else = FALSE)
```
AshesITR, line 910: Isn't this a false negative?
I'd expect `return_style = "implicit"` to require
```r
if (TRUE) {
  bar
} else {
  NULL
}
```

tests/testthat/test-return_linter.R
```diff
@@ -882,3 +882,117 @@ test_that("return_linter lints `message`, `warning` and `stopifnot`", {
...
+  )
+})
+
+test_that("allow_implicit_else = FALSE + explicit returns skips side-effect functions like .onLoad", {
+  expect_lint(
+    trim_some("
+      .onAttach <- function(libname, pkgname) {
+        if (TRUE) return(foo())
```
AshesITR, line 992: Shouldn't this be without `return()` to test what the description says?

MEO265:
I think there is still a bug in the marking in the new functionality.
The lint is displayed for line 1 in this example:
```R
foo <- function(x, y = 3) {
  if(x) {
    return(x)
  }
}
```
with `return_linter(return_style = "explicit", allow_implicit_else = FALSE)`

tests/testthat/test-return_linter.R
MEO265: I think with `"implicit"` it should be allowed to use `return`s in control statements, especially in `if` and `switch`.  Maybe instead of more flags for control, one could introduce even more types like `“strict_implicit”` in which this would also be checked.

This would also fit with the original idea of ​​this PR to bring in the functionality as a new type.

MEO265:
> Also, the message implies explicit returns. Should be reworded to be agnostic of the return style if possible.

@AshesITR unfortunately, I don't understand how this parameter should be understood in implicit style.  Can you give an example that should be linted without this parameter but not?

I actually found the idea of ​​@MichaelChirico with the different types with gradations of explicitness very elegant

R/shared_constants.R
MichaelChirico: Ah, meant to revert this file, thanks

tests/testthat/test-return_linter.R
MichaelChirico: > I think with "implicit" it should be allowed to use returns in control statements

Disagree with this, and it's not the style followed by Tidyverse or base R. I agree it looks like a false negative to me, investigating. We can file a request to support mixed implicit/explicit returns as a separate FR.

MichaelChirico: Corrected this test, and added the incorrect test as a new regression test.

That said, the new fix is more expedient than perfect -- it'll throw an incorrect lint here:

```r
foo <- function() {
  if (TRUE) {
    if (TRUE) {
      return(1)
    }
    2
  } else {
    3
  }
```

(the `return()` is a valid early return within that branch).

But as this issue is not relevant to the implicit `else` returns feature, I'm pushing to follow-up.

AshesITR:
> > Also, the message implies explicit returns. Should be reworded to be agnostic of the return style if possible.
> 
> @AshesITR unfortunately, I don't understand how this parameter should be understood in implicit style.  Can you give an example that should be linted without this parameter but not?
> 
> I actually found the idea of ​​@MichaelChirico with the different types with gradations of explicitness very elegant

```r
function(x) {
  if (x) x
}
```

Should lint with implicit return style if implicit else is disallowed.
The lint-free version would be

```r
function(x) {
  if (x) x else NULL
}
```

MichaelChirico:
> > > Also, the message implies explicit returns. Should be reworded to be agnostic of the return style if possible.
> > 
> > 
> > @AshesITR unfortunately, I don't understand how this parameter should be understood in implicit style.  Can you give an example that should be linted without this parameter but not?
> > I actually found the idea of ​​@MichaelChirico with the different types with gradations of explicitness very elegant
> 
> ```r
> function(x) {
>   if (x) x
> }
> ```
> 
> Should lint with implicit return style if implicit else is disallowed. The lint-free version would be
> 
> ```r
> function(x) {
>   if (x) x else NULL
> }
> ```

👍 that looks like how I've implemented it. so basically there are 4 styles available (existing issues with implementation notwithstanding):

 - return_style="implicit", allow_implicit_else=TRUE (default), only early returns should be explicit, 'else NULL' not required
 - return_style="implicit", allow_implicit_else=FALSE, only early returns should be explicit, 'else NULL' required
 - return_style="explicit", allow_implicit_else=TRUE, all returns should be explicit, 'else NULL' not required
 - return_style="explicit", allow_implicit_else=FALSE, all returns should be explicit, 'else NULL' required

MEO265:
```R
lint(
  trim_some("
  function(x, y) {
    if(x) {
      1
    } else if(y) {
      2
    }
  }
  "),
  return_linter(return_style = "implicit", allow_implicit_else = FALSE)
)

# No lint, but should lint
```

I think this combination should return a lint on line 1 (the first `if`) (or alternatively at least on line 3 on the second if`)

MichaelChirico:
> ```r
> lint(
>   trim_some("
>   function(x, y) {
>     if(x) {
>       1
>     } else if(y) {
>       2
>     }
>   }
>   "),
>   return_linter(return_style = "implicit", allow_implicit_else = FALSE)
> )
> 
> # No lint, but should lint
> ```
> 
> I think this combination should return a lint on line 1 (the first `if`) (or alternatively at least on line 3 on the second if`)

I think that's out of scope for now. We don't handle arbitrary nesting with any part of this linter's logic at the moment. The "outer" `if` _does_ have its corresponding `else`.

Note that the parser treats the "missing" last `else` as part of a sub-expression:

```r
substitute(if (A) 1 else if (B) 2)[[4]]
# if (B) 2
```

i.e. the parser treats the expression equivalently to

```r
if (A) {
  1
} else {
  if (B) {
    2
  }
}
```

There is _an_ implicit `else` but it's nested.

MichaelChirico:
> That is true. Is there a way to formulate a recursive XPath for that? The problem is, in the parse tree, an `if () else if () ...` chain can get arbitrarily deep.
> 
> Maybe something like `//IF[not(ELSE)]/ancestor-or-self::expr[IF and last()]` would land on the first if of a chain? It might mess up with multiply nested if blocks...

I think it's possible with an XPath, but yes, quite tricky. I'm thinking a better approach might be to use {xml2} functions recursively. But as noted, I think that's better reserved for future improvement.

MEO265:
I think this double lint is not correct either you see the `if` after an `else` as a new expression, then only the second one should be marked, or you understand an `if` with a lot of `else` as one expression (that's how I would see it) and then only the first one would have to be marked.

```R
lint(
  trim_some("
  function(x, y) {
    if(x) {
      1
    } else if(y) {
      2
    }
  }
  "),
  return_linter(return_style = "explicit", allow_implicit_else = FALSE)
)
```
```

<text>:2:9: style: [return_linter] All functions must have an explicit return().
  if(x) {
        ^
<text>:4:10: style: [return_linter] All functions must have an explicit return().
  } else if(y) {
         ^~~~~~~
```

I'm not entirely sure if it's a bug of this PR or the first one. If it's one of the first, feel free to convert this message into an issue, because then I'll probably have a (simple) solution to fix it and properly handle control statements of any depth.

MEO265:
> I think that's out of scope for now. We don't handle arbitrary nesting with any part of this linter's logic at the moment. The "outer" `if` _does_ have its corresponding `else`.

Then I would want to take care of it in a separate PR, with the bug from my last message

> Note that the parser treats the "missing" last `else` as part of a sub-expression

I know, but thanks. I would try to recreate the functionality of XPath 2.0 by recursively applying parts of the current XPath and hope to handle it correctly.

MichaelChirico:
> I think this double lint is not correct either you see the `if` after an `else` as a new expression, then only the second one should be marked, or you understand an `if` with a lot of `else` as one expression (that's how I would see it) and then only the first one would have to be marked.
> 
> ```r
> lint(
>   trim_some("
>   function(x, y) {
>     if(x) {
>       1
>     } else if(y) {
>       2
>     }
>   }
>   "),
>   return_linter(return_style = "explicit", allow_implicit_else = FALSE)
> )
> ```
> 
> ```
> 
> <text>:2:9: style: [return_linter] All functions must have an explicit return().
>   if(x) {
>         ^
> <text>:4:10: style: [return_linter] All functions must have an explicit return().
>   } else if(y) {
>          ^~~~~~~
> ```
> 
> I'm not entirely sure if it's a bug of this PR or the first one. If it's one of the first, feel free to convert this message into an issue, because then I'll probably have a (simple) solution to fix it and properly handle control statements of any depth.

this is WAI, as noted we don't support arbitrary nesting yet. R sees this like so:

```r
if (x) {
  1
} else {
  if (y) {
    2
  }
}
```

The current linter requires (1) a `return()` under the `isTRUE(x)` branch and (2) a `return()` under the `!isTRUE(x)` branch. Neither is present, and it marks the lint in a useful place in both cases. Perhaps the first lint could land on `1` instead of `{` -- better deferred to a follow-up PR.

MichaelChirico:
> I think this double lint is not correct either you see the `if` after an `else` as a new expression, then only the second one should be marked, or you understand an `if` with a lot of `else` as one expression (that's how I would see it) and then only the first one would have to be marked.
> 
> ```r
> lint(
>   trim_some("
>   function(x, y) {
>     if(x) {
>       1
>     } else if(y) {
>       2
>     }
>   }
>   "),
>   return_linter(return_style = "explicit", allow_implicit_else = FALSE)
> )
> ```
> 
> ```
> 
> <text>:2:9: style: [return_linter] All functions must have an explicit return().
>   if(x) {
>         ^
> <text>:4:10: style: [return_linter] All functions must have an explicit return().
>   } else if(y) {
>          ^~~~~~~
> ```
> 
> I'm not entirely sure if it's a bug of this PR or the first one. If it's one of the first, feel free to convert this message into an issue, because then I'll probably have a (simple) solution to fix it and properly handle control statements of any depth.

Thanks @MEO265, arbitrary nesting is now handled in #2363 and adapted here -- your example now catches 3 lints (two missing `return()` calls, one missing `else`).

R/return_linter.R
```diff
@@ -55,13 +71,19 @@
...
 return_linter <- function(
     return_style = c("implicit", "explicit"),
+    allow_implicit_else = TRUE,
     return_functions = NULL,
     except = NULL) {
   return_style <- match.arg(return_style)
 
+  if (!allow_implicit_else || return_style == "explicit") {
+    except_xpath <- glue("parent::expr[not(
+      preceding-sibling::expr/SYMBOL[{ xp_text_in_table(union(special_funs, except)) }]
```
MichaelChirico, line 81: I was a bit worried about the `preceding-sibling::expr` and `parent::expr` here in the context of `=` or `->` assignment. Added a test for the former; the latter is not relevant I think, because of operator precedence:

```r
function() {
  ...
} -> .onLoad
```

Does not work as "intended" -- it actually parses like:

```r
function() ( foo <- { } )
```

So right-assignment of a function would actually have to look like:

```r
(function() {
  ...
}) -> .onLoad
```

That seems pretty unlikely to me, so ignoring it for now unless there's a user request.

MichaelChirico: I do see a handful of such assignments, but I'm happy to ignore them, e.g.

https://github.com/cran/symbol.equation.gpt/blob/4c2d44d5f175f5ea5bbd540a1fd56d8cf0736f4f/R/aaa.R#L15-L22

https://github.com/cran/symbol.equation.gpt/blob/4c2d44d5f175f5ea5bbd540a1fd56d8cf0736f4f/R/main.R#L14-L20

https://github.com/search?q=lang%3AR+%2F%5B%7D%5D%5B%29%5D%5Cs*%5B-%5D%5B%3E%5D%2F+%2F%5E%5B%28%5D%5Cs*function%5Cs*%5B%28%5D%2F&type=code

NEWS.md
```diff
@@ -19,15 +19,19 @@
...
+* New default linter `return_linter()` for the style guide rule that terminal returns should be left implicit (#1100, #2354, and #2356, @MEO265 and @MichaelChirico).
 
 ## New and improved features
 
 * More helpful errors for invalid configs (#2253, @MichaelChirico).
 * `library_call_linter()` is extended
   + to encourage all packages to be attached with `library(symbol)`, not `library("symbol", character.only = TRUE)` or "vectorized" approaches looping over package names (part of #884, @MichaelChirico).
   + to discourage many consecutive calls to `suppressMessages()` or `suppressPackageStartupMessages()` (part of #884, @MichaelChirico).
-* `return_linter()` also has an argument `return_style` (`"implicit"` by default) which checks that all functions confirm to the specified return style of `"implicit"` or `"explicit"` (part of #884, @MichaelChirico, @AshesITR and @MEO265).
+* `return_linter()` also has arguments for fine-tuning which functions get linted:
```
AshesITR, line 30: I'm wondering whether the issue numbers related to `return_linter` should be condensed to a single location.

MichaelChirico: Shall we punt to later? There are a slew more issues related to `return_linter()` I'd like to land before next release, maybe best to address them all at once in a sweep over the NEWS in prep for release, WDYT?

AshesITR (approved):

PR #2326, opened 2023-11-21
Remove excess nesting

MichaelChirico:
Caught by #2302. Follow-up to #2303. Missed before as it got mixed in with `allow_assignment = TRUE` false positives.

### DIFF
--- a/tests/testthat/test-pipe_continuation_linter.R
+++ b/tests/testthat/test-pipe_continuation_linter.R
@@ -181,9 +181,7 @@ local({
   )
   patrick::with_parameters_test_that(
     "valid nesting is handled",
-    {
-      expect_lint(code_string, NULL, linter)
-    },
+    expect_lint(code_string, NULL, linter),
     .test_name = valid_code,
     code_string = valid_code
   )


AshesITR (commented):
I'm not sure this actually improves readability. Note the parallelism between `patrick::with_parameters_test_that()` and `testthat::test_that()`.

MichaelChirico:
> I'm not sure this actually improves readability. Note the parallelism between `patrick::with_parameters_test_that()` and `testthat::test_that()`.

difference is `test_that()` is usually like

```r
test_that("descr", {
  ...
})
```

vs. because of the parameter arguments `with_parameters_test_that` is usually like

```r
with_parameters_test_that(
  "descr",
  {
    ...
  },
  ...
)
```

AshesITR:
Yes and that highlights the actual test code very well.

IndrajeetPatil:
I agree with @AshesITR here—the braced expressions does highlight the test code and draws a parallel with testthat. 

Is this really flagged as excessive nesting by the new linter?

AshesITR:
I can see a similar problem occur with shiny.
`reactive({})`, `observe({})`, `observeEvent(., {})` and `render*({})` can potentially cause unwanted lints.

MichaelChirico:
Let's discuss exceptions needed in #2302 or its follow-ups. `test_that()` also lacks any exception at the moment.

PR #2328, opened 2023-11-21
Apply all but some linters

MichaelChirico:
This will make it easier to apply the panoply of linters we author to ourselves without having to add them to a growing list one-by-one.


AshesITR:
`all_linters()` accepts customizations directly via `...`.

MichaelChirico:
> `all_linters()` accepts customizations directly via `...`.

TIL. I admit it looks strange to me this way. But may as well use the feature we've written.

MichaelChirico:
Oh, I see, we need to supply `packages=`? That feels even stranger. Should we move `packages=` after `...` in the `all_linters()` signature?

AshesITR:
Sounds good. Any usages in the wild of this? I suspect not so a breaking change won't have much impact.

MichaelChirico:
I don't see any:

https://github.com/search?q=lang%3AR%20%2Fall_linters%5C(%2F&type=code

https://github.com/search?q=path%3A.lintr+%2Fall_linters%5C%28%2F&type=code

AshesITR:
cc @Bisaloo it seems you are using `all_linters()` in your work.

MichaelChirico:
> cc @Bisaloo it seems you are using `all_linters()` in your work.

In the public repos, `package=` is always supplied by name, so the proposed re-ordering of the signature of `all_linters()` would not affect you unless you have a different practice private repos/projects off GitHub.

MichaelChirico:
Draft while #2332 is pending.

AshesITR:
I would be interested in the performance implications of this change.

Bisaloo:
> cc @Bisaloo it seems you are using all_linters() in your work.

Thanks for pinging :pray:! As @MichaelChirico pointed out, the `packages` argument is named so nothing should break. FWIW, it would feel very strange to me to not name this argument no matter its position since it's an outlier compared to the following/preceding arguments.

MichaelChirico:
> I would be interested in the performance implications of this change.

Good call-out. I think it's slower, but not substantially:

```r
system.time(lint_package())

# this branch
#    user  system elapsed 
# 306.824   4.011 311.853 

# main
#    user  system elapsed 
# 256.827   2.310 259.812 
```

MichaelChirico:
Here are the timings of individual linters:

```
           cyclocomp_linter     object_overwrite_linter         object_usage_linter          indentation_linter 
                     84.264                      19.455                       8.364                       7.711 
            backport_linter        object_length_linter          object_name_linter        package_hooks_linter 
                      7.590                       4.897                       4.432                       4.336 
               paste_linter      string_boundary_linter        keyword_quote_linter      list_comparison_linter 
                      4.322                       4.319                       3.875                       3.847 
  unnecessary_lambda_linter                brace_linter trailing_blank_lines_linter          fixed_regex_linter 
                      3.409                       3.287                       3.053                       3.019 
```

`backport_linter()` is a bit surprising. I think this is worth a few follow-ups.

AshesITR:
20% slowdown seems harsh.

I'm surprised by object_overwrite_linter.
Maybe not worth precompiling disallowed names into the XPath? It shouldn't be slower than object_name_linter.

MichaelChirico:
> 20% slowdown seems harsh.

not sure that's the right framing. many of these are linters we applied to ourselves but just aren't in the suite. I think more importantly it still runs faster than many of the other GHA checks so we won't really notice.

Eventually, parallelism can help.

MichaelChirico:
Current timings:

```
# current main
#    user  system elapsed 
# 123.475   0.092 123.567 

# this branch
#    user  system elapsed 
# 183.452   0.960 184.412
```

Actually (still) 50% slower on this machine! But the absolute time is only 3 minutes. I think we should just merge & keep working on some performance issues going forward, if we really notice the `lint` GHA being a blocker we can revisit.

AshesITR:
I agree. Are those timings from the same machine as before?

.lintr
```diff
@@ -1,45 +1,35 @@
...
-    unnecessary_nested_if_linter(),
-    unnecessary_lambda_linter(),
+    object_overwrite_linter(
+      allow_names = c("end", "line", "lines", "new", "offset", "pipe", "start", "symbols")
+    ),
+    undesirable_function_linter(modify_defaults(
+      defaults = default_undesirable_functions,
+      Sys.setenv = NA_character_,
+      mapply = NA_character_,
+      structure = NA_character_,
```
AshesITR, line 11: While we're at it, this could use some explanation, why we don't like `structure()`.

.lintr
```diff
@@ -1,45 +1,35 @@
...
+    library_call_linter = NULL,
+    nonportable_path_linter = NULL,
+    todo_comment_linter = NULL,
+    # TODO(#2327): Enable this.
+    unreachable_code_linter = NULL
   )
 exclusions: list(
     "inst/doc/creating_linters.R" = 1,
     "inst/example/bad.R",
+    "tests/testthat.R" = list(unused_import_linter = Inf),
```
AshesITR, line 32: Where does `unused_import_linter()` provide a benefit at the moment? In `.dev`?

MichaelChirico: That won't get hit by `lint_package()`. Shouldn't it work for vignettes though?

I'm also open to just disabling it.

MichaelChirico:
> I agree. Are those timings from the same machine as before?

no. best guess is the faster absolute timings come from a machine with the faster dev version of {xmlparsedata} installed. Won't be able to reproduce that any time soon though.

.lintr
MichaelChirico: Oops, these are already `default_undesirable_functions`, this just served to give us a crappier lint message. Removed them.

.lintr
AshesITR: It's fine, we can keep it.

.lintr
```diff
@@ -1,45 +1,32 @@
...
-    redundant_equals_linter(),
-    redundant_ifelse_linter(),
-    sort_linter(),
-    sprintf_linter(),
-    strings_as_factors_linter(),
-    undesirable_function_linter(c(Sys.setenv = NA_character_, mapply = NA_character_, structure = NA_character_)),
-    unnecessary_nested_if_linter(),
-    unnecessary_lambda_linter(),
+    object_overwrite_linter(
+      allow_names = c("end", "line", "lines", "new", "offset", "pipe", "start", "symbols")
```
AshesITR, line 5: Where are the lints corresponding to the exceptions? Maybe we can nolint some sections instead?

R/zzz.R
```diff
@@ -172,9 +174,7 @@ all_undesirable_functions <- modify_defaults(
   untrace = paste(
     "remove this likely leftover from debugging.",
     "It is only useful for interactive debugging with trace()"
-  ),
-  structure =
```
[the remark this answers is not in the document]
MichaelChirico: The rest of the object is sorted, I think we tacked this on the end by mistake. Now it's sorted again.

(we can't really regression test this because `modify_defaults()` will return it sorted -- the benefit is only for ease of reading the source code)

.lintr
MichaelChirico: They're all over:

```r
                       filename line_number            object
 1:  R/get_source_expressions.R         638        stats::end
 2:  R/get_source_expressions.R         642        stats::end
 3:              R/path_utils.R         147        stats::end
 4:  R/get_source_expressions.R         205       stats::line
 5:  R/get_source_expressions.R         236       stats::line
 6:  R/get_source_expressions.R         239       stats::line
 7:                    R/lint.R         650       stats::line
 8:  R/get_source_expressions.R         564   graphics::lines
 9:                    R/lint.R          43   graphics::lines
10:                    R/lint.R         362   graphics::lines
11:                   R/utils.R         182   graphics::lines
12:      R/xml_nodes_to_lints.R          69   graphics::lines
13:      R/xml_nodes_to_lints.R          70   graphics::lines
14: tests/testthat/test-cache.R           7   graphics::lines
15:                    R/lint.R         338      methods::new
16:                    R/lint.R         341      methods::new
17:  R/get_source_expressions.R         587     stats::offset
18:    R/one_call_pipe_linter.R          73        base::pipe
19:        R/pipe_call_linter.R          34        base::pipe
20:  R/get_source_expressions.R         633      stats::start
21:              R/path_utils.R         146      stats::start
22:        R/namespace_linter.R          97 graphics::symbols
23:     R/object_usage_linter.R         138 graphics::symbols
24:        R/shared_constants.R         294 graphics::symbols
                       filename line_number            object
```

MichaelChirico: I took another look and renamed `start`, `end`, and `#nolint`'d `offset`.

I think the other names are good to ignore everywhere. `line`/`lines`/`pipe`/`symbols` all will continue to come up reasonably often for us.

MichaelChirico: OK, also renamed the arguments for internal function `lintr_deprecated` to avoid `new`.

AshesITR (approved):

PR #2331, opened 2023-11-21
exclude %<>% from one_call_pipe_linter

### LINKED ISSUE #2330: Should `one_call_pipe_linter()` apply only to `%>%`?
I can see how `as.character(x)` can be preferable over `x %>% as.character()`, but I don't think the alternatives for other pipes work as well.

Consider `%<>%` (e.g.)

This example lints:

``` r
library(lintr)
lint(text = "data %<>% mutate(n.obs = n)", linters = one_call_pipe_linter())
#> <text>:1:2: warning: [one_call_pipe_linter] Expressions with only a single call shouldn't use pipe %<>%.
#>  data %<>% mutate(n.obs = n)
#>  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

<sup>Created on 2023-11-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

To avoid this, I instead need to do `data <- mutate(data, n.obs = n)`, and avoiding repetition of `data` is the entire _raison d'etre_ of `%<>%`. So when `{lintr}` says one shouldn't use this pipe only with a single call, it's basically saying don't use this pipe for the very thing for which it was built. That's not the case for `%>%`/`|>`. I hope I am being cogent here.

I could imagine a similar argument for other pipes. 

---

So I am wondering if the following should be changed only to cover `%>%`/`|>`:

https://github.com/r-lib/lintr/blob/209e250d499630cd6a5aab7319696d725e4b81b1/R/one_call_pipe_linter.R#L48

MichaelChirico:
Closes #2330

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,7 @@
 * `nested_pipe_linter()` for discouraging pipes within pipes, e.g. `df1 %>% inner_join(df2 %>% select(a, b))` (part of #884, @MichaelChirico).
 * `nrow_subset_linter()` for discouraging usage like `nrow(subset(x, conditions))` in favor of something like `with(x, sum(conditions))` which doesn't require a full subset of `x` (part of #884, @MichaelChirico).
 * `pipe_return_linter()` for discouraging usage of `return()` inside a {magrittr} pipeline (part of #884, @MichaelChirico).
+* `one_call_pipe_linter()` for discouraging one-step pipelines like `x |> as.character()` (#2330 and part of #884, @MichaelChirico).
 
 ### Lint accuracy fixes: removing false positives
 
--- a/R/one_call_pipe_linter.R
+++ b/R/one_call_pipe_linter.R
@@ -31,13 +31,20 @@
 #'   linters = one_call_pipe_linter()
 #' )
 #'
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -39,6 +39,7 @@
 * `nested_pipe_linter()` for discouraging pipes within pipes, e.g. `df1 %>% inner_join(df2 %>% select(a, b))` (part of #884, @MichaelChirico).
 * `nrow_subset_linter()` for discouraging usage like `nrow(subset(x, conditions))` in favor of something like `with(x, sum(conditions))` which doesn't require a full subset of `x` (part of #884, @MichaelChirico).
 * `pipe_return_linter()` for discouraging usage of `return()` inside a {magrittr} pipeline (part of #884, @MichaelChirico).
+* `one_call_pipe_linter()` for discouraging one-step pipelines like `x |> as.character()` (#2330 and part of #884, @MichaelChirico).
```
AshesITR, line 42: Did we miss the bullet for the first PR? 😳

MichaelChirico: yes 🥲

AshesITR (approved):

PR #2340, opened 2023-11-23
exit early from package_hooks_linter() if no hooks are found

### LINKED ISSUE #2336: Improve performance of package_hooks_linter()
`package_hooks_linter()` is slow because of how many steps it takes.

But on the vast majority of source files it'll encounter, there will be no `.onLoad()` or `.onAttach()` hooks and it could just exit early.

MichaelChirico:
Closes #2336

Also move to `"file"`-level linting, I think that makes more sense for this linter.

With this, `lint_package()` goes from taking 11s to 8s on my machine.

For reference, `make_linter_from_xpath("false", "xx")` also takes about 8s, i.e., most of that 11s is the overhead of loading the files.


AshesITR:
Can you provide timing info?
Code LGTM. Just interested in how much we shave by this change.

MichaelChirico:
> Can you provide timing info? Code LGTM. Just interested in how much we shave by this change.

do you want more detail than given in the main comment?

> With this, `lint_package()` goes from taking 11s to 8s on my machine.
> 
> For reference, `make_linter_from_xpath("false", "xx")` also takes about 8s, i.e., most of that 11s is the overhead of loading the files.

AshesITR:
Sorry, I missed the timings in OP. LGTM.

AshesITR (approved):

Issue #2337, opened 2023-11-22
Find ways to improve performance of `object_overwrite_linter()`


MichaelChirico:
I think we know it's bound to be slow in general since there are so many exports by default (I see 2,591 names to check in R4.1.3).

Still #2328 shows it's faster only than `cyclocomp_linter()` in the current suite.

One concrete idea is to try sorting. Branch prediction intuition tells me if we sort the set of functions to search for, it will be faster. I _assume_ we need to sort in C order, though, which is a bit more subtle & requires some investigation.

MichaelChirico:
Quick experiment:

```r
system.time(replicate(5, lint_package(linters = l)))

# No sorting
   user  system elapsed 
128.161   1.103 129.687 

# sort() in system locale 
   user  system elapsed 
128.227   1.047 129.665 

# force C locale sorting
   user  system elapsed 
127.524   1.068 128.912 
```

So, looks like a _very small_ benefit from sorting _in C locale only_. Not sure it's worth pursuing, unless we want it to be something `xp_text_in_table()` does by default more generally. If so, I'd like it to be controlled by an option since we can imagine cases where `xp_text_in_table()` in put is intentionally ordered to exit early on the likeliest input.

MichaelChirico:
#2341 and #2341 have two small chips at the problem.

Another idea would be to try and ignore S3 generics, or to offer an argument like `name_regex` that filters exports for matching certain patterns e.g. "all letters" which could default to e.g. `^[a-zA-Z0-9]+` i.e. excluding `[._]`.

```r
sum(grepl(paste0("^(", paste(Filter(\(x) make.names(x) == x, .base_s3_generics), collapse = "|"), ")[.]"), pkg_exports$name))
[1] 373sum(grepl(paste0("^(", paste(Filter(\(x) make.names(x) == x, .base_s3_generics), collapse = "|"), ")[.]"), pkg_exports$name))
# [1] 373
```

i.e. 373 (~15%) among defaults look like methods to base S3 generics

AshesITR:
I think the main problem is we compile a giant XPath instead of taking an approach similar to object_name_linter and extracting relevant assigned symbols in the actual code.

MichaelChirico:
maybe this should just be combined with object_name_linter?

AshesITR:
[...]

PR #2341, opened 2023-11-23
Drop duplicates from list of names in object_overwrite_linter()

MichaelChirico:
Related to #2337

This does not have much impact (there's only 3 overwrites in the default packages), but should be done anyway.


IndrajeetPatil (approved):
LGTM

PR #2342, opened 2023-11-23
Drop non-syntactic names from list in object_overwrite_linter()

MichaelChirico:
Related to #2337 

Such names (e.g. `base::body<-`) need special handling anyway that's not yet covered.

We could toggle whether to include such names, but I think we can leave that for if it's actually requested.

About 7% of exports from default packages fit this filter, so there should be some improvement, though I don't expect anything dramatic.

### DIFF
--- a/R/object_overwrite_linter.R
+++ b/R/object_overwrite_linter.R
@@ -70,7 +70,7 @@ object_overwrite_linter <- function(
     # .__C__ etc.: drop 150+ "virtual" names since they are very unlikely to appear anyway
     !grepl("^[.]__[A-Z]__", pkg_exports$name) &
     # exclude non-syntactic names. this might be supported upon user request.
-    make.names(pkg_exports$name) == pkg_exports$name
+    make.names(pkg_exports$name) == pkg_exports$name,
   ]
 
   # test that the symbol doesn't match an argument name in the function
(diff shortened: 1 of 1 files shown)


AshesITR:
This will be obsolete with #2344 I think, since I use `get_r_string()` there so overwrites of non-syntactic names should also be caught.

This reminds me that could use some tests.

MichaelChirico:
> This will be obsolete with #2344 I think, since I use `get_r_string()` there so overwrites of non-syntactic names should also be caught.

Not really -- note that `object_name_linter()` (IINM) uses the `symbols` regex for this:

```r
parse(text = "`+` <- 2") |>
  xmlparsedata::xml_parse_data() |>
  xml2::read_xml() |>
  xml2::xml_find_all("//SYMBOL") |>
  get_r_string()
# [1] "`+`"

"+" %in% getNamespaceExports("base")
# [1] TRUE
```

I intentionally left out tests because I'd rather leave this as an implementation detail. If you think it's still worth catching overwrites like this, please file an issue & I'll close this PR.

AshesITR:
Filed #2346

AshesITR:
#2347 makes this obsolete for good, now. Closing.

PR #2344, opened 2023-11-23
Speed up object_overwrite_linter

AshesITR:
```r
system.time(lint_package(linters = object_overwrite_linter()))
# this
#   user  system elapsed 
#  5.744   0.010   5.752 

# main
#   user  system elapsed 
# 18.869   0.026  18.896 
```

While working on it, I wondered, if the lint message could be made more elegant.
WDYT about

`"'fun' overwrites 'pkg::fun'. Avoid overwriting symbols from other packages."`

### DIFF
--- a/R/object_overwrite_linter.R
+++ b/R/object_overwrite_linter.R
@@ -104,8 +104,10 @@ object_overwrite_linter <- function(
     assigned_symbols <- get_r_string(assigned_exprs, "SYMBOL")
     is_bad <- assigned_symbols %in% pkg_exports$name
     source_pkg <- pkg_exports$package[match(assigned_symbols[is_bad], pkg_exports$name)]
-    lint_message <-
-      sprintf("'%s' is an exported object from package '%s'. Avoid re-using such symbols.", bad_symbol, source_pkg)
+    lint_message <- sprintf(
+      "'%s' is an exported object from package '%s'. Avoid re-using such symbols.",
+      assigned_symbols[is_bad], source_pkg
+    )
 
     xml_nodes_to_lints(
       assigned_exprs[is_bad],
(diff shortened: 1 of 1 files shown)


IndrajeetPatil:
Maybe `Avoid overwriting symbols from base/recommended R packages.`?

"Other packages" leaves too much room for interpretation.

AshesITR:
`packages=` is configurable. Maybe `Avoid overwriting symbols from this package.`?

IndrajeetPatil:
Yes, that sounds much better to me.

AshesITR:
@MichaelChirico WDYT? We can do this in one PR or file for follow-up.

R/object_overwrite_linter.R
```diff
@@ -101,14 +100,17 @@ object_overwrite_linter <- function(
...
-    bad_symbol <- xml_text(xml_find_first(bad_expr, "SYMBOL"))
-    source_pkg <- pkg_exports$package[match(bad_symbol, pkg_exports$name)]
-    lint_message <-
-      sprintf("'%s' is an exported object from package '%s'. Avoid re-using such symbols.", bad_symbol, source_pkg)
+    assigned_exprs <- xml_find_all(xml, xpath_assignments)
+    assigned_symbols <- get_r_string(assigned_exprs, "SYMBOL")
+    is_bad <- assigned_symbols %in% pkg_exports$name
+    source_pkg <- pkg_exports$package[match(assigned_symbols[is_bad], pkg_exports$name)]
+    lint_message <- sprintf(
+      "'%s' is an exported object from package '%s'. Avoid re-using such symbols.",
```
MichaelChirico, line 108: Copying here to put it all side-by-side:

> 'fun' overwrites 'pkg::fun'. Avoid overwriting symbols from other packages.

I don't like 'overwrites' which is misleading. 'masks' might be better. OTOH I don't see all that much wrong with the current message. Alternative:

> '%s' is exported by package '%s'. Avoid re-using function names for local symbols.

MichaelChirico (approved):
Approving now & we can continue discussion on the message in follow-up if needed (but feel free to add more commits to this PR)

PR #2347, opened 2023-11-23
fix detection of non-syntactic names in `object_overwrite_linter()`

### LINKED ISSUE #2346: Support non-syntactic names in `object_overwrite_linter()`
This will be obsolete with #2344 I think, since I use `get_r_string()` there so overwrites of non-syntactic names should also be caught.

This reminds me that could use some tests.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2342#issuecomment-1823892508_

AshesITR:
fixes #2346


R/object_overwrite_linter.R
```diff
@@ -102,6 +102,8 @@ object_overwrite_linter <- function(
 
     assigned_exprs <- xml_find_all(xml, xpath_assignments)
     assigned_symbols <- get_r_string(assigned_exprs, "SYMBOL")
+    is_quoted <- startsWith(assigned_symbols, "`")
```
MichaelChirico, line 105: this won't catch assignments with ' or "

AshesITR: fixed. Won't bother with raw strings though.

AshesITR: Those weren't caught by the XPath before, because they are `//STR_CONST`. `get_r_string()` handles removal of `"`, `'` and even `r"()"`. I added tests for the first two cases.

MichaelChirico: > even r"()"

The whole original point of `get_r_string()` was to account for that! :smile: 

Almost surely there's at least one other linter using the same logic to find assigned symbols, probably that should be kept in sync. But also I'm sure there's dozens of such cases in our code base by now of XPaths with 99% overlap that should be relegated to helpers.

R/object_overwrite_linter.R
```diff
@@ -76,7 +76,7 @@ object_overwrite_linter <- function(
   # test that the symbol doesn't match an argument name in the function
   # NB: data.table := has parse token LEFT_ASSIGN as well
   xpath_assignments <- glue("
-    //SYMBOL[
+    (//SYMBOL | //STR_CONST)[
```
MichaelChirico, line 79: I wonder about the performance of `//STR_CONST` vs. `//LEFT_ASSIGN`. Actually per the original issue about avoiding `//expr`, they appear at roughly the same frequency, so probably it's not much difference.

https://github.com/r-lib/lintr/issues/1358#issue-1260855389

MichaelChirico (approved):

MichaelChirico:
I reproduce the `link-rot` issue with `ess.r-project.org` on my local browser, but I assume this is a transient issue. Ignoring for now.

R/object_overwrite_linter.R
AshesITR: Yeah, from what I remember, only `//expr` stood out to be exceptionally bad cost-wise.

PR #2352, opened 2023-11-24
Add linter_level Attribute to `Linter()`

### LINKED ISSUE #2351: Add `linter_level` to `Linter()` metadata.
Here's what cost > 1s in `lint()` during `lint_package()` currently:
image

We can see 30% of the time is spent in `xml_find_all()`.
What surprised me was `is_lint_level()` taking a whole 8s. That may be worth converting to an attribute of `Linter()`?

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/issues/2350#issuecomment-1825195389_

AshesITR:
fixes #2351

```r
system.time(lint_package())

# here
#   user  system elapsed 
# 76.062   0.196  76.302 

# main
#   user  system elapsed 
# 85.855   0.126  86.092 
```

### DIFF
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -44,10 +44,6 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
   replacement_map <- c(T = "TRUE", F = "FALSE")
 
   Linter(function(source_expression) {
-    if (!is_lint_level(source_expression, "expression")) {
-      return(list())
-    }
-
     bad_usage <- xml_find_all(source_expression$xml_parsed_content, usage_xpath)
     bad_assignment <- xml_find_all(source_expression$xml_parsed_content, assignment_xpath)
 
@@ -68,5 +64,5 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
       make_lints(bad_usage, "Use %s instead of the symbol %s."),
       make_lints(bad_assignment, "Don't use %2$s as a variable name, as it can break code relying on %2$s being %1$s.")
     )
-  })
+  }, linter_level = "expression")
 }
--- a/R/T_and_F_symbol_linter.R
+++ b/R/T_and_F_symbol_linter.R
@@ -44,8 +44,10 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
   replacement_map <- c(T = "TRUE", F = "FALSE")
 
   Linter(function(source_expression) {
-    bad_usage <- xml_find_all(source_expression$xml_parsed_content, usage_xpath)
-    bad_assignment <- xml_find_all(source_expression$xml_parsed_content, assignment_xpath)
+    xml <- source_expression$xml_parsed_content
+    if (is.null(xml)) return(list())
+    bad_usage <- xml_find_all(xml, usage_xpath)
+    bad_assignment <- xml_find_all(xml, assignment_xpath)
 
     make_lints <- function(expr, fmt) {
       symbol <- xml_text(expr)
--- a/R/is_lint_level.R
+++ b/R/is_lint_level.R
@@ -24,3 +24,23 @@ is_lint_level <- function(source_expression, level = c("expression", "file")) {
   required_key <- paste0(if (level == "file") "full_", "parsed_content")
   required_key %in% names(source_expression)
 }
+
+#' Determine whether a linter needs to run for a given source_expression level
+#'
+#' Used by [lint()] to avoid unneccessary calls to linters.
+#'
+#' @param linter A linter.
+#' @param level Which level of expression is being tested? `"expression"`
+#'   means an individual expression, while `"file"` means all expressions
+#'   in the current file are available.
+#'
+#' @keywords internal
+#' @noRd
+is_linter_level <- function(linter, level = c("expression", "file")) {
+  linter_level <- attr(linter, "linter_level", exact = TRUE)
+  if (is.null(linter_level) || is.na(linter_level)) {
+    return(TRUE)
+  }
+  level <- match.arg(level)
+  identical(linter_level, level)
+}
--- a/R/lint.R
+++ b/R/lint.R
@@ -70,10 +70,18 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings = 
     return(exclude(lints, lines = lines, linter_names = names(linters), ...))
   }
 
+  file_linter_names <- names(linters)[vapply(linters, is_linter_level, logical(1L), "file")]
+  expression_linter_names <- names(linters)[vapply(linters, is_linter_level, logical(1L), "expression")]
+
   lints <- list()
   if (!is_tainted(source_expressions$lines)) {
     for (expr in source_expressions$expressions) {
-      for (linter in names(linters)) {
+      if (is_lint_level(expr, "expression")) {
+        necessary_linters <- expression_linter_names
+      } else {
+        necessary_linters <- file_linter_names
+      }
+      for (linter in necessary_linters) {
         # use withCallingHandlers for friendlier failures on unexpected linter errors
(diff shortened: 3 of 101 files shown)


MichaelChirico:
FYI I won't be available for detailed review for some days

AshesITR:
Thanks for the heads-up

R/utils.R
```diff
@@ -158,15 +158,22 @@ reset_lang <- function(old_lang) {
...
 #' Lints produced by the linter will be labelled with `name` by default.
+#' @param linter_level Which level of expression is the linter working with?
+#'   `"expression"` means an individual expression in `xml_parsed_content`, while `"file"` means all expressions
+#'   in the current file are available in `full_xml_parsed_content`.
+#'   `NA` means the linter will be run with both, expression-level and file-level source expressions.
+#'
 #' @return The same function with its class set to 'linter'.
 #' @export
-Linter <- function(fun, name = linter_auto_name()) { # nolint: object_name.
+Linter <- function(fun, name = linter_auto_name(), linter_level = c(NA_character_, "file", "expression")) { # nolint: object_name, line_length.
```
IndrajeetPatil, line 168: Nit: Instead of turning off the line length linter, why not just have one argument per line here?

IndrajeetPatil (approved):
This looks great! Definitely an appreciable speed-up.

(I've a nitpicky comment, but don't feel strongly either way.)

R/utils.R
AshesITR: I felt it looked less cluttered here because all that exceeds the line are nolint statements.
I worry that a `# nolint next` would mess up roxygen here. Maybe time for a `# nolint prev`? 😁

R/T_and_F_symbol_linter.R
```diff
@@ -68,5 +66,5 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
       make_lints(bad_usage, "Use %s instead of the symbol %s."),
       make_lints(bad_assignment, "Don't use %2$s as a variable name, as it can break code relying on %2$s being %1$s.")
     )
-  })
+  }, linter_level = "expression")
```
MichaelChirico, line 69: I'm not a fan of this style, WDYT about rearranging instead:

```r
  ...
})
```

R/is_lint_level.R
```diff
@@ -24,3 +24,23 @@ is_lint_level <- function(source_expression, level = c("expression", "file")) {
...
+#' @param linter A linter.
+#' @param level Which level of expression is being tested? `"expression"`
+#'   means an individual expression, while `"file"` means all expressions
+#'   in the current file are available.
+#'
+#' @keywords internal
+#' @noRd
+is_linter_level <- function(linter, level = c("expression", "file")) {
+  linter_level <- attr(linter, "linter_level", exact = TRUE)
+  if (is.null(linter_level) || is.na(linter_level)) {
```
MichaelChirico, line 41: `if` will give a bad error if `length(linter_level) != 1L`, worth adjusting? `if (is.null(linter_level) || (length(linter_level) == 1L && is.na(linter_level)))` or `if (is.null(linter_level) && identical(linter_level, NA))`

R/is_lint_level.R
```diff
@@ -24,3 +24,23 @@ is_lint_level <- function(source_expression, level = c("expression", "file")) {
   required_key <- paste0(if (level == "file") "full_", "parsed_content")
```
MichaelChirico, line 24: Should `is_lint_level()` be deprecated?

R/T_and_F_symbol_linter.R
AshesITR: SGTM I'll handle this in the next pass.

R/lint.R
```diff
@@ -70,10 +70,18 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
     return(exclude(lints, lines = lines, linter_names = names(linters), ...))
   }
 
+  file_linter_names <- names(linters)[vapply(linters, is_linter_level, logical(1L), "file")]
+  expression_linter_names <- names(linters)[vapply(linters, is_linter_level, logical(1L), "expression")]
```
MichaelChirico, line 74: what happens to linters lacking `linter_level`, exactly?  Won't they be missing from `necessary_linters` in all cases?

R/is_lint_level.R
AshesITR: I think we can assume that the attribute isn't "hacked in". `Linter()` already takes care of validation.

R/is_lint_level.R
AshesITR: Not at the moment. We need it for linters that run on both levels.

AshesITR: But it could be sped up by avoiding a `paste0()` call.

R/lint.R
```diff
@@ -70,10 +70,18 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
...
+
   lints <- list()
   if (!is_tainted(source_expressions$lines)) {
     for (expr in source_expressions$expressions) {
-      for (linter in names(linters)) {
+      if (is_lint_level(expr, "expression")) {
```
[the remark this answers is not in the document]
AshesITR: It does, but the tests in `get_source_expression()` don't.
We could discuss removing them.

R/lint.R
AshesITR: They will be in both sets. That's the `is.null() || is.na()` branch in `is_linter_level()`.

R/utils.R
MichaelChirico: > I worry that a # nolint next would mess up roxygen here. 

It wouldn't, plain comments are simply skipped:

> Blocks ignore empty lines, including lines made up of non-roxygen comments.

https://roxygen2.r-lib.org/articles/rd.html#basics

R/utils.R
```diff
@@ -158,15 +158,22 @@ reset_lang <- function(old_lang) {
...
   if (!is.function(fun) || length(formals(args(fun))) != 1L) {
     stop("`fun` must be a function taking exactly one argument.", call. = FALSE)
   }
+  linter_level <- match.arg(linter_level)
```
MichaelChirico, line 172: I see, this ensures `length(linter_level) == 1`. Resolving the above.

R/utils.R
AshesITR: Good to know, thanks!

R/utils.R
MichaelChirico, line 172: NB:

```r
match.arg(NA, NA_character_)
# Error in match.arg(NA, NA_character_) : 
#   'arg' must be NULL or a character vector
```

Maybe `match.arg(as.character(linter_level))` is better?

R/utils.R
```diff
@@ -158,15 +158,22 @@ reset_lang <- function(old_lang) {
 #' @param fun A function that takes a source file and returns `lint` objects.
 #' @param name Default name of the Linter.
 #' Lints produced by the linter will be labelled with `name` by default.
+#' @param linter_level Which level of expression is the linter working with?
+#'   `"expression"` means an individual expression in `xml_parsed_content`, while `"file"` means all expressions
+#'   in the current file are available in `full_xml_parsed_content`.
+#'   `NA` means the linter will be run with both, expression-level and file-level source expressions.
```
MichaelChirico, line 164: We should mention that `NA` is really only available for back-compatibility, and using one of the others is preferable?

Also, note from below: unless we support (logical) `NA`, we need to clarify `NA_character_` here.

R/utils.R
AshesITR: Note that can omit `linter_level=` if you want `NA`. But yes, would be nicer to allow logical `NA` as well.

R/lint.R
MichaelChirico: Oh, got it.

Issue #2191, opened 2023-09-19
Lint consistency fuzz testing?


MichaelChirico:
One concern I have about the consistency cleanups in this release (#2039, #2046, #2190) is there's no mechanism to ensure future improvements/new linters continue to keep "up to code" & enforce the same consistency.

One approach that seems doable is to have a GHA that [fuzzes](https://en.wikipedia.org/wiki/Fuzzing) our test suite. For example, for #2190, that means:

 - Go through the suite and identify all `expect_lint()` calls using "function(...)" or `\(...)` in the `content=` argument (text matching should be enough, but we could also use `getParseData()` here...).
 - Randomly (or exhaustively?) switch all such usages to their opposite: `s/function/\/g` and `s/\/function/g`
 - Run the test suite
 - Any test failure means a linter is inconsistently handling `function(...)` and `\(...)`

Similar fuzzes could apply in the other cases:

 - Randomly swap `%>%` and `|>`
 - Randomly swap `$` and `@`

Some exceptions will need to be carved out, e.g. for cases where intentionally only one alternative is meant to be linted, and we might need some work to make the fuzzing robust to changes in an expected lint message. But overall this approach seems pretty doable to me.

If run with randomness, I'd also run it periodically, not on push/merge, since a failure might not be caught up-front & surfacing that failure on "someone else's" PR would be bad form.

AshesITR:
Interesting idea. Things to consider:

* Lint metadata may also change, especially expected column ranges.
* `|>` and `%>%` can't be swapped in all contexts; they also have different placeholder semantics (`_` vs. `.`)

MichaelChirico:
True. Let's see how relevant those turn out in practice. Another idea is to allow for a `# nofuzz` escape to skip certain tests if they're too onerous to accomodate.

MichaelChirico:
we can also work on #2221 here by fuzzing the lint metadata & making sure tests fail.

MichaelChirico:
Another idea for the fuzz/metatesting suite, based on #2402 -- inject comments in random places in the test code and ensure the linters keep working. Have to exclude linters where `<COMMENT>` is meaningful, of course.

AshesITR:
Another fuzzer idea: swap `=`, `<-` and `->`, and maybe the double-assignment arrows.

IndrajeetPatil:
Of interest: there is also a new package for mutation testing in R (https://jakubsobolewski.com/muttest/)

Bisaloo:
[...]

PR #2357, opened 2023-11-27
add `xml_find_function_calls()` helper to source expressions

### LINKED ISSUE #2338: Investigate performance of backport_linter()
#2328 found `backport_linter()` to be among the slowest in the suite. This is surprising to me, it seems like a relatively simple linter.

### LINKED ISSUE #2350: Re-Use (and cache?) common XPaths
Inspired by a comment by @MichaelChirico which I can't find at the moment.

Some linters share common XPath logic. This is currently in no way cached, so all linters have to re-evaluate the common XPath logic.
It might be worth to

1. Figure out which XPaths could be made common.
2. Figure out a way to cache the result of applying the XPaths to expressions / files so they are only evaluated once.

Related to #963

AshesITR:
fixes #2350
fixes #2338

- [x] Implement helper
- [x] Document
- [x] Add NEWS bullet
- [x] Use in our linters
- [x] Compare performance
- [x] Add section in custom linters vignette

Filing for early feedback and discussion


R/get_source_expressions.R
```diff
@@ -18,32 +18,36 @@
...
 #'     \item{`full_parsed_content` (`data.frame`) as given by
-#'           [utils::getParseData()] for the full content}
+#'           [utils::getParseData()] for the full content.}
 #'     \item{`full_xml_parsed_content` (`xml_document`) the XML parse tree of all
-#'           expressions as given by [xmlparsedata::xml_parse_data()]}
+#'           expressions as given by [xmlparsedata::xml_parse_data()].}
 #'     \item{`terminal_newline` (`logical`) records whether `filename` has a terminal
-#'           newline (as determined by [readLines()] producing a corresponding warning)}
+#'           newline (as determined by [readLines()] producing a corresponding warning).}
+#'     \item{`xml_find_function_calls(function_names)` (`function`) a function that returns all `SYMBOL_FUNCTION_CALL`
```
MichaelChirico, line 49: Not sure I like the approach of making this a factory, WDYT about just making this an entry with the named `xml_find_all()` output?

Another suggestion is making this a 3rd "linter level" (file>expression>call). Since these are pointers to nodes in the XML tree the overhead should not be crazy. OTOH copying the subsetted `getParseData()` data.frame will blow up the object size, but we could also pursue a breaking change whereby instead of fully subsetting the file-wide data.frame, the nested levels just consist of row indices on the full data.frame. So e.g. instead of a data.frame at expression level, we have an integer vector like `idx=c(1, 8, 12, 15)`, and `FullDF[idx,]` gives the subset we used to store earlier. Storing a bunch of integer vectors should be more efficient (and we hardly use this object anymore anyway). This is not a fully-fleshed-out idea but something kicking around in my head.

Would also like to start getting a sense of performance comparison. Normal (any random R file) and extreme (tools/R/QC.R) cases.

AshesITR: The named call list has the disadvantage of exposing the implementation detail.
If we ever find a faster way to do it, or a fast way to add additional conditions (as often is done), we can't extend this feature.

re Performance: Sure, atm it's not integrated in any linter so will be slightly slower. Trying to suss out the API before creating a huge diff.

AshesITR: NB a back-compatible change for smaller data structures could be using `rsample::resample()`. Not sure we should do that, tough.

MichaelChirico:
this should probably get a section in the custom linters vignette, I don't see that on your TODO right now

AshesITR:
so far 118s on main vs 113s here for
```r
devtools::load_all()
system.time(lintr::lint_package())
```

So 5 secs savings ATM with only some linters converted.

AshesITR:
We have quite a few `//SYMBOL_FUNCTION_CALL[text() = ...]/path/1 | //SYMBOL_FUNCTION_CALL[text() = ...]/path/2` XPaths.
ATM, I convert them using `self::` and the union of the two function calls examined.

It might be worth splitting these into to separate paths with separate calls to `xml_find_function_call()`. Not sure about the performance gains of that.

AshesITR:
@MichaelChirico PTAL for another round.
The technical part of this PR is done I think.

If you approve of the changes so far, I would flesh out the documentation.

MichaelChirico:
> @MichaelChirico PTAL for another round. The technical part of this PR is done I think.
> 
> If you approve of the changes so far, I would flesh out the documentation.

I was also hoping for the vignette change to serve as a high-level overview for guiding the review 😅

It would help to leave that in the main PR comment anyway for future reference on such a big change.

For now, I'll try reading the tests and see if that suffices.

tests/testthat/test-get_source_expressions.R
```diff
@@ -218,7 +218,10 @@ test_that("returned data structure is complete", {
 
   for (i in seq_along(lines)) {
     expr <- exprs$expressions[[i]]
-    expect_named(expr, c("filename", "line", "column", "lines", "parsed_content", "xml_parsed_content", "content"))
+    expect_named(expr, c(
+      "filename", "line", "column", "lines", "parsed_content", "xml_parsed_content", "content",
+      "xml_find_function_calls"
```
MichaelChirico, line 223: WDYT about naming `xml_` vs. `xp_` here? I don't have an immediate preference but we do have `xp_call_name()` exported so worth considering.

tests/testthat/test-get_source_expressions.R
```diff
@@ -245,6 +249,47 @@ test_that("returned data structure is complete", {
...
+  )
+
+  expect_length(exprs$expressions[[2L]]$xml_find_function_calls("foo"), 0L)
+  expect_length(exprs$expressions[[2L]]$xml_find_function_calls("bar"), 1L)
+
+  expect_length(exprs$expressions[[4L]]$xml_find_function_calls("foo"), 2L)
+  expect_length(exprs$expressions[[4L]]$xml_find_function_calls("bar"), 1L)
+  expect_length(exprs$expressions[[4L]]$xml_find_function_calls(c("foo", "bar")), 3L)
+
+  expect_length(exprs$expressions[[5L]]$xml_find_function_calls("foo"), 4L)
```
MichaelChirico, line 272: this is the file-level expression, right? worth a comment I think

R/sprintf_linter.R
```diff
@@ -105,9 +104,7 @@ sprintf_linter <- function() {
   }
 
   Linter(linter_level = "file", function(source_expression) {
-    xml <- source_expression$full_xml_parsed_content
```
MichaelChirico, line 108: why drop this local variable?

I would keep `xml_calls <- .$xml_find_function_calls(.)`. Having this variable stored is very helpful for debugging+development IME.

R/source_utils.R
```diff
@@ -0,0 +1,21 @@
+#' Build the `xml_find_function_calls()` helper for a source expression
+#'
+#' @param xml The XML parse tree as an XML object (`xml_parsed_content` or `full_xml_parsed_content`)
+#'
+#' @return A fast function to query
+#' `xml_find_all(xml, glue::glue("//SYMBOL_FUNCTION_CALL[{ xp_text_in_table(function_names) }]"))`.
```
MichaelChirico, line 6: NB: `xp_text_in_table()` isn't (yet?) exported. Not important b/c this is `@noRd` right?

R/source_utils.R
```diff
@@ -0,0 +1,21 @@
...
+#'
+#' @noRd
+build_xml_find_function_calls <- function(xml) {
+  function_call_cache <- xml_find_all(xml, "//SYMBOL_FUNCTION_CALL")
+  names(function_call_cache) <- get_r_string(function_call_cache)
```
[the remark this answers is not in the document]
MichaelChirico: I think `xml` is still there:

```r
foo <- function(x) { function() x }
foo(1)()
# [1] 1
```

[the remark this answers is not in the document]
MichaelChirico: Oh, nvm, I see it's needed every time `!is.null(function_names)`, which is a lot of the point of this functionality, so better to calculate up front. Resolving.

MichaelChirico:
> so far 118s on main vs 113s here for
> 
> ```r
> devtools::load_all()
> system.time(lintr::lint_package())
> ```
> 
> So 5 secs savings ATM with only some linters converted.

Overall change looks pretty good, I think we're close enough to ready that we can re-run the benchmark + start on user-facing docs.

R/source_utils.R
AshesITR: Added another example without `xp_text_in_table()` just to be sure.

tests/testthat/test-get_source_expressions.R
AshesITR: I thought about that, but we're replacing `xml_find_all()`, so I went with the `xml_` prefix.

AshesITR:
Another performance timing on lintr itself is 117s [main] vs 101s [here] => 14% Speedup 🥳

LMK if you want more comparisons before merging. Otherwise, I think it's good to go.

tests/testthat/test-get_source_expressions.R
MichaelChirico: SGTM, thanks

NEWS.md
```diff
@@ -40,6 +40,7 @@
 * `string_boundary_linter()` recognizes regular expression calls like `grepl("^abc$", x)` that can be replaced by using `==` instead (#1613, @MichaelChirico).
 * `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
 * `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
+* The `source_expression` passed to linters gains a fast way to query function call nodes using `source_expression$xml_find_function_calls()`. Use this to speed up linters using XPaths that start with `//SYMBOL_FUNCTION_CALL`. (@AshesITR, #2357)
```
MichaelChirico, line 43: no mention of the new exported function?

MichaelChirico: the NEWS should mention "See creating linters vignette for more details". It might help to have an illustrative benchmark as well.

MichaelChirico:
> Another performance timing on lintr itself is 117s [main] vs 101s [here] => 14% Speedup 🥳
> 
> LMK if you want more comparisons before merging. Otherwise, I think it's good to go.

Quite nice! My initial reaction was "Hmm, doesn't seem like all that much for such a big effort", but we must keep in mind this is an end-to-end benchmark. It would also be good to include one or two "focused" benchmarks, e.g. before/after for a few linters. I guess the most improvement comes in a case like `object_overwrite_linter()`? And cite a representative number in the NEWS.

NEWS.md
MichaelChirico: I would rework the NEWS bullet generally, it would be good to share some motivation for the change as it's pretty significant! We re-worked a large number of our exported linters to use this approach -- it's a bit of a paradigm shift for how downstreams should use {lintr} to write their own linters/XPaths. So motivating it well is important to drive adoption.

To me there's two big things (1) `//SYMBOL_FUNCTION_CALL` is easily the most common XPath component we see (2) It does not scale well for finding a medium-to-large number of functions since the `or` operator is not optimized in `libxml2`, as we saw for `backports_linter()`. Those insights make this new functionality a natural direction for new development, rather than us just releasing new features for the fun of it :)

R/any_duplicated_linter.R
```diff
@@ -35,8 +35,7 @@
 #' @export
 any_duplicated_linter <- function() {
   any_duplicated_xpath <- "
-  //SYMBOL_FUNCTION_CALL[text() = 'any']
-    /parent::expr
+  parent::expr
```
MichaelChirico, line 38: WDYT about `xml_find_function_calls()` landing on `//SYMBOL_FUNCTION_CALL/parent::expr`? It seems like most XPaths start there.

R/backport_linter.R
```diff
@@ -45,12 +45,13 @@ backport_linter <- function(r_version = getRversion(), except = character()) {
...
   names(backport_index) <- unlist(backport_blacklist)
 
-  names_xpath <- "//SYMBOL | //SYMBOL_FUNCTION_CALL"
-
   Linter(linter_level = "expression", function(source_expression) {
     xml <- source_expression$xml_parsed_content
 
-    all_names_nodes <- xml_find_all(xml, names_xpath)
+    all_names_nodes <- combine_nodesets(
+      source_expression$xml_find_function_calls(NULL),
```
MichaelChirico, line 52: may as well build in the idea from #2365 here? then we can mark that PR as superseded+close when this is merged.

R/backport_linter.R
```diff
@@ -45,12 +45,13 @@ backport_linter <- function(r_version = getRversion(), except = character()) {
   backport_index <- rep(names(backport_blacklist), times = lengths(backport_blacklist))
   names(backport_index) <- unlist(backport_blacklist)
 
-  names_xpath <- "//SYMBOL | //SYMBOL_FUNCTION_CALL"
-
   Linter(linter_level = "expression", function(source_expression) {
     xml <- source_expression$xml_parsed_content
 
-    all_names_nodes <- xml_find_all(xml, names_xpath)
+    all_names_nodes <- combine_nodesets(
```
MichaelChirico, line 51: might suggest this as a helper for {xml2} directly, it seems generally useful

R/boolean_arithmetic_linter.R
```diff
@@ -35,30 +35,30 @@ boolean_arithmetic_linter <- function() {
...
   ")
-  any_xpath <- paste(length_xpath, "|", sum_xpath)
 
   Linter(linter_level = "expression", function(source_expression) {
-    xml <- source_expression$xml_parsed_content
-
-    any_expr <- xml_find_all(xml, any_xpath)
+    length_calls <- source_expression$xml_find_function_calls(c("which", "grep"))
+    sum_calls <- source_expression$xml_find_function_calls("sum")
+    any_expr <- c(
```
MichaelChirico, line 58: Oh, I thought `c()` wouldn't work like this, why can we use `c()` here but need `combine_nodesets()` in `backport_linter()`?

R/condition_message_linter.R
```diff
@@ -44,9 +44,8 @@
 condition_message_linter <- function() {
   translators <- c("packageStartupMessage", "message", "warning", "stop")
   xpath <- glue("
-  //SYMBOL_FUNCTION_CALL[
-    ({xp_text_in_table(translators)})
-    and not(preceding-sibling::OP-DOLLAR or preceding-sibling::OP-AT)
+  self::SYMBOL_FUNCTION_CALL[
+    not(preceding-sibling::OP-DOLLAR or preceding-sibling::OP-AT)
```
MichaelChirico, line 48: WDYT about a new argument to `xml_find_function_calls()` to drop calls in extractions? I would do so by default. That would be another way to get at #2191 by only including extracted calls explicitly, since we usually don't want them.

R/expect_identical_linter.R
```diff
@@ -74,17 +73,18 @@ expect_identical_linter <- function() {
...
     /following-sibling::expr[1][expr[1]/SYMBOL_FUNCTION_CALL[text() = 'identical']]
     /parent::expr
   "
-  xpath <- paste(expect_equal_xpath, "|", expect_true_xpath)
-
   Linter(linter_level = "expression", function(source_expression) {
-    xml <- source_expression$xml_parsed_content
+    expect_equal_calls <- source_expression$xml_find_function_calls("expect_equal")
+    expect_true_calls <- source_expression$xml_find_function_calls("expect_true")
+    bad_expr <- c(
```
MichaelChirico, line 83: Another `c()` inconsistent with `combine_nodesets()`

R/expect_not_linter.R
```diff
@@ -22,10 +22,10 @@
 #' @evalRd rd_tags("expect_not_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-expect_not_linter <- make_linter_from_xpath(
+expect_not_linter <- make_linter_from_function_xpath(
```
MichaelChirico, line 25: Is it too much overload to just add an argument to `make_linter_from_xpath()` to handle this instead of a new function?

R/expect_s4_class_linter.R
```diff
@@ -26,19 +26,18 @@ expect_s4_class_linter <- function() {
...
-    /parent::expr
+  parent::expr
     /following-sibling::expr[1][count(expr) = 3 and expr[1][SYMBOL_FUNCTION_CALL[text() = 'is']]]
     /parent::expr[not(SYMBOL_SUB[text() = 'info' or text() = 'label'])]
   "
 
   Linter(linter_level = "expression", function(source_expression) {
-    xml <- source_expression$xml_parsed_content
-
     # TODO(michaelchirico): also catch expect_{equal,identical}(methods::is(x), k).
```
MichaelChirico, line 35: Just noticed this TODO again, checked on GitHub... there are plenty of hits but it's not epidemic:

https://github.com/search?q=lang%3AR%20%2Fexpect_(equal%7Cidentical)%5B(%5D%5Cs*(methods%3A%3A)%3Fis%5B(%5D%2F&type=code

MichaelChirico: #2423

MichaelChirico:
```suggestion
    # TODO(#2423): also catch expect_{equal,identical}(methods::is(x), k).
```

R/get_source_expressions.R
MichaelChirico: Happy with the current implementation, resolving.

R/missing_argument_linter.R
```diff
@@ -44,16 +44,16 @@ missing_argument_linter <- function(except = c("alist", "quote", "switch"), allo
...
-    //SYMBOL_FUNCTION_CALL[not({ xp_text_in_table(except) })]
-      /parent::expr
+    parent::expr
       /parent::expr[count(*) > 3]
       /*[{xp_or(conds)}]
   ")
 
   Linter(linter_level = "file", function(source_expression) {
-    xml <- source_expression$full_xml_parsed_content
+    xml_targets <- source_expression$xml_find_function_calls(NULL, keep_names = TRUE)
```
MichaelChirico, line 53: Maybe cleaner to have a parameter `drop=` or `exclude=` for `xml_find_function_calls()`?

R/string_boundary_linter.R
```diff
@@ -119,8 +117,8 @@ string_boundary_linter <- function(allow_grepl = FALSE) {
     list(lint_expr = expr[can_replace], lint_type = lint_type)
   }
 
-  substr_xpath_parts <- glue("
-  //{ c('EQ', 'NE') }
+  substr_xpath <- glue("
```
MichaelChirico, line 120: 😎

R/system_file_linter.R
```diff
@@ -28,16 +28,19 @@ system_file_linter <- function() {
...
     /parent::expr
   ")
-  xpath <- paste(xpath_parts, collapse = " | ")
 
   Linter(linter_level = "expression", function(source_expression) {
-    xml <- source_expression$xml_parsed_content
-
-    bad_expr <- xml_find_all(xml, xpath)
+    xml_calls <- list(
+      source_expression$xml_find_function_calls(funs[1L]),
```
MichaelChirico, line 37: readability has gone down a wee bit here, I don't think `funs` is likely to change in the future, with the new functionality I think it'll be clearer to just write out `"system.file"`/`"file.path"` more explicitly?

vignettes/creating_linters.Rmd
```diff
@@ -274,6 +274,21 @@ and so they've been tested and demonstrated their utility already.
    extract the string exactly as R will see it. This is especially important to make
    your logic robust to R-4-style raw strings like `R"-(hello)-"`, which is otherwise
    difficult to express, for example as an XPath.
+ * `xml_find_function_calls()`: Whenever your linter needs to query R function calls,
```
MichaelChirico, line 277: Maybe note this in the Wiki too?

https://github.com/r-lib/lintr/wiki/Writing-robust,-performant-linters

MichaelChirico:
Really beautiful stuff! Quite excited about this new feature!!

R/missing_argument_linter.R
AshesITR: What would this parameter do?

R/boolean_arithmetic_linter.R
AshesITR: `xml_nodes_to_lints()` accepts plain lists too. The others use xml2 functions which are only defined as a generic for `xml_nodeset`

R/system_file_linter.R
AshesITR: okay, will do.

vignettes/creating_linters.Rmd
AshesITR: Done.

R/any_duplicated_linter.R
AshesITR: It makes the docs harder to read and makes it more clunky to add additional conditions to the `SYMBOL_FUNCTION_CALL` node.
At the moment, migrating is "remove `//SYMBOL_FUNCTION_CALL[text() = foo]/` and run the xpath on `xml_find_function_calls("foo")`. If you need an additional condition, replace `//SYMBOL_FUNCTION_CALL[text() = 'foo' and cond]` by `self::SYMBOL_FUNCTION_CALL[cond]`.

R/backport_linter.R
AshesITR: I'd suggest `c.xml_nodeset` to be implemented. No need for a custom function IMO.

AshesITR: https://github.com/r-lib/xml2/issues/433

R/boolean_arithmetic_linter.R
AshesITR: I'd replace `combine_nodesets()` by `c()` when https://github.com/r-lib/xml2/issues/433 is resolved instead.

R/condition_message_linter.R
AshesITR: Sounds good for follow-up.
Need to think a bit about the best way to store additional metadata for function calls.
Usage in pipes seems interesting to track, too.

R/expect_identical_linter.R
AshesITR: See above, I'd only use `combine_nodesets()` if truly necessary as I expect `c()` to actually work for all use cases some time in the future.

R/expect_not_linter.R
AshesITR: It would require casework in the `Linter()` itself, harming performance. Or some `eval(bquote())` trickery.
I'd rather not do either of those.

R/any_duplicated_linter.R
MichaelChirico: > makes it more clunky to add additional conditions to the SYMBOL_FUNCTION_CALL node.

A little bit, but it's only needed rarely, e.g. a handful of times in our codebase. More common was conditions on `parent::expr`, and from what I saw many of those could be migrated as well, e.g. instead of `[SYMBOL_FUNCTION_CALL]/parent::expr[following-sibling::expr[...]]` it is `[SYMBOL_FUNCTION_CALL/parent::expr]/parent::expr[expr[...]]`. Often that `parent::expr` is part of the XPath already so the result is even more concise.

> It makes the docs harder to read

Yes, a little, but I wonder if it's not more natural to land on `<expr>` nodes in general. They're the basic atoms of the AST, I know we used to write `//expr[SYMBOL_FUNCTION_CALL[...]]` by default until we realized that's inefficient & got used to `//SYMBOL_FUNCTION_CALL`.

BTW one might infer from the name `find_function_calls()` that we actually land on the _next_ outer `<expr>`, i.e. `//SYMBOL_FUNCTION_CALL/parent::expr/parent::expr`, since that node is where the full "call" is found. I don't have a better naming suggestion.

R/missing_argument_linter.R
MichaelChirico: conjugate `function_names`, i.e. `function_names` sets which calls to _include_, this sets which calls to _exclude_. Then in this step we can do `$xml_find_function_calls(exclude = except)`, done.

We could also use it in `return_linter()`:

https://github.com/r-lib/lintr/blob/0f0b17c367e305ba2a3316f5cc532e1e2fa1e0c5/R/return_linter.R#L94

R/boolean_arithmetic_linter.R
MichaelChirico: Got it... it will work to use `combine_nodesets()` anyway, right? I might do so just for consistency' sake.

NEWS.md
```diff
@@ -40,7 +40,11 @@
 * `string_boundary_linter()` recognizes regular expression calls like `grepl("^abc$", x)` that can be replaced by using `==` instead (#1613, @MichaelChirico).
 * `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
 * `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
-* The `source_expression` passed to linters gains a fast way to query function call nodes using `source_expression$xml_find_function_calls()`. Use this to speed up linters using XPaths that start with `//SYMBOL_FUNCTION_CALL`. (@AshesITR, #2357)
+* The full linter suite is roughly 14% faster due to caching of the frequently used `//SMBOL_FUNCTION_CALL` XPath to examine function calls. (@AshesITR, #2357)
```
MichaelChirico, line 43: (unrelated to PR)

Oh, @AshesITR I think I see what's happening. The `suggestion` UI button doesn't appear on this page, I guess because it's for a specific commit and not the overall PR. That's probably why it wasn't showing up on mobile.

I have seen the UI button on mobile when looking at the overall PR page.

NEWS.md
```diff
@@ -40,6 +40,11 @@
 * `string_boundary_linter()` recognizes regular expression calls like `grepl("^abc$", x)` that can be replaced by using `==` instead (#1613, @MichaelChirico).
 * `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
 * `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
+* The full linter suite is roughly 14% faster due to caching of the frequently used `//SMBOL_FUNCTION_CALL` XPath to examine function calls. (@AshesITR, #2357)
```
MichaelChirico, line 43:
```suggestion
* The full linter suite is roughly 14% faster due to caching of the frequently used `//SYMBOL_FUNCTION_CALL` XPath to examine function calls. (@AshesITR, #2357)
```

R/missing_argument_linter.R
AshesITR: Would file that for follow-up to keep focused, if that is okay with you.

R/boolean_arithmetic_linter.R
AshesITR: Yes, it will.

R/backport_linter.R
AshesITR: done.

NEWS.md
AshesITR: better?

NEWS.md
```diff
@@ -40,6 +40,11 @@
 * `string_boundary_linter()` recognizes regular expression calls like `grepl("^abc$", x)` that can be replaced by using `==` instead (#1613, @MichaelChirico).
 * `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
 * `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
+* The full linter suite is roughly 14% faster due to caching of the frequently used `//SYMBOL_FUNCTION_CALL` XPath to examine function calls. (@AshesITR, #2357)
```
MichaelChirico, line 43: this still reads like a note to ourselves rather than an announcement of a cool new feature. How about:

```
* New function node caching for big efficiency gain to most linters (e.g. up to xx%; #2357, @AshesITR). Most linters are written around usage of functions, and XPath performance searching for many functions is poor. The new `xml_find_function_calls()` entry in the `get_source_expressions()` output caches all function call nodes instead. See the vignette on creating linters for more details on how to use it. {lintr} itself saw an overall 14% improvement for running the full linter suite as a result of migrating linters to this style.
```

AshesITR: Excellent. Feel free to commit that.

R/backport_linter.R
MichaelChirico: Great. If not, we might want to export our helper to facilitate downstreams using it since it comes up a fair amount in our suite.

NEWS.md
MichaelChirico: I need an `xx`, could you help with that? I had in mind comparing `object_overwrite_linter()` specifically.

NEWS.md
MichaelChirico: (threading got lost, alas, continued: https://github.com/r-lib/lintr/pull/2357#discussion_r1424908130. resolving this one)

R/boolean_arithmetic_linter.R
MichaelChirico: Just saw your latter comment, agree + resolving.

R/xp_utils.R
```diff
@@ -124,3 +124,14 @@ xpath_comment_re <- rex::rex(
   ":)"
 )
 xp_strip_comments <- function(xpath) rex::re_substitutes(xpath, xpath_comment_re, "", global = TRUE)
+
+#' Combine two or more nodesets to a single nodeset
+#'
+#' Useful for calling `{xml2}` functions on a combined set of nodes obtained using different XPath searches.
+#'
+#' @noRd
+combine_nodesets <- function(...) {
```
MichaelChirico, line 133:
```suggestion
TODO(r-lib/xml2#433): remove this and just use c()
combine_nodesets <- function(...) {
```

MichaelChirico: Alternatively, we could just go ahead and rename this `c.xml_nodeset()` and register the S3 method, right?

PS FWIW it looks like package {do} defines this method doing something totally different:

https://github.com/search?q=lang%3AR%20%2Fc%5B.%5Dxml_nodeset%2F&type=code

NEWS.md
AshesITR: Would we count the cache building time as well? `get_source_expressions()` got slower to get the speedup.

R/condition_message_linter.R
MichaelChirico: #2428

R/expect_not_linter.R
MichaelChirico: SGTM, thanks

NEWS.md
MichaelChirico: I think `lint_package(linters = object_overwrite_linter())` is the best comparison

R/xp_utils.R
AshesITR: Not "totally". It's a by-foot version of `c()` where attributes of `..1` are kept in tact.

R/missing_argument_linter.R
MichaelChirico: #2429

R/any_duplicated_linter.R
AshesITR: Not sure how we want to proceed here. Maybe follow-up?
I want to avoid risking merge conflicts on this huge diff.

R/missing_argument_linter.R
AshesITR: #2430

NEWS.md
AshesITR: 7s main, 8s here. The cache really only excels in E2E comparison for many linters using it.

AshesITR: Let's just drop the up to xx% then.

MichaelChirico: Oh, duh. That linter is not a good choice -- it's only looking at `//SYMBOL`, not `//SYMBOL_FUNCTION_CALL` 🤦. How about `backport_linter()`? Maybe not a good choice since this branch makes two changes, what about comparing to the #2365 implementation?

Or maybe `unnecessary_lambda_linter()`, since that one is searching for 10s of functions, right?

R/any_duplicated_linter.R
MichaelChirico: OK, SG

MichaelChirico: #2431

NEWS.md
AshesITR: The fundamental problem is that the caching effect is a fixed cost upfront so the more linters are active the greater the improvement.

Maybe dev/compare_branches gives a better basis to reason on?

MichaelChirico: I see, right then it's easy to cheat and find really impressive numbers but those comparisons wouldn't be fair. I think the 14% number is best, then, let me think how to re-word the NEWS.

MichaelChirico (dismissed):
Might be worth benchmarking on base R as well, here's a snippet I tried:

```r
system.time(withr::with_dir(
  "~/svn/r-devel/src/library",
  for (pkg in list.files()) if (file.exists(file.path(pkg, "DESCRIPTION")))
    lint_package(pkg, linters = linters)
))
```

But `linters` includes `cyclocomp_linter()` which probably isn't a good idea, wheels were spinning interminably. So maybe a benchmark on a "reasonable" set of linters instead of the full suite.

Approved now if you want to forego that.

MichaelChirico:
OK, I'm seeing 27% improvement on r-devel using our suite (minus `cyclocomp_linter()`):

### Setup

```r
get_linters <- \() modify_defaults(
  eval(parse(text = read.dcf(".lintr", "linters"))),
  cyclocomp_linter = NULL
)
linters <- get_linters(); system.time(withr::with_dir(
  "~/svn/r-devel/src/library",
  for (pkg in list.files()) if (file.exists(file.path(pkg, "DESCRIPTION")))
    lint_package(pkg, linters = linters)
))
```

### Timings

```
# main
    user   system  elapsed 
2239.148  327.508 2666.586 

# current branch
    user   system  elapsed 
1828.817   88.002 1949.443 
```

MichaelChirico (approved):

MichaelChirico (approved):

AshesITR:
Thanks @MichaelChirico for taking this to the finish line 👍🏻

R/make_linter_from_xpath.R
```diff
@@ -38,3 +40,36 @@ make_linter_from_xpath <- function(xpath,
     })
   }
 }
+
+#' @rdname make_linter_from_xpath
+#' @param function_names Character vector, names of functions whose calls to examine..
+#' @export
+# nolint next: object_length.
+make_linter_from_function_xpath <- function(function_names,
```
MichaelChirico, line 48: looks like this snuck in without test coverage

Issue #2343, opened 2023-11-23
return_linter() should allow switch() where every entry has an exit call


MichaelChirico:
@MichaelChirico I don't think we should give a lint here. In my opinion this is a false positive, as there is no way an implicit return is given.
My first implementation handled this accordingly. Maybe we could do that again.

_Originally posted by @MEO265 in https://github.com/r-lib/lintr/pull/2271#discussion_r1401041242_

PR #2362, opened 2023-11-28
handle nested terminal if/else

### LINKED ISSUE #2354: return_linter() for implicit returns has a false positive on nested early returns
Corrected this test, and added the incorrect test as a new regression test.

That said, the new fix is more expedient than perfect -- it'll throw an incorrect lint here:

```r
foo <- function() {
  if (TRUE) {
    if (TRUE) {
      return(1)
    }
    2
  } else {
    3
  }
```

(the `return()` is a valid early return within that branch).

But as this issue is not relevant to the implicit `else` returns feature, I'm pushing to follow-up.

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/2321#discussion_r1405571314_

### LINKED ISSUE #2356: Handling control statements of any depth in `return_linter`
I think this double lint is not correct either you see the `if` after an `else` as a new expression, then only the second one should be marked, or you understand an `if` with a lot of `else` as one expression (that's how I would see it) and then only the first one would have to be marked.

```R
lint(
  trim_some("
  function(x, y) {
    if(x) {
      1
    } else if(y) {
      2
    }
  }
  "),
  return_linter(return_style = "explicit", allow_implicit_else = FALSE)
)
```
```

<text>:2:9: style: [return_linter] All functions must have an explicit return().
  if(x) {
        ^
<text>:4:10: style: [return_linter] All functions must have an explicit return().
  } else if(y) {
         ^~~~~~~
```

I'm not entirely sure if it's a bug of this PR or the first one. If it's one of the first, feel free to convert this message into an issue, because then I'll probably have a (simple) solution to fix it and properly handle control statements of any depth.

_Originally posted by @MEO265 in https://github.com/r-lib/lintr/issues/2321#issuecomment-1827240302_

MichaelChirico:
Closes #2354, Closes #2356


MichaelChirico:
I was also a bit worried about performance -- would overhead to recursing in R push us to pursue this as pure XPath?

I am satisfied with the R-recursion approach -- timing is not meaningfully different (faster, if anything) & should be much more readable. I linted all base R packages as a stress test:

```r
setwd("src/library")
f = list.files()
f = f[file.exists(file.path(f, "DESCRIPTION"))]
lint_all = \(pkgs, linters) for (pkg in pkgs) lint_package(pkg, linters = linters)

# 'old' = current main, 'new' = current this PR

system.time(lint_all(f, old_explicit))
#    user  system elapsed 
#  72.638   1.162  74.112 
system.time(lint_all(f, old_implicit))
#    user  system elapsed 
#  66.546   0.819  67.682 
system.time(lint_all(f, new_explicit))
#    user  system elapsed 
#  72.178   1.058  73.604 
system.time(lint_all(f, new_implicit))
#    user  system elapsed 
#  62.563   0.871  63.674 
```

R/return_linter.R
```diff
@@ -94,80 +86,58 @@ return_linter <- function(
...
-    #   we don't apply recursive logic to check every branch, only that the
-    #   two top level branches have at least two return()s
-    # because of special 'in' syntax for 'for' loops, the condition is
-    #   tagged differently than for 'if'/'while' conditions (simple PAREN)
-    xpath <- glue("
+    body_xpath <- glue("
     (//FUNCTION | //OP-LAMBDA)[parent::expr[not(
       preceding-sibling::expr[SYMBOL[{ xp_text_in_table(except) }]]
     )]]
       /following-sibling::expr[OP-LEFT-BRACE and expr[last()]/@line1 != @line1]
```
AshesITR, line 93: Could this condition be pushed up and changed to `@line1 != @line2` for the function body expression?

AshesITR: Or is this meant to allow multi-line functions consisting of only a single expression?

R/return_linter.R
```diff
@@ -94,80 +86,58 @@ return_linter <- function(
...
+  if (length(child_expr) == 0L) {
+    return(list())
+  }
+  child_node <- xml_name(child_expr)
 
-    xml_nodes_to_lints(
-      xml_nodes,
+  if (child_node[1L] == "OP-LEFT-BRACE") {
+    expr_idx <- which(child_node %in% c("expr", "equal_assign", "expr_or_assign_or_help"))
+    if (length(expr_idx) == 0L) { # empty brace expression {}
```
AshesITR, line 123: IIUC, this will permit `if (a) { return(b) } else {}` with explicit return style, right?

R/return_linter.R
```diff
@@ -94,80 +86,58 @@ return_linter <- function(
...
+      return(list())
+    }
+    nested_return_lints(child_expr[[tail(expr_idx, 1L)]], xpath, source_expression, lint_message)
+  } else if (child_node[1L] == "IF") {
+    expr_idx <- which(child_node %in% c("expr", "equal_assign", "expr_or_assign_or_help"))
+    c(
```
AshesITR, line 129:
```r
unlist(lapply(
  child_expr[expr_idx[-1L]], 
  nested_return_lints, 
  xpath = xpath,
  lint_message = lint_message
), recursive = FALSE)
```
should work too. Not sure which is faster or more readable.

R/return_linter.R
```diff
@@ -94,80 +86,58 @@ return_linter <- function(
...
+      # TRUE condition
+      nested_return_lints(child_expr[[expr_idx[2L]]], xpath, source_expression, lint_message),
+      # FALSE condition, if present
+      if (length(expr_idx) > 2L) nested_return_lints(child_expr[[expr_idx[3L]]], xpath, source_expression, lint_message)
+    )
+  } else {
+    list(xml_nodes_to_lints(
```
AshesITR, line 136: Why is `list()` needed?

tests/testthat/test-return_linter.R
```diff
@@ -882,3 +882,272 @@ test_that("return_linter lints `message`, `warning` and `stopifnot`", {
...
+            return(TRUE)
+          }
+        }
+      }
+    "),
+    NULL,
+    explicit_linter
+  )
+
+  mixed_lines <- trim_some("
```
AshesITR, line 958: Very nice test!

R/return_linter.R
MichaelChirico: I'll stick with the `c()` approach, but may use this for #2343

R/return_linter.R
MichaelChirico: otherwise `c()` is used on the lints and they break, i.e. `c(xml_nodes_to_lints(...), xml_nodes_to_lints(...))` doesn't work. Not sure if it should but this is simple enough.

R/return_linter.R
MichaelChirico: I think you're right, which is bad. I'll add a test & fix if needed.

R/return_linter.R
MichaelChirico: Hmm I'll look into it but prefer as a follow-up, there's enough going on in this CL already.

MichaelChirico: #2363

R/return_linter.R
MichaelChirico: Added tests... the recurser gets a new argument, which is a bit ugly, but couldn't see a better way to handle it.

tests/testthat/test-return_linter.R
MichaelChirico: I'm more excited we finally have a working version of this feature, it's been low-key bugging me for years that we didn't handle it properly!

R/return_linter.R
MichaelChirico: (I guess that's not exactly true, we have plenty of examples of building lints like `c(lintsA, lintsB, lintsC)`. I guess it's the recursion that's breaking things in this case? What's certain is I originally tried without `list()` and it didn't work (I got back a list looking like `c(as.list(lintsA), as.list(lintsB))` with duplicate names and the wrong class), but does work with `list()`, so I'm content :)

MichaelChirico: OK it bugged me. The issue comes down to the `c(...)` approach, which breaks when `else` is missing, e.g. `c(xml_nodes_to_lints(...), NULL)` strips the `lint` attribute, breaking things.

A solution is to take your advice from the [other comment](https://github.com/r-lib/lintr/pull/2362#discussion_r1408185541) and use `lapply()` instead of `c()`. It's actually much cleaner now that I've condensed the recurser signature, too. `unlist()` is not needed -- `flatten_lints()` handles that later on.

PR #2364, opened 2023-11-29
New 'function_body' test helper

MichaelChirico:
Noticed we have a lot of tests on linters that apply specifically to function bodies where we have to waste space+mental effort writing the boilerplate `foo <- function(...) {`.

This helper helps us focus on what actually matters -- the function body.

Applied it here to only one test script to get a sense of how it looks in practice.

If we agree this is useful I'll apply it more broadly.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -53,9 +53,9 @@ function_body <- function(x, signature = "", assign = TRUE) {
   } else {
     symbol <- ""
   }
-  x <- paste0("      ", unlist(strsplit(trim_some(x), "\n", fixed = TRUE)), collapse = "\n")
-  fmt <- "    %sfunction(%s) {\n%s\n    }"
-  trim_some(sprintf(fmt, symbol, signature, x))
+  x <- paste0("   ", unlist(strsplit(trim_some(x), "\n", fixed = TRUE)), collapse = "\n")
+  fmt <- "%sfunction(%s) {\n%s\n}"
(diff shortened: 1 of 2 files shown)


AshesITR:
To me it reads less fluent with the helper TBH.
Especially because the signature comes after the body, I think.

MichaelChirico:
PTAL with `signature=` first

AshesITR:
Sorry, still feels off.
The lines are way longer and `foo <- function(...) {` just reads much more natural.
Savings in terms of characters aren't that huge either because many functions need a signature.

MichaelChirico:
> Savings in terms of characters aren't that huge either because many functions need a signature.

That was really to keep fidelity with the original test, but that's a distraction in this case. The signature doesn't matter actually. The point is to focus attention as much as possible on the core logic of the test.

We could keep the newlines, but condensing those helps even more to focus on what's different between the tests:

image

AshesITR:
Now it looks better. Two thoughts:

1. rename to `function_code()`? It doesn't "do" anything with the body.
2. drop the usage for the complex function (`lines_subassignment`)? I feel that one is still worse to read than with `trim_some()`. The biggest benefit appears to be with single-line-in-body functions that fit into a single line with this helper.

MichaelChirico:
> The biggest benefit appears to be with single-line-in-body functions that fit into a single line with this helper.

In that case there's not much benefit at all. OK giving up.

PR #2367, opened 2023-11-30
skip #nocov end and other comments generally in unreachable_code_linter()

### LINKED ISSUE #2327: unreachable_code_linter carve-out for 'nolint end' should also apply to 'nocov end'
This is currently linted:

https://github.com/r-lib/lintr/blob/0815b2a1e70d8b8df47850d0309a18f0bb145329/R/get_source_expressions.R#L281

MichaelChirico:
Closes #2327


R/unreachable_code_linter.R
```diff
@@ -37,24 +50,31 @@
...
+#' lint(
+#'   text = code_lines,
+#'   linters = unreachable_code_linter(allow_comment_regex = "# end skip")
+#' )
+#'
 #' @evalRd rd_tags("unreachable_code_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-unreachable_code_linter <- function() {
+unreachable_code_linter <- function(allow_comment_regex = getOption("covr.exclude_end")) {
```
AshesITR, line 77: explicit default if that option is unset?

MichaelChirico: NULL is the default which I intended, upstream also writes the same, so I wouldn't write it out (do we have a linter proposed for supplying explicit arguments that match defaults?)

https://github.com/r-lib/covr/blob/54de614f5c17a42bfaef2c87dbf5723ca701326a/R/exclusions.R#L55-L57

AshesITR: Actually, the defaults are set in `.onLoad`
https://github.com/r-lib/covr/blob/main/R/zzz.R

So the feature won't work unless we `loadNamespace("covr")`.
I suggest to default to this regex for nocov end.

MichaelChirico: Hmm good point, thanks. We'll have to note about that in the docs.

AshesITR: Sorry, I suggested to use the regex as default for `getOption(...)`. That way, we pick up custom exclusion patterns automatically while not depending on covr to be loaded for the default case.

MichaelChirico: oh duh, that's obviously a way better option!

NEWS.md
```diff
@@ -31,7 +31,8 @@
...
 * `backport_linter()` is slightly faster by moving expensive computations outside the linting function (#2339, #2348, @AshesITR and @MichaelChirico).
 * `Linter()` has a new argument `linter_level` (default `NA`). This is used by `lint()` to more efficiently check for expression levels than the idiom `if (!is_lint_level(...)) { return(list()) }` (#2351, @AshesITR).
-* `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are `# nocov end` by default.
+* `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
+* `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
```
IndrajeetPatil, line 35: This NEWS item is expected here? I also see it in NEWS on the `main` branch.

MichaelChirico: Not sure I follow your comment, could you elaborate?

IndrajeetPatil: Ah, never mind. Read the git diff incorrectly. Sorry.

PR #2368, opened 2023-11-30
Also handle 'single-quoted' library calls in sapply() for libary_call_linter()

### LINKED ISSUE #2319: Catch single-quotes FUN when looking for character.only=TRUE in library_call_linter()
This is to catch `lapply(pkgs, "library", character.only = TRUE)`?
Is that worth it? If yes, we should catch `lapply(pkgs, 'library', character.only = TRUE)` too.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2306#discussion_r1398406589_

MichaelChirico:
Closes #2319


tests/testthat/test-library_call_linter.R
```diff
@@ -233,50 +233,51 @@ test_that("skips allowed usages of library()/character.only=TRUE", {
...
-
   expect_lint(
     'purr::map("data.table", library, character.only = TRUE)',
-    rex::rex("Call library() directly, not vectorized with map()"),
+    rex::rex(direct_stub, "map()"),
     linter
   )
 })
 
+test_that("character.only=TRUE is caught in *apply functions passed as strings", {
```
IndrajeetPatil, line 270: Is it worth switching to parametrized tests here?

MichaelChirico: Hmm my sense is it's a lot less readable:

```r
local({
  linter <- library_call_linter()

  cases <- expand.grid(
    call = c("library", "require"),
    quote = c("'", '"'),
    stringsAsFactors = FALSE
  )

  cases$lint_msg <- vapply(paste("Call", cases$call, "directly, not vectorized with sapply()"), rex::rex, character(1L))
  cases$.test_name <- sprintf("Attach with %s, quote with %s", cases$call, cases$quote)

  patrick::with_parameters_test_that(
    "character.only=TRUE is caught in *apply functions passed as strings",
    {
      expect_lint(
        sprintf("sapply(pkgs, %s, character.only = TRUE)", sQuote(call, quote)), 
        lint_msg,
        linter
      )
    },
    .cases = cases
  )
})
```

IndrajeetPatil (approved):

PR #2369, opened 2023-11-30
new width= parameter for output methods

### LINKED ISSUE #1884: Provide an easy way to format lint messages with user-specified width
No matter where lint messages are printed (e.g. GitHub annotations, RStudio markers, Markdown in GitHub issues, etc.), currently, there is no straightforward way to print them with a certain width. This means that the user always needs to scroll horizontally to see many of the lint messages. 

``` r
options(width = 50)
library(lintr)

lint(
  text = "x <- c();",
  linters = unnecessary_concatenation_linter()
)
#> <text>:1:6: style: [unnecessary_concatenation_linter] Unneeded concatenation without arguments. Replace the "c" call by NULL or, whenever possible, vector() seeded with the correct type and/or length.
#> x <- c();
#>      ^~~
```

<sup>Created on 2023-01-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

A typical project that starts using linters might throw hundreds of messages, and that's a lot of scrolling. This might be mildly annoying for an able-bodied individual like me, but must be extremely challenging and disappointing for people with mobility impairment. Almost every single lint asks them to scroll to see the full lint message. 

We should anticipate this and provide a way out.

MichaelChirico:
Closes #1884

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -1,5 +1,5 @@
 #' @export
-format.lint <- function(x, ...) {
+format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
   if (requireNamespace("cli", quietly = TRUE)) {
     color <- switch(x$type,
       warning = cli::col_magenta,
@@ -15,7 +15,7 @@ format.lint <- function(x, ...) {
     # nocov end
   }
 
-  paste0(
+  annotated_msg <- paste0(
     emph(
       x$filename, ":",
       as.character(x$line_number), ":",
@@ -24,7 +24,15 @@ format.lint <- function(x, ...) {
     ),
     color(x$type, ": ", sep = ""),
     "[", x$linter, "] ",
-    emph(x$message), "\n",
+    emph(x$message)
+  )
+
+  if (!is.null(width)) {
+    annotated_msg <- paste(strwrap(annotated_msg, exdent = 4L, width = width), collapse = "\n")
+  }
+
+  paste0(
+    annotated_msg, "\n",
     # swap tabs for spaces for #528 (sorry Richard Hendricks)
     chartr("\t", " ", x$line), "\n",
     highlight_string(x$message, x$column_number, x$ranges),
@@ -34,7 +42,7 @@ format.lint <- function(x, ...) {
 
 #' @export
 print.lint <- function(x, ...) {
-  cat(format(x))
+  cat(format(x, ...))
   invisible(x)
 }
 
@@ -68,8 +76,8 @@ markdown <- function(x, info, ...) {
 }
 
 #' @export
-format.lints <- function(x, ...) {
-  paste(vapply(x, format, character(1L)), collapse = "\n")
+format.lints <- function(x, ..., width = getOption("lintr.format_width")) {
+  paste(vapply(x, format, character(1L), width = width), collapse = "\n")
 }
 
 #' @export
--- a/tests/testthat/test-methods.R
+++ b/tests/testthat/test-methods.R
@@ -157,3 +157,41 @@ test_that("as.data.table.list is _not_ dispatched directly", {
   lints <- lint(text = "a = 1", linters = assignment_linter())
   expect_identical(nrow(data.table::as.data.table(lints)), 1L)
 })
+
+local({
+  # avoid impact of CLI mark-up on strwrap output.
+  #   (testthat, or cli, already do so, but force it explicitly here for emphasis)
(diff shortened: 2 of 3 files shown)


R/methods.R
```diff
@@ -1,5 +1,5 @@
 #' @export
-format.lint <- function(x, ...) {
+format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
```
MichaelChirico, line 2: I think we could write this as lintr_option("format_width") but that would not be user-friendly IINM.

IndrajeetPatil: Yes, I agree. 

This convention also makes us consistent with how other tidyverse packages provide options (e.g. [pillar](https://pillar.r-lib.org/reference/pillar_options.html), [usethis](https://usethis.r-lib.org/reference/usethis_options.html), etc.).

AshesITR: We should also respect `getOption("width")`, no?

tests/testthat/test-methods.R
```diff
@@ -157,3 +157,43 @@ test_that("as.data.table.list is _not_ dispatched directly", {
...
+  withr::local_options(c(cli.num_colors = 0L))
+  # force "default" print method even on GHA
+  withr::local_envvar(c(GITHUB_ACTIONS = NA))
+
+  test_linter <- make_linter_from_xpath("*[1]", lint_message = "The quick brown fox jumps over the lazy dog.")
+
+  lints <- lint(text = "a", linters = test_linter())
+  lint <- lints[[1L]]
+
+  wrapped_strings <- c(
```
IndrajeetPatil, line 173: I had to open my IDE to check these tests locally because I found it so hard to _see_ if the output indeed looked as expected. And this is making me wonder if we have finally found a place where we should use snapshot tests? WDYT?

R/methods.R
MichaelChirico: > We should also respect `getOption("width")`, no?

I think not really. `cat()` handles that -- I don't see a need to change the existing behavior in this regard. Normal terminal wrapping has worked fine so far for interactive usage (IMO). I think that would be a separate FR.

tests/testthat/test-methods.R
IndrajeetPatil: I could get the snapshot tests to work locally, and it seems to work, except for `format()`. 
If I do `cat(format(...))`, I get the following error:

```r
── Error (test-methods.R:183:7): format.lint, format.lints, print.lint, print.lints support optional message wrapping 80 ──
Error in `trace_back(top = getOption("testthat_topenv"), bottom = trace_env)`: Can't find `bottom` on the call tree.
Backtrace:
    ▆
 1. ├─rlang::eval_tidy(code, args)
 2. └─testthat::expect_snapshot(cat(format(lint, width = 20L))) at test-methods.R:183:7
 3.   └─testthat:::expect_snapshot_helper(...)
 4.     └─testthat::expect(...)
 5.       └─rlang::trace_back(top = getOption("testthat_topenv"), bottom = trace_env)
 6.         └─rlang:::trace_find_bottom(bottom, frames)
 7.           └─rlang::abort(msg, call = call)
```

If I don't `cat()`, then, of course, the snapshotted output isn't any better than just having the wrapped strings.

I can investigate this further after work, unless you already have some ideas why this might not be working.

MichaelChirico: Seems like a bug (not sure if {testthat} or maybe even {rlang}).

It _does_ seem less appropriate to test the `format()` output with snapshots. But understand it's a bit awkward to mix snapshots with ugly formatted string outputs.

Easiest way to proceed would seem to be (1) file an upstream issue while we wait to figure out why snapshots won't work (2) leave a TODO here (3) merge with the working (but hard-to-read) tests we have already

IndrajeetPatil: I'd prefer to go for the third option: merge as is and then create a follow-up issue (I can take care of this) for using snapshot tests.

I will approve this, but let's wait for @AshesITR to also resolve his comment before we merge.

IndrajeetPatil: Done: https://github.com/r-lib/lintr/issues/2370

IndrajeetPatil (approved):

R/methods.R
IndrajeetPatil: @AshesITR I am going to merge this. Please create a follow-up if you feel strongly that `getOption("width")` should be respected here.

Issue #2359, opened 2023-11-27
Tests in `get_source_expression()` no longer relevant to actual `lint()` behavior.


AshesITR:
Some linters require explicit NULL checks only because of a test in `get_source_expressions()` that ensures all linters accept all types of `xml_missing()` source expressions.
We could think about removing that test to save a bunch of `if (is.null(xml)) return(list())` lines.

PR #2378, opened 2023-12-02
Follow tidyverse style guide for inline statements

IndrajeetPatil:
https://style.tidyverse.org/syntax.html#inline-statements

> Function calls that affect control flow (like [`return()`](https://rdrr.io/r/base/function.html), [`stop()`](https://rdrr.io/r/base/stop.html) or `continue`) should always go in their own `{}` block


AshesITR:
We should let this simmer a nit because the `if (is.null(xml)) return(list())` code might become obsolete if we change a test in `test-get_source_expressions.R`

IndrajeetPatil:
Sounds good to me!

Converting to draft until https://github.com/r-lib/lintr/issues/2359 is resolved.

PR #2380, opened 2023-12-03
Add more details to `T_and_F_symbol_linter()` docs

IndrajeetPatil:


R/T_and_F_symbol_linter.R
```diff
@@ -1,6 +1,7 @@
 #' `T` and `F` symbol linter
 #'
-#' Avoid the symbols `T` and `F`, and use `TRUE` and `FALSE` instead.
+#' Although they are synonyms, avoid the symbols `T` and `F`, and use `TRUE` and `FALSE`, respectively, instead.
```
MichaelChirico, line 3:
```suggestion
#' Although they can be synonyms, avoid the symbols `T` and `F`, and use `TRUE` and `FALSE`, respectively, instead.
```

maybe?

PR #2381, opened 2023-12-03
Run non-UTF8 locale R CMD check on R-release

IndrajeetPatil:
As decided in https://github.com/r-lib/lintr/issues/2295


MichaelChirico (approved):
Thanks!

PR #2396, opened 2023-12-05
feat: `use_lintr`adds .lintr to .Rbuildignore

### LINKED ISSUE #1805: `use_lintr()` should add `.lintr` to `.Rbuildignore`

MEO265:
Closes #1805 

I know that a PR (#1895 ) already exists, but it doesn't seem to be moving forward anymore.

Please give me your opinion on whether information should be output or additional parameters should be added.

### DIFF
--- a/R/use_lintr.R
+++ b/R/use_lintr.R
@@ -1,6 +1,6 @@
 #' Use lintr in your project
 #'
-#' Create a minimal lintr config file as a starting point for customization
+#' Create a minimal lintr config file as a starting point for customization and add it to the .Rbuildignore
 #'
 #' @param path Path to project root, where a `.lintr` file should be created.
 #' If the `.lintr` file already exists, an error will be thrown.
@@ -25,7 +25,7 @@
 #'   lintr::lint_dir()
 #' }
 use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
-  config_file <- normalizePath(file.path(path, lintr_option("linter_file")), mustWork = FALSE)
+  config_file <- normalizePath(file.path(path, lintr_option("linter_file")), mustWork = FALSE, winslash = "/")
   if (file.exists(config_file)) {
     stop("Found an existing configuration file at '", config_file, "'.")
   }
@@ -43,5 +43,21 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
     )
   )
   write.dcf(the_config, config_file, width = Inf)
+
+  # Check if config_file is in package i.e. lintr_option("linter_file") != "../.lintr"
+  pkg_path <- normalizePath(path, mustWork = FALSE, winslash = "/")
+  if(startsWith(config_file, prefix = pkg_path)) {
+    # Skip a extra character for the leading `/`
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+    ignore_path <- file.path(pkg_path, ".Rbuildignore")
+    if (!file.exists(ignore_path)) file.create(ignore_path)
+    # Follow the same procedure as base R to see if the file is already ignored
+    ignore <- trimws(readLines(ignore_path, warn = FALSE))
+    ignore <- ignore[nzchar(ignore)]
+    if(!any(vapply(ignore, \(x) grepl(rel_path, pattern = x, perl = TRUE, ignore.case = TRUE), logical(1L)))) {
+      cat(file = ignore_path, paste0("^", rel_path, "$"), "\n", append = TRUE)
+    }
+  }
+
   invisible(config_file)
 }
--- a/man/use_lintr.Rd
+++ b/man/use_lintr.Rd
@@ -21,7 +21,7 @@ These are suitable for following \href{https://style.tidyverse.org/}{the tidyver
 Path to the generated configuration, invisibly.
 }
 \description{
-Create a minimal lintr config file as a starting point for customization
+Create a minimal lintr config file as a starting point for customization and add it to the .Rbuildignore
 }
 \examples{
 if (FALSE) {


MEO265:
Here you can see how base R handles the .Rbuildignore:
https://github.com/wch/r-source/blob/1d4f7aa1dac427ea2213d1f7cd7b5c16e896af22/src/library/tools/R/build.R#L85-L87

R/use_lintr.R
```diff
@@ -43,5 +43,21 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+  if(startsWith(config_file, prefix = pkg_path)) {
+    # Skip a extra character for the leading `/`
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+    ignore_path <- file.path(pkg_path, ".Rbuildignore")
+    if (!file.exists(ignore_path)) file.create(ignore_path)
+    # Follow the same procedure as base R to see if the file is already ignored
+    ignore <- trimws(readLines(ignore_path, warn = FALSE))
+    ignore <- ignore[nzchar(ignore)]
+    if(!any(vapply(ignore, \(x) grepl(rel_path, pattern = x, perl = TRUE, ignore.case = TRUE), logical(1L)))) {
+      cat(file = ignore_path, paste0("^", rel_path, "$"), "\n", append = TRUE)
```
AshesITR, line 58: This should be `rex::rex(start, rel_path, end)` to handle escapes correctly.

R/use_lintr.R
```diff
@@ -43,5 +43,21 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
   )
   write.dcf(the_config, config_file, width = Inf)
+
+  # Check if config_file is in package i.e. lintr_option("linter_file") != "../.lintr"
+  pkg_path <- normalizePath(path, mustWork = FALSE, winslash = "/")
+  if(startsWith(config_file, prefix = pkg_path)) {
+    # Skip a extra character for the leading `/`
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+    ignore_path <- file.path(pkg_path, ".Rbuildignore")
+    if (!file.exists(ignore_path)) file.create(ignore_path)
```
AshesITR, line 53: This should only run when we are in a package (Check presence of a DESCRIPTION file?)

R/use_lintr.R
```diff
@@ -43,5 +43,21 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
     )
   )
   write.dcf(the_config, config_file, width = Inf)
+
+  # Check if config_file is in package i.e. lintr_option("linter_file") != "../.lintr"
+  pkg_path <- normalizePath(path, mustWork = FALSE, winslash = "/")
+  if(startsWith(config_file, prefix = pkg_path)) {
+    # Skip a extra character for the leading `/`
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
```
AshesITR, line 51: Not 100% sure but I remember package tools has some helpers for these operations.

R/use_lintr.R
```diff
@@ -25,7 +25,7 @@
 #'   lintr::lint_dir()
 #' }
 use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
-  config_file <- normalizePath(file.path(path, lintr_option("linter_file")), mustWork = FALSE)
+  config_file <- normalizePath(file.path(path, lintr_option("linter_file")), mustWork = FALSE, winslash = "/")
```
AshesITR, line 28: What is the reason for this change?

AshesITR (commented):
Is there a test that can be added?

MEO265:
> Is there a test that can be added?

Surely.  I'll take care of that too.  Just wanted to hear beforehand whether additional functionality would be desired.  In order not to have to adapt the tests again and again.

R/use_lintr.R
MEO265: The only thing I can find is this `tools::file_path_as_absolute(".")`, but it's just a wrapper around `normalizePath`. If you know something specific, I'd be happy to include it.

R/use_lintr.R
MEO265: The documentation of the function says:
```R
#' @param path Path to project root, where a `.lintr` file should be created.
#' If the `.lintr` file already exists, an error will be thrown.
```
and in the function we use `lintr_option("linter_file")` to specify the path of the ".lintr" file relative to the project path and the ".lintr" file only makes sense in a package. So I think it would be ok to assume that the path is that of a package.

If you want me to change it, you could also add a parameter "add_to_ignore" which is set to `file.exists(file.path(path, "DESCRIPTION"))` by default.

AshesITR: You can use a `.lintr` config for non-package R projects too.
I don't think an argument is necessary, but feel free to add it anyway. AFAIR, `usethis` doesn't provide an option like this, but is transparent about what changes are made.

R/use_lintr.R
AshesITR: NVM I misremembered. They are in {fs} https://search.r-project.org/CRAN/refmans/fs/html/path_math.html

Since we don't import it, let's stick to base R. Thanks for checking.

R/use_lintr.R
```diff
@@ -43,5 +43,21 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+  pkg_path <- normalizePath(path, mustWork = FALSE, winslash = "/")
+  if(startsWith(config_file, prefix = pkg_path)) {
+    # Skip a extra character for the leading `/`
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+    ignore_path <- file.path(pkg_path, ".Rbuildignore")
+    if (!file.exists(ignore_path)) file.create(ignore_path)
+    # Follow the same procedure as base R to see if the file is already ignored
+    ignore <- trimws(readLines(ignore_path, warn = FALSE))
+    ignore <- ignore[nzchar(ignore)]
+    if(!any(vapply(ignore, \(x) grepl(rel_path, pattern = x, perl = TRUE, ignore.case = TRUE), logical(1L)))) {
```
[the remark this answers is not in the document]
AshesITR: You don't need to use an anonymous function at all if you provide `x = rel_path` by name too.

R/use_lintr.R
```diff
@@ -43,5 +43,25 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+    rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+    ignore_path <- file.path(pkg_path, ".Rbuildignore")
+    if (!file.exists(ignore_path)) file.create(ignore_path)
+    # Follow the same procedure as base R to see if the file is already ignored
+    tryCatch({
+      ignore <- trimws(readLines(ignore_path, warn = FALSE))
+    }, warning = function(e) cat(file = ignore_path, "\n", append = TRUE)
+    )
+    ignore <- ignore[nzchar(ignore)]
+    if (!any(vapply(ignore, function(x) grepl(rel_path, pattern = x, perl = TRUE, ignore.case = TRUE), logical(1L)))) {
```
AshesITR, line 60:
```r
vapply(ignore, function(x) grepl(rel_path, pattern = x, perl = TRUE, ignore.case = TRUE), logical(1L))
```

Is equivalent to and slightly slower than

```r
vapply(ignore, grepl, x = rel_path, perl = TRUE, ignore.case = TRUE, logical(1L))
```

R/use_lintr.R
```diff
@@ -43,5 +43,37 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
     )
   )
   write.dcf(the_config, config_file, width = Inf)
+
+  if (file.exists(file.path(path, "DESCRIPTION"))) {
+    # Some OS can only normalize a path if the associated file or folder exists, so the path needs to be re-normalized
+    tryCatch({
```
AshesITR, line 49: I don't understand this part.
What do these systems return with `normalizePath("foo", mustWork = FALSE)` if foo doesn't exist?

R/use_lintr.R
```diff
@@ -43,5 +43,37 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+      pkg_path <- normalizePath(path, mustWork = TRUE, winslash = "/")
+      config_file <- normalizePath(file.path(path, lintr_option("linter_file")), mustWork = TRUE, winslash = "/")
+    }, error = function(e) {
+      stop("No entry could be added to the .Rbuildignore.", call. = FALSE)
```
AshesITR, line 53: This sounds like a warning is in order. The main side effect was caused at this point anyway.

R/use_lintr.R
```diff
@@ -43,5 +43,37 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+    })
+    # Check if config_file is in package i.e. lintr_option("linter_file") != "../.lintr"
+    if (startsWith(config_file, prefix = pkg_path)) {
+      # Skip a extra character for the leading `/`
+      rel_path <- substring(config_file, first = nchar(pkg_path) + 2L, last = nchar(config_file))
+      ignore_path <- file.path(pkg_path, ".Rbuildignore")
+      if (!file.exists(ignore_path)) file.create(ignore_path)
```
AshesITR, line 60: This branch can be a `writeLines(..., ignore_path)` with immediate return.

R/use_lintr.R
```diff
@@ -43,5 +43,37 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+      # Follow the same procedure as base R to see if the file is already ignored
+      ignore <- tryCatch({
+        trimws(readLines(ignore_path))
+      }, warning = function(e) {
+        cat(file = ignore_path, "\n", append = TRUE)
```
AshesITR, line 65: Why are you silently fixing the missing terminal newline here?

R/use_lintr.R
```diff
@@ -43,5 +43,37 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
+        trimws(readLines(ignore_path))
+      })
+      ignore <- ignore[nzchar(ignore)]
+      already_ignored <-
+        any(vapply(ignore, FUN = grepl, x = rel_path, perl = TRUE, ignore.case = TRUE, FUN.VALUE = logical(1L)))
+      if (!already_ignored) {
+        cat(file = ignore_path, rex::rex(start, rel_path, end), sep = "\n", append = TRUE)
```
AshesITR, line 72: IINM the `sep=` is ignored because you only pass one entry into cat.

tests/testthat/test-use_lintr.R
```diff
@@ -35,3 +35,56 @@ test_that("use_lintr with type = full also works", {
...
+test_that("use_lintr creates the correct regex", {
+  tmp <- withr::local_tempdir()
+  file.create(file.path(tmp, "DESCRIPTION"))
+  ignore <- file.path(tmp, ".Rbuildignore")
+  file.create(ignore)
+  cat(file = ignore, "^fu$", "^bar$", sep = "\n")
+
+  expect_message({
+    lintr_file <- use_lintr(path = tmp, type = "full")
+  }, regexp = "Adding .* to .Rbuildignore")
```
AshesITR, line 75: Nit: the `.` in front of Rbuildignore matches any character.

NEWS.md
```diff
@@ -36,10 +36,12 @@
...
 * `backport_linter()` is slightly faster by moving expensive computations outside the linting function (#2339, #2348, @AshesITR and @MichaelChirico).
 * `Linter()` has a new argument `linter_level` (default `NA`). This is used by `lint()` to more efficiently check for expression levels than the idiom `if (!is_lint_level(...)) { return(list()) }` (#2351, @AshesITR).
+* `use_lintr()` adds the created `.lintr` file to the `.Rbuildignore` if run in a package (#1805, @MEO265).
 * `string_boundary_linter()` recognizes regular expression calls like `grepl("^abc$", x)` that can be replaced by using `==` instead (#1613, @MichaelChirico).
 * `unreachable_code_linter()` has an argument `allow_comment_regex` for customizing which "terminal" comments to exclude (#2327, @MichaelChirico). `# nolint end` comments are always excluded, as are {covr} exclusions (e.g. `# nocov end`) by default.
 * `format()` and `print()` methods for `lint` and `lints` classes get a new option `width` to control the printing width of lint messages (#1884, @MichaelChirico). The default is controlled by a new option `lintr.format_width`; if unset, no wrapping occurs (matching earlier behavior).
 
+
```
AshesITR, line 44: Superfluous change

R/use_lintr.R
```diff
@@ -1,6 +1,6 @@
 #' Use lintr in your project
 #'
-#' Create a minimal lintr config file as a starting point for customization
+#' Create a minimal lintr config file as a starting point for customization and add it to the .Rbuildignore
```
AshesITR, line 3: It should be `add it to .Rbuildignore`, without a "the". Same in other places.

PR #2397, opened 2023-12-05
Remove `xml_missing()` tests

### LINKED ISSUE #2359: Tests in `get_source_expression()` no longer relevant to actual `lint()` behavior.

IndrajeetPatil:
Closes #2359


AshesITR:
The test should instead be refactored to honor `is_linter_level()`, I think.

AshesITR (approved):

AshesITR (approved):

IndrajeetPatil:
Thanks, @AshesITR, for taking this to the finish line. I've made one minor change.

AshesITR (approved):

MichaelChirico:
This is beautiful 🥹

PR #2401, opened 2023-12-06
Lint message cites which argument is missing

### LINKED ISSUE #2389: duplicate_argument_linter() could cite which argument is duplicated
Remove is not always correct. The argument might also be misnamed.

e.g.
`list(a = 1, a = 2)` may as well be intended to be `list(a = 1, b = 2)`

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2385#discussion_r1414408042_

MichaelChirico:
Closes #2389

### DIFF
--- a/tests/testthat/test-missing_argument_linter.R
+++ b/tests/testthat/test-missing_argument_linter.R
@@ -145,6 +145,17 @@ test_that("lints vectorize", {
       foo(1,)
       bar(,2)
     }"),
+    list(
+      list("Missing argument 2", line_number = 2L),
+      list("Missing argument 1", line_number = 3L)
+    ),
+    linter
+  )
+  expect_lint(
+    trim_some("{
+      foo(1,)
+      bar(,2)
+    }"),
     list("Missing argument 1", line_number = 3L),
     linter_trailing
   )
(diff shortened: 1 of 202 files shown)


tests/testthat/test-missing_argument_linter.R
```diff
@@ -18,14 +18,21 @@ test_that("missing_argument_linter skips allowed usages", {
 
 test_that("missing_argument_linter blocks disallowed usages", {
   linter <- missing_argument_linter()
-  lint_msg <- rex::rex("Missing argument in function call.")
+  lint_msg1 <- rex::rex("Missing argument 1 in function call.")
```
AshesITR, line 21: "Missing positional argument 1" would read more clearly I think. Worth changing?

MichaelChirico: I would leave it as is. I worry about beginners being confused if they don't know this term. also, it's somewhat ambiguous -- for `foo = function(x,y)` and call `foo(y = 1, )` is the first positional argument missing, or the second?

AshesITR: Hmm okay. I was stumped because `1` also appeared (unrelatedly) in the line `foo(, a = 1)`.

AshesITR (approved):

tests/testthat/test-missing_argument_linter.R
MichaelChirico: > Hmm okay. I was stumped because `1` also appeared (unrelatedly) in the line `foo(, a = 1)`.

I think the lint's column_number should help with that

PR #2407, opened 2023-12-08
Mention saving the output of order() in sort_linter() message

### LINKED ISSUE #2009: Mention saving order(.) if reused in sort_linter()
A common violation of `sort_linter()` I'm seeing is:

```r
y <- y[order(x)]
x <- x[order(x)]
```

I think `sort_linter()` still applies, but a better approach is to save `order(x)` to avoid expensive sorting twice:

```r
x_order <- order(x)
y <- y[x_order]
x <- x[x_order]
```

We could mention something along those lines in the lint message.

MichaelChirico:
Closes #2009.

Hopefully this is an improvement... I'm skeptical it's worthwhile to try and only supply this message if we  _detect_ such a re-use of `order()`.


R/sort_linter.R
```diff
@@ -126,7 +126,10 @@ sort_linter <- function() {
     order_lints <- xml_nodes_to_lints(
       order_expr,
       source_expression = source_expression,
-      lint_message = paste0(new_call, " is better than ", orig_call, "."),
+      lint_message = paste0(
+        new_call, " is better than ", orig_call, ". ",
+        "If the output of order() is re-used, save it to a local variable."
```
AshesITR, line 131: s/is/can be/?

If it were re-used by storing in a variable,  we wouldn't lint.

MichaelChirico: I still think the current phrasing is OK. Maybe to clarify, there's two types of places this `sort_linter()` will throw:

```r
x <- x[order(x)]
# ... more code
```

```r
y <- y[order(x)]
x <- x[order(x)]
# ... more code
```

The issue is the latter case, the lint message at HEAD would recommend doing:

```r
y <- y[order(x)]
x <- sort(x)
```

That's pretty awkward, and doesn't get around the real issue here which is trying to sort `x` two times inefficiently.

So "is" has in mind this latter case where `order(x)` already _is_ re-used.

AshesITR: I understand what you mean, but technically, the result of `order(x)` is not reused, it's computed twice. Not sure how to phrase that elegantly.

MichaelChirico: > Note that it's always preferable to save the output of order() as a local variable than to re-compute it for the same vector.

Better?

AshesITR (approved):

PR #2411, opened 2023-12-10
Handle comparison to negative non-integer in expect_identical_linter

### LINKED ISSUE #2410: `expect_identical_linter()` doesn't skip cases with negative numeric constant
``` r
# comparison with positive non-integer constant is skipped
lintr::lint(text = "expect_equal(x, 1.04)", linters = lintr::expect_identical_linter())

# comparison with negative non-integer constant is skipped
lintr::lint(text = "expect_equal(x, -1.04)", linters = lintr::expect_identical_linter())
#> <text>:1:1: warning: [expect_identical_linter] Use expect_identical(x, y) by default; resort to expect_equal() only when needed, e.g. when setting ignore_attr= or tolerance=.
#> expect_equal(x, -1.04)
#> ^~~~~~~~~~~~~~~~~~~~~~
```

<sup>Created on 2023-12-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Bisaloo:
Fix #2410

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 
 ## Bug fixes
 
-* `expect_identical_linter()` no longer lints `expect_equal()` comparison to negative non-integers (#2411, @Bisaloo).
+* `expect_identical_linter()` also skips `expect_equal()` comparison to _negative_ non-integers like `-1.034` (#2411, @Bisaloo). This is a parity fix since _positive_ reals have always been skipped because "high-precision" comparisons are typically done to get tests within `tolerance`, so `expect_identical()` is not a great substitution.
 * `object_name_linter()` no longer errors when user-supplied `regexes=` have capture groups (#2188, @MichaelChirico).
 * `.lintr` config validation correctly accepts regular expressions which only compile under `perl = TRUE` (#2375, @MichaelChirico). These have always been valid (since `rex::re_matches()`, which powers the lint exclusion logic, also uses this setting), but the new up-front validation in v3.1.1 incorrectly used `perl = FALSE`.
 * `.lintr` configs set by option `lintr.linter_file` or environment variable `R_LINTR_LINTER_FILE` can point to subdirectories (#2512, @MichaelChirico).
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 
 ## Bug fixes
 
+* `expect_identical_linter()` no longer lints `expect_equal()` comparison to negative non-integers (#2411, @Bisaloo).
 * `object_name_linter()` no longer errors when user-supplied `regexes=` have capture groups (#2188, @MichaelChirico).
 * `.lintr` config validation correctly accepts regular expressions which only compile under `perl = TRUE` (#2375, @MichaelChirico). These have always been valid (since `rex::re_matches()`, which powers the lint exclusion logic, also uses this setting), but the new up-front validation in v3.1.1 incorrectly used `perl = FALSE`.
 * `.lintr` configs set by option `lintr.linter_file` or environment variable `R_LINTR_LINTER_FILE` can point to subdirectories (#2512, @MichaelChirico).
--- a/R/expect_identical_linter.R
+++ b/R/expect_identical_linter.R
@@ -60,33 +60,30 @@ expect_identical_linter <- function() {
   #   - skip cases like expect_equal(x, 1.02) or the constant vector version
   #     where a numeric constant indicates inexact testing is preferable
   #   - skip calls using dots (`...`); see tests
-  expect_equal_xpath <- "
+  non_integer <- "NUM_CONST[contains(text(), '.')]"
+  non_integer_negative <- glue::glue("
+    OP-MINUS
+    and count(expr) = 1
+    and expr[{non_integer}]
+  ")
+
+  expect_equal_xpath <- glue::glue("
   parent::expr[not(
       following-sibling::EQ_SUB
       or following-sibling::expr[
         (
           expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]
           and expr[
-            NUM_CONST[contains(text(), '.')]
-            or (
-              OP-MINUS
-              and count(expr) = 1
-              and expr[
-                NUM_CONST[contains(text(), '.')]
-              ]
-            )
+            {non_integer} or {non_integer_negative}
           ]
         ) or (
-          NUM_CONST[contains(text(), '.')]
+          {non_integer} or {non_integer_negative}
         ) or (
           OP-MINUS
           and count(expr) = 1
           and expr[
-            NUM_CONST[contains(text(), '.')]
-            or (
-              expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]
-              and expr[NUM_CONST[contains(text(), '.')]]
-            )
+            expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]
+            and expr[{non_integer}]
           ]
         ) or (
           SYMBOL[text() = '...']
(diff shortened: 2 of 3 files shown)


R/expect_identical_linter.R
```diff
@@ -69,6 +69,10 @@ expect_identical_linter <- function() {
         and expr[NUM_CONST[contains(text(), '.')]]
       ]
       or following-sibling::expr[NUM_CONST[contains(text(), '.')]]
+      or following-sibling::expr[
```
MichaelChirico, line 72: IINM this will skip any subtraction with a number on one side, e.g. x-1.01 or 1.01-x

R/expect_identical_linter.R
```diff
@@ -69,6 +69,10 @@ expect_identical_linter <- function() {
...
+        OP-MINUS
+        and expr[NUM_CONST[contains(text(), '.')]]
+      ]
```
Bisaloo, line 75: I see two potential solutions:

```suggestion
      or following-sibling::expr[
        OP-MINUS
        and not(descendant::SYMBOL)
        and expr[NUM_CONST[contains(text(), '.')]]
      ]
```

OR 

```suggestion
      or following-sibling::expr[
        OP-MINUS
        and count(expr) = 1
        and expr[NUM_CONST[contains(text(), '.')]]
      ]
```

The latter would lint cases such as `expect_equal(x, 1.01-0.01)` but maybe that's what we want :thinking:.

AshesITR: then `count(expr) = 1` version seems more intuitive.
I think the idea of detecting constant expressions (like `1.004 - 1.003`, `2.0 * 42.0`, `10L + 2^5`) is best done with a more general approach.

MichaelChirico: related: #963

NEWS.md
```diff
@@ -16,6 +16,7 @@
 
 ## Bug fixes
 
+* `expect_identical_linter()` no longer lints `expect_equal()` comparison to negative non-integers (#2411, @Bisaloo).
```
MichaelChirico, line 19: I think this description requires a lot of context to understand, I would mention this is for parity with positive non-integers, maybe:

```suggestion
* `expect_identical_linter()` also skips `expect_equal()` comparison to _negative_ non-integers like `-1.034` (#2411, @Bisaloo). This is a parity fix since _positive_ reals have always been skipped because "high-precision" comparisons are typically done to get tests within `tolerance`, so `expect_identical()` is not a great substitution.
```

R/expect_identical_linter.R
```diff
@@ -69,6 +69,11 @@ expect_identical_linter <- function() {
         and expr[NUM_CONST[contains(text(), '.')]]
```
MichaelChirico, line 69: Hmm I saw this too -- so we haven't fully restored parity yet in this PR.

```r
expect_equal(x, 1.02)     # no lint
expect_equal(x, c(1.02))  # no lint
expect_equal(x, -1.02)    # no lint
expect_equal(x, c(-1.02)) # lint!
```

Let's at least get parity in this PR. I would put all of the conditions on `following-sibling::expr[]` together btw, like:

```xpath
following-sibling::expr[
  (
    expr[1]/SYMBOL_FUNCTION_CALL[text() = 'c']
    and expr/NUM_CONST[contains(text(), '.')]
  ) or (
    NUM_CONST[contains(text(), '.')]
  ) or (
    OP-MINUS ...
  )
]
```

Since `NUM_CONST[contains(text(), '.')]` is getting repeated so much let's extract it to a variable too to save space.

MichaelChirico: BTW, could you also test if the following lints?

```r
expect_equal(x, c(1.02, y))
```

I think it's something to possible address in follow-up. I'm wondering if maybe it was intentional not to lint `c(1.02, y)` because there's _some_ part of the expectation that uses an "exact" number so we "should" skip linting on the same principle as just `1.02`? In which case we shouldn't lint `1.01-y` either 🤔

@AshesITR any thoughts? This cut-out does feel a bit arbitrary, probably by necessity at some level. Wondering how to close the loop here for this PR, and what can be deferred to a more holistic follow-up.

Bisaloo:
``` r
lintr::lint(
  text = "expect_equal(x, c(1.02, y))", 
  linters = lintr::expect_identical_linter()
)
```

<sup>Created on 2023-12-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

It doesn't lint.

R/expect_identical_linter.R
```diff
@@ -69,6 +69,11 @@ expect_identical_linter <- function() {
...
       ]
       or following-sibling::expr[NUM_CONST[contains(text(), '.')]]
+      or following-sibling::expr[
+        OP-MINUS
+        and count(expr) = 1
+        and expr[NUM_CONST[contains(text(), '.')]]
```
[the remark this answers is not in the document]
MichaelChirico: I think the last condition belongs in `[]` -- otherwise it can get hard to visually parse operator presence of `=` vs. nearby logical operators like `and`/`or`. IINM occasionally we have to let `/text()` "out" e.g. to compare to the `text()` of another node, but it's best avoided.

We don't really have an XPath style guide, so these are just my thoughts from writing many of these, you or other lintr authors may disagree.

Bisaloo:
All tests are now passing. I'm not super happy about the complexity of this xpath but I can't think of something simpler at the moment. Please let me know if you have a better way in mind.

Bisaloo:
https://github.com/r-lib/lintr/issues/963 may be a good way to simplify the code eventually.

tests/testthat/test-expect_identical_linter.R
```diff
@@ -30,6 +30,10 @@ test_that("expect_identical_linter skips cases likely testing numeric equality",
   lint_msg <- rex::rex("Use expect_identical(x, y) by default; resort to expect_equal() only when needed")
 
   expect_lint("expect_equal(x, 1.034)", NULL, linter)
+  expect_lint("expect_equal(x, -1.034)", NULL, linter)
+  expect_lint("expect_equal(x, c(-1.034))", NULL, linter)
+  expect_lint("expect_equal(x, -c(1.034))", NULL, linter)
```
AshesITR, line 35: Is this case relevant in practice?

R/expect_identical_linter.R
```diff
@@ -60,18 +60,38 @@ expect_identical_linter <- function() {
...
+            {non_integer} or {non_integer_negative}
+          ]
+        ) or (
+          {non_integer} or {non_integer_negative}
+        ) or (
+          OP-MINUS
+          and count(expr) = 1
+          and expr[
+            expr[1][SYMBOL_FUNCTION_CALL[text() = 'c']]
+            and expr[{non_integer}]
```
AshesITR, line 86: While not usual, this could also be made to catch `-c(-4.2)`, enabling a refactor of `non_integer` and reducing repetition.

tests/testthat/test-expect_identical_linter.R
Bisaloo: I'm on the fence.

On the one hand, it should be caught be `unnecessary_concatenation_linter()` and converted to the standard case. On the other hand, it's not just a false negative; not treating this case would lead us to produce incorrect advice.

R/expect_identical_linter.R
Bisaloo: If I understood correctly your comment, it should be fixed by https://github.com/r-lib/lintr/pull/2411/commits/36a8c0c74be73b3f8303416f7d58e78a48fbe97c.

AshesITR (approved):

PR #2413, opened 2023-12-11
New parameters for if_switch_linter() to optionally skip lints on "complex" if/else usage

### LINKED ISSUE #2322: Parameterize if_switch_linter() not to require switch() for "highly-complicated" clauses
`if_switch_linter()` is too strict -- in some cases,using `switch()` is actually quite awkward.

```r
if (x == 'a') {
  do_1()
  do_2()
  do_3()
  if (y > 2) {
    do_4()
    do_5()
  }
} else if (x == 'b') {
  do_6()
  do_7()
  do_8()
  do_9()
  do_10()
  for (i in 1:10) {
    do_11()
    do_12(0
  }
} else if (x == 'c') {
  do_13()
  do_14()
  do_15()
  do_16()
  do_17()
} else if (x == 'd') {
  do_18()
  do_19()
  do_20()
  do_21()
  do_22()
}
```

Would become

```r
switch(x,
  a = {
    do_1()
    do_2()
    do_3()
    if (y > 2) {
      do_4()
      do_5()
    }
  },
  b = {
    do_6()
    do_7()
    do_8()
    do_9()
    do_10()
    for (i in 1:10) {
      do_11()
      do_12(0
    }
  },
  c = {
    do_13()
    do_14()
    do_15()
    do_16()
    do_17()
  },
  d = {
    do_18()
    do_19()
    do_20()
    do_21()
    do_22()
  }
)
```

Note the increased nesting, and the distance between `switch(x` and the `x` value `b = {`.

It would be good to come up with a rule to weaken `if_switch_linter()` to not apply to cases where the branches are "overly complicated". The trouble is coming up with a good rule for what that means (e.g. `cyclocomp_linter()` only cares about branching, not the number of expressions).

One simple option is to just count the number of "child expressions" in each branch, and not lint if, say `max(expressions) > complexity` where `complexity=` is an argument set by the user, perhaps with a sensible default like `3L`.

MichaelChirico:
Closes #2322


AshesITR:
Should this lint `switch()` calls that violate the `max_*` bounds, too?

MichaelChirico:
> Should this lint `switch()` calls that violate the `max_*` bounds, too?

Interesting... possibly? I would wait for user request though, WDYT?

AshesITR:
I feel the linter should enforce some consistency.
The reason _why_ the lints would be suppressed are also valid reasons why the `switch()` should be broken up into an `if () ... else if () ...` chain.

AshesITR:
This seems to have a mingled diff with the devcontainers PR...

MichaelChirico:
> This seems to have a mingled diff with the devcontainers PR...

yes it was branched from there but since diverged. let's merge that PR so that I can launch codespaces from main to avoid this

AshesITR:
Looks good, I would make some of the tests "unbalanced" in terms of lines/exprs per branch to ensure the largest/most complex branch dictates behavior.

AFAICT there is no such test for `switch()`.
Also, is there a metadata test for switch()?

MichaelChirico:
> AFAICT there is no such test for `switch()`.

IIUC those are https://github.com/r-lib/lintr/pull/2413/files#diff-5a5a1de0b476f8f7c4ffcfbf4f4307c8cdfbbfa85367eccf61060e48e433c9e5R337-R340 and https://github.com/r-lib/lintr/pull/2413/files#diff-5a5a1de0b476f8f7c4ffcfbf4f4307c8cdfbbfa85367eccf61060e48e433c9e5R361-R364

> Also, is there a metadata test for switch()?

No, adding now

AshesITR:
> > AFAICT there is no such test for `switch()`.
> 
> 
> 
> IIUC those are https://github.com/r-lib/lintr/pull/2413/files#diff-5a5a1de0b476f8f7c4ffcfbf4f4307c8cdfbbfa85367eccf61060e48e433c9e5R337-R340 and https://github.com/r-lib/lintr/pull/2413/files#diff-5a5a1de0b476f8f7c4ffcfbf4f4307c8cdfbbfa85367eccf61060e48e433c9e5R361-R364
> 

In those tests, all switch branches have the same complexity (loc & expr count). That's what I meant by "balanced".

I'd add something like

```r
switch(x,
  a = 1
  b = {
    2
    3
    4; 5; 6
  },
  c = {
    7
  }
)
```

to the mix

> 
> > Also, is there a metadata test for switch()?
> 
> 
> 
> No, adding now

MichaelChirico:
Thanks for clarifying, added some tests unbalanced across branches

R/if_switch_linter.R
```diff
@@ -14,13 +14,77 @@
 #'   approach is roughly linear in the number of conditions that need to
 #'   be evaluated, here up to 3 times).
 #'
+#' @param max_branch_lines,max_branch_expr Integer, default 0 indicates "no maximum".
```
AshesITR, line 17: Plural the expr arg too? `max_branch_exprs`?

MichaelChirico: that sounds unnatural to me, I would pluralize as "several expr nodes" or "several expressions", never "several exprs".

would you prefer max_branch_expressions?

AshesITR: `max_branch_expressions` sounds good as well.
We might even drop the `_branch_` in both arg names for brevity, WDYT?

I read expr as a contraction of expression, so I don't have a problem with "exprs", but I can see the confusion with <expr> nodes.

MichaelChirico: Thinking again, "expr" anchors too much on thinking in terms of XPath, which probably isn't the most user-friendly approach. "expressions" is more universal / understood without knowing any internals of the AST/its XML representation.

I'd keep `branch` for clarity.

AshesITR (approved):

Issue #2252, opened 2023-10-31
Support cli functions in `extract_glued_symbols()`


Bisaloo:
``` r
# glue::glue passes
lintr::lint(
  text = "foo <- function() { x <- 1; glue::glue('{x}') }",
  linters = lintr::object_usage_linter()
)

# cli functions lint
lintr::lint(
  text = "foo <- function() { x <- 1; cli::cli_h1('{x}') }",
  linters = lintr::object_usage_linter()
)
#> <text>:1:21: warning: [object_usage_linter] local variable 'x' assigned but may not be used
#> foo <- function() { x <- 1; cli::cli_h1('{x}') }
#>                     ^

# This is valid cli usage
x <- 1
cli::cli_h1('{x}')
#> 
#> ── 1 ───────────────────────────────────────────────────────────────────────────
```

<sup>Created on 2023-10-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico:
Didn't realize how highly requested this is!

One major issue is that {cli}'s mark-up syntax is much richer than {glue}'s, and there's currently no equivalent for extracting the variables besides custom logic we'd have to maintain & keep in sync:

https://github.com/r-lib/cli/issues/750

In the short term, we could just treat cli exactly like glue, at least capturing simple glue-like interpolations, would that be useful enough as a first step?

PR #2418, opened 2023-12-12
Use `{cli}` for formatting condition messages

### LINKED ISSUE #2386: Use `{cli}` to build messages for conditions/exceptions
https://cli.r-lib.org/articles/semantic-cli.html

IndrajeetPatil:
Closes #2386


MichaelChirico:
> using snapshot tests for tests capturing warnings and errors

My experience is this sets us up for a ton of toil maintaining brittle tests. Using `expect_warning()`/`expect_error()` is IMO far superior

 - The entire test is self-contained, no need to reference a different file to find the expected output when reading/reviewing
 - The expectation can be focused -- no need to write out the full condition message in the test, just the parts that matter. This increases clarity & reduces fragility.
 - We only test things that we own. Any time {cli} changes, we are at risk of our snapshot tests breaking. Whereas the condition message fragments come directly from our sources.

R/addins.R
```diff
@@ -1,11 +1,11 @@
 # nocov start
 addin_lint <- function() {
   if (!requireNamespace("rstudioapi", quietly = TRUE)) {
-    stop("'rstudioapi' is required for add-ins.", call. = FALSE)
+    cli::cli_abort("{.pkg rstudioapi} is required for add-ins.")
   }
   filename <- rstudioapi::getSourceEditorContext()
   if (filename$path == "") {
-    return("Current source has no path. Please save before continue")
+    return("Current source has no path. Please save before continue.")
```
MichaelChirico, line 8:
```suggestion
return("Current source has no path. Please save before continuing.")
```

R/backport_linter.R
```diff
@@ -90,7 +90,7 @@ normalize_r_version <- function(r_version) {
     version_names <- c("devel", "release", paste0("oldrel-", seq_len(length(minor_versions) - 2L)))
     if (!r_version %in% version_names) {
       # This can only trip if e.g. oldrel-99 is requested
-      stop("`r_version` must be a version number or one of ", toString(sQuote(version_names)), call. = FALSE)
+      cli::cli_abort("{.var r_version} must be a version number or one of {toString(sQuote(version_names))}")
```
MichaelChirico, line 93: Use `{.val {sQuote(version_names)}}` instead? Is there a {cli}-specific way to do `sQuote()` too?

https://cli.r-lib.org/articles/semantic-cli.html#inline-text-formatting

R/addins.R
```diff
@@ -1,11 +1,11 @@
 # nocov start
 addin_lint <- function() {
   if (!requireNamespace("rstudioapi", quietly = TRUE)) {
-    stop("'rstudioapi' is required for add-ins.", call. = FALSE)
+    cli::cli_abort("{.pkg rstudioapi} is required for add-ins.")
```
MichaelChirico, line 4: I find `cli::cli_` redundant, let's properly import these in R/lintr-package.R

MichaelChirico:
Directional greenlight that using {cli} signals is a go for me. @AshesITR, you?

PS we should also add `stop()`/`warning()`/`message()` as undesirable functions in our `.lintr` afterwards to prevent backsliding.

AshesITR:
I agree with both of your comments.

IndrajeetPatil:
Btw, instead of just converting the old error messages to use `{cli}`, I am also using some new features that `{cli}` affords to improve the error messages. I think this is desirable, but you can let me know if you see any shortcomings to pursuing this further.

## On main branch

``` r
library(lintr)
backport_linter("oldrel-99")
#> Error: `r_version` must be a version number or one of 'devel', 'release', 'oldrel-1', 'oldrel-2', 'oldrel-3', 'oldrel-4', 'oldrel-5', 'oldrel-6', 'oldrel-7', 'oldrel-8', 'oldrel-9'
```

<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## On feature branch

``` r
library(lintr)
backport_linter("oldrel-99")
#> Error in `normalize_r_version()` at lintr/R/backport_linter.R:37:3:
#> ! `r_version` is not valid:
#> ℹ It must be a version number or one of 'devel', 'release', 'oldrel-1',
#>   'oldrel-2', 'oldrel-3', 'oldrel-4', 'oldrel-5', 'oldrel-6', 'oldrel-7',
#>   'oldrel-8', 'oldrel-9'
#> ✖ You entered 'oldrel-99' instead
#> Backtrace:
#>     ▆
#>  1. └─lintr::backport_linter("oldrel-99")
#>  2.   └─lintr:::normalize_r_version(r_version) at lintr/R/backport_linter.R:37:3
#>  3.     └─cli::cli_abort(...) at lintr/R/backport_linter.R:96:7
#>  4.       └─rlang::abort(...)
```

<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/test-xml_nodes_to_lints.R
```diff
@@ -110,14 +110,13 @@ test_that("it doesn't produce invalid lints", {
         range_start_xpath = xp_invalid,
         range_end_xpath = xp_range_end
       )
-    },
-    rex::rex("Could not find range start for lint. Defaulting to start of line.")
```
IndrajeetPatil, line 114: If we are committed to testing this full message, I am not sure if there is a cleaner way than just using the snapshot test. Let me know if you'd much prefer instead to only test some part of the message and not use snapshots.

R/lint.R
```diff
@@ -446,23 +451,27 @@ check_ranges <- function(ranges, max_col) {
...
     if (!is_number(range, 2L)) {
-      stop("`ranges` must only contain length 2 integer vectors without NAs.", call. = FALSE)
+      cli_abort("{.arg ranges} must only contain length 2 integer vectors without {.code NA}s.")
     } else if (!is_valid_range(range, max_col)) {
-      stop(sprintf(
-        "All entries in `ranges` must satisfy 0 <= range[1L] <= range[2L] <= nchar(line) + 1 (%d).", max_col
-      ), call. = FALSE)
+      cli_abort(c(
+        x = "Invalid range specified.",
+        i = "All entries in {.arg ranges} must satisfy 0 <= range[1L] <= range[2L] <= nchar(line) + 1 ({.val {max_col}})."
```
IndrajeetPatil, line 466: @MichaelChirico I'm having a hard time understanding this message. We talk about `ranges`, but the condition to be specified refers to `range`, which is confusing. Can we improve it somehow?

MichaelChirico: Per docs, `ranges` is a list of ranges, i.e. `ranges[[ii]]` needs to be a range.

But we've found `ranges[[ii]][1L] < 0 || ranges[[ii]][1L] > ranges[[ii]][2L] || ranges[[ii][2L] > max_col`, hence the error. We can probably `seq_along()` instead to provide info like:

> ranges[[5]] (2, 1) is not a valid range, which must satisfy 0 <= ranges[1L] <= ranges[2L] <= nchar(line) + 1 (5)

Otherwise we could refactor to something like:

```r
if (!is_number(...)) {
  stop(...)
}
tryCatch(check_valid_range(range,max_col),
  negative_column_error = \(c) stop("ranges[[5]] has lower bound %d < 0"),
  range_decreasing_error = \(c) stop("ranges[[5]] has mixed bounds ranges[1]=%d > %d=ranges[2]"),
  overwide_column_error = \(c) stop("ranges[[5]] has upper bound %d > {max_col}")
)
# or, simpler
tryCatch(check_valid_range(range,max_col),
  range_validation_error = \(c) stop("ranges[[5]]: ", conditionMessage(c))
)
```

IndrajeetPatil: Thanks for the explanation. 

Out of those options, I like this:

```r
tryCatch(check_valid_range(range,max_col),
  range_validation_error = \(c) stop("ranges[[5]]: ", conditionMessage(c))
)
```

R/exclude.R
```diff
@@ -127,9 +127,12 @@ parse_exclusions <- function(file,
 
   if (length(starts) > 0L) {
     if (length(starts) != length(ends)) {
-      starts_msg <- line_info(starts, type = "start")
-      ends_msg <- line_info(ends, type = "end")
-      stop(file, " has ", starts_msg, " but only ", ends_msg, " for exclusion from linting!", call. = FALSE)
+      starts_msg <- line_info(starts, type = "start") # nolint: object_usage_linter.
```
IndrajeetPatil, line 130: All newly introduced nolint directives are due to https://github.com/r-lib/lintr/issues/2252.

IndrajeetPatil:
@olivroy If you have time, I would also love to get your feedback on this PR :)

R/addins.R
```diff
@@ -1,23 +1,23 @@
 # nocov start
 addin_lint <- function() {
   if (!requireNamespace("rstudioapi", quietly = TRUE)) {
-    stop("'rstudioapi' is required for add-ins.", call. = FALSE)
+    cli_abort("{.pkg rstudioapi} is required for add-ins.")
   }
   filename <- rstudioapi::getSourceEditorContext()
   if (filename$path == "") {
-    return("Current source has no path. Please save before continue")
+    return("Current source has no path. Please save before continuing.")
```
AshesITR, line 8: Shouldn't this output a warning and return empty `Lints()` instead?

R/backport_linter.R
```diff
@@ -93,7 +93,11 @@ normalize_r_version <- function(r_version) {
     version_names <- c("devel", "release", paste0("oldrel-", seq_len(length(minor_versions) - 2L)))
     if (!r_version %in% version_names) {
       # This can only trip if e.g. oldrel-99 is requested
-      stop("`r_version` must be a version number or one of ", toString(sQuote(version_names)), call. = FALSE)
+      cli_abort(c(
+        "{.arg r_version} is not valid:",
+        i = "It must be a version number or one of {toString(sQuote(version_names))}.",
+        x = "You entered {toString(sQuote(r_version))} instead."
```
AshesITR, line 99: `r_version` should be scalar, so no toString needed.

R/exclude.R
AshesITR: It seems wise to mark them with the issue number then so they are cleaned up in a PR addressing this issue.

R/exclude.R
```diff
@@ -127,9 +127,12 @@ parse_exclusions <- function(file,
...
+      ends_msg <- line_info(ends, type = "end") # nolint: object_usage_linter.
+      cli_abort(c(
+        i = "Equal number of line starts and ends expected for exclusion from linting.",
+        x = "{.file {file}} has {.strong {starts_msg}}, but only {.strong {ends_msg}}."
```
AshesITR, line 134: nit: For starts < ends, the message is weird (2 starts but only 3 ends)

R/addins.R
IndrajeetPatil: You mean `return(list())`, right?

R/exclude.R
IndrajeetPatil: Good catch! We have no test for this condition, AFAICT, and so didn't think about that scenario. Changed the message.

R/addins.R
AshesITR: A `list()` with class `"lints"` would be better (so that the print method is the correct one).

R/addins.R
```diff
@@ -1,23 +1,24 @@
...
 
 addin_lint_package <- function() {
   if (!requireNamespace("rstudioapi", quietly = TRUE)) {
-    stop("'rstudioapi' is required for add-ins.", call. = FALSE)
+    cli_abort("{.pkg rstudioapi} is required for add-ins.")
   }
   project <- rstudioapi::getActiveProject()
   if (is.null(project)) {
-    message("No project found, passing current directory")
+    cli_inform("No configuration found. Using default linters.")
```
AshesITR, line 21: Copy-paste error?

NEWS.md
```diff
@@ -28,6 +28,7 @@
 
 ## New and improved features
 
+* All user-facing messages are now prepared using the [`{cli}`](https://github.com/r-lib/cli) package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
```
MichaelChirico, line 31: I thought I remembered {pkgdown} automatically linking packages like `{pkg}`, but the current NEWS does not do so, any idea?

R/backport_linter.R
```diff
@@ -93,7 +93,11 @@ normalize_r_version <- function(r_version) {
     version_names <- c("devel", "release", paste0("oldrel-", seq_len(length(minor_versions) - 2L)))
     if (!r_version %in% version_names) {
       # This can only trip if e.g. oldrel-99 is requested
-      stop("`r_version` must be a version number or one of ", toString(sQuote(version_names)), call. = FALSE)
+      cli_abort(c(
+        "{.arg r_version} is not valid:",
+        i = "It must be a version number or one of {sQuote(version_names)}.",
```
MichaelChirico, line 98: how does {cli} handle aggregation? before we needed `toString()` for the vector `version_names`.

MichaelChirico: PS I'm not generally a fan of `sQuote()` since it's inconsistent across platforms, WDYT about making this `{.val r_version}` instead?

R/exclude.R
```diff
@@ -200,13 +203,11 @@ add_exclusions <- function(exclusions, lines, linters_string, exclude_linter_sep
...
-        bad <- excluded_linters[!matched]
-        warning(
-          "Could not find linter", if (length(bad) > 1L) "s" else "", " named ",
-          glue_collapse(sQuote(bad), sep = ", ", last = " and "),
-          " in the list of active linters. Make sure the linter is uniquely identified by the given name or prefix.",
-          call. = FALSE
-        )
+        bad <- excluded_linters[!matched] # nolint: object_usage_linter.
+        cli_warn(c(
+          x = "Could not find linter{?s} named {.field {sQuote(bad)}} in the list of active linters.",
```
MichaelChirico, line 208: same Q for aggregating `bad`, does {cli} use `glue_collapse()`?

R/exclude.R
```diff
@@ -248,22 +249,20 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
...
   if (any(unnamed)) {
-    # must be character vectors of length 1
     bad <- vapply(
       seq_along(x),
       function(i) {
         unnamed[i] && (!is.character(x[[i]]) || length(x[[i]]) != 1L)
       },
       logical(1L)
     )
+    bad_idx <- which(bad) # nolint: object_usage_linter.
```
MichaelChirico, line 259: why define this instead of doing `which(bad)` inside the glue expression below? If we're going to define it, at least `any(bad)` should become `length(bad_idx) > 0L`.

R/expect_lint.R
```diff
@@ -41,10 +41,12 @@
...
-    stop( # nocov start
-      "'expect_lint' is designed to work within the 'testthat' testing framework, but 'testthat' is not installed.",
-      call. = FALSE
-    ) # nocov end
+    # nocov start
+    cli_abort(c(
+      i = "{.fun expect_lint} is designed to work within the {.pkg testthat} testing framework.",
+      x = "{.pkg testthat} is not installed."
+    ))
+    # nocov end
```
MichaelChirico, line 49: this message is way too verbose IMO. R/addins.R is much better: `{.pkg testthat} is required to use {.fun expect_lint}`. The old message was OK because it's just a sentence, I think it's a lot harder to process all the visual elements introduced by {cli}:

```
Error:
ℹ `expect_lint()` is designed to work within the testthat testing
  framework.
✖ testthat is not installed.
Run `rlang::last_trace()` to see where the error occurred.
```

vs

```
Error: 'expect_lint' is designed to work within the 'testthat' testing framework, but 'testthat' is not installed.
```

R/expect_lint.R
```diff
@@ -106,11 +108,10 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
...
           }
           if (!is.logical(ok)) {
-            stop(
-              "Invalid regex result, did you mistakenly have a capture group in the regex? ",
-              "Be sure to escape parenthesis with `[]`",
-              call. = FALSE
-            )
+            cli_abort(c(
+              x = "Invalid regex result. Did you mistakenly have a capture group in the regex?",
+              i = "Be sure to escape parenthesis with `[]`."
```
MichaelChirico, line 113:
```suggestion
              i = "Be sure to escape parentheses with `[]`."
```

R/lint.R
```diff
@@ -335,10 +340,10 @@ validate_linter_object <- function(linter, name) {
     return(linter)
   }
   if (!is.function(linter)) {
-    stop(gettextf(
-      "Expected '%s' to be a function of class 'linter', not a %s of class '%s'",
-      name, typeof(linter), class(linter)[[1L]]
-    ), call. = FALSE)
+    cli_abort(c(
+      i = "Expected {.fn {name}} to be a function of class {.cls linter}.",
+      x = "Instead, it is of class {.cls {class(linter)[[1L]]}}."
```
MichaelChirico, line 345: I would assign this outside the error, too many delimiters going on in there:

```r
linter_class <- class(linter)[1L]
... {.cls linter_class}
```

MichaelChirico: actually, elsewhere we just use `class()` inside `{.cls` and {cli} handles it well to give a full class summary:

```r
cli_warn("{.cls {class(Sys.time())}}")
# Warning message:
# <POSIXct/POSIXt>
```

So let's just use that.

R/lint.R
```diff
@@ -446,23 +454,27 @@ check_ranges <- function(ranges, max_col) {
...
     if (!is_number(range, 2L)) {
-      stop("`ranges` must only contain length 2 integer vectors without NAs.", call. = FALSE)
+      cli_abort("{.arg ranges} must only contain {.cls integer} vectors of length 2 without {.code NA}s.")
     } else if (!is_valid_range(range, max_col)) {
-      stop(sprintf(
-        "All entries in `ranges` must satisfy 0 <= range[1L] <= range[2L] <= nchar(line) + 1 (%d).", max_col
-      ), call. = FALSE)
+      cli_abort(c(
+        x = "Invalid range specified.",
+        i = "Argument {.arg ranges} must satisfy 0 <= range[1L] <= range[2L] <= nchar(line) + 1 ({.val {max_col}})."
```
MichaelChirico, line 469: aside: in #2551 we specifically stopped `range[1L]==0`, but here it's valid. Does that mean we solved that issue incorrectly? Maybe it's something that should be handled in the SARIF code specifically instead? cc @AshesITR

R/linter_tags.R
```diff
@@ -117,13 +117,10 @@ validate_linter_db <- function(available, package) {
...
   if (!all(c("linter", "tags") %in% colnames(available))) {
-    warning(
-      "`linters.csv` must contain the columns 'linter' and 'tags'.\nPackage '",
-      package, "' is missing ",
-      paste0("'", setdiff(c("linter", "tags"), names(available)), "'", collapse = " and "),
-      ".",
-      call. = FALSE
-    )
+    cli_warn(c(
+      i = "{.file linters.csv} must contain the columns {.str linter} and {.str tags}.",
```
MichaelChirico, line 121: could this be a `{.field linter}` instead of `{.str linter}`?

R/linter_tags.R
```diff
@@ -117,13 +117,10 @@ validate_linter_db <- function(available, package) {
...
+      x = "Package {.pkg package} is missing {.str {setdiff(c('linter', 'tags'), names(available))}}."
```
MichaelChirico, line 122: again I'd move this out of the error string, too much nesting

tests/testthat/test-Lint-builder.R
```diff
@@ -2,26 +2,32 @@ test_that("Lint() errors on invalid input", {
...
   expect_error(
     Lint("dummy.R", ranges = c(1L, 3L)),
-    rex::rex("`ranges` must be NULL or a list.")
+    "`ranges` must be `NULL` or a <list>",
+    fixed = TRUE
   )
   expect_error(
     Lint("dummy.R", ranges = list(1L)),
-    rex::rex("`ranges` must only contain length 2 integer vectors without NAs.")
+    "`ranges` must only contain <integer> vectors of length 2 without `NA`s.",
```
MichaelChirico, line 20: this looks really fragile TBH. What happens when {cli} changes the formatting of any of the `{.cls}` `{.fn}` `{.str}` etc markers? Maybe we need a helper like `cli_style()` that can match what {cli} does to the strings? Does {cli} itself offer anything like this?

R/lint.R
AshesITR: IINM Jump-to-source in RStudio accepts 0 as start-of-line while 1 jumps to after the first character. I see two options: Disallow 0 in general or handle it in sarif output.

If I indeed recall correctly about the special behavior for column 0, handling this case in sarif output seems acceptable. Do you have the time to test this?

R/tree_utils.R
```diff
@@ -10,10 +10,10 @@ generate_top_level_map <- function(pc) {
     tl_parent[i_not_assigned] <- pc$parent[match(tl_parent[i_not_assigned], pc$id)]
     i_not_assigned <- which(!tl_parent %in% tl_ids)
     if (length(i_not_assigned) >= prev_length) { # nocov start
-      stop(
-        "Logical error: unassigned set did not shrink. Check file syntax and please report as a lintr bug.",
-        call. = FALSE
-      )
+      cli_abort(c(
+        x = "Logical error: unassigned set did not shrink.",
+        i = "Check file syntax and please report as a {.pkg lintr} bug."
```
AshesITR, line 15: Add the issues URL here too for consistency?

R/namespace_linter.R
```diff
@@ -84,11 +84,10 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
     if (any(failed_namespace)) {
-      stop(
-        "Failed to retrieve namespaces for one or more of the packages used with `::` or `:::`. ",
-        "Please report the issue at https://github.com/r-lib/lintr/issues.",
-        call. = FALSE
-      )
+      cli_abort(c(
+        x = "Failed to retrieve namespaces for one or more of the packages used with `::` or `:::`. ",
+        i = "Please report the issue at {.url https://github.com/r-lib/lintr/issues}."
+      ))
```
olivroy, line 90:
```suggestion
      cli_abort(
        "Failed to retrieve namespaces for one or more of the packages used with `::` or `:::`. ",
        .internal = TRUE
     )
```

R/linter_tags.R
olivroy: There is `{.val {c('linter', 'tags')}` if you want quoting + color

R/lint.R
olivroy: There is also `{.obj_type_friendly {Sys.time()}}`, but in dev cli, will only show the first class.

olivroy (commented):
Feel free to disregard!

Disclaimer: I am not a heavy user of lintr, just the basic, so my comments may be invalid.

I also think it is not worth fiddling with the default of `cli::cli_abort()`, showing a "!" bullet for the first seems like a good default?

Since the review is getting a bit heavy, maybe some of that may be tackled in a follow-up?

I generally like how the tidyverse errors about type error (like in standalone type check from rlang

arg must be a <expected>, not <actual>.

So, while reviewing error messages, I think it is worth standardising over this.

```r
# a good format
cli::cli_abort(c(
  "What happened",
  i = "What you should do"
))

# if internal 
cli::cli_abort(c(
  "What happened",
  i = "What you should do / more info"
  ),
  .internal = TRUE
)
```

I would consider also maybe taking an rlang dependency to take advantage of `check_installed()` for Suggests dependency + potentially the standalone type check messages, `arg_match0()`, etc.

R/addins.R
IndrajeetPatil: Good catch!

NEWS.md
IndrajeetPatil: Oh, {pkgdown} does link it, but I had more in mind someone who might be reading the NEWS on CRAN.

But, on a second thought, maybe it's better to leave out the URL since it's not that hard to someone just look for cli on CRAN.

R/backport_linter.R
IndrajeetPatil: I don't like it either, but I wanted to preserve the previous behaviour. Happy to change that.

Just one thing to keep in mind: Although the value is highlighted in the console and so it stands out:

```r
r_version <- getRversion()

f <- \() cli::cli_abort("Version {.val {r_version}} is not valid")
g <- \() cli::cli_abort("Version {.val {sQuote(r_version)}} is not valid")

f()
g()
```

Screenshot 2024-04-13 at 15 57 53

That is not the case when we need to write unit tests:

```r
expect_error(f(), "Version 4.3.3 is not valid")
expect_error(g(), "Version '4.3.3' is not valid")
```

But I don't think it's too big a deal.

R/linter_tags.R
IndrajeetPatil: I wanted the column names to be quoted, this is why I didn't use `.field` here.

R/shared_constants.R
```diff
@@ -274,14 +274,16 @@ extract_glued_symbols <- function(expr, interpret_glue) {
 }
 
 unexpected_glue_parse_error <- function(cond) {
-  stop("Unexpected failure to parse glue call, please report: ", conditionMessage(cond), call. = FALSE) # nocov
+  cli_abort(c(
+    x = "Unexpected failure to parse glue call.",
+    i = "Please report: {conditionMessage(cond)}"
+  )) # nocov
```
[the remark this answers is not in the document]
olivroy: Gets added automatically in package code when a bug report field is present. just add this small function in lintr and call `load_all()`. It should show!

```r
f <- function() {
cli::cli_abort("a", .internal = TRUE)
}
```

R/backport_linter.R
olivroy:
```r
f <- \() cli::cli_abort("Version {.str {r_version}} is not valid")
```

would quote it!

R/shared_constants.R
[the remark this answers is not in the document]
MichaelChirico: Confirming `.internal` already works in our namespace:

```
> f()
Error in `f()`:
! a
ℹ This is an internal error that was detected in the lintr package.
  Please report it at <https://github.com/r-lib/lintr/issues> with a reprex and the full backtrace.
Run `rlang::last_trace()` to see where the error occurred.
```

R/backport_linter.R
MichaelChirico: > I wanted to preserve the previous behaviour.

Here's a case where I'm happy to break with old behavior, part of the advantage of moving to {cli} is easier consistency on things like this where we can get the same formatting of arguments, function names, packages etc. by off-loading that to {cli}. Things like `sQuote()` I see as our piecemeal attempt at this kind of formatting, and can be abandoned now.

R/lint.R
MichaelChirico: It's more that I don't really use RStudio these days, not sure how to test it (or at least, don't want to spend that much time setting up/figuring out testing).

I don't see anything about 0/1 mentioned in the `insertText()` documentation, though "start of document" is referred to as `c(1,1)`:

https://rstudio.github.io/rstudioapi/reference/rstudio-documents.html

R/shared_constants.R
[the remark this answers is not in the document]
IndrajeetPatil: Wonderful!

R/addins.R
IndrajeetPatil: Okay, I am going to create a new issue for this because currently we are returning just the empty list across many linters and so this change would be out of scope: https://github.com/r-lib/lintr/issues/2556

R/expect_lint.R
```diff
@@ -41,10 +41,9 @@
 #' @export
 expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
   if (!requireNamespace("testthat", quietly = TRUE)) {
-    stop( # nocov start
-      "'expect_lint' is designed to work within the 'testthat' testing framework, but 'testthat' is not installed.",
-      call. = FALSE
-    ) # nocov end
+    # nocov start
+    cli_abort("{.fun expect_lint} is designed to work within the {.pkg testthat} testing framework, which is not installed.")
```
MichaelChirico, line 45:
```suggestion
    cli_abort(
      "{.fun expect_lint} is designed to work within the {.pkg testthat} testing framework, which is not installed."
    )
```

NEWS.md
```diff
@@ -29,6 +29,7 @@
 
 ## New and improved features
 
+* All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
```
MichaelChirico, line 32: I would put this under a `Notes` section

R/expect_lint.R
```diff
@@ -88,10 +89,10 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
         itr <<- itr + 1L
         lapply(names(check), function(field) {
           if (!field %in% lint_fields) {
-            stop(sprintf(
-              "check #%d had an invalid field: \"%s\"\nValid fields are: %s\n",
-              itr, field, toString(lint_fields)
-            ), call. = FALSE)
+            cli_abort(c(
+              x = "Check {.val {itr}} has an invalid field: {.field {field}}.",
+              i = "Valid fields are: {.field {toString(lint_fields)}}."
```
MichaelChirico, line 94:
```suggestion
              i = "Valid fields are: {.field {lint_fields}}."
```

{cli} auto-aggregates, see https://cli.r-lib.org/articles/pluralization.html#use-the-length-of-character-vectors

After this change, I get:

```
Error:
✖ Check 1 has an invalid field: abc.
ℹ Valid fields are: filename, line_number, column_number, type, message, line, ranges, and linter.
Run `rlang::last_trace()` to see where the error occurred.
```

The coloring is also nicer, instead of coloring the commas that `toString()` produces, on the actual `lint_field` text is colored.

MichaelChirico: Note that by default {cli} adds `...` to truncate vectors >20 elements, if we want to avoid that it's pretty clunky:

https://cli.r-lib.org/reference/inline-markup.html#collapsing-inline-vectors

R/expect_lint.R
```diff
@@ -106,11 +107,10 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
...
           }
           if (!is.logical(ok)) {
-            stop(
-              "Invalid regex result, did you mistakenly have a capture group in the regex? ",
-              "Be sure to escape parenthesis with `[]`",
-              call. = FALSE
-            )
+            cli_abort(c(
+              x = "Invalid regex result. Did you mistakenly have a capture group in the regex?",
+              i = "Be sure to escape parentheses with `[]`."
```
MichaelChirico, line 112:
```suggestion
              i = "You can match parentheses with a character class, i.e. inside `[]`."
```

R/linter_tags.R
IndrajeetPatil: TBH, I think I started to retain this kind of nesting—consciously or unconsciously—after I realized that every new object to be consumed in `cli_*()` calls needs to have a `nolint: object_usage_linter` directive. 

Do you think we should close https://github.com/r-lib/lintr/issues/2252 first before finalizing this PR?

R/expect_lint.R
IndrajeetPatil: Will that be an issue? Do we expect any relevant vector in this context to contain more than 20 elements? 

The only one that I have encountered that has _many_ elements is related to supported R versions, which has length of 11. Otherwise, nothing comes to mind.

R/exclude.R
IndrajeetPatil: I am actually thinking that maybe we should fix #2252 before considering the current PR for a merge.

The current PR will be a good crucible to test it.

WDYT, @MichaelChirico?

tests/testthat/test-Lint-builder.R
IndrajeetPatil: I completely agree with you that this makes the tests fragile, and I can see why the tidyverse projects use snapshots for testing exceptions.

I think the good news here is that, because a lot of revisions in `{cli}` have already happened based on the feedback it received during its adoption across many tidyverse/r-lib repos, its formatting API is quite stable. Thus, it is quite unlikely that the formatting for any of these markers will change because that will lead to test failures across many repositories. 

Should we find some comfort in that knowledge? Or do you think that would instead be complacency?

MichaelChirico: >  it is quite unlikely that the formatting for any of these markers will change because that will lead to test failures across many repositories.

that's not quite true because snapshots are still not tested on CRAN, so breakages are not reported there. the volume of breakages of dev packages is not (AFAIK) monitored. we could ignore breakages in dev commit-by-commit and update snapshots as part of release, at the cost of not noticing real breakages when they happen.

R/linter_tags.R
MichaelChirico: sure, feel free to work on that -- I am not prioritizing it at the moment.

tests/testthat/test-Lint-builder.R
IndrajeetPatil: Most of the repos in question are under active development, and CI run snapshot tests almost on a daily basis as part of each build.
 
So, yes, if `{cli}` were to change its formatting API, there would be a big spike in failed CI builds across many repos.

I am not advocating for snapshot tests, but I am not too concerned about fragility of tests due to the changes in `{cli}` formatting either.

R/expect_lint.R
MichaelChirico: I agree it's not a major concern, just something to be aware of.

IndrajeetPatil:
Should we also refactor `lintr_deprecated()` to use `{cli}`?

I think it's fine without such an upgrade but maybe the team thinks otherwise.

R/exclude.R
MichaelChirico: SGTM, do you want to have a go at implementing?

AshesITR:
I would go for consistency. Let's also add `message()`, `warning()` and `stop()` to our undesirable functions to prevent regressions.

.lintr
```diff
@@ -14,8 +14,11 @@ linters: all_linters(
     undesirable_function_linter(modify_defaults(
       defaults = default_undesirable_functions,
       library = NULL,
-      options = NULL
-    )),
+      options = NULL,
+      message = "use cli::cli_inform()",
+      warning = "use cli::cli_warn()",
+      stop = "use cli::cli_abort()"
+    ), symbol_is_undesirable = FALSE),
```
MichaelChirico, line 21: did `message` as a symbol only come up the once? If so I'd rather just `#nolint` that one case. If this was hitting several times, please ignore :)

R/exclude.R
```diff
@@ -261,12 +261,10 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
...
     if (any(bad)) {
-      stop(
-        "Full file exclusions must be character vectors of length 1. items: ",
-        toString(which(bad)),
-        " are not!",
-        call. = FALSE
-      )
+      cli_abort(c(
+        i = "Full file exclusions must be {.cls character} vectors of length 1.",
+        x = "Items at following indexes are not: {.val {which(bad)}}."
```
MichaelChirico, line 266:
```suggestion
        x = "Items at the following indices are not: {.val {which(bad)}}."
```

R/exclude.R
```diff
@@ -281,12 +279,10 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
...
     if (any(bad)) {
-      stop(
-        "Full line exclusions must be numeric or integer vectors. items: ",
-        toString(which(bad)),
-        " are not!",
-        call. = FALSE
-      )
+      cli_abort(c(
+        i = "Full line exclusions must be {.cls numeric} or {.cls integer} vectors.",
+        x = "Items at following indexes are not: {.val {which(bad)}}."
```
MichaelChirico, line 284:
```suggestion
        x = "Items at the following indices are not: {.val {which(bad)}}."
```

R/lint.R
```diff
@@ -335,10 +340,10 @@ validate_linter_object <- function(linter, name) {
     return(linter)
   }
   if (!is.function(linter)) {
-    stop(gettextf(
-      "Expected '%s' to be a function of class 'linter', not a %s of class '%s'",
-      name, typeof(linter), class(linter)[[1L]]
-    ), call. = FALSE)
+    cli_abort(c(
+      i = "Expected {.fn {name}} to be a function of class {.cls linter}.",
+      x = "Instead, it is {.obj_type_friendly {linter}}."
```
MichaelChirico, line 345: This gives us a {cli} version dependency right?

https://github.com/r-lib/cli/blob/main/NEWS.md#cli-340

R/lint.R
MichaelChirico: Forgot to link this back: https://github.com/rstudio/rstudioapi/issues/299

R/namespace_linter.R
```diff
@@ -84,11 +84,7 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
 
     # nocov start
     if (any(failed_namespace)) {
-      stop(
-        "Failed to retrieve namespaces for one or more of the packages used with `::` or `:::`. ",
-        "Please report the issue at https://github.com/r-lib/lintr/issues.",
-        call. = FALSE
-      )
+      cli_abort_internal("Failed to retrieve namespaces for one or more of the packages used with `::` or `:::`. ")
```
MichaelChirico, line 87: this function is only used once, any advantage over just using `cli_abort("...", .internal=TRUE)` here?

R/settings.R
```diff
@@ -98,7 +98,10 @@ read_config_file <- function(config_file) {
   if (endsWith(config_file, ".R")) {
     load_config <- function(file) sys.source(file, config, keep.source = FALSE, keep.parse.data = FALSE)
     malformed <- function(e) {
-      stop("Malformed config file, ensure it is valid R syntax\n  ", conditionMessage(e), call. = FALSE)
+      cli_abort(c(
+        "Malformed config file, ensure it is valid R syntax.",
+        conditionMessage(e)
```
MichaelChirico, line 103: Should we use `i = ` here as well as below?

R/settings.R
```diff
@@ -153,10 +163,8 @@ read_config_file <- function(config_file) {
 validate_config_file <- function(config, config_file, defaults) {
   matched <- names(config) %in% names(defaults)
   if (!all(matched)) {
-    warning(
-      "Found unused settings in config '", config_file, "': ", toString(names(config)[!matched]),
-      call. = FALSE
-    )
+    unused_settings <- names(config)[!matched] # nolint: object_usage_linter.
+    cli_warn("Found unused settings in config {.str config_file}: {toString(unused_settings)}")
```
MichaelChirico, line 167: do we need `toString()` here?

R/tree_utils.R
```diff
@@ -9,12 +9,11 @@ generate_top_level_map <- function(pc) {
...
     i_not_assigned <- which(!tl_parent %in% tl_ids)
-    if (length(i_not_assigned) >= prev_length) { # nocov start
-      stop(
-        "Logical error: unassigned set did not shrink. Check file syntax and please report as a lintr bug.",
-        call. = FALSE
-      )
-    } # nocov end
+    # nocov start
+    if (length(i_not_assigned) >= prev_length) {
+      cli_abort("Logical error: unassigned set did not shrink. Check file syntax.")
```
MichaelChirico, line 14: Please retain "and report as a bug", we want our logic to catch all such cases & hitting this bug means we have something new to address.

R/with.R
```diff
@@ -44,12 +44,10 @@ modify_defaults <- function(defaults, ...) {
...
     bad_nms <- setdiff(nms[to_null], names(defaults))
-    is_are <- if (length(bad_nms) > 1L) "are" else "is"
-    warning(
-      "Trying to remove ", glue_collapse(sQuote(bad_nms), sep = ", ", last = " and "),
-      ", which ", is_are, " not in `defaults`.",
-      call. = FALSE
-    )
+    is_are <- if (length(bad_nms) > 1L) "are" else "is" # nolint: object_usage_linter.
+    cli_warn(c(
+      i = "Trying to remove {.field {bad_nms}}, which {is_are} not in {.arg defaults}."
```
MichaelChirico, line 49: We should use {cli} pluralization here, something like `{?is/are}`:

https://cli.r-lib.org/articles/pluralization.html

R/with.R
```diff
@@ -211,10 +208,11 @@ call_linter_factory <- function(linter_factory, linter_name, package) {
   linter <- tryCatch(
     linter_factory(),
     error = function(e) {
-      stop(
-        "Could not create linter with ", package, "::", linter_name, "(): ", conditionMessage(e),
-        call. = FALSE
-      )
+      pkg_fn <- paste0(package, "::", linter_name) # nolint: object_name_linter
+      cli_abort(c(
+        "Could not create linter with {.fun pkg_fn}.",
```
MichaelChirico, line 213: This works IINM

```suggestion
        "Could not create linter with {.fun {package}::{linter_name}}.",
```

R/xp_utils.R
```diff
@@ -83,10 +83,9 @@ xp_call_name <- function(expr, depth = 1L, condition = NULL) {
   )
   is_valid_expr <- is_node(expr) || is_nodeset(expr)
   if (!is_valid_expr) {
-    stop(
-      "Expected an xml_nodeset or an xml_node, instead got an object of class(es): ",
-      toString(class(expr)),
-      call. = FALSE
+    cli_abort(
+      "{.arg expr} must be an {.cls xml_nodeset} or an {.cls xml_node}, \\
```
MichaelChirico, line 87: ah this is a weird feature of the parser I'd rather not use TBH, I've not seen it documented (e.g. it's not in `?Quotes`) and don't see it used very often.

R/namespace_linter.R
IndrajeetPatil: This function _needs_ to be internally defined for this feature to work, AFAIK.

R/lint.R
IndrajeetPatil: Good point! I have updated DESCRIPTION to reflect this.

R/xp_utils.R
IndrajeetPatil: I am not sure what "this" refers to here. The use of `\\`?

R/exclude.R
IndrajeetPatil: I have instead added TODO linking to this issue to address.

The PR has already been pending for a while, and I don't want to delay its merge even further while trying to find time to work on a different feature.

R/lint.R
IndrajeetPatil: I don't consider this to be a blocker for this PR. If something needs to be changed about the message here, that's a separate issue than using `{cli}` for formatting messages.

R/tree_utils.R
IndrajeetPatil: Using `cli_abort_internal()` here to retain this.

tests/testthat/test-Lint-builder.R
IndrajeetPatil: @MichaelChirico Can I resolve this comment thread? Or do you have further thoughts here?

R/settings.R
IndrajeetPatil: Good catch. That was unintentional. I have gotten rid of it from this file, and also elsewhere in the context of cli.

R/settings.R
IndrajeetPatil: The problem is that when the condition message (in `e`) itself is formatted with `{cli}`, either the formatting looks messed up, or part of the condition message is dropped.

E.g. notice that part of the condition message is dropped here:

``` r
foo <- function(expr) {
  cli::cli_abort(c(
    i = "{.arg expr} must be an {.cls xml_nodeset} or an {.cls xml_node}.",
    x = "Instead, it is {.obj_type_friendly {expr}}."
  ))
}

tryCatch(
  foo(1),
  error = function(e) {
    print(e)
    cat("---------------------------\n")
    cli::cli_inform(c(
      i = "Something went wrong here.",
      x = e$message
    ))
  }
)
#> <error/rlang_error>
#> Error in `foo()`:
#> ℹ `expr` must be an <xml_nodeset> or an <xml_node>.
#> ✖ Instead, it is a number.
#> ---
#> Backtrace:
#>     ▆
#>  1. ├─base::tryCatch(...)
#>  2. │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>  3. │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  4. │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
#>  5. └─global foo(1)
#> ---------------------------
#> ℹ Something went wrong here.
#> `expr` must be an <xml_nodeset> or an <xml_node>.
```

<sup>Created on 2024-05-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

@olivroy Do you know how to deal with this?

olivroy: I think it is better to use the following:

```r
cli::cli_abort(
  "message",
  parent = e
)
```

If you want me to do it, you can leave it for now, open a new issue and I can open a PR to address that post-merge!

R/with.R
IndrajeetPatil: Ah, we are using pluralization, but I had missed this instance somehow. Fixed it.

R/settings.R
IndrajeetPatil: Thanks! I think, for now, I am going to leave this message as is.

I have created a new issue for a follow-up: https://github.com/r-lib/lintr/issues/2588. We can work on that once the current PR is merged.

olivroy (approved):

IndrajeetPatil:
I am going to wait for all three reviewers to approve this before I merge it.

MichaelChirico:
Done the R/ code, it looks great!

I will get to the tests/ changes in the next pass. Getting close!

tests/testthat/test-Lint-builder.R
MichaelChirico: are you ok being responsible with another pass over the tests to remove the cli formatting if we do get hit by such a breaking change? note that it would likely mean CRAN failing so would need a relatively fast turnaround time.

IndrajeetPatil: Yes, I can take responsibility of updating the tests in case of formatting changes in `{cli}`.

tests/testthat/test-parse_exclusions.R
```diff
@@ -150,7 +150,7 @@ test_that("it throws an error if start and end are unpaired", {
     a
     test
   "))
-  expect_error(lintr:::parse_exclusions(t1), "but only")
+  expect_error(lintr:::parse_exclusions(t1), "and")
```
MichaelChirico, line 153: this looks too weak, "and" could appear in any number of error messages, care to revise?

MichaelChirico (approved):

PR #2420, opened 2023-12-12
Customize lint message for nzchar_linter()

### LINKED ISSUE #2383: Confusing (misleading?) lint message from `nzchar_linter()`
The current message makes it seem like the `nzchar()` is a drop-in replacement for  `nchar(x) == 0`; at least that's my read of it.

``` r
library(lintr)
lint(
  text = 'if (nchar(x) == 0L) print("x is empty")',
  linters = nzchar_linter()
)
#> <text>:1:5: warning: [nzchar_linter] Instead of comparing nchar(x) to 0, use nzchar(). Whenever missing data is possible, please take care to use nzchar(., keepNA = TRUE); nzchar(NA) is TRUE by default.
#> if (nchar(x) == 0L) print("x is empty")
#>     ^~~~~~~~~~~~~~
```

But, instead, the `nzchar()` result needs to be negated to get the equivalent output:

```r
x <- ""
identical(
  nchar(x) == 0L,
  !nzchar(x)
)
#> [1] TRUE
```

<sup>Created on 2023-12-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Can we refine the lint message to reflect this?

MichaelChirico:
Closes #2383


tests/testthat/test-nzchar_linter.R
```diff
@@ -25,26 +25,26 @@ test_that("nzchar_linter skips as appropriate for other nchar args", {
...
-    rex::rex("Use nzchar() instead of comparing nchar(x) to 0"),
+    rex::rex("Use !nzchar(x) instead of nchar(x) == 0"),
     linter
   )
 })
 
 test_that("nzchar_linter blocks simple disallowed usages", {
   linter <- nzchar_linter()
-  lint_msg_quote <- rex::rex('Use nzchar() instead of comparing strings to ""')
+  lint_msg_quote <- rex::rex('Use !nzchar(x) instead of x == "')
```
AshesITR, line 35: missing a double quote

AshesITR (commented):
nice!

tests/testthat/test-nzchar_linter.R
MichaelChirico:
```suggestion
  lint_msg_quote <- rex::rex('Use !nzchar(x) instead of x == ""')
```

IndrajeetPatil:
Overall, this looks good! Thanks.

While you are at `nzchar_linter()`, can you please also update the [examples](https://lintr.r-lib.org/dev/reference/nzchar_linter.html#ref-examples) to follow the same pattern we are following in other linters (parallel examples that do and don't lint)?

IndrajeetPatil (approved):

PR #2422, opened 2023-12-13
handle terminal switch() in return_linter()

### LINKED ISSUE #2343: return_linter() should allow switch() where every entry has an exit call

MichaelChirico:
Closes #2343.

cc @MEO265


MichaelChirico:
😆 Definitely my most unexpected time getting hit with the `commented_code_linter()` 🔨

```r
switch(x, ...) | expr[1]: switch; expr[2]: x
```

Pretty impressive that parses as valid code!

AshesITR:
> 😆 Definitely my most unexpected time getting hit with the `commented_code_linter()` 🔨
> 
> ```r
> switch(x, ...) | expr[1]: switch; expr[2]: x
> ```
> 
> Pretty impressive that parses as valid code!

I like to prepend the code by `#>`, which generally kills parsing due to a missing LHS of `>` ;)

MichaelChirico:
> > 😆 Definitely my most unexpected time getting hit with the `commented_code_linter()` 🔨
> > ```r
> > switch(x, ...) | expr[1]: switch; expr[2]: x
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Pretty impressive that parses as valid code!
> 
> I like to prepend the code by `#>`, which generally kills parsing due to a missing LHS of `>` ;)

right, but I would reserve that for code that's actually intended as code. In this case, I was writing a comment to explain the XPath

tests/testthat/test-return_linter.R
```diff
@@ -1521,3 +1471,231 @@ test_that("= assignments are handled correctly", {
...
+  all_return_lines <- trim_some("
+    foo <- function(x) {
+      switch(x,
+        a = return(1),
+        b = return(2)
+      )
+    }
+  ")
+  expect_lint(all_return_lines, list(implicit_msg, implicit_msg), implicit_linter)
+  expect_lint(all_return_lines, NULL, explicit_linter)
```
AshesITR, line 1523: Should the behavior of `explicit_linter` here depend on `allow_implicit_else`, i.e. force a `return()` in the default case if the latter is `FALSE`?

MichaelChirico: interesting Q... I lean towards yes. @MEO265 any thoughts?

AshesITR: Ditto actually for implicit return style requiring an unnamed arg to `switch()`.

MichaelChirico: OK, `allow_implicit_else` also covers whether the default case is required in a terminal `switch()` statement.

AshesITR (approved):
Awesome!

AshesITR (approved):

MichaelChirico:
roxygen test handy already 🤭

AshesITR (approved):

PR #2433, opened 2023-12-13
Support pattern-based exclusion in return_linter()

### LINKED ISSUE #2335: Extend `return_linter` to allow prefix exclusions
What about adding rudimentary globbing support to `except=`? It could detect `*` at the end of exceptions and convert those entries to prefix-exclusions.

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2271#discussion_r1401286415_

MichaelChirico:
Closes #2335.

Implementation is maybe not ideal since it requires running the `//FUNCTION | //OP-LAMBDA` XPath twice -- once to find the names of assigned functions, a second time to check for lints.

The workaround I could think of was to (1) find all `//FUNCTION | //OP-LAMBDA` nodes (2) partition into assigned vs. unassigned (3) trim `except|except_regex` from the assigned group (4) `combine_nodesets()` (4) check for lints. But IINM this would require refactoring a lot of other code as well, I'm not sure it's worth it -- `//FUNCTION | //OP-LAMBDA` should be relatively fast.


R/return_linter.R
```diff
@@ -73,13 +74,17 @@ return_linter <- function(
...
-  if (!allow_implicit_else || return_style == "explicit") {
-    except_xpath <- glue("parent::expr[not(
+  check_except <- !allow_implicit_else || return_style == "explicit"
+
+  if (check_except) {
+    except_xpath_fmt <- "parent::expr[not(
       preceding-sibling::expr/SYMBOL[{ xp_text_in_table(union(special_funs, except)) }]
-    )]")
+    )]"
+    if (is.null(except_regex)) except_xpath <- glue(except_xpath_fmt)
```
AshesITR, line 87: Is simultaneous use of `except` and `except_regex` precluded? Otherwise, I don't understand this condition. If it is, that should be documented.

R/return_linter.R
```diff
@@ -130,6 +141,12 @@ return_linter <- function(
 
   Linter(linter_level = "expression", function(source_expression) {
     xml <- source_expression$xml_parsed_content
+    if (check_except && !is.null(except_regex)) {
```
AshesITR, line 144: Is it possible to do static exclusion for `except` and post-process resulting lints to filter out regex matches? This approach is hard for me to understand due to the many branches.

R/return_linter.R
MichaelChirico: No, using both is fine. If there's `except_regex`, defer compiling this string until run-time when we know the list of possible assignments.

Anyway your other comment is probably the better way to handle this, so resolving.

R/return_linter.R
MichaelChirico: Hmm actually I think it may be hard due to the recursion.

```r
ExcludeMe <- function() {
  lapply(seq_len(10), \(i) {
    i_need_a_return()
  })
  i_dont()
}
```

Not sure how to handle this without adding an unwind step for the recursion which is probably too much.

R/return_linter.R
MichaelChirico: I tried to clarify things with `defer_except` and a comment, hopefully that makes it easier to follow.

R/return_linter.R
```diff
@@ -73,16 +74,25 @@ return_linter <- function(
...
     return_functions = NULL,
-    except = NULL) {
+    except = NULL,
+    except_regex = NULL) {
   return_style <- match.arg(return_style)
 
-  if (!allow_implicit_else || return_style == "explicit") {
-    except_xpath <- glue("parent::expr[not(
+  check_except <- !allow_implicit_else || return_style == "explicit"
+  # We defer building the XPath strings in this case since we can't build the
```
AshesITR, line 82: Add a TODO referencing the XPath 2.0 issue?

R/return_linter.R
```diff
@@ -73,16 +74,25 @@ return_linter <- function(
...
+  #   pattern-based "except" logic directly into the XPath (because of v1.0)
+  defer_except <- check_except && !is.null(except_regex)
+
+  if (check_except) {
+    except_xpath_fmt <- "parent::expr[not(
       preceding-sibling::expr/SYMBOL[{ xp_text_in_table(union(special_funs, except)) }]
-    )]")
+    )]"
+    # nolint next: object_usage_linter. Hidden by dynamic glue.
```
AshesITR, line 90: Most of our `#nolint` comments skip the `_linter` suffix.

R/return_linter.R
```diff
@@ -130,6 +146,12 @@ return_linter <- function(
 
   Linter(linter_level = "expression", function(source_expression) {
     xml <- source_expression$xml_parsed_content
+    if (defer_except) {
+      assigned_functions <- xml_text(xml_find_all(xml, function_name_xpath))
+      except <- union(except, assigned_functions[re_matches(assigned_functions, except_regex)])
+      except_xpath <- glue(except_xpath_fmt)
+      body_xpath <- glue(body_xpath_fmt)
```
AshesITR, line 153: Using named dots for glue like `glue::glue("{a}{b}", a = 1, b = 2)` would make the implicit dependency of `body_xpath` on `except_xpath` explicit (also, this would remove the object usage false positives). WDYT about that?

R/return_linter.R
MichaelChirico: Great idea, but I think there are a bunch of places with the same issue... #2453

R/return_linter.R
MichaelChirico: Great idea, much improved. Hopefully I didn't screw it up in the browser :)

R/return_linter.R
AshesITR: Thanks, the new flow reads nicely. The `glue(..., bla = bla)` also really helps.

R/return_linter.R
```diff
@@ -73,16 +74,25 @@ return_linter <- function(
...
+  if (check_except) {
+    except_xpath_fmt <- "parent::expr[not(
+      preceding-sibling::expr/SYMBOL[{ xp_text_in_table(except) }]
+    )]"
+    except <- union(special_funs, except)
+    if (!defer_except) except_xpath <- glue(except_xpath_fmt, except = except)
   }
 
   if (return_style == "implicit") {
+    # nolint next: object_usage. False positive.
```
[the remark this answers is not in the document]
AshesITR: (Apparently, yes - but I'm surprised)

AshesITR (approved):

AshesITR:
Very nice feature - makes the linter much more attractive for larger code bases.

Issue #2047, opened 2023-08-05
todo_comment_linter could get an argument for valid TODO patterns


MichaelChirico:
This can allow `TODO` _only if_ they fit certain patterns.

This could be used to enforce a rule like "All TODO must be tied to a bug, not a person".

```r
## BAD
# TODO(michaelchirico): fix this
# TODO: add testing

## GOOD
# TODO(#nnn): fix this
# TODO(#mmm): add testing
```

Maybe an argument like `allow_regex` or `allow_style_regex`, in this case `allow_regex = "TODO[(][#][0-9]+[)]:"` could work.

AshesITR:
Maybe just expose the regex as an optional argument and call it `regex=`?

`todo_comment_linter(regex = "# TODO\\(#\\d+\\):")` seems fine to me.

MichaelChirico:
The question then is how to interact this with the existing `todo` argument:

https://github.com/r-lib/lintr/blob/191e3b7638218920f90688dc6b96fe69c05f4a8a/R/comment_linters.R#L164

Should we be looking for matches of `paste0(todo, regex)`? Then the call would be `todo_comment_linter(regex = "\\(#\\d+\\):")`.

AshesITR:
I would let `regex` take precedence with a warning if both arguments are present.

MichaelChirico:
After #2433, I'm thinking the approach here will be:

```r
function(todo = c("todo", "fixme"), except_regex = NULL)
```

Where `except_regex` is used to filter out lints matched by `todo`. It also makes sense verbally: Don't use `todo` _unless_ it matches one of `except_regex`.

I think we should support case-sensitive matching too -- I would want `TODO(#1234)` to be OK but not `todo(#1234)`. Probably letting `todo=` be case-insenstive but `except_regex=` case-sensitive makes sense, otherwise we could add a third argument.

AshesITR:
I don't think we need a third argument. You can simply turn on ignore-case by prepending `except_regex` with `"(?i)"` if necessary.

PR #2437, opened 2023-12-14
Split comment_linters into two separate files for each linter

MichaelChirico:
Towards #2047 

This is for consistency with the rest of the code base, and to match the style of `R/$FILE` being tested in `tests/testthat/test-$FILE`.

I followed https://stackoverflow.com/a/44036771/3576984 to try and ensure code history is retained going forward.


IndrajeetPatil (approved):
Good idea!

PR #2438, opened 2023-12-14
Convert todo_comment_linter to XPath approach

MichaelChirico:
Follow-up to #2437, towards #2047.


R/todo_comment_linter.R
```diff
@@ -42,23 +42,18 @@
 #' @export
 todo_comment_linter <- function(todo = c("todo", "fixme")) {
   todo_comment_regex <- rex(one_or_more("#"), any_spaces, or(todo))
```
AshesITR, line 44: While at this: Do we want to lint rpxygen TODO comments as well?

`#' TODO` wouldn't lint ATM.

MichaelChirico: makes sense, though it's a bit odd for people to be leaving these comments in their documentation. 

OTOH, it's common: https://github.com/search?q=lang%3AR%20%2F%23%27%5Cs*TODO%2F&type=code

Let's push to follow-up though as the real fix (#2439) is based here and I'd rather not deal with more merge conflicts 😂

R/todo_comment_linter.R
```diff
@@ -42,23 +42,18 @@
...
-          line_number = token[["line1"]],
-          column_number = token[["col1"]],
-          type = "style",
-          message = "Remove TODO comments.",
-          line = source_expression[["lines"]][[as.character(token[["line1"]])]],
-          ranges = list(c(token[["col1"]], token[["col2"]]))
-        )
-      }
+
+  Linter(linter_level = "file", function(source_expression) {
```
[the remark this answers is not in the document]
MichaelChirico: I was thinking there's not much benefit to being expression-level for what's basically a simple text-match linter. But there might be a ton of comments in a huge file, so maybe caching would help.

R/todo_comment_linter.R
MichaelChirico: #2447

R/todo_comment_linter.R
```diff
@@ -42,23 +42,18 @@
...
-          column_number = token[["col1"]],
-          type = "style",
-          message = "Remove TODO comments.",
-          line = source_expression[["lines"]][[as.character(token[["line1"]])]],
-          ranges = list(c(token[["col1"]], token[["col2"]]))
-        )
-      }
+
+  Linter(linter_level = "expression", function(source_expression) {
+    xml <- source_expression$full_xml_parsed_content
```
MichaelChirico, line 47:
```suggestion
    xml <- source_expression$xml_parsed_content
```

R/todo_comment_linter.R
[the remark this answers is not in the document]
AshesITR: I'm thinking we should specify the lowest cacheable level, which should be expression here.

I have an idea cooking on how to speed up expression-level linters by running them on the file-level if they support it and being smart about the caching. But that's not completely ripe yet.

AshesITR (approved):

R/todo_comment_linter.R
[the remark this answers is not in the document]
MichaelChirico: sounds exciting. I think something like this is definitely possible and would be a big boon to anyone who never caches (like me 😅)

PR #2440, opened 2023-12-14
Mention and link to correct fct in use_lintr()

Bisaloo:
Noticed this while browsing `use_lintr()` docs.


MichaelChirico (approved):
nice!

PR #2446, opened 2023-12-15
Drop support for Travis, Jenkins, and Werker CI tools posting GitHub comments

### LINKED ISSUE #2148: Use {httr2}, not {httr}
{httr} is superseded in favor of {httr2}. We don't have much usage, so it should be "easy" to refactor in favor of the newer/better-supported package:

https://github.com/r-lib/lintr/blob/203be65e86ea843500abcc171fc76c54fcbc81b8/R/comments.R#L85-L115

MichaelChirico:
Closes #2148

(see comment history for an earlier implementation using {httr2}, but that's removed entirely now)


AshesITR:
Let's try (the new code) by committing a config here that would cause a comment?

R/comments.R
```diff
@@ -101,16 +101,10 @@ github_comment <- function(text, info = NULL, token = settings$comment_token) {
...
-    body = list(body = jsonlite::unbox(text)),
-    query = list(access_token = token),
-    encode = "json"
-  )
-
-  if (httr::status_code(response) >= 300L) {
-    message(httr::http_condition(response, "error", task = httr::content(response, as = "text")))
-  }
+  req <- httr2::request("https://api.github.com")
+  req <- httr2::req_url_path_append(req, file.path("repos", info$user, info$repo, api_subdir, "comments"))
```
mgirlich, line 105: You can of course use `file.path()` but `c()` also works

```suggestion
  req <- httr2::req_url_path_append(req, c("repos", info$user, info$repo, api_subdir, "comments"))
```

or even without `c()` it works

```
  req <- httr2::req_url_path_append(req, "repos", info$user, info$repo, api_subdir, "comments")
```

MichaelChirico: Awesome! I like the latter since it totally off-loads handling to {httr2}. Unlikely but `c()` could maybe eventually cause some issue by type coercion that's avoided using `...` instead.

MichaelChirico:
> Let's try (the new code) by committing a config here that would cause a comment?

I _think_ this is how it's meant to work?

https://github.com/r-lib/lintr/blob/24286466bd286de6d269b65a1c3b096e2531c009/R/methods.R#L98-L108
https://github.com/r-lib/lintr/blob/24286466bd286de6d269b65a1c3b096e2531c009/R/zzz.R#L331

But I don't know how to check

MichaelChirico:
The check warning shows up in the `Files` pane, but I don't think that's what this functionality is about, it's supposed to be posting on this `Discussion` pane, right? I haven't seen any of those comments in ages.

image

It reminds me of #854... should we just remove this functionality instead? The annotations in the File pane seem to work perfectly fine.

hadley:
You also might want to consider using gh, since it's a higher level wrapper around the GitHub API with better built in handling for rate-limiting, errors etc.

MichaelChirico:
OK looked more carefully. Why we don't get the bot is that we have GHA set up -- we'd need to turn off GHA and on Travis/Wercker/Jenkins to test it out:

https://github.com/r-lib/lintr/blob/4b59aac514c4c67873248e16c4db127c8627dfdc/R/methods.R#L95-L108

https://github.com/r-lib/lintr/blob/4b59aac514c4c67873248e16c4db127c8627dfdc/R/comments.R#L1-L3

We haven't gotten any bugs about the current setup not working... I see {pander} using .lintr with Travis... @daroczig are you aware if .lintr is still posting comments as expected for you?

daroczig:
sorry, @MichaelChirico, I don't have experience on using `lintr` to comment back on GH -- U was only using it report errors in the terminal (and fail the tests)

MichaelChirico:
Test failure seems unrelated but it's persistent, hmm wonder what's going on.

MichaelChirico:
OK I give up. Genuinely no idea what's going on. Skipping it.

MichaelChirico:
Bump for review :)

MichaelChirico:
Friendly bump for review :)

PR #2455, opened 2023-12-15
Sort default_undesirable docs robustly

MichaelChirico:
Have been getting some spurious diffs vs. `main` on this. Is anyone else seeing it? I'm a bit surprised because the `default*` objects are coming from `modify_defaults()`, which uses `platform_independent_order()` already:

https://github.com/r-lib/lintr/blob/fdbb9bfa1ea868ad66c6fe9a30b4ae872c03f56e/R/with.R#L59

Maybe this is a call to improve `platform_independent_order()`? That's pretty rudimentary indeed:

https://github.com/r-lib/lintr/blob/fdbb9bfa1ea868ad66c6fe9a30b4ae872c03f56e/R/utils.R#L206-L207

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -16,6 +16,7 @@ linters: all_linters(
     ),
     undesirable_function_linter(modify_defaults(
       defaults = default_undesirable_functions,
+      Sys.setlocale = NULL,
       library = NULL,
       options = NULL
     )),
--- a/.lintr
+++ b/.lintr
@@ -16,7 +16,6 @@ linters: all_linters(
     ),
     undesirable_function_linter(modify_defaults(
       defaults = default_undesirable_functions,
-      Sys.setlocale = NULL,
       library = NULL,
       options = NULL
     )),
--- a/R/utils.R
+++ b/R/utils.R
@@ -201,14 +201,9 @@ release_bullets <- function() {
 }
 # nocov end
 
-# see issue #923, PR #2455 -- some locales ignore _ when running sort(), others don't.
-#   we want to consistently treat "_" < "n" = "N"
-platform_independent_order <- function(x) {
-  old <- Sys.getlocale("LC_COLLATE")
-  Sys.setlocale("LC_COLLATE", "C")
-  on.exit(Sys.setlocale("LC_COLLATE", old))
-  order(tolower(x))
-}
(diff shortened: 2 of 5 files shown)


IndrajeetPatil:
> Is anyone else seeing it?

Yes, indeed, I was seeing these diffs in this `.Rd` file while working on some other PRs.

AshesITR:
Me too. I manually roll those changes back before committing. Similarly for some snapshot tests (Windows recreates them with CRLF endings on at least one of my machines)

MichaelChirico:
OK, switched to using `LC_COLLATE` for full robustness.

Per https://github.com/r-lib/withr/issues/179, it's also possible to see some interference with `LC_COLLATE` envvar, but not adding that complication yet since this is only needed for dev.

Bisaloo:
In case that's useful, `method = "radix"` will always use C locale for character strings. It can be a convenient way to get locale-independent sorting without directly tweaking the LOCALE settings

IndrajeetPatil:
@Bisaloo This is great! 

## With the default sorting method

``` r
callr::r(
  function() withr::with_collate("C", sort(c("a", "_a", "A"))),
  env = c(LC_COLLATE = "C")
)
#> [1] "A"  "_a" "a"

callr::r(
  function() withr::with_collate("en_US", sort(c("a", "_a", "A"))),
  env = c(LC_COLLATE = "en_US")
)
#> [1] "_a" "a"  "A"
```

## With the radix sorting method

```r
callr::r(
  function() withr::with_collate("C", sort(c("a", "_a", "A"), method = "radix")),
  env = c(LC_COLLATE = "C")
)
#> [1] "A"  "_a" "a"

callr::r(
  function() withr::with_collate("en_US", sort(c("a", "_a", "A"), method = "radix")),
  env = c(LC_COLLATE = "en_US")
)
#> [1] "A"  "_a" "a"
```

<sup>Created on 2023-12-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico:
> In case that's useful, `method = "radix"` will always use C locale for character strings. It can be a convenient way to get locale-independent sorting without directly tweaking the LOCALE settings

awesome idea! thank you {data.table} 😅

IndrajeetPatil (approved):

Issue #2317, opened 2023-11-19
unnecessary_nesting_linter() can catch else { if () } where the braces could be dropped


MichaelChirico:
```r
if (x == 'a') 1 else { if (x == 'b') 2 else 3 }
```

`else {` is not considered by the current logic.

AshesITR:
Maybe instead add a linter that lints `else { if () ... }` if there are no expressions after the `if` expr-block?
Might fit into `brace_linter()`"do not put `{` between `else` and `if` if the `if` branch is the only statement in the `else` block." or something more readable with the same intent...

MichaelChirico:
could also appropriate for unnecessary_nested_if_linter(), or unnecessary_nesting_linter()?

(we might want to merge those two both as unnecessary_nesting_linter?)

MichaelChirico:
@IndrajeetPatil WDYT about subsuming `unnecessary_nested_if_linter()` into `unnecessary_nesting_linter()`?

IndrajeetPatil:
That's a great idea!

PR #2458, opened 2023-12-17
Subsume unnecessary_nested_if_linter() into unnecessary_nesting_linter()

MichaelChirico:
As discussed in #2317


tests/testthat/test-unnecessary_nested_if_linter.R
```diff
@@ -174,7 +183,9 @@ test_that("unnecessary_nested_if_linter skips allowed usages", {
 
 test_that("unnecessary_nested_if_linter blocks disallowed usages", {
   lint_message <- rex::rex("Don't use nested `if` statements")
-  linter <- unnecessary_nested_if_linter()
+  expect_warning({
```
AshesITR, line 186: Which is better, to test for the warning multiple times, or just once and then use `suppressWarnings()`?

MichaelChirico: I followed the approach taken for extraction_operator_linter(). Other than that I'm indifferent

AshesITR: Consistency should be the way to go here.
I don't have a preference either.

tests/testthat/test-unnecessary_nesting_linter.R
```diff
@@ -345,3 +345,255 @@ test_that("lints vectorize", {
     unnecessary_nesting_linter()
   )
 })
+
+test_that("unnecessary_nesting_linter skips allowed usages", {
+  linter <- unnecessary_nesting_linter()
+
+  expect_lint(
+    trim_some("
```
AshesITR, line 353: These usages are inconsistent WRT indentation and content of the last line across the file.
Do you mind cleaning up so we have
```r
trim_some("
  two-spaces-indented
  code
  here
") # but not here
```
everywhere in the file?

AshesITR (approved):

PR #2459, opened 2023-12-17
Get basic devcontainer setup working

### LINKED ISSUE #2479: Handle native pipe '|>' and placeholder '_' in unnecessary_placeholder_linter()
It should be a simple mimicking of the current XPath:

https://github.com/r-lib/lintr/blob/72eeefd87596ea7fa72ee1a8b14bcb82d322c381/R/unnecessary_placeholder_linter.R#L37-L50

I just haven't been on a machine with recent enough R to confirm+test yet.

MichaelChirico:
1. Fix typo (devcontainers->devcontainer)
 2. Add git

Also closes #2479.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM rocker/r-base
+
+RUN apt-get -qq update \
+  && apt-get install -y --no-install-recommends \
+  git libxml2-dev
+
+RUN Rscript -e "
+  read.dcf('DESCRIPTION', 'Imports') |>
```
AshesITR, line 8: Maybe `remotes::install_deps(dependencies=TRUE)`?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+  strsplit(',[\\n ]*') |>
+  unlist() |>
+  gsub(pattern = '[\\n ]*[(].*', replacement = '') |>
+  setdiff(tools:::.get_standard_package_names()$base) |>
+  c('pkgload', 'testthat') |>
```
AshesITR, line 13: What about `patrick` if we stick to this approach?

.devcontainer/Dockerfile
MichaelChirico: It does require installing remotes, which is lightweight but has to be run separately (so `install.packages()` calculates the deps graph twice). Seems fine though.

`dependencies=TRUE` seems too much for dev, I think the default (just `Imports` for us) is better. If you need the others, `install.packages()` works inside the container.

We might also add `Config/needs/testing: patrick` to our DESCRIPTION, then we could combine the 2nd+3rd steps.

MichaelChirico: OK, don't think it'll get simpler than what we have currently. while still using remotes.

A simpler {base} version is this:

```r
read.dcf("DESCRIPTION", "imports") |>
  tools:::.split_dependencies() |>
  names() |>
  c("patrick") |>
  install.packages()
```

That looks pretty good to me and gets us back to one `install.packages()` call, WDYT?

MichaelChirico:
Ugh, docker is the worst. It looks like there's no way to reference the DESCRIPTION file without hard-copying (symlink is not possible, nor is `../DESCRIPTION`) to refer to it during the `RUN` phase of build. Probably the best bet is using the CRAN DESCRIPTION instead? Since that won't change so much. I'll do that for now and ask around about a better approach here.

MichaelChirico:
OK, finally hit on something reasonable by diving into the devcontainer docs. We can set `"context": ".."` so that the `Dockerfile` is run from the package directory, where `DESCRIPTION` is available. Looking good.

.devcontainer/Dockerfile
AshesITR: Seems a bit convoluted, especially using a `:::` function. Is a separate `install.packages()` call that problematic?
Also, is there a more generic way to specify minimal dev deps? Maybe via some `Config/...` key in DESCRIPTION?

MichaelChirico: FWIW, `.split_dependencies()` hasn't changed substantially in 15+ years:

image

I really wish they would export it! In fact I filed https://bugs.r-project.org/show_bug.cgi?id=18639.

> Is a separate install.packages() call that problematic?

Not catastrophic, just nice to avoid. In playing around on #2446 yesterday, I noticed the default Travis script does exactly that (install {remotes}, then use it to finish installing deps).

> Also, is there a more generic way to specify minimal dev deps? Maybe via some Config/... key in DESCRIPTION?

Yes, see first comment in this thread (crossed out now). Just wanted to avoid editing DESCRIPTION.

What would you like to change, (1) switch to {remotes} (2) use Config/needs/testing (3) both?

AshesITR: Both SGTM if {remotes} takes a flag. This should also lower the entry barrier for new contributors, and we could add a stub our README on how to get a dev environment setup properly.

MichaelChirico: > we could add a stub our README on how to get a dev environment setup properly.

That part, I'm less sure -- IIUC the devcontainers will all run against my billing account 😂 don't necessarily want everyone piling in :)

MichaelChirico: OK, PTAL.

AshesITR: > > we could add a stub our README on how to get a dev environment setup properly.
> 
> That part, I'm less sure -- IIUC the devcontainers will all run against my billing account 😂 don't necessarily want everyone piling in :)

I was thinking about how to install the dev deps, not how to use the container 🤣

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM rocker/r-base
+
+RUN apt-get -qq update \
+  && apt-get install -y --no-install-recommends \
+  git libxml2-dev
+
+COPY DESCRIPTION .
+
+# NB: patrick also pulls in testthat and pkgload
```
AshesITR, line 9: This comment is no longer relevant, right?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# hard-copied from DESCRIPTION
+RUN Rscript -e '                           \
+  install.packages("remotes");             \
```
AshesITR, line 12: Very nice and clean to read 👍🏻

.devcontainer/Dockerfile
MichaelChirico: right; axed

R/unnecessary_placeholder_linter.R
```diff
@@ -33,7 +33,7 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 unnecessary_placeholder_linter <- function() {
-  # TODO(michaelchirico): handle R4.2.0 native placeholder _ as well
+  # NB: Native placeholder '_' must be used with a named argument, so it's not relevant here.
```
AshesITR, line 36: Sporadic diff?

MichaelChirico: no, you merged #2495 into this PR 🙈

MichaelChirico: Sorry for the confusing choice of merge target, but I was using the codespace to get an R4.3.0+ version running, so it had to include all the changes here and it was easier to just diff against this branch than to remove all the changes.

I split off the change to #2499 now.

AshesITR (approved):

PR #2460, opened 2023-12-17
New meta-level test that roxygen2 has been run

MichaelChirico:
We've been bitten a number of times by some small change to the doc mark-up sneaking through to `main` without be `roxygenize()`'d, winding up included in some unrelated PR.

This simple workflow should prevent that issue from recurring.


MichaelChirico:
Confirmed WAI:

image

AshesITR:
Hmm this might be a problem for OS dependent Rd.
E.g. I can't generate the current man page for `default_undesirable_functions` without changing sort order.

MichaelChirico:
even after #2455?

AshesITR:
Should be fixed by that but a similar problem could occur in the future.
So if we put it in place, it should run at least on linux + windows + linux with non-standard locale to catch any potentially problematic markup ahead of time.

MichaelChirico:
> Should be fixed by that but a similar problem could occur in the future. So if we put it in place, it should run at least on linux + windows + linux with non-standard locale to catch any potentially problematic markup ahead of time.

IMO that will catch some rare (but annoying!) issues, but the current script will catch most issues. Not sure it's worth extra compute resources for the rest.

AshesITR:
How much resources does it cost to run?
Maybe run it twice in the same process with different locales as a compromise?

I'm willing to give it a try (it won't be blocking anyway, so might just be a minor annoyance if encountered).

Perhaps output a git diff of old vs. new for changed files to ease debugging?

MichaelChirico:
> How much resources does it cost to run? Maybe run it twice in the same process with different locales as a compromise?

different locales is a good idea, that'll have low overhead. setting up different containers/OS was my issue.

> Perhaps output a git diff of old vs. new for changed files to ease debugging?

I pondered this but figured it'd be pretty clear why it fails (oops, I forgot to document). let me add a one liner anyway.

AshesITR:
> > Perhaps output a git diff of old vs. new for changed files to ease debugging?
> 
> I pondered this but figured it'd be pretty clear why it fails (oops, I forgot to document). let me add a one liner anyway.

Usually, yes. But if and when `devtools::document()` won't fix the run, it will be very frustrating to debug.

.dev/roxygen_test.R
```diff
@@ -0,0 +1,54 @@
...
+    
+  for (file in new_files) {
+    old_file <- file.path(old_dir, file)
+    new_file <- file.path(new_dir, file)
+    if (rd_equal(old_file, new_file)) {
+      next
+    }
+    cat(sprintf("roxygenize() output differs from saved output for %s.\n", file))
+    cat("Here's the Rd2txt() output for what's saved:\n")
+    Rd2txt(old_file)
```
AshesITR, line 41: Will this show macros etc.?

.dev/roxygen_test.R
```diff
@@ -0,0 +1,54 @@
...
+    cat("Here's the Rd2txt() output for roxygenize():\n")
+    Rd2txt(new_file)
+    stop(call. = FALSE)
```
AshesITR, line 44: Output the value of `LOCALE` here?

.dev/roxygen_test.R
MichaelChirico: Per the help, the default is the `"render"` stage which expands macros:

https://rdrr.io/r/tools/Rd2HTML.html

AshesITR: IMO the raw contents of the Rd file are more helpful since those are the files that are checked in.

MichaelChirico: `tools::Rdiff()` will compare the raw contents and show the diff, here's the output when I changed `<-` to `=` in `assignment_linter.Rd`:

```
34c34
< code_lines <- "1 -> x\n2 ->> y"
---
> code_lines = "1 -> x\n2 ->> y"
```

.dev/roxygen_test.R
```diff
@@ -38,10 +38,8 @@ check_roxygenize_idempotent <- function(LOCALE) {
       next
     }
     cat(sprintf("roxygenize() output differs from saved output for %s.\n", file))
-    cat("Here's the Rd2txt() output for what's saved:\n")
-    Rd2txt(old_file)
-    cat("Here's the Rd2txt() output for roxygenize():\n")
-    Rd2txt(new_file)
+    cat("Here's the tools::Rdiff() comparison of the two files:\n")
+    tools::Rdiff(old_file, new_file)
```
AshesITR, line 42:
``` r
tmp1 <- withr::local_tempfile(lines = c("a", "b"))
tmp2 <- withr::local_tempfile(lines = c("a", "b", "c"))
tools::Rdiff(tmp1, tmp2)
#> files differ in number of lines
```

<sup>Created on 2023-12-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

That's not very helpful, either...

MichaelChirico: Ah, that's annoying, thanks for checking. Maybe we should use `testthat::expect_identical()` to handle giving a summarized diff instead?

AshesITR: What about
``` r
tmp1 <- withr::local_tempfile(lines = c("a", "b"))
tmp2 <- withr::local_tempfile(lines = c("a", "b", "c"))
system2("diff", c("--unified", normalizePath(tmp1), normalizePath(tmp2)))
```

<sup>Created on 2023-12-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

MichaelChirico: > ```r
> tmp1 <- withr::local_tempfile(lines = c("a", "b"))
> tmp2 <- withr::local_tempfile(lines = c("a", "b", "c"))
> tools::Rdiff(tmp1, tmp2)
> #> files differ in number of lines
> ```
> 
> Created on 2023-12-19 with [reprex v2.0.2](https://reprex.tidyverse.org)
> 
> That's not very helpful, either...

Hmm, this truncated the output:

```
files differ in number of lines:
2a3
> c
```

Note that `tools::Rdiff()` is calling `diff` under the hood, just with different options (`-bw` vs. your `--unified`):

https://github.com/r-devel/r-svn/blob/c786526c27271c4bb04b621d2bde5cdd93033bc0/src/library/tools/R/testing.R#L290-L293

Vs. testthat:

```
testthat::expect_identical(readLines(tmp1), readLines(tmp2))
> Error: readLines(tmp1) (`actual`) not identical to readLines(tmp2) (`expected`).

`actual`:   "a" "b"    
`expected`: "a" "b" "c"
```

(with color if we enable it)

AshesITR: Oh, weird. I don't get the output from Rdiff in either my IDE or reprex.
I prefer `--unified` because that's basically `git diff`, which we're very used to reading.

MichaelChirico: What about with `Rdiff(Log = TRUE)` maybe? (just trying to avoid writing our own `system2()` which feels fragile)

AshesITR: Let's try it by forcing a failure?

MichaelChirico: Indeed plain `Rdiff()` output is junk:

image

Trying with Log=TRUE now.

MichaelChirico: OK, `--unified` looks like the way to go. One last small bit is to try and make this part comprehensible:

image

AshesITR: I wouldn't worry too much about that, unless there is a very simple option for diff to fake this path.

MichaelChirico: more importantly we need to make clear which output comes from which source. handled already at current commit.

.dev/roxygen_test.R
```diff
@@ -0,0 +1,54 @@
...
+    new_file <- file.path(new_dir, file)
+    if (rd_equal(old_file, new_file)) {
+      next
+    }
+    cat(sprintf("roxygenize() output differs from saved output for %s.\n", file))
+    cat("Here's the tools::Rdiff() comparison of the two files:\n")
+    cat("  [---]: saved output in man/ directory\n")
+    cat("  [+++]: roxygenize() output of R/ sources\n")
+    system2("diff", c("--unified", old_file, new_file))
+    stop("Failed in LOCALE=", LOCALE, ".", call. = FALSE)
```
AshesITR, line 45: WDYT about going over all files even when a diff is encountered? That would show all changes in the event of an error, potentially accelerating bug squashing.

I don't think we need to check more locales if we find an error to prevent spam, though.

AshesITR (commented):
NB cli is available if we want prettier messages.

.dev/roxygen_test.R
MichaelChirico: I think overkill for now -- the main use case for this check IMO is "oops I forgot to document()". so failing at first file should be "good enough". Happy to revisit if we encounter a case where it would have helped to see everything at once.

AshesITR (approved):

PR #2462, opened 2023-12-18
Catch 'else { if' where 'else if' is preferable in unnecessary_nesting_linter()

### LINKED ISSUE #2317: unnecessary_nesting_linter() can catch else { if () } where the braces could be dropped

MichaelChirico:
Closes #2317


AshesITR:
Sorry, too fast for GH to change target again...

R/shared_constants.R
```diff
@@ -34,6 +34,7 @@ rx_static_token <- local({
 rx_unescaped_regex <- paste0("(?s)", rex(start, zero_or_more(rx_non_active_char), end))
 rx_static_regex <- paste0("(?s)", rex(start, zero_or_more(rx_static_token), end))
 rx_first_static_token <- paste0("(?s)", rex(start, zero_or_more(rx_non_active_char), rx_static_escape))
+rx_escapable_tokens <- "^${}().*+?|[]\\<>=:;/_-!@#%&,~"
```
AshesITR, line 37: How did this diff get in here?

AshesITR (approved):
LGTM. The diff in `R/shared_constants.R` seems unrelated but sufficiently minor.

R/shared_constants.R
MichaelChirico: This is newly linted:

https://github.com/r-lib/lintr/blob/72eeefd87596ea7fa72ee1a8b14bcb82d322c381/R/shared_constants.R#L105-L111

I was originally going to move the comment to the end of the line, so pulled out this constant to keep things within 120. In the end opted to move the comment as shown.

Kept this change because I still think it's better -- abstracts away the logic for readability.

PR #2499, opened 2023-12-21
Add note about '_' to unnecessary_placeholder_linter()

### LINKED ISSUE #2479: Handle native pipe '|>' and placeholder '_' in unnecessary_placeholder_linter()

MichaelChirico:
Closes #2479 (for real this time 😄)


MichaelChirico:
I'm not reproducing the link rot check (which looks garbled/unusable anyway?) using `tools:::check_url_db()` from current r-devel. Hopefully it will just go away.

MichaelChirico:
Neither with current dev version of {urlchecker}:

```
urlchecker::url_check("lintr")
✔ All URLs are correct!
```

MichaelChirico:
Honestly I would suggest dropping the check-link-rot GHA. It seems to be more often a source of false positives than anything.

I would replace it with a release bullet point -- we only need to check URL rot once per cycle IMO. WDYT @IndrajeetPatil

IndrajeetPatil (approved):

Issue #2466, opened 2023-12-19
Vectorize the loop for equal-assignment content in get_source_expressions()


MichaelChirico:
Vectorize this loop away. The tricky part is, this loop doesn't execute on most R versions (we tried 3.6.3 and 4.2.0), so it likely requires some GHA print debugging or otherwise installing a few different versions of R -- tedious :)

OTOH, if it only runs on certain old versions of R, maybe it's best to just wait it out and drop the loop once our R dependency is "new" enough.

https://github.com/r-lib/lintr/blob/72eeefd87596ea7fa72ee1a8b14bcb82d322c381/R/get_source_expressions.R#L647-L654

MichaelChirico:
See #2501. This loop is not run on any version of R that we support, so I'll simply delete that code.

PR #2501, opened 2023-12-21
Check which versions of R touch the while() loop in get_source_expressions()

MichaelChirico:
For #2466


MichaelChirico:
Closing, this has served its purpose

PR #2504, opened 2023-12-21
Skip one more test on a specific R version on Windows

MichaelChirico:
Also found incidentally by #2501.

Also encapsulate the `skip()` logic by dropping cases, rather than formally `skip()`ing in a loop (the test code was getting pretty distracting).


tests/testthat/test-fixed_regex_linter.R
```diff
@@ -300,51 +300,59 @@ robust_non_printable_unicode <- function() {
 }
 
 # styler: off
-patrick::with_parameters_test_that("fixed replacements are correct", {
-  skip_if(
```
AshesITR, line 304: Maybe `skip_if(regex_expr %in% skip_cases, ...)` to still keep the skips visible while also making the code easier to read?

MichaelChirico: see https://github.com/google/patrick/issues/19, I wanted to avoid running skip_if() every iteration. I guess we can restore that approach in the meantime for readability.

AshesITR: Yes. With the conditional factored out, this will be more readable than subsetting I think.

AshesITR (approved):

PR #2506, opened 2023-12-22
Catch unreadable files early with warning

### LINKED ISSUE #1939: lint_dir() should be cognizant of read permissions
`lint_dir()` currently just looks at `dir(recursive = TRUE)` and attempts to run `lint()` on all the included files.

This is usually fine for packages, but less so for maintaining an analysis codebase where ownership permissions may be mixed.

The current behavior is clunky because linting can fail mid-way & lose progress. It would be better to skip (possibly warning) files where the correct read permissions are not available:

```r
lint_dir()
# ...........................Error in file(con, "r") : cannot open the connection
# In addition: Warning message:
# In file(con, "r") :
#   cannot open file '...': Permission denied
```

MichaelChirico:
Closes #1939

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,6 +47,7 @@
 * `todo_comment_linter()` has a new argument `except_regex` for setting _valid_ TODO comments, e.g. for forcing TODO comments to be linked to GitHub issues like `TODO(#154)` (#2047, @MichaelChirico).
 * `vector_logic_linter()` is extended to recognize incorrect usage of scalar operators `&&` and `||` inside subsetting expressions like `dplyr::filter(x, A && B)` (#2166, @MichaelChirico).
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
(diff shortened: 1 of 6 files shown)


AshesITR:
Why should we just warn instead of stopping?

MichaelChirico:
> Why should we just warn instead of stopping?

what I had in mind is a directory of scripts where one doesn't have the right permissions. Better to just skip the one file than stop midway.

AshesITR:
Seems to me that access rights should then be checked by lint_dir instead of lint.

MichaelChirico:
> Seems to me that access rights should then be checked by lint_dir instead of lint.

maybe, but after #2135 it will wind up back in lint() I reckon.

Shall we table this PR for now then?

AshesITR:
Not sure that will happen, actually.
But sounds okay if there is no immediate pain.

Even with vectorized lint I feel access checking doesn't belong there. `lint_dir(path, skip_inaccessible = TRUE)`, which should be non-default imo, could do `files <- files[vapply(files, \(x) file.access(x, 4L) != -1, logical(1L))]` if requested by the caller.

PR #2510, opened 2024-01-13
Bump roxygen2 version

IndrajeetPatil:
Re-documenting with the new version of `{roxygen2}` (7.3.0) seems to generate a non-parsable `NAMESPACE` file.

I am wondering if this is a bug upstream.

### DIFF
--- a/.github/workflows/repo-meta-tests.yaml
+++ b/.github/workflows/repo-meta-tests.yaml
@@ -24,7 +24,7 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: |
-            r-lib/roxygen2
+            any::roxygen2
 
       - name: Ensure lint metadata is tested
         run: |
--- a/.github/workflows/repo-meta-tests.yaml
+++ b/.github/workflows/repo-meta-tests.yaml
@@ -24,7 +24,7 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: |
-            any::roxygen2
+            r-lib/roxygen2
 
       - name: Ensure lint metadata is tested
         run: |
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,7 @@ Config/Needs/development: pkgload, cli, testthat, patrick
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.0.9000
+RoxygenNote: 7.3.1
 Collate:
     'make_linter_from_xpath.R'
     'xp_utils.R'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,7 @@ Config/Needs/development: pkgload, cli, testthat, patrick
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.0
 Collate:
     'make_linter_from_xpath.R'
     'xp_utils.R'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,7 @@ Config/Needs/development: pkgload, cli, testthat, patrick
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.0
+RoxygenNote: 7.3.0.9000
 Collate:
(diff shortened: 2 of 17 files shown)


MichaelChirico:
cc @hadley is this an {roxygen2} regression? Here's the markup leading to this garbled output:

https://github.com/r-lib/lintr/blob/bad1632f23b77f41cec13fb6f1c883302f64f506/R/lintr-package.R#L20-L24

MichaelChirico:
`git bisect` points to https://github.com/r-lib/roxygen2/commit/1ea43c9237e9d10bdf70ae023956396862f87d3c as the culprit

AshesITR:
@MichaelChirico thanks for the pointer. I looked at the diff and found the problem:

`update_namespace_imports()` sorts the lines, which will break any multi-line `@rawNamespace` directive.

MichaelChirico:
> update_namespace_imports

Thanks for the pointer; filed something that works in https://github.com/r-lib/roxygen2/pull/1573.

IndrajeetPatil:
Thank you, both, for getting to the bottom of this so quickly and for filing a fix!

IndrajeetPatil:
Can confirm that devel-roxygen2 fixes the issue. 

The new warning in the examples looks legit, though, and something we should deal with.

MichaelChirico:
> The new warning in the examples looks legit, though, and something we should deal with.

Can you point to the warning? I'm seeing various issues & can't figure out what's the problem, exactly. Maybe we're seeing something different, though.

IndrajeetPatil:
I see:

```r
> ### Name: get_r_string
> ### Title: Extract text from 'STR_CONST' nodes
> ### Aliases: get_r_string
> 
> ### ** Examples
> 
> ## Don't show: 
> if (requireNamespace("withr", quietly = TRUE)) (if (getRversion() >= "3.4") withAutoprint else force)({ # examplesIf
+ ## End(Don't show)
+ tmp <- withr::local_tempfile(lines = "c('a', 'b')")
+ expr_as_xml <- get_source_expressions(tmp)$expressions[[1L]]$xml_parsed_content
+ writeLines(as.character(expr_as_xml))
+ get_r_string(expr_as_xml, "expr[2]") # "a"
+ get_r_string(expr_as_xml, "expr[3]") # "b"
+ 
+ # more importantly, extract strings under R>=4 raw strings
+ ## Don't show: 
+ }) # examplesIf
> tmp <- withr::local_tempfile(lines = "c('a', 'b')")
> expr_as_xml <- get_source_expressions(tmp)$expressions[[1L]]$xml_parsed_content
Error in file(con, "r") : 
  (converted from warning) cannot open file '/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T//Rtmpk8Dcx3/file640749057d03': No such file or directory
Run `rlang::last_trace()` to see where the error occurred.
ℹ Loading lintr
> rlang::last_trace()
<error/rlang_error>
Error:
! No such file or directory
---
Backtrace:
     ▆
  1. └─devtools::run_examples(run_donttest = TRUE, run_dontrun = TRUE)
  2.   └─base::lapply(...)
  3.     └─pkgload (local) FUN(X[[i]], ...)
  4.       └─base::source(tmp, echo = !quiet, local = env, max.deparse.length = Inf)
  5.         ├─base::withVisible(eval(ei, envir))
  6.         └─base::eval(ei, envir)
  7.           └─base::eval(ei, envir)
  8.             ├─(if (getRversion() >= "3.4") withAutoprint else force)(...) at /Rtmpk8Dcx3/file6407739f7efc.R:8:1
  9.             │ └─base::source(...)
 10.             │   ├─base::withVisible(eval(ei, envir))
 11.             │   └─base::eval(ei, envir)
 12.             │     └─base::eval(ei, envir)
 13.             └─lintr::get_source_expressions(tmp)
 14.               └─lintr:::read_lines(filename) at lintr/R/get_source_expressions.R:69:5
 15.                 ├─base::withCallingHandlers(...) at lintr/R/utils.R:182:3
 16.                 └─base::readLines(file, warn = TRUE, ...)
 17.                   └─base::file(con, "r")
```

MichaelChirico:
Hmm I'm not seeing that. Maybe transient. Closing + reopening to see if 7.3.1 gets picked up by CI yet.

MichaelChirico:
#2518 proves (IINM) the issue is not related to {roxygen2}, since running with {pkgload} skips the `document=TRUE` step {devtools} does.

The new error might then be related to the {pkgload} update. Anyway, merging this.

MichaelChirico (approved):

IndrajeetPatil:
No longer relevant after #2519 merge

PR #2514, opened 2024-01-22
Skip basename() call when finding configs

### LINKED ISSUE #2512: settings not WAI in package sub-directory?
In https://github.com/Rdatatable/data.table/pull/5908/commits/5e36615783f846964418085d8dcf55f16e7c90f2, I tried to put the `.lintr` config into a non-standard location and point to it in CI with the `env` approach mentioned in `?read_settings`: `R_LINTR_LINTER_FILE: .ci/.lintr`. Also tried with `.ci/.lintr.R` (the exact file, not relying on auto-guessing the R extension).

However, this failed to get picked up, as seen by the 1000s of lints picked up in the associated job: https://github.com/Rdatatable/data.table/actions/runs/7523183440/job/20476211212?pr=5908

Am I doing something wrong?

I've also tried running `options(lintr.linter_file = ".ci/.lintr")` (and `.ci/.lintr.R`) to no avail. I've also print-debugged that `file.exists(".ci/.lintr.R")`, so the file is not in the wrong location / omitted for being in a hidden directory.

Currently, I'm needing to run `lintr:::read_settings(".ci/lintr"); lint_package(parse_settings = FALSE)` to accomplish what I intended. Really not pretty.

PS it might help for `verbose=TRUE` to report some info in `read_settings()` as well, e.g. which file was chosen, maybe even which settings are read as which values.

MichaelChirico:
Closes #2512. Didn't break any existing tests, so assume it's extraneous as posited.

BTW, I do find it a bit odd that `R_LINTR_LINTER_FILE` can _only_ be set at the R session level (that env is only read in `.onLoad()`):

https://github.com/r-lib/lintr/blob/bad1632f23b77f41cec13fb6f1c883302f64f506/R/zzz.R#L294-L296

Is that WAI? I am not sure the documentation reflects this well, if so:

https://github.com/r-lib/lintr/blob/bad1632f23b77f41cec13fb6f1c883302f64f506/R/settings.R#L10-L11

Therefore to test that `R_LINTR_LINTER_FILE` is working under this PR we'd need to invoke a subprocess, that felt like overkill.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,7 @@
 
 * `object_name_linter()` no longer errors when user-supplied `regexes=` have capture groups (#2188, @MichaelChirico).
 * `.lintr` config validation correctly accepts regular expressions which only compile under `perl = TRUE` (#2375, @MichaelChirico). These have always been valid (since `rex::re_matches()`, which powers the lint exclusion logic, also uses this setting), but the new up-front validation in v3.1.1 incorrectly used `perl = FALSE`.
+* `.lintr` configs set by option `lintr.linter_file` or environment variable `R_LINTR_LINTER_FILE` can point to specific package directories (#2512, @MichaelChirico).
 
 ## Changes to default linters
 
--- a/R/settings_utils.R
+++ b/R/settings_utils.R
@@ -75,7 +75,7 @@ find_config <- function(filename) {
   # may exist in subsequent directories are ignored
   file_locations <- c(
     # Local (incl. parent) directories
-    find_local_config(path, basename(linter_file)),
+    find_local_config(path, linter_file),
     # User directory
     # cf: rstudio@bc9b6a5 SessionRSConnect.R#L32
     file.path(Sys.getenv("HOME", unset = "~"), linter_file),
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -19,6 +19,7 @@
 
 * `object_name_linter()` no longer errors when user-supplied `regexes=` have capture groups (#2188, @MichaelChirico).
 * `.lintr` config validation correctly accepts regular expressions which only compile under `perl = TRUE` (#2375, @MichaelChirico). These have always been valid (since `rex::re_matches()`, which powers the lint exclusion logic, also uses this setting), but the new up-front validation in v3.1.1 incorrectly used `perl = FALSE`.
+* `.lintr` configs set by option `lintr.linter_file` or environment variable `R_LINTR_LINTER_FILE` can point to specific package directories (#2512, @MichaelChirico).
```
AshesITR, line 22: ... can point to subdirectories?

NB the change also affects search from HOME, not only from working directories.

MichaelChirico: not 100% sure I captured your comment, PTAL

AshesITR (approved):

AshesITR (approved):

PR #2515, opened 2024-01-22
workaround for roxygen2 bug

MichaelChirico:
See #2510, https://github.com/r-lib/roxygen2/issues/1572.

This workaround fits the full `@rawNamespace` directive on one line, thus impervious to the `sort()` issue.

We can revert this to the more readable version either at release or when {roxygen2} releases again, but for now it's causing some of our CI tests to fail e.g.

https://github.com/r-lib/lintr/actions/runs/7607480388/job/20714878154?pr=2514


MichaelChirico:
{roxygen2} published to CRAN already, this PR obsolete.

PR #2518, opened 2024-01-25
Run examples with pkgload to set quiet=TRUE

MichaelChirico:
Current output has a ton of noise. This helps by suppressing a lot of output, though all of the output from `writeLines()`/`cat()` is still shown.

I'm still not sure why we're seeing the new error.


IndrajeetPatil (approved):
LGTM, thanks!

PR #2519, opened 2024-01-25
Remove local_tempfile() usage in examples following breakage by withr 3.0.0

MichaelChirico:
Per discussion here:

https://github.com/r-lib/withr/issues/256

This is an expected failure, and it seems `local_tempfile()` never really behaved as I'd've thought in examples anyway (namely, to clean up the file after each example runs).

There are ways to keep using {withr} if we really want, e.g. I think adding a line to set `withr.hook_source` option in the `check-all-examples` GHA would work, but now I think it's best to just stick with base functionality for these examples.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,7 @@ Config/Needs/development: pkgload, cli, testthat, patrick
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.1
 Collate:
     'make_linter_from_xpath.R'
     'xp_utils.R'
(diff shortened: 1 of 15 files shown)


MichaelChirico:
Hmm I think this is a {withr} issue with the 3.0.0 update. I am able to reproduce locally now.

MichaelChirico:
Per discussion here:

https://github.com/r-lib/withr/issues/256

This is an expected failure, and it seems `local_tempfile()` never really behaved as I'd've thought in examples anyway (namely, to clean up the file after each example runs).

There are ways to keep using {withr} if we really want, e.g. I think adding a line to set `withr.hook_source` option in the `check-all-examples` GHA would work, but now I think it's best to just stick with base functionality for these examples.

IndrajeetPatil (approved):
Thanks for digging into the source of this problem and making the fixes.

PR #2529, opened 2024-03-20
Make Michael the new maintainer

jimhester:
cc @MichaelChirico


MichaelChirico (dismissed):

MichaelChirico:
Used `grep -Fir "hester" . | grep -Fv "/revdep/" | grep -Fv "/.dev/"` to find more hits & made two more changes. I think we leave you as named party in LICENSE and COPYING as original author, right?

MichaelChirico (approved):

MichaelChirico:
Thanks again for your long stewardship + authorship @jimhester!!!

PR #2531, opened 2024-03-25
Include patch NEWS entry

MichaelChirico:
Moves the items from dev to release that are now on CRAN:

https://cran.r-project.org/web/packages/lintr/news/news.html


MichaelChirico:
NEWS-only & required to get the repo in a "proper" state after the release, so merging right away

PR #2534, opened 2024-03-26
Remove reference to specific functions in help

### LINKED ISSUE #2532: docs: return() is not undesirable
In https://lintr.r-lib.org/reference/undesirable_function_linter.html

> Report the use of undesirable functions (e.g. [base::return()](https://rdrr.io/r/base/function.html), [base::options()](https://rdrr.io/r/base/options.html), or [base::sapply()](https://rdrr.io/r/base/lapply.html)) and suggest an alternative.

But return() is not an undesirable function?

MichaelChirico:
Closes #2532. No sense keeping specific functions named here, which might change over time; better to just link `?default_undesirable_functions`, which is already done.

### DIFF
--- a/R/undesirable_function_linter.R
+++ b/R/undesirable_function_linter.R
@@ -1,7 +1,6 @@
 #' Undesirable function linter
 #'
-#' Report the use of undesirable functions (e.g. [base::return()], [base::options()], or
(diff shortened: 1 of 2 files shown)


IndrajeetPatil (approved):
Yes, I agree. The examples list all the functions on the website anyway.

PR #2538, opened 2024-03-30
Bump package version to devel

IndrajeetPatil:
Since CRAN version is `3.1.2`, the default branch's "devel" version is lagging CRAN version.


MichaelChirico (approved):
oh, nice catch, thanks

PR #2540, opened 2024-03-31
Get rid of outdated comments from GHA for R CMD check

IndrajeetPatil:
> `next` corresponds to the next version of R, either R-patched, or R-alpha, R-beta, R-rc, or R-prerelease

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -28,14 +28,12 @@ jobs:
         config:
           - {os: macOS-latest,   r: 'release'}
 
-          - {os: windows-latest, r: 'next'}
           - {os: windows-latest, r: 'release'}
           # Use 3.6 to trigger usage of RTools35
           - {os: windows-latest, r: '3.6'}
           # use 4.1 to check with rtools40's older compiler
           - {os: windows-latest, r: '4.1'}
 
-          - {os: ubuntu-latest,  r: 'next'}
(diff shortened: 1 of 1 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
           # use 4.1 to check with rtools40's older compiler
           - {os: windows-latest, r: '4.1'}
 
-          # Use older ubuntu to maximise backward compatibility
```
IndrajeetPatil, line 37: This is an outdated comment. We use `ubuntu-latest`.

.github/workflows/R-CMD-check.yaml
```diff
@@ -72,7 +73,6 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
-      # `{patrick}` imports `{purrr}`, and so needs to be ignored on R < 3.5
```
IndrajeetPatil, line 75: Not relevant anymore.

MichaelChirico:
Is there any particular reason we need to run on `next` version for every change?

We already have "run on every R version" as part of the release procedure, is that not sufficient?

(changes to obsolete comments look good anyway)

IndrajeetPatil:
> We already have "run on every R version" as part of the release procedure, is that not sufficient?

Yeah, that's fair. It does sound like an overkill to run it for every commit.

MichaelChirico (approved):

PR #2543, opened 2024-04-01
Run styler

IndrajeetPatil:


R/missing_argument_linter.R
```diff
@@ -37,7 +37,8 @@ missing_argument_linter <- function(except = c("alist", "quote", "switch"), allo
     "self::EQ_SUB[following-sibling::*[not(self::COMMENT)][1][self::OP-RIGHT-PAREN or self::OP-COMMA]]"
   )
   if (!allow_trailing) {
-    conds <- c(conds,
+    conds <- c(
```
MichaelChirico, line 40: is this a style guide violation? I like having both `conds` in the same line, I think it improves readability

R/settings.R
```diff
@@ -159,7 +159,8 @@ validate_config_file <- function(config, config_file, defaults) {
     )
   }
 
-  validate_regex(config,
+  validate_regex(
```
MichaelChirico, line 162: I'm more indifferent here but also didn't think this is a violation

R/missing_argument_linter.R
IndrajeetPatil: IMO, the point here is not whether the code violates the style guide or not, but rather that we don't fight against the tool (the formatter). 

If we start picking-and-choosing which changes to keep and which to discard, we lose the benefits of having a formatter: time saved and consistency achieved is offset by time lost in discussing some formatting changes and inconsistency introduced by personal preferences. 

Of course, we can always request changes upstream in the formatter if the team doesn't agree (e.g. we are already doing this for `switch()` statements).

MichaelChirico: Don't really agree here TBH, this change actively makes the code less readable to me. "Don't fight the formatter" can apply where it's purely stylistic, e.g. the change in R/settings.R.

Here {styler} is actively violating what's allowed by the style guide, so yes it needs to be filed with them:

[list.files(path, pattern = "\\.[RrSsQq]$")](https://style.tidyverse.org/syntax.html#long-lines)

IndrajeetPatil: Thanks, I have manually reverted these two changes.

MichaelChirico (approved):
thanks! hopefully there's an upstream update before the next time we run styler for a final decision

PR #1938, opened 2023-03-26
Check if tests are coupled

IndrajeetPatil:
cf. https://github.com/r-lib/lintr/issues/1937

PR #2555, opened 2024-04-13
Make `cyclocomp_linter()` optional

### LINKED ISSUE #2554: Make `cyclocomp_linter()` optional
As Michael mentioned in https://github.com/r-lib/lintr/issues/2539:

> there's not anything specific in the style guide about meeting certain complexity requirements & that linter is quite heavy, we could make it optional.

IndrajeetPatil:
Closes #2554

Creating a draft PR. Not sure in which release we would like to include this change.

TODO:
- [ ] update NEWS


R/cyclocomp_linter.R
```diff
@@ -23,6 +23,12 @@
 #' @export
 cyclocomp_linter <- function(complexity_limit = 15L) {
   Linter(linter_level = "expression", function(source_expression) {
+    # nocov start
+    if (!requireNamespace("cyclocomp", quietly = TRUE)) {
+      cli::cli_abort("'cyclocomp' package needs to be installed for this linter.")
```
AshesITR, line 28: Maybe make the message include the function name for clarity?

NEWS.md
```diff
@@ -15,6 +15,7 @@
 * `extraction_operator_linter()` is deprecated. Although switching from `$` to `[[` has some robustness benefits for package code, it can lead to non-idiomatic code in many contexts (e.g. R6 classes, Shiny applications, etc.) (#2409, @IndrajeetPatil). To enable the detection of the `$` operator for extraction through partial matching, use `options(warnPartialMatchDollar = TRUE)`.
 * `unnecessary_nested_if_linter()` is deprecated and subsumed into the new/more general `unnecessary_nesting_linter()`.
 * Drop support for posting GitHub comments from inside GitHub comment bot, Travis, Wercker, and Jenkins CI tools (spurred by #2148, @MichaelChirico). We rely on GitHub Actions for linting in CI, and don't see any active users relying on these alternatives. We welcome and encourage community contributions to get support for different CI system going again.
+* `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. New users will need to install the needed `{cyclocomp}` package and, if relevant, also install it in the CI/CD pipelines.
```
MichaelChirico, line 18:
```suggestion
* `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. Note that users with `cyclocomp_linter()` in their configs may now need to install {cyclocomp} intentionally, in particular in CI/CD pipelines.
```

MichaelChirico (dismissed):
LGTM, I'm assuming this is ready to merge?

IndrajeetPatil:
> LGTM, I'm assuming this is ready to merge?

Yes!

tests/testthat/test-cache.R
```diff
@@ -434,6 +434,8 @@ test_that("it works outside of a package", {
 })
 
 test_that("cache = TRUE workflow works", {
+  skip_if_not_installed("cyclocomp")
```
AshesITR, line 437: Why not drop it from the test package instead?

IndrajeetPatil: Sorry, don't follow you. Why would we want to do that?

MichaelChirico: Hmm, I am not seeing why this test should fail absent cyclocomp, can you please record how this test fails without cyclocomp installed?

IndrajeetPatil: Wow, this wasn't needed at all. 

I am not sure why I had added it; my hunch is it was a different issue that affected this test, and I mistook it to be related to cyclocomp linter removal and added this skip.

IndrajeetPatil: Hmm, the tests fail on CI:

```r
i Please install the needed cyclocomp package.
Backtrace:
     ▆
  1. ├─lintr::lint_package(pkg, cache = TRUE) at test-cache.R:445:3
  2. │ └─lintr::lint_dir(...) at lintr/R/lint.R:263:3
  3. │   ├─lintr:::flatten_lints(...) at lintr/R/lint.R:185:3
  4. │   │ └─lintr:::flatten_list(x, class = "lint") at lintr/R/utils.R:18:3
  5. │   │   └─lintr (local) assign_item(x) at lintr/R/utils.R:36:3
  6. │   └─base::lapply(...) at lintr/R/lint.R:185:3
  7. │     └─lintr (local) FUN(X[[i]], ...)
  8. │       └─lintr::lint(file, ..., parse_settings = FALSE, exclusions = exclusions) at lintr/R/lint.R:189:7
  9. │         ├─base::withCallingHandlers(...) at lintr/R/lint.R:88:9
 10. │         └─lintr:::get_lints(...) at lintr/R/lint.R:88:9
 11. │           ├─lintr:::flatten_lints(linter_fun(expr)) at lintr/R/lint.R:305:5
 12. │           │ └─lintr:::flatten_list(x, class = "lint") at lintr/R/utils.R:18:3
 13. │           │   └─lintr (local) assign_item(x) at lintr/R/utils.R:36:3
 14. │           └─lintr (local) linter_fun(expr) at lintr/R/lint.R:305:5
 15. │             └─cli::cli_abort(...) at lintr/R/cyclocomp_linter.R:28:7
 16. │               └─rlang::abort(...)
 17. │                 └─rlang:::signal_abort(cnd, .file)
 18. │                   └─base::signalCondition(cnd)
 19. └─lintr (local) `<fn>`(`<rlng_rrr>`)
```

MichaelChirico: Interesting... I see cyclocomp referenced in that package, but only in comments:

```shell
grep -Fr cyclocomp tests/testthat/dummy_packages/package
# tests/testthat/dummy_packages/package/inst/data-raw/default_linter_testcode.R:# cyclocomp
# tests/testthat/dummy_packages/package/data-raw/default_linter_testcode.R:# cyclocomp
# tests/testthat/dummy_packages/package/R/default_linter_testcode.R:# cyclocomp
```

MichaelChirico: Maybe there's a cache issue in CI? I confirm it passes locally:

```
remove.packages("cyclocomp")
# Removing package from ‘/home/michael/R/x86_64-pc-linux-gnu-library/4.4’
# (as ‘lib’ is unspecified)
pkgload::load_all()
# ℹ Loading lintr
testthat::test_file('tests/testthat/test-cache.R')
# ══ Testing test-cache.R ══════════════════════════════════════════════════════════════
# [ FAIL 0 | WARN 0 | SKIP 0 | PASS 62 ] Done!
```

MichaelChirico:
`R CMD check` throws a `NOTE`:

```
* checking Rd cross-references ... NOTE
Package unavailable to check Rd xrefs: ‘cyclocomp’
```

Should we change that link (it won't cause any CRAN issue since {cyclocomp} will be available)? I lean towards "not an issue".

tests/testthat/test-cache.R
MichaelChirico: Actually, I _do_ reproduce the error locally under `R CMD check`:

```
* checking tests ...
  Running ‘testthat.R’
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
    9. │         ├─base::withCallingHandlers(...)
   10. │         └─lintr:::get_lints(...)
   11. │           ├─lintr:::flatten_lints(linter_fun(expr))
   12. │           │ └─lintr:::flatten_list(x, class = "lint")
   13. │           │   └─lintr (local) assign_item(x)
   14. │           └─lintr (local) linter_fun(expr)
   15. │             └─cli::cli_abort(...)
   16. │               └─rlang::abort(...)
   17. │                 └─rlang:::signal_abort(cnd, .file)
   18. │                   └─base::signalCondition(cnd)
   19. └─lintr (local) `<fn>`(`<rlng_rrr>`)
  
  [ FAIL 1 | WARN 0 | SKIP 346 | PASS 5901 ]
  Error: Test failures
  Execution halted
```

tests/testthat/test-get_source_expressions.R
```diff
@@ -414,6 +414,8 @@ param_df$.test_name <- with(param_df, sprintf("%s on expression %d", linter, exp
 patrick::with_parameters_test_that(
   "linters pass with xml_missing() content",
   {
+    skip_if_not_installed("cyclocomp")
```
MichaelChirico, line 417: this actually induces 340 skips, we only need one though right?

MichaelChirico: How about

```
if (linter == "backport_linter") {
  ...
} else {
  if (linter == "cyclocomp_linter") {
    skip_if_not_installed("cyclocomp")
  }
  linter <- eval(call(linter))
}
```

MichaelChirico: Relevant again :smile: 

https://github.com/google/patrick/issues/19

IndrajeetPatil: Good call!

IndrajeetPatil:
The note in rcmdcheck should be fixed. 

But the failing test continues to befuddle me. Can have a look tomorrow.

MichaelChirico:
> The note in rcmdcheck should be fixed.

I'm really not sure -- wanted to flag for vis. I especially have in mind that base R itself throws the same "missing cross-reference" `NOTE`s a fair amount, e.g. in `?cor.test`:

https://github.com/r-devel/r-svn/blob/b5582bb6dc31e988de5c33ebb28932901c2206ae/src/library/stats/man/cor.test.Rd#L131-L140

We also link to Suggests packages elsewhere, e.g.:

https://github.com/r-lib/lintr/blob/79d105998d06d0d731f19466439dfa7f1cfe4968/man/expect_comparison_linter.Rd#L9-L15

Do you have a proposed workaround in mind? Maybe there's a good alternative I haven't hit on yet.

tests/testthat/test-cache.R
MichaelChirico: I don't follow _precisely_ the issue, but the error did ring a bell. I think the wrong config was getting picked up, but I'm not sure which config / not sure why cyclocomp was involved.

This test is better off specifying the right lintr config anyway, and doing so made the tests pass (locally at least), so I'm OK to proceed.

MichaelChirico:
Should we make this function simpler & remove the cyclocomp reference?

https://github.com/r-lib/lintr/blob/79d105998d06d0d731f19466439dfa7f1cfe4968/tests/testthat/default_linter_testcode.R#L18

IndrajeetPatil:
> We also link to Suggests packages elsewhere, e.g.:

This is because we are cheating a bit. We don't treat all suggested dependencies equally while running R CMD check in "no Suggests" mode:

https://github.com/r-lib/lintr/blob/79d105998d06d0d731f19466439dfa7f1cfe4968/.github/workflows/R-CMD-check-hard.yaml#L55-L60

If we were to also add `{cyclocomp}` here, we can get rid of that NOTE, but then our CI will never treat this dependency as truly a suggested dependency.

IndrajeetPatil:
> Should we make this function simpler & remove the cyclocomp reference?

Yes, we should remove those references. 

I don't have a good sense of how much work making that function simpler will be, but I can have a look in the evening, after work, unless you get to it before me.

tests/testthat/test-cache.R
IndrajeetPatil: Thanks for fixing it!

I am fairly certain that this is also the reason the tests fail [in random order](https://github.com/r-lib/lintr/pull/1938).

PR #2557, opened 2024-04-17
Return an empty list with `<lints>` class in addin

### LINKED ISSUE #2556: Return an empty list with `<lints>` class in addin
Instead of returning `return(list())`, we should be returning:

              A `list()` with class `"lints"` would be better (so that the print method is the correct one).

_Originally posted by @AshesITR in https://github.com/r-lib/lintr/pull/2418#discussion_r1554990377_

IndrajeetPatil:
Closes #2556

### DIFF
--- a/R/addins.R
+++ b/R/addins.R
@@ -5,7 +5,8 @@ addin_lint <- function() {
   }
   filename <- rstudioapi::getSourceEditorContext()
   if (filename$path == "") {
-    return("Current source has no path. Please save before continue")
+    warning("Current source has no path. Please save before continuing.", call. = FALSE)
+    return(flatten_lints(list()))
   }
 
   lint(filename$path)


R/addins.R
```diff
@@ -5,7 +5,8 @@ addin_lint <- function() {
   }
   filename <- rstudioapi::getSourceEditorContext()
   if (filename$path == "") {
-    return("Current source has no path. Please save before continue")
+    warning("Current source has no path. Please save before continuing.", call. = FALSE)
```
MichaelChirico, line 8: shall we wait for the {cli} PR first?

IndrajeetPatil: I think this issue is independent of that PR. We can just convert that warning to cli warning in the PR once this change is merged to main.

MichaelChirico (approved):

PR #2561, opened 2024-04-21
Change URL order in DESCRIPTION to be pkgdown-friendly

IndrajeetPatil:
- On main:

```r
downlit:::autolink_curly("{lintr}")
#> "<a href='https://github.com/r-lib/lintr'>lintr</a>"
```

- On current branch:

```r
downlit:::autolink_curly("{lintr}")
#> "<a href='https://lintr.r-lib.org'>lintr</a>"
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,7 +18,7 @@ Description: Checks adherence to a given style, syntax errors and possible
     'RStudio IDE', 'Emacs', 'Vim', 'Sublime Text', 'Atom' and 'Visual
     Studio Code'.
 License: MIT + file LICENSE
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Confirming the real workhorse here:

https://github.com/r-lib/downlit/blob/c0a8f645e21a03e258b7c1684901f84279cc706a/R/link.R#L338-L345
https://github.com/r-lib/downlit/blob/c0a8f645e21a03e258b7c1684901f84279cc706a/R/metadata.R#L98

I.e., it literally pulls the 1st entry from the URL key in DCF.

I was wondering if there might be a more specific key we could use instead, e.g. `Config/pkgdown/url` or something.

MichaelChirico (approved):

PR #2568, opened 2024-05-04
Suppress coercion warnings

### LINKED ISSUE #2566: literal_coercion_linter surfaces coercion warnings
Here we intentionally run `as.integer('a')` to pull out the coercion warning message to later match against:

https://github.com/Rdatatable/data.table/blob/8c34563c6390d9acb0b4067a9e1d260b57937c24/inst/tests/tests.Rraw#L186

That's fine and can be `#nolint`'d, but still it throws a warning:

```r
l <- lint(text = 'as.integer("a")', linters = literal_coercion_linter())
# Warning message:
# In eval(expr) : NAs introduced by coercion
```

The resulting lint message still looks right to me:

```
<text>:1:1: warning: [literal_coercion_linter] Use NA_integer_ instead of as.integer("a"), i.e., use literals directly where possible, instead of coercion.
as.integer("a")
^~~~~~~~~~~~~~~
```

So IMO we can solve this by just suppressing warnings in `eval()` here:

https://github.com/r-lib/lintr/blob/28166bf697312c749c68f9a857fdb394937d4ba7/R/literal_coercion_linter.R#L102

MichaelChirico:
Closes #2566

### DIFF
--- a/tests/testthat/test-literal_coercion_linter.R
+++ b/tests/testthat/test-literal_coercion_linter.R
@@ -1,5 +1,5 @@
 test_that("literal_coercion_linter skips allowed usages", {
-  linter <- line_length_linter()
+  linter <- literal_coercion_linter()
 
   # naive xpath includes the "_f0" here as a literal
   expect_lint('as.numeric(x$"_f0")', NULL, linter)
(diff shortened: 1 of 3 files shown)


tests/testthat/test-literal_coercion_linter.R
```diff
@@ -1,5 +1,5 @@
 test_that("literal_coercion_linter skips allowed usages", {
-  linter <- line_length_linter()
+  linter <- literal_coercion_linter()
```
IndrajeetPatil, line 2: Good catch! 🙈

IndrajeetPatil (approved):

PR #2574, opened 2024-05-10
Make `scalar_in_linter()` configurable

F-Noelle:
Make scalar_in_linter configurable to allow projects to define additional %in% style functions like %notin%.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@
 * `unnecessary_nested_if_linter()` is deprecated and subsumed into the new/more general `unnecessary_nesting_linter()`.
 * Drop support for posting GitHub comments from inside GitHub comment bot, Travis, Wercker, and Jenkins CI tools (spurred by #2148, @MichaelChirico). We rely on GitHub Actions for linting in CI, and don't see any active users relying on these alternatives. We welcome and encourage community contributions to get support for different CI system going again.
 * `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. Note that users with `cyclocomp_linter()` in their configs may now need to install {cyclocomp} intentionally, in particular in CI/CD pipelines.
+* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
 
 ## Bug fixes
 
@@ -54,7 +55,6 @@
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
-* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
 
 ### New linters
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,7 +54,7 @@
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
-* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. `%chin%` is no longer linted by default. (@F-Noelle)
+* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
 
 ### New linters
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,6 +54,7 @@
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
+* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. `%chin%` is no longer linted by default. (@F-Noelle)
 
 ### New linters
 
--- a/R/scalar_in_linter.R
+++ b/R/scalar_in_linter.R
@@ -6,7 +6,8 @@
 #' `scalar %in% vector` is OK, because the alternative (`any(vector == scalar)`)
 #'   is more circuitous & potentially less clear.
 #'
-#' @param in_operators Character vector of additional infix operators that behave like the `%in%` operator
+#' @param in_operators Character vector of additional infix operators that behave like the `%in%` operator,
+#'   e.g. `{data.table}`'s `%chin%` operator.
 #'
 #' @examples
 #' # will produce lints
--- a/R/scalar_in_linter.R
+++ b/R/scalar_in_linter.R
@@ -1,13 +1,12 @@
 #' Block usage like x %in% "a"
 #'
 #' `vector %in% set` is appropriate for matching a vector to a set, but if
-#'   that set has size 1, `==` is more appropriate. `%chin%` from `{data.table}`
-#'   is matched as well.
+#'   that set has size 1, `==` is more appropriate.
 #'
 #' `scalar %in% vector` is OK, because the alternative (`any(vector == scalar)`)
 #'   is more circuitous & potentially less clear.
 #'
-#' @param add_in_operators Character vector of additional functions that behave like the %in% operator
+#' @param in_operators Character vector of additional functions that behave like the `%in%` operator
 #'
 #' @examples
 #' # will produce lints
@@ -18,7 +17,7 @@
 #'
 #' lint(
 #'   text = "x %chin% 'a'",
-#'   linters = scalar_in_linter()
+#'   linters = scalar_in_linter(in_operators = "%chin%")
 #' )
 #'
 #' # okay
@@ -30,13 +29,11 @@
 #' @evalRd rd_tags("scalar_in_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-scalar_in_linter <- function(add_in_operators = "%chin%") {
+scalar_in_linter <- function(in_operators = NULL) {
   # TODO(#2085): Extend to include other cases where the RHS is clearly a scalar
   # NB: all of logical, integer, double, hex, complex are parsed as NUM_CONST
-  special <- paste(paste0("text() = '", c("%in%", add_in_operators), "'"), collapse = " or ")
-
-  xpath <- paste0("
-  //SPECIAL[", special, "]
+  xpath <- glue("
+  //SPECIAL[{xp_text_in_table(c('%in%', {in_operators}))}]
     /following-sibling::expr[NUM_CONST[not(starts-with(text(), 'NA'))] or STR_CONST]
     /parent::expr
   ")
@@ -46,13 +43,19 @@ scalar_in_linter <- function(add_in_operators = "%chin%") {
 
     bad_expr <- xml_find_all(xml, xpath)
     in_op <- xml_find_chr(bad_expr, "string(SPECIAL)")
-    lint_msg <-
-      paste0("Use == to match length-1 scalars, not ", in_op, ". Note that == preserves NA where ", in_op, " does not.")
+    msg <- ifelse(
+      in_op == "%in%",
+      "Use == to match length-1 scalars instead of %in%. Note that == preserves NA where %in% does not.",
+      glue(
+        "{in_op} behaves similar to %in%. ",
+        "Use operators like == to match length-1 scalars instead of operators like %in%."
+      )
+    )
 
     xml_nodes_to_lints(
       bad_expr,
       source_expression = source_expression,
-      lint_message = lint_msg,
+      lint_message = msg,
       type = "warning"
     )
   })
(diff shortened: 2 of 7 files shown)


R/scalar_in_linter.R
```diff
@@ -7,6 +7,8 @@
 #' `scalar %in% vector` is OK, because the alternative (`any(vector == scalar)`)
 #'   is more circuitous & potentially less clear.
 #'
+#' @param add_in_operators Character vector of additional functions that behave like the %in% operator
```
MichaelChirico, line 10: not sure about the name, how about just `in_operators`?

R/scalar_in_linter.R
```diff
@@ -28,14 +30,16 @@
 #' @evalRd rd_tags("scalar_in_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-scalar_in_linter <- function() {
+scalar_in_linter <- function(add_in_operators = "%chin%") {
   # TODO(#2085): Extend to include other cases where the RHS is clearly a scalar
   # NB: all of logical, integer, double, hex, complex are parsed as NUM_CONST
-  xpath <- "
-  //SPECIAL[text() = '%in%' or text() = '%chin%']
+  special <- paste(paste0("text() = '", c("%in%", add_in_operators), "'"), collapse = " or ")
```
MichaelChirico, line 36:
```suggestion
  special <- xp_text_in_table(c("%in%", add_in_operators))
```

R/scalar_in_linter.R
```diff
@@ -28,14 +30,16 @@
...
+
+  xpath <- paste0("
```
MichaelChirico, line 38: we are usually using `glue()` once an XPath is parameterized, e.g.

```r
XPath <- glue::glue("
  //SPECIAL[{ special }]
    ...
")
```

R/scalar_in_linter.R
MichaelChirico: Also, not a fan of `special` as a variable name here, how about `operators` instead?

MichaelChirico:
Thanks for the PR & great suggestion! Please add a NEWS item & be sure to cite yourself!

I'm also tempted to say the default should be `character()`, there's no particular reason to make `%chin%` special. WDYT @AshesITR / @IndrajeetPatil, since that would be a breaking change (require downstream to supply `%chin%` if they need it)

R/scalar_in_linter.R
F-Noelle: Shortend to `in_operators`

R/scalar_in_linter.R
F-Noelle: With `xp_text_in_table` - which I completely overlooked before - I just put everything together with glue in the xpath assignment.

R/scalar_in_linter.R
F-Noelle: And now I use `glue` as well

F-Noelle:
I also tried to make the lint message for `%in%` like operators a bit more generic and updated the tests accordingly.

NEWS.md
```diff
@@ -54,6 +54,7 @@
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
+* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. `%chin%` is no longer linted by default. (@F-Noelle)
```
IndrajeetPatil, line 57: Since the default argument is `NULL` for this parameter, the second sentence is no longer needed here.

tests/testthat/test-scalar_in_linter.R
```diff
@@ -15,16 +15,16 @@ test_that("scalar_in_linter skips allowed usages", {
 })
 
 test_that("scalar_in_linter blocks simple disallowed usages", {
-  linter <- scalar_in_linter()
-  lint_in_msg <- rex::rex("Use == to match length-1 scalars, not %in%.")
-  lint_chin_msg <- rex::rex("Use == to match length-1 scalars, not %chin%.")
+  linter <- scalar_in_linter(in_operators = "%chin%")
```
IndrajeetPatil, line 18: Can we add a test with a vector of operators passed as an argument? 
You can use some additional binary infix operators from `{data.table}`.

NEWS.md
F-Noelle: The linter previously did lint the `%chin%` in its unconfigurable state. As this is no longer the case due to the default changing to NULL this is a notable change. If there is a better way to phrase that I am all ears.

IndrajeetPatil: The larger point of the change you are making is allowing to configure this linter. `%chin%` was just _your_ use case, and the developers who are reading this NEWS item don't need to know that. For users of some other package, it would have been a different operator (say `%nin%`).

I personally don't use `{data.table}` much, so I wasn't even aware of this infix operator; as a result, this part of the NEWS item was confusing for me because I thought this was an operator in base that I didn't know about.

tests/testthat/test-scalar_in_linter.R
F-Noelle: Use a vector of in operators to the second test

NEWS.md
F-Noelle: So you might misunderstand something here. `%chin%` was never my use case. I know of `{data.tables}`, but never used it in any of my projects. I just through it would be easy to make this linter configurable to work with my use cases, hence the pull request.

Previously the linter in its unconfigurable form linted `%in%` and `%chin%`, nothing else. When I made the update I wanted to make it backwards compatible to its original use case, so I opted to make `%chin%` the default for the new parameter. 

@MichaelChirico then suggested to demote `%chin%` from its special status making this PR a breaking change, which I am completely on board with. But having a breaking change is NEWS worthy in my opinion, because people might have relied on the old behaviour aka `%chin%` being linted. 

As this is no longer the case I want to highlight that simple fact as part of the NEWS. Just because I don't rely on it doesn't mean that nobody else did. If there is a better way of phrasing this I am all ears.

IndrajeetPatil: I have nothing more to add here over my previous reply. 

I will let @MichaelChirico decide the best phrasing for this news item, if he thinks it merits being phrased differently.

tests/testthat/test-scalar_in_linter.R
```diff
@@ -15,16 +15,18 @@ test_that("scalar_in_linter skips allowed usages", {
...
-  lint_in_msg <- rex::rex("Use == to match length-1 scalars, not %in%.")
-  lint_chin_msg <- rex::rex("Use == to match length-1 scalars, not %chin%.")
+  linter <- scalar_in_linter(in_operators = c("%chin%", "%notin%"))
+  lint_in_msg <- rex::rex("Use == to match length-1 scalars instead of %in%.")
+  lint_chin_msg <- rex::rex("%chin% behaves similar to %in%.")
+  lint_notin_msg <- rex::rex("%notin% behaves similar to %in%.")
 
   expect_lint("x %in% 1", lint_in_msg, linter)
   expect_lint("x %chin% 'a'", lint_chin_msg, linter)
+  expect_lint("x %notin% 1", lint_notin_msg, linter)
```
AshesITR, line 25: Add a negative test case (no lints) here?

NEWS.md
AshesITR: If we do make a breaking change, the corresponding bullet should be in its own section to highlight the change more clearly.

F-Noelle:
> > Thanks for the PR & great suggestion! Please add a NEWS item & be sure to cite yourself!
> > I'm also tempted to say the default should be `character()`, there's no particular reason to make `%chin%` special. WDYT @AshesITR / @IndrajeetPatil, since that would be a breaking change (require downstream to supply `%chin%` if they need it)? I do see a handful of users:
> > https://github.com/search?q=%2Fscalar_in_linter%2F+-path%3ANEWS&type=code
> 
> I would like to see a rough idea of the impact, using our branch comparison script for example. Slightly leaning towards a non-breaking change because false negatives (missing lints for %chin%) are generally hard to make out so lintr users basically have to read the news to find out they need to change their config.

I mean we could make the "%chin%" the default for in_operators again and push the change of defaults to a later point. Essentially saying the current default is "%chin%" but it is planned to switch it to NULL in the future so please make this explicit in your defaults if you rely on this functionality.

NEWS.md
MichaelChirico: I agree with @F-Noelle about the NEWS item, it looks good to me. Maybe we can emphasize better that it's a {data.table} operator.

@IndrajeetPatil see here, I'm not quite sure your point:

https://github.com/r-lib/lintr/blob/fc2268fb4b8609ac16fdba0b112319008529e11c/R/scalar_in_linter.R#L35

 - On current `main`, `x %chin% 'a'` will lint
 - On this PR, `x %chin% 'a'` will only lint under `scalar_in_linter("%chin%")` or similar.

MichaelChirico, line 57:
```suggestion
* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
```

R/scalar_in_linter.R
```diff
@@ -1,12 +1,13 @@
...
 #'
 #' `vector %in% set` is appropriate for matching a vector to a set, but if
-#'   that set has size 1, `==` is more appropriate. `%chin%` from `{data.table}`
-#'   is matched as well.
+#'   that set has size 1, `==` is more appropriate.
 #'
 #' `scalar %in% vector` is OK, because the alternative (`any(vector == scalar)`)
 #'   is more circuitous & potentially less clear.
 #'
+#' @param in_operators Character vector of additional functions that behave like the `%in%` operator
```
MichaelChirico, line 9:
```suggestion
#' @param in_operators Character vector of additional infix operators that behave like the `%in%` operator
```

"function" might imply any `foo(.)` can be linted, but the XPath logic is very specific to `//SPECIAL`, so we should emphasize that.

tests/testthat/test-scalar_in_linter.R
```diff
@@ -3,28 +3,48 @@ test_that("scalar_in_linter skips allowed usages", {
...
   expect_lint("x %in% NA_character_", NULL, linter)
 })
 
 test_that("scalar_in_linter blocks simple disallowed usages", {
-  linter <- scalar_in_linter()
-  lint_in_msg <- rex::rex("Use == to match length-1 scalars, not %in%.")
-  lint_chin_msg <- rex::rex("Use == to match length-1 scalars, not %chin%.")
+  linter <- scalar_in_linter(in_operators = c("%chin%", "%notin%"))
+
+  lint_in_msg <- rex::rex("Use == to match length-1 scalars instead of %in%.")
```
MichaelChirico, line 19: I think no need to create these temp variables unless they'll be re-used, here it will be IMO more readable for the tested message to be written in-line with the test.

NEWS.md
F-Noelle: Suggestion added

R/scalar_in_linter.R
F-Noelle: Suggestion added

tests/testthat/test-scalar_in_linter.R
F-Noelle: Made the message more generic, so it can be repeated.

R/scalar_in_linter.R
```diff
@@ -28,27 +29,33 @@
...
 
     bad_expr <- xml_find_all(xml, xpath)
     in_op <- xml_find_chr(bad_expr, "string(SPECIAL)")
-    lint_msg <-
-      paste0("Use == to match length-1 scalars, not ", in_op, ". Note that == preserves NA where ", in_op, " does not.")
+    msg <- ifelse(
+      in_op == "%in%",
+      "Use == to match length-1 scalars instead of %in%. Note that == preserves NA where %in% does not.",
+      glue(
+        "{in_op} behaves similar to %in%. ",
```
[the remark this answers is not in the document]
F-Noelle: I think this is now a sensible `lint_msg` and I also renamed the variable.

NEWS.md
```diff
@@ -54,6 +54,7 @@
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
+* `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
```
MichaelChirico, line 57: sorry, i'm having trouble making edits myself since this is made in your `main` branch, I think you need to create a dedicated branch for me to make edits.

all this to say -- could you please move this item to the "breaking changes" section at the top :)

R/scalar_in_linter.R
```diff
@@ -1,12 +1,13 @@
...
 #'
 #' `vector %in% set` is appropriate for matching a vector to a set, but if
-#'   that set has size 1, `==` is more appropriate. `%chin%` from `{data.table}`
-#'   is matched as well.
+#'   that set has size 1, `==` is more appropriate.
 #'
 #' `scalar %in% vector` is OK, because the alternative (`any(vector == scalar)`)
 #'   is more circuitous & potentially less clear.
 #'
+#' @param in_operators Character vector of additional infix operators that behave like the `%in%` operator
```
MichaelChirico, line 9:
```suggestion
#' @param in_operators Character vector of additional infix operators that behave like the `%in%` operator,
#'   e.g. `{data.table}`'s `%chin%` operator.
```

(and the `roxygenize()` it 😄)

R/scalar_in_linter.R
```diff
@@ -28,22 +29,24 @@
...
   Linter(linter_level = "expression", function(source_expression) {
     xml <- source_expression$xml_parsed_content
 
     bad_expr <- xml_find_all(xml, xpath)
     in_op <- xml_find_chr(bad_expr, "string(SPECIAL)")
-    lint_msg <-
-      paste0("Use == to match length-1 scalars, not ", in_op, ". Note that == preserves NA where ", in_op, " does not.")
+    lint_msg <- glue(
+      "Use comparison operators (e.g. ==, !=, etc.) to match length-1 scalars instead of {in_op} ",
+      "as these operators preserve NA where {in_op} does not."
```
MichaelChirico, line 48: Sorry, I'm afraid we've over-corrected 😄

"as" implies that preserving NA _is the reason_ for the lint. But that's not true... actually this is a somewhat ambiguous thing. Some authors might prefer this behavior and thus disfavor this linter -- hence we're highlighting this aspect as a tradeoff users should be aware of.

So let's go with

> Use comparison operators... to match length-1 scalars instead of {in_op}. Note that comparison operators preserve NA where {in_op} does not.

NEWS.md
F-Noelle: moved

MichaelChirico (approved):
Looks good, thanks for your patience!

PR #2579, opened 2024-05-20
Document known false positive for as.POSIXct

### LINKED ISSUE #2567: False positive for inner_combine_linter() on timestamps with different formats
compare

```r
c(as.POSIXct("2015-01-01",tz="UTC"), as.POSIXct("2015-01-02 01:02:03",tz="UTC"))
# [1] "2015-01-01 00:00:00 UTC" "2015-01-02 01:02:03 UTC"
as.POSIXct(c("2015-01-01", "2015-01-02 01:02:03"), tz="UTC")
# [1] "2015-01-01 UTC" "2015-01-02 UTC"
```

Without `format=` specified, `as.POSIXct()` guesses the format based on the first element --> infers `%F` in the latter case, in error.

It's maybe not the best default for `as.POSIXct()`, but IIRC it's as documented. The user can still get the desired behavior by giving matched formats on the input:

```r
as.POSIXct(c("2015-01-01 00:00:00", "2015-01-02 01:02:03"), tz="UTC")
# [1] "2015-01-01 00:00:00 UTC" "2015-01-02 01:02:03 UTC"
```

But we also lint in dynamic cases:

```r
lintr::lint(text = 'c(as.POSIXct(x), as.POSIXct(y))', linters = lintr::inner_combine_linter())
# <text>:1:1: warning: [inner_combine_linter] Combine inputs to vectorized functions first to take full advantage of vectorization, e.g., as.POSIXct(c(x, y)) only runs the more expensive as.POSIXct() once as compared to c(as.POSIXct(x), as.POSIXct(y)).
# c(as.POSIXct(x), as.POSIXct(y))
# ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

What's the best to do here? Just document the potential issue? Skip lints if there are non-`CONST` entries for timestamp converters? Then if there _are_ `CONST` entries, try and run the comparison ourselves before throwing a lint?

MichaelChirico:
Closes #2567

Does this warrant a NEWS item?

### DIFF
--- a/R/inner_combine_linter.R
+++ b/R/inner_combine_linter.R
@@ -6,6 +6,13 @@
 #'   preferred so that the most expensive part of the operation ([as.Date()])
 #'   is applied only once.
 #'
+#' Note that [strptime()] has one idiosyncrasy to be aware of, namely that
+#'   auto-detected `format=` is set by the first matching input, which means
+#'   that a case like `c(as.POSIXct("2024-01-01"), as.POSIXct("2024-01-01 01:02:03"))`
+#'   gives different results to `as.POSIXct(c("2024-01-01", "2024-01-01 01:02:03"))`.
+#'   This false positive is rare; a workaround where possible is to use
+#'   consistent formatting, i.e., `"2024-01-01 00:00:00"` in the example.
+#'
 #' @examples
 #' # will produce lints
 #' lint(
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
> Does this warrant a NEWS item?

Should we also add a test? 

In case this behaviour changes in R itself, we could update the current note in the docs.

MichaelChirico:
> Should we also add a test?

That would be committing us to a false positive, so I wouldn't do that. I think it's unlikely the R behavior will change because of the serious back-compatibility implications.

IndrajeetPatil (approved):

Issue #2580, opened 2024-05-20
New function expect_no_lint() akin to expect_no_warning() etc


MichaelChirico:
Is there by the way an `expect_no_lint` function? I would have expected a wrapper for `expect_lint(content, NULL, ...)`.

_Originally posted by @F-Noelle in https://github.com/r-lib/lintr/pull/2574#discussion_r1600474890_

F-Noelle:
This issue can be considered closed. Especially since #2585 can be considered a follow up issue.

PR #2582, opened 2024-05-23
Add expect_no_lint

F-Noelle:
This pull request is meant to address issue #2580.

This is probably the easiest way to add such a function. I assume with time the tests should be updated to use one over the other.


AshesITR:
Thanks for this. I would suggest adding a NEWS bullet crediting yourself and citing the issue number, as well as updating our test suite to use the new function.

MichaelChirico:
> updating our test suite to use the new function.

I would hold this for follow-up actually, that will be several-hundred-lines change, better to keep this PR focused on the minimal change to get this added.

To that end we probably want a `expect_no_lint_linter()` 😄

F-Noelle:
Added NEWS item and yeah the test suit update will be massive.

I actually thought about an `expect_no_lint_linter()`. It kind of is obvious when you think about it, but it is also super silly.😄

AshesITR:
Follow up PR for the migration sounds fine.

R/expect_lint.R
```diff
@@ -1,6 +1,7 @@
 #' Lint expectation
 #'
-#' This is an expectation function to test that the lints produced by `lint` satisfy a number of checks.
+#' These are expectation function to test that the lints produced by `lint` satisfy a number of checks.
+#' `expect_no_lint` is equivalent to `expect_lint` with checks set to NULL.
```
AshesITR, line 4: I would phrase this in an active way (asserts that the specified linter generates no lints on the code).

R/expect_lint.R
```diff
@@ -23,6 +24,7 @@
 #' @examples
 #' # no expected lint
 #' expect_lint("a", NULL, trailing_blank_lines_linter())
```
AshesITR, line 26: Maybe remove this undesirable alternative?

F-Noelle: Makes sense, I will also remove it from the tests.

R/expect_lint.R
F-Noelle: Not entirely sure how to best phrase that. Here would my my attempt. 
```
#' These are expectation functions to test specified linters on sample code in the `testthat` framework.
#' * `expect_lint` asserts that specified lints are generated.
#' * `expect_no_lint` asserts that no lints are generated. 
#' `expect_no_lint` is equivalent to `expect_lint` with checks set to NULL, and used for better readability.
```
Better? worse? Something diffrent?

R/expect_lint.R
F-Noelle: And added this to the vignette.

R/expect_lint.R
AshesITR: Sounds good, but I would leave out the last sentence. This makes it easier to turn passing `NULL` into a warning / an error down the road when `expect_no_lint()` is established.

F-Noelle: Makes sense, give me a minute.

F-Noelle: And in ...

AshesITR (approved):
LGTM, nice!

NEWS.md
```diff
@@ -54,7 +54,8 @@
 * `vector_logic_linter()` is extended to recognize incorrect usage of scalar operators `&&` and `||` inside subsetting expressions like `dplyr::filter(x, A && B)` (#2166, @MichaelChirico).
 * `any_is_na_linter()` is extended to catch the unusual usage `NA %in% x` (#2113, @MichaelChirico).
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
-* `paste_linter()`  is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
+* `paste_linter()` is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
+* `expect_no_lint()` was added as new function to cover the typical use case of expecting no lint message (#2580, @F-Noelle).
```
MichaelChirico, line 58:
```suggestion
* `expect_no_lint()` was added as new function to cover the typical use case of expecting no lint message, akin to the recent {testthat} functions like `expect_no_warning()` (#2580, @F-Noelle).
```

R/expect_lint.R
```diff
@@ -42,7 +44,8 @@
 expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
   if (!requireNamespace("testthat", quietly = TRUE)) {
     stop( # nocov start
-      "'expect_lint' is designed to work within the 'testthat' testing framework, but 'testthat' is not installed.",
+      "'expect_lint' and 'expect_no_lint' are designed to work within the 'testthat' testing framework, ",
```
MichaelChirico, line 47: ideally this will just reference the actual call for clarity, but we can save that for later.

MichaelChirico: #2585

MichaelChirico (approved):
Thank you!

PR #2583, opened 2024-05-23
Use the latest major `{withr}` version

IndrajeetPatil:
To benefit from performance and other enhancements in this release: https://www.tidyverse.org/blog/2024/01/withr-3-0-0/

We have already fixed issues stemming from this version: https://github.com/r-lib/lintr/pull/2519

We didn't bump the version because it was quite new at the time, but we can do so now.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,7 +45,7 @@ Suggests:
     testthat (>= 3.2.1),
     tibble,
     tufte,
-    withr (>= 2.5.0)
+    withr (>= 3.0.0)
 Enhances:
     data.table
 VignetteBuilder: 


MichaelChirico:
Sorry, I don't see why we should increase the minimum version dependency unless we actually use something from that version?

If some user can't upgrade {withr}, why should we prevent them from using {lintr} unless there's something specific in the new version that we "need" for {lintr}?

I must be missing something...

IndrajeetPatil:
> If some user can't upgrade {withr}, why should we prevent them from using {lintr}

But it's not a runtime dependency of {lintr}. Why would it prevent them from using {lintr}?

MichaelChirico:
> But it's not a runtime dependency of {lintr}. Why would it prevent them from using {lintr}?

True, but all the more reason not to enforce a minimum version IMO.

If there's some actual code changes you want to make that use updated functionality, I am happy to proceed.

PR #2593, opened 2024-06-06
Use public interface to test `logical_env()`

IndrajeetPatil:
cf. https://github.com/r-lib/lintr/issues/1692

### DIFF
--- a/tests/testthat/test-ci.R
+++ b/tests/testthat/test-ci.R
@@ -24,16 +24,20 @@ test_that("GitHub Actions functionality works in a subdirectory", {
   )
 })
 
-test_that("GitHub Actions - error on lint works", {
-  withr::local_envvar(list(GITHUB_ACTIONS = "true", LINTR_ERROR_ON_LINT = "true"))
-  withr::local_options(lintr.rstudio_source_markers = FALSE)
-  tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")
+patrick::with_parameters_test_that(
+  "GitHub Actions - error on lint works",
+  {
+    withr::local_envvar(list(GITHUB_ACTIONS = "true", LINTR_ERROR_ON_LINT = env_var_value))
+    withr::local_options(lintr.rstudio_source_markers = FALSE)
+    tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")
 
-  l <- lint(tmp)
+    l <- lint(tmp)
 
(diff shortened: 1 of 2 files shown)


tests/testthat/test-ci.R
```diff
@@ -1,5 +1,4 @@
 test_that("GitHub Actions functionality works", {
-  # imitate being on GHA whether or not we are
```
IndrajeetPatil, line 2: Comment repeats code.

tests/testthat/test-ci.R
```diff
@@ -26,8 +24,21 @@ test_that("GitHub Actions functionality works in a subdirectory", {
...
+  {
+    withr::local_envvar(list(GITHUB_ACTIONS = "true", LINTR_ERROR_ON_LINT = env_var_value))
+    withr::local_options(lintr.rstudio_source_markers = FALSE)
+    tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")
+
+    l <- lint(tmp)
+
+    expect_output(print(l), "::warning file", fixed = TRUE)
+  },
+  env_var_value = c("", "F", NA),
```
IndrajeetPatil, line 38: We were also testing `NULL`; I am not sure how to incorporate that in this parametrized test suite.

AshesITR: We are no longer testing `"true"` in this branch, are we?

AshesITR: Is NULL the equivalent of Sys.unsetenv()?
If so, there is an observable difference in behavior to `""`, so I wouldn't remove the test case.

IndrajeetPatil: Sounds good. Added a test case for `NULL` as well.

tests/testthat/test-ci.R
```diff
@@ -26,14 +24,31 @@ test_that("GitHub Actions functionality works in a subdirectory", {
...
-  l <- lint(tmp)
+    l <- lint(tmp)
 
-  local_mocked_bindings(quit = function(...) cat("Tried to quit.\n"))
-  expect_output(print(l), "::warning file", fixed = TRUE)
-})
+    local_mocked_bindings(quit = function(...) cat("Tried to quit.\n"))
+    expect_output(print(l), "::warning file", fixed = TRUE)
+  },
+  env_var_value = list("T", "true")
```
IndrajeetPatil, line 39: Also checking the variants of true for a good measure.

PR #2601, opened 2024-06-11
Remove `R 3.6` checks from CI

IndrajeetPatil:
Since neither the vignette builder (knitr) nor the testing framework (testthat) is available for this R version: they (via dependency `{evaluate}`) need `>= R 4.0`.


MichaelChirico (approved):
ok, might as well bump ourselves to 4.0 at this point I guess

PR #2602, opened 2024-06-11
Use cli error chaining + add clickable link to bad config

### LINKED ISSUE #2588: Use bulleted list for `{cli}`-formatted messages involving `conditionMessage()`
For more context, see discussion in the PR: https://github.com/r-lib/lintr/pull/2418#discussion_r1616367865

olivroy:
fix #2588, follows-up to #2418 

Thanks @IndrajeetPatil for merging #2418. I think it is a good improvement.

Taking advantage of this to send this up.

Basically uses `parent` from cli to format error messages.

Add a link to config file.

Tweak some wordings that didn't seem clear to me and use https://style.tidyverse.org/error-messages.html as a reference.

Screenshots of how this looks: (I had more but I erased my console output) (config is a clickable link to .lintr)
image

Changed wording of this one, but made the note about the warning that will disappear a little less prominent
image

Removed some `{.cls}`. Imo, it decreases readability, by adding `<` `>`, I think it reads better as a sentence.


R/lint.R
```diff
@@ -410,16 +410,17 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
   }
   max_col <- max(nchar(line) + 1L, 1L, na.rm = TRUE)
   if (!is_number(column_number) || column_number < 0L || column_number > max_col) {
-    cli_abort(c(
-      i = "{.arg column_number} must be an integer between {.val {0}} and {.code nchar(line) + 1} ({.val {max_col}})",
-      x = "Instead, it was {.val {column_number}}."
-    ))
+    cli_abort(
+      "{.arg column_number} must be an integer between \\
```
MichaelChirico, line 414: please don't use the end-of-line `\\`, I find the behavior pretty confusing & it's not a well-documented part of the R parser.

MichaelChirico: Note that this repo uses line width=120, so you also  don't need to squeeze your code so tightly :)

R/settings.R
```diff
@@ -94,54 +94,59 @@ read_config_file <- function(config_file) {
...
+        "Malformed {config_link} file, ensure it is valid R syntax.",
+        parent = e
+      )
     }
   } else {
     load_config <- function(file) {
       dcf_values <- read.dcf(file, all = TRUE)
       for (setting in names(dcf_values)) {
-        parsed_setting <- tryCatch(
+        parsed_setting <- withCallingHandlers(
```
MichaelChirico, line 115: why `withCallingHandlers()` over `tryCatch()`?

R/settings.R
```diff
@@ -94,54 +94,59 @@ read_config_file <- function(config_file) {
...
   config <- new.env()
   if (endsWith(config_file, ".R")) {
     load_config <- function(file) sys.source(file, config, keep.source = FALSE, keep.parse.data = FALSE)
     malformed <- function(e) {
-      cli_abort(c(
-        "Malformed config file, ensure it is valid R syntax.",
-        conditionMessage(e)
-      ))
+      cli_abort(
+        "Malformed {config_link} file, ensure it is valid R syntax.",
```
MichaelChirico, line 107: This will look like `Malformed <a>config</a> file`, right? WDYT about doing `Malformed config file <a>basename(file)</a>` instead? (ditto elsewhere)?

R/with.R
```diff
@@ -31,8 +31,12 @@
 #' names(my_undesirable_functions)
 #' @export
 modify_defaults <- function(defaults, ...) {
-  if (missing(defaults) || !is.list(defaults) || !all(nzchar(names2(defaults)))) {
-    cli_abort("{.arg defaults} must be a named list.")
+  if (missing(defaults)) {
+    cli_abort("{.arg defaults} must be a named list, not absent.")
```
MichaelChirico, line 35:
```suggestion
    cli_abort("Missing required argument {.arg defaults} should be a named list.")
```

R/with.R
```diff
@@ -177,10 +183,14 @@ all_linters <- function(..., packages = "lintr") {
...
+      "This warning will be removed when {.fun with_defaults} is fully deprecated."
     cli_warn(c(
-      "Did you mean {.arg defaults}?",
-      x = "{.arg default} is not an argument to {.fun linters_with_defaults}.",
-      i = "This warning will be removed when {.fun with_defaults} is fully deprecated."
+      x = "{.arg default} is not an argument to \\
+           {.help [{.fn linters_with_defaults}](lintr::linters_with_defaults)}.",
+      i = "Did you mean {.arg defaults}?",
+      # make it more subtle
+      cli::col_silver(deprecation_advice)
```
MichaelChirico, line 193:
```suggestion
      cli::col_silver("This warning will be removed when {.fun with_defaults} is fully dep)
```

MichaelChirico:
Thanks for the PR! I'll let @IndrajeetPatil add final review.

R/settings.R
olivroy: Prodcues better stack trace? I don't mind reverting

R/with.R
olivroy: I created it for line width. you reassured me with the 120 chr limit

R/lint.R
IndrajeetPatil: @MichaelChirico Out of curiosity, what's the kosher way of doing readable, multiline strings? That is, one that doesn't use `\\`? 

Because this is what I do as well 😅

R/settings.R
```diff
@@ -94,54 +94,59 @@ read_config_file <- function(config_file) {
...
         setting_value <- withCallingHandlers(
           tryCatch(
             eval(parsed_setting),
             error = function(e) {
-              cli_abort(c(
-                "Error from config setting {.code {setting}} in {.code {format(conditionCall(e))}}:",
-                conditionMessage(e)
-              ))
+              cli_abort(
+                "Error from {config_link} setting {.code {setting}} in {.code {format(conditionCall(e))}}:",
```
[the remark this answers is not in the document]
olivroy: I had not looked closely enough at this one.

Here is what it looks like with my changes
image

I think it could still read a bit better
image
(Fr missing argument with no default

R/settings.R
olivroy: I proposed something. Only show the link in the `malformed error msg to avoid repetition.

R/lint.R
olivroy: I ended up not putting `\\` but still keeping line break. cli doesn't care about line breaks (or double spaces) like glue does. With cli, you have to put `f` to force a line break.

Lmk if you want another readjustment!

MichaelChirico: > Out of curiosity, what's the kosher way of doing readable, multiline strings?

`paste()`, `paste0()`

R/settings.R
MichaelChirico: For non-error signals the difference is clear, agree it's more ambiguous the difference between `tryCatch(error=...)` and `withCallingHandlers(error=...)`.

Since the stack is likely to be helpful here, I agree to use `withCallingHandlers()`.

R/settings.R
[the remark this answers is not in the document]
MichaelChirico: Agree it's nicer to surface 'lint_dir()' as the error source up front.

R/lint.R
```diff
@@ -410,16 +410,16 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
...
   max_col <- max(nchar(line) + 1L, 1L, na.rm = TRUE)
   if (!is_number(column_number) || column_number < 0L || column_number > max_col) {
-    cli_abort(c(
-      i = "{.arg column_number} must be an integer between {.val {0}} and {.code nchar(line) + 1} ({.val {max_col}})",
-      x = "Instead, it was {.val {column_number}}."
-    ))
+    cli_abort(
+      "{.arg column_number} must be an integer between {.val {0}} and {.val {max_col}} ({.code nchar(line) + 1}),
+       not {.obj_type_friendly {column_number}}."
+    )
```
MichaelChirico, line 416:
```suggestion
    cli_abort("
      {.arg column_number} must be an integer between {.val {0}} and {.val {max_col}} ({.code nchar(line) + 1}),
      not {.obj_type_friendly {column_number}}.
    ")
```

R/lint.R
```diff
@@ -410,16 +410,16 @@ Lint <- function(filename, line_number = 1L, column_number = 1L, # nolint: objec
...
   }
   if (!is_number(line_number) || line_number < 1L) {
-    cli_abort(c(
-      i = "{.arg line_number} must be a positive integer.",
-      x = "Instead, it was {.val {line_number}}."
-    ))
+    cli_abort(
+      "{.arg line_number} must be a positive integer,
+       not {.obj_type_friendly {line_number}}."
+    )
```
MichaelChirico, line 422:
```suggestion
    cli_abort("{.arg line_number} must be a positive integer, not {.obj_type_friendly {line_number}}.")
```

R/settings.R
```diff
@@ -89,59 +89,67 @@ read_settings <- function(filename) {
   }
 }
 
-read_config_file <- function(config_file) {
+read_config_file <- function(config_file, call = parent.frame()) {
+  # call is passed to malformed to ensure linear trace.
```
MichaelChirico, line 93:
```suggestion
#' @param call Passed to malformed to ensure linear trace.
#' @noRd
read_config_file <- function(config_file, call = parent.frame()) {
```

R/with.R
```diff
@@ -90,7 +93,8 @@ modify_defaults <- function(defaults, ...) {
 #' @export
 linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "deprecated") {
   if (!is.character(tags) && !is.null(tags)) {
-    cli_abort("{.arg tags} must be a {.cls character} vector, or {.code NULL}.")
+    cli_abort("{.arg tags} must be a character vector, or {.code NULL}, \\
+               not {.obj_type_friendly {tags}}.")
```
MichaelChirico, line 97:
```suggestion
    cli_abort("{.arg tags} must be a character vector, or {.code NULL}, not {.obj_type_friendly {tags}}.")
```

R/with.R
```diff
@@ -102,8 +106,9 @@ linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "dep
       if (!all(available$linter %in% ns_exports)) {
         missing_linters <- setdiff(available$linter, ns_exports) # nolint: object_usage_linter. TODO(#2252).
         cli_abort(c(
-          i = "Linters {.fn {missing_linters}} are advertised by {.fn available_linters}.",
-          x = "But these linters are not exported by package {.pkg {package}}."
+          x = "Can't find linters {.fn {missing_linters}}.",
+          i = "These linters are advertised by {.fn available_linters}, but \\
+               are not exported by package {.pkg {package}}."
```
MichaelChirico, line 111:
```suggestion
          i = "These are advertised by {.fn available_linters}, but are not exported by package {.pkg {package}}."
```

R/with.R
```diff
@@ -178,9 +183,11 @@ linters_with_defaults <- function(..., defaults = default_linters) {
   dots <- list(...)
   if (missing(defaults) && "default" %in% names(dots)) {
     cli_warn(c(
-      "Did you mean {.arg defaults}?",
-      x = "{.arg default} is not an argument to {.fun linters_with_defaults}.",
-      i = "This warning will be removed when {.fun with_defaults} is fully deprecated."
+      x = "{.arg default} is not an argument to
+           {.help [{.fn linters_with_defaults}](lintr::linters_with_defaults)}.",
```
MichaelChirico, line 187:
```suggestion
      x = "
        {.arg default} is not an argument to {.help [{.fn linters_with_defaults}](lintr::linters_with_defaults)}.
      ",
```

R/settings.R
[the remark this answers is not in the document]
olivroy: Oh, sorry. In 07a96a0, I changed this back to Error in `read_settings()` : since `read_settings()` is exported by lintr. I could add a `call` argument to `read_settings()` if it makes sense to do so, but I didn't want to change anything in the user-facing functions. I can open an issue to do this in a follow-up? Or do it in this PR?

R/with.R
```diff
@@ -178,9 +181,12 @@ linters_with_defaults <- function(..., defaults = default_linters) {
...
     cli_warn(c(
-      "Did you mean {.arg defaults}?",
-      x = "{.arg default} is not an argument to {.fun linters_with_defaults}.",
-      i = "This warning will be removed when {.fun with_defaults} is fully deprecated."
+      x = "
+        {.arg default} is not an argument to {.help [{.fn linters_with_defaults}](lintr::linters_with_defaults)}.
+      ",
+      i = "Did you mean {.arg defaults}?",
+      # make message more subtle
+      cli::col_silver("This warning will be removed when {.fun with_defaults} is fully deprecated.")
```
IndrajeetPatil, line 189: Neat! I like it :)

R/with.R
```diff
@@ -31,8 +31,11 @@
 #' names(my_undesirable_functions)
 #' @export
 modify_defaults <- function(defaults, ...) {
-  if (missing(defaults) || !is.list(defaults) || !all(nzchar(names2(defaults)))) {
-    cli_abort("{.arg defaults} must be a named list.")
+  if (missing(defaults)) {
+    cli_abort("Missing required argument {.arg defaults} should be a named list.")
```
IndrajeetPatil, line 35: I think this error message should only mention that the required argument is missing. 
"Missing required argument" sounds a bit odd, and combines the error message for two different checks into one.

```r
cli_abort("{.arg defaults} is a required argument, but is missing. [It should be a named list.]") # second sentence is optional
```

olivroy: I agree, it would mimic more what `rlang::check_required()` provides. I can make this change. and adjust test.

```r
! `x` is absent but must be supplied.
```

R/settings.R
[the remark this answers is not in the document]
MichaelChirico: Hmm, `read_settings()` is not exported though 🤔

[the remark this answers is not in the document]
olivroy: Oh, okay! It is just a forgotten `#' @noRd`  that triggered build failure. I will add this and unrevert my change.

R/settings.R
```diff
@@ -164,7 +174,11 @@ validate_config_file <- function(config, config_file, defaults) {
   matched <- names(config) %in% names(defaults)
   if (!all(matched)) {
     unused_settings <- names(config)[!matched] # nolint: object_usage_linter. TODO(#2252).
-    cli_warn("Found unused settings in config {.str config_file}: {.field unused_settings}")
+    config_link <- cli::style_hyperlink( # nolint: object_usage_linter. TODO(#2252).
+      cli::col_blue(basename(config_file)),
+      url = paste0("file://", config_file)
+    )
```
IndrajeetPatil, line 180: Should we create a helper to create these clickable links? 

We are already doing it twice in `settings.R`, and maybe will need to do it a few more times in future, and I don't wish for these `nolint` directives to proliferate.

olivroy: Moved it to helper. However, still need the nolint directive until #2252 is fixed. It is that the variable created in unused, except in the error message.
The nolint can only be removed if the function is called directly in the cli message

IndrajeetPatil (approved):
Michael had already thoroughly reviewed it, so I have no further concerns. 

Thanks!

PR #2606, opened 2024-06-13
More adaptation to 4.0+ features

MichaelChirico:
Last follow-up to #2603


R/utils.R
```diff
@@ -220,7 +220,6 @@ platform_independent_sort <- function(x) x[platform_independent_order(x)]
 #' unlink(tmp)
 #'
 #' # more importantly, extract strings under R>=4 raw strings
```
AshesITR, line 222: Maybe also reword the example?

MichaelChirico: I originally thought it's still useful to call this out as an R4+ thing, but you're right, eventually this will look out of place, so might as well make the cut now.

AshesITR (approved):

PR #2613, opened 2024-06-19
Normalize output to Forward slash

olivroy:
Minor annoyance, but thought I'd share the PR in case there is interest

it is just to fix a minor inconsistency between Unix and Windows.

```r
x <- lint(path)
names(x)
```

This basically just changes that and makes sure that the names of the linted object are the same on both platforms with forward slash

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 * Drop support for posting GitHub comments from inside GitHub comment bot, Travis, Wercker, and Jenkins CI tools (spurred by #2148, @MichaelChirico). We rely on GitHub Actions for linting in CI, and don't see any active users relying on these alternatives. We welcome and encourage community contributions to get support for different CI system going again.
 * `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. Note that users with `cyclocomp_linter()` in their configs may now need to install {cyclocomp} intentionally, in particular in CI/CD pipelines.
 * `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
+* `lint()` and friends now normalizes paths to forward slashes on Windows.
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@
 * Drop support for posting GitHub comments from inside GitHub comment bot, Travis, Wercker, and Jenkins CI tools (spurred by #2148, @MichaelChirico). We rely on GitHub Actions for linting in CI, and don't see any active users relying on these alternatives. We welcome and encourage community contributions to get support for different CI system going again.
 * `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. Note that users with `cyclocomp_linter()` in their configs may now need to install {cyclocomp} intentionally, in particular in CI/CD pipelines.
 * `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
-* `lint()` and friends now normalizes paths to forward slashes on Windows.
+* `lint()` and friends now normalize paths to forward slashes on Windows (@olivroy, #2613).
 
 ## Bug fixes
 
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -292,7 +292,7 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
 
   paths <- names(x)
   rel_path <- !is_absolute_path(paths)
-  paths[rel_path] <- file.path(root, paths[rel_path], fsep = "/")
+  paths[rel_path] <- file.path(root, paths[rel_path])
 
   is_dir <- dir.exists(paths)
   if (any(is_dir)) {
@@ -307,7 +307,7 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
           pattern = pattern,
           recursive = TRUE
         )
-        file.path(dir, files, fsep = "/") # non-normalized relative paths
+        file.path(dir, files) # non-normalized relative paths
       }
     ))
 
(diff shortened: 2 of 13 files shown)


NEWS.md
```diff
@@ -17,6 +17,7 @@
 * Drop support for posting GitHub comments from inside GitHub comment bot, Travis, Wercker, and Jenkins CI tools (spurred by #2148, @MichaelChirico). We rely on GitHub Actions for linting in CI, and don't see any active users relying on these alternatives. We welcome and encourage community contributions to get support for different CI system going again.
 * `cyclocomp_linter()` is no longer part of the default linters (#2555, @IndrajeetPatil) because the tidyverse style guide doesn't contain any guidelines on meeting certain complexity requirements. Note that users with `cyclocomp_linter()` in their configs may now need to install {cyclocomp} intentionally, in particular in CI/CD pipelines.
 * `scalar_in_linter` is now configurable to allow other `%in%` like operators to be linted. The data.table operator `%chin%` is no longer linted by default; use `in_operators = "%chin%"` to continue linting it. (@F-Noelle)
+* `lint()` and friends now normalizes paths to forward slashes on Windows.
```
MichaelChirico, line 20:
```suggestion
* `lint()` and friends now normalize paths to forward slashes on Windows.
```

MichaelChirico:
Thanks! makes sense to me. admittedly, it feels weird to override the platform settings and force '/', but I don't know of any OS in current use that doesn't understand '/' so it's kinda moot.

I think more importantly, this PR suffers from threat of future regression -- there's nothing stopping us from using file.path()/normalizePath() the "old" way going forward.

Could you file a follow-up FR about a linter for this purpose? we can hash out design details there. It needn't impede this PR from being merged but needs to be recorded before slipping through the cracks.

olivroy:
Thanks for looking into it! I did this quickly to see if it could work (and if there was interest). 

I will look into removing changes, and see which ones are actually necessary. If there are <10, this would be less scary, and it could be documented where this is needed and not?

R/exclude.R
```diff
@@ -323,10 +323,10 @@ normalize_exclusions <- function(x, normalize_path = TRUE,
     paths <- names(x)
     # specify relative paths w.r.t. root
     rel_path <- !is_absolute_path(paths)
-    paths[rel_path] <- file.path(root, paths[rel_path])
+    paths[rel_path] <- file.path(root, paths[rel_path], fsep = "/")
     names(x) <- paths
     x <- x[file.exists(paths)] # remove exclusions for non-existing files
-    names(x) <- normalizePath(names(x)) # get full path for remaining files
+    names(x) <- normalizePath(names(x), winslash = "/") # get full path for remaining files
```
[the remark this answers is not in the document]
IndrajeetPatil: But we don't use `{purrr}` in `{lintr}`, and I am not sure how much work this would be in base-R.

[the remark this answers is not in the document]
MichaelChirico:
```r
\(...) normalizePath(..., winslash="/")
```

seems good enough for internal usage...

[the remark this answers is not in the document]
olivroy: @MichaelChirico that is what I did. Pushing the changes shortly.

olivroy:
Some tests feel redundant now (they are just testing the equivalency of paths). The only problem is that `withr::local_tempfile()` seems to create files with `\\`, so maybe a helper is needed to normalize path to fslash

```r
local_tempfile_f <- function(...) {
  file <- withr::local_tempfile()
  normalize_path(file)
}
```

R/lint.R
```diff
@@ -163,14 +163,17 @@ lint_dir <- function(path = ".", ...,
...
-    full.names = TRUE
-  ))
+  files <- normalize_path(
+    dir(
+      path,
+      pattern = pattern,
+      recursive = TRUE,
+      full.names = TRUE
+    )
+  )
```
MichaelChirico, line 176:
```suggestion
  files <- normalize_path(dir(
    path,
    pattern = pattern,
    recursive = TRUE,
    full.names = TRUE
  ))
```

R/path_utils.R
```diff
@@ -133,6 +133,16 @@ split_path <- function(dirs, prefix) {
   dirs[nzchar(dirs)]
 }
 
+# https://github.com/r-lib/lintr/pull/2613
+# Simple wrapper around normalizePath but ensures forward slash on Windows
+normalize_path <- function(path, winslash = "/", mustWork = NA) { # nolint: object_name_linter
+  normalizePath( # nolint: undesirable_function_linter
```
MichaelChirico, line 139:
```suggestion
normalize_path <- function(path, winslash = "/", mustWork = NA) { # nolint: object_name_linter.
  normalizePath( # nolint: undesirable_function_linter.
```

R/path_utils.R
```diff
@@ -133,6 +133,16 @@ split_path <- function(dirs, prefix) {
   dirs[nzchar(dirs)]
 }
 
+# https://github.com/r-lib/lintr/pull/2613
+# Simple wrapper around normalizePath but ensures forward slash on Windows
```
MichaelChirico, line 137:
```suggestion
#' Simple wrapper around normalizePath to ensure forward slash on Windows
#' https://github.com/r-lib/lintr/pull/2613
#' @noRd
```

R/path_utils.R
```diff
@@ -133,6 +133,16 @@ split_path <- function(dirs, prefix) {
...
+normalize_path <- function(path, winslash = "/", mustWork = NA) { # nolint: object_name_linter
```
MichaelChirico, line 138: I would simply drop `winslash=` from the signature, we don't have any plans of ever using it right?

olivroy: Good idea! Indeed, but could always be added back if needed.

MichaelChirico:
The approach of just using `normalizePath()` as an undesirable function looks good enough to me for now. We can think about a new linter upon user request later.

R/path_utils.R
MichaelChirico: i think if needed you would just use `normalizePath()` directly :)

R/path_utils.R
```diff
@@ -133,6 +133,17 @@ split_path <- function(dirs, prefix) {
...
 
+#' Simple wrapper around normalizePath to ensure forward slash on Windows
+#' https://github.com/r-lib/lintr/pull/2613
+#' @noRd
+normalize_path <- function(path, mustWork = NA) { # nolint: object_name_linter.
+  normalizePath( # nolint: undesirable_function_linter.
+    path = path,
+    winslash = "/",
+    mustWork = mustWork
+  )
```
MichaelChirico, line 144:
```suggestion
  # nolint next: undesirable_function_linter.
  normalizePath(path = path, winslash = "/", mustWork = mustWork)
```

R/path_utils.R
```diff
@@ -133,6 +133,14 @@ split_path <- function(dirs, prefix) {
   dirs[nzchar(dirs)]
 }
 
+#' Simple wrapper around normalizePath to ensure forward slash on Windows
+#' https://github.com/r-lib/lintr/pull/2613
+#' @noRd
+normalize_path <- function(path, mustWork = NA) { # nolint: object_name_linter.
+  # nolint next: undesirable_function_linter.
+  normalizePath(path = path, winslash = "/", mustWork = mustWork)
+}
```
MichaelChirico, line 142:
```suggestion
# nolint next: undesirable_function_linter, object_name_linter.
normalize_path <- function(path, mustWork = NA) normalizePath(path = path, winslash = "/", mustWork = mustWork)
```

MichaelChirico (approved):

PR #2618, opened 2024-06-21
Add potential `sequence()` lints to `seq_linter()`

### LINKED ISSUE #2595: New linter to recommend `sequence()` base R function
As an alternative to `unlist(lapply(x, seq_len))`.

A proof of concept implementation is available at https://github.com/r-lib/lintr/commit/32049afb75aa6b950d459a2797da0e059353ddac and I'd be happy to turn it into a PR, and add tests, etc.

Anything I should take into account first? Should it be merged into `seq_linter()`?

Bisaloo:
Fix #2595


R/seq_linter.R
```diff
@@ -42,6 +47,11 @@
 #'   linters = seq_linter()
 #' )
 #'
+#' lint(
+#'   text = "sapply(x, seq_len)",
```
IndrajeetPatil, line 51: Shouldn't this example use `sequence()`?

NEWS.md
```diff
@@ -32,6 +32,7 @@
 
 ## New and improved features
 
+* `seq_linter()` now includes lints to inform about missed opportunities to use the `sequence()` base R function (#2618, @Bisaloo)
```
MichaelChirico, line 35: I think `sequence()` is not a very well-known function ([<500 calls on CRAN vs more than 80K for `seq()`](https://github.com/search?q=org%3Acran+lang%3AR+%2F%5B%5E0-9a-z._%5Dsequence%5B%28%5D%2F&type=code)), here is a good opportunity to quickly introduce it:

"to use the `sequence()` function, e.g. `unlist(lapply(ints, seq))`"

R/seq_linter.R
```diff
@@ -66,6 +76,15 @@ seq_linter <- function() {
     ]
   ")
 
+  map_funcs <-  xp_text_in_table(c("sapply", "lapply", "map"))
+  sequence_xpath <- glue("
+  //SYMBOL_FUNCTION_CALL[ { map_funcs } ]
+    /parent::expr/parent::expr[
+      preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'unlist']
```
MichaelChirico, line 83: nit: I would try and keep the logic in one direction on the AST: check sapply, then check seq_len, then check unlist, roughly

```
//SYMBOL_FUNCTION_CALL/parent::expr[following-sibling::expr/SYMBOL]/parent::expr/parent::expr[expr/SYMBOL_FUNCTION_CALL]
```

Also, please switch to use the new `find_function_calls` approach, which is appropriate here, e.g.

https://github.com/r-lib/lintr/blob/6ed78c2373ea435932151a5ae2090a3efc346e25/R/condition_call_linter.R#L83

tests/testthat/test-seq_linter.R
```diff
@@ -123,6 +123,24 @@ test_that("reverse seq is ok", {
   )
 })
 
+test_that("finds potential sequence() replacements", {
+  linter <- seq_linter()
+  lint_msg <- rex::rex("Use sequence()")
+
+  expect_lint(
+    "unlist(lapply(x, seq_len))",
```
MichaelChirico, line 131: The lint now only covers `unlist(lapply(x, seq_len))`.

What about

 - `unlist(lapply(x, seq))`
 - `unlist(lapply(x, seq, by = 2))`, and other variants equivalent to using the `from=` and/or `by=` arguments of `sequence()`?
 - I _don't_ think it's worth trying to find examples like `lapply(x, \(xi) seq(xi))` which would be covered by `unnecessary_lambda_linter()`. But are there any examples that _do_ require looking into a lambda to replicate a `sequence()`-able call?

(it would be nice to get a sense of how many hits these are getting before investing too much time -- and also feel free to earmark some of this as a TODO in a follow-up issue, though `lapply(x,seq)` is trivial enough to be done here)

Bisaloo: > `unlist(lapply(x, seq, by = 2))`, and other variants equivalent to using the `from=` and/or `by=` arguments of sequence()?

I don't think there is a direct equivalence:

- `from` has a difference behaviour in `seq()` and `sequence()`

  ``` r
  unlist(lapply(1:5, seq, from = 2))
  #>  [1] 2 1 2 2 3 2 3 4 2 3 4 5

  sequence(1:5, from = 2)
  #>  [1] 2 2 3 2 3 4 2 3 4 5 2 3 4 5 6
  ```

  <sup>Created on 2024-06-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

- As far as I can tell, `by=` cannot be used on its own without `from=`, which brings us back to the previous case

  ``` r
  seq(20, by = 2)
  #> Error in seq.default(20, by = 2): wrong sign in 'by' argument
  ```

  <sup>Created on 2024-06-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I have edited the xpath to make sure this type of call doesn't lint.

R/seq_linter.R
Bisaloo: No preference really. I'm always a bit confused by negative examples because they could be almost anything really :shrug:

Happy to hear any guidelines on this or just follow your instinct if you think that's better.

IndrajeetPatil: We are trying to follow a one-to-one mapping between examples with-to-without lints in such a way that the without version follows the provided advice in with version. So if the with version mentions that the user is better off using `sequence()`, it would be preferable to follow that advice in the without version.

Does that make sense? Open to breaking that pattern if you feel strongly about it.

Bisaloo: Ooooh, yes, that makes sense. I thought they were just false negative but didn't realize they were actually the result of the linter recommendation. I have changed it

IndrajeetPatil: Thanks!

IndrajeetPatil (dismissed):
Let's also wait for MC's approval.

@MichaelChirico Can we add @Bisaloo to the repo as a contributor? Of course, only if he wishes to. 😅

MichaelChirico:
> @MichaelChirico Can we add @Bisaloo to the repo as a contributor? Of course, only if he wishes to. 😅

Invited!

tests/testthat/test-seq_linter.R
MichaelChirico: From `?sequence`, I think there _is_ another case, I just don't know if we'll ever see it in the wild :)

```r
unlist(mapply(\(nvec, from, by) seq(from, by=by, length.out=nvec), nvces, froms, bys))
```

There are a few possibilities here, anyway I don't think we need to include those yet:

https://github.com/search?q=lang%3AR+%2Funlist%5B%28%5D%5B+%5Cn%5D*map%2F+%2Fseq%5B%28%5D%2F&type=code

More worthy of consideration is `unlist(lapply(i, "seq"))`, but again can be done later.

PR #2629, opened 2024-07-11
Add new section for grouped linters in website reference index

IndrajeetPatil:
Currently, we are dumping way too many functions in the [meta-tooling section](https://lintr.r-lib.org/dev/reference/index.html#meta-tooling). This increases the possibility of the users missing out on important functions, like `all_linters()` or `default_linters()`.

### DIFF
--- a/_pkgdown.yaml
+++ b/_pkgdown.yaml
@@ -26,7 +26,7 @@ reference:
     contents:
       - ends_with("linter")
 
-  - title: Grouped linters
+  - title: Groups of linters
     contents:
       - ends_with("linters")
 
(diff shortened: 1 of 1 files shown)


_pkgdown.yaml
```diff
@@ -26,6 +26,10 @@ reference:
     contents:
       - ends_with("linter")
 
+  - title: Grouped linters
```
IndrajeetPatil, line 29: I am not sure about this title. But I also can't think of a better one.

MichaelChirico: "Groups of linters"?

MichaelChirico (approved):

AshesITR:
I see this is merged already, but I'd suggest moving these functions, and also `use_lintr()`, to the existing configuration section which already contains `linters_with_tags()`.

PR #2633, opened 2024-07-18
Don't include GHA log output in pkgdown website

IndrajeetPatil:
If I locally render the site, I see:

Screenshot 2024-07-18 at 17 36 48

## Minimal reprex:

``` r
library(lintr)

withr::local_options(lintr.rstudio_source_markers = FALSE)
withr::local_envvar(list(GITHUB_ACTIONS = "true"))

tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")

withr::local_envvar(list(IN_PKGDOWN = "true"))
lint(tmp)
#> /private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpwfSzjb/filedd1064e6f6e8:1:6: warning: [seq_linter] Use seq_len(nrow(...)) instead of 1:nrow(...), which is likely to be wrong in the empty edge case.
#> x <- 1:nrow(y)
#>      ^~~~~~~~~

withr::local_envvar(list(IN_PKGDOWN = "false"))
lint(tmp)
#> ::warning file=/private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpwfSzjb/filedd1064e6f6e8,line=1,col=6::file=/private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpwfSzjb/filedd1064e6f6e8,line=1,col=6,[seq_linter] Use seq_len(nrow(...)) instead of 1:nrow(...), which is likely to be wrong in the empty edge case.

withr::local_envvar(list(IN_PKGDOWN = ""))
lint(tmp)
#> ::warning file=/private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpwfSzjb/filedd1064e6f6e8,line=1,col=6::file=/private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpwfSzjb/filedd1064e6f6e8,line=1,col=6,[seq_linter] Use seq_len(nrow(...)) instead of 1:nrow(...), which is likely to be wrong in the empty edge case.
```

<sup>Created on 2024-07-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/tests/testthat/test-ci.R
+++ b/tests/testthat/test-ci.R
@@ -52,7 +52,6 @@ patrick::with_parameters_test_that(
   env_var_value = list("", "F", NA, NULL)
 )
 
-
 test_that("GitHub Actions log is skipped in pkgdown websites", {
   withr::local_envvar(list(GITHUB_ACTIONS = "true", IN_PKGDOWN = "true"))
   withr::local_options(lintr.rstudio_source_markers = FALSE)
--- a/tests/testthat/test-ci.R
+++ b/tests/testthat/test-ci.R
@@ -1,5 +1,5 @@
 test_that("GitHub Actions functionality works", {
-  withr::local_envvar(list(GITHUB_ACTIONS = "true"))
+  withr::local_envvar(list(GITHUB_ACTIONS = "true", IN_PKGDOWN = "false"))
   withr::local_options(lintr.rstudio_source_markers = FALSE)
   tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")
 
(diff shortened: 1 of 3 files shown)


tests/testthat/test-ci.R
```diff
@@ -51,3 +51,13 @@ patrick::with_parameters_test_that(
   },
   env_var_value = list("", "F", NA, NULL)
 )
+
+
```
MichaelChirico, line 55:
```suggestion
```

tests/testthat/test-ci.R
```diff
@@ -51,3 +51,12 @@ patrick::with_parameters_test_that(
   },
   env_var_value = list("", "F", NA, NULL)
 )
+
+test_that("GitHub Actions log is skipped in pkgdown websites", {
+  withr::local_envvar(list(GITHUB_ACTIONS = "true", IN_PKGDOWN = "true"))
+  withr::local_options(lintr.rstudio_source_markers = FALSE)
+  tmp <- withr::local_tempfile(lines = "x <- 1:nrow(y)")
+
+  l <- lint(tmp)
```
MichaelChirico, line 60: Might it be better to set `linters=seq_linter()` explicitly? Feels more future-proof & intentional.

MichaelChirico (approved):
LGTM! Minor tidy-up requests only.

MichaelChirico (approved):

MichaelChirico:
Does this warrant a NEWS entry? Not sure any downstreams care about running {lintr} inside {pkgdown}...

IndrajeetPatil:
> Does this warrant a NEWS entry? Not sure any downstreams care about running {lintr} inside {pkgdown}...

I would say no. I am not sure how many users are including lintr examples in their package's pkgdown websites, but I would bet that that's going to be a vanishingly small number.

PR #2635, opened 2024-07-24
Fix missing DCF code block on website

IndrajeetPatil:
My guess is since knitr doesn't support DCF code blocks (cf. `names(knitr::knit_engines$get())`), pkgdown doesn't either.

- Current status:

https://lintr.r-lib.org/dev/reference/read_settings.html

Screenshot 2024-07-24 at 07 31 33

- With this PR (locally)

Screenshot 2024-07-24 at 07 31 51

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,7 @@ Config/Needs/development: pkgload, cli, testthat, patrick
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.1
+RoxygenNote: 7.3.2
 Collate:
     'make_linter_from_xpath.R'
     'xp_utils.R'
--- a/R/settings.R
+++ b/R/settings.R
@@ -12,7 +12,7 @@
 #' This file is a DCF file, see [base::read.dcf()] for details.
 #' Here is an example of a `.lintr` file:
 #'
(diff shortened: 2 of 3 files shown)


MichaelChirico:
Strange, I wouldn't have expected {knitr} to get involved here at all -- that's a pandoc/markdown code gate, IIUC {knitr} only considers ```` ```{<engine> ```` gates.

Possibly, the {pkgdown} rendering tool sees `dcf` as unrecognized & just fails to emit anything -- I wrote that expecting it would just ignore the language "hint" in this case (actually, I fully didn't expect any markdown renderer to have a specific DCF highlighter, just left it in as a hint in case one "eventually" shows up...)

MichaelChirico:
Happy to accept the change, just wondering if this might be also reported with {pkgdown}?

MichaelChirico (approved):
Or maybe it's the `class="sourceCode dcf"` part that's at fault? since that's already HTML, i.e. past markdown rendering already.

Also note that this is not a {knitr} thing at all IIUC, it's really {roxygen2}->HTML-in-Rd->rendered HTML-in-{pkgdown}.

So, possibly {roxygen2} should be vetting that? 🤔

IndrajeetPatil:
You are right! 

This is indeed seems to be coming via roxygen2.

Feel free to create an issue upstream if you wish. :)

PR #2643, opened 2024-08-04
Print a message when no lints found

### LINKED ISSUE #2640: Provide feedback when no lints are found
I think it would be good if `{lintr}` provided some feedback for the success case where there are no lints found instead of returning haunting silence. I am not saying we should go full [`{praise}`](https://cran.r-project.org/web/packages/praise/index.html)-mode, like `{testthat}`. It could simply be a message.

---

``` r
library(withr)
library(lintr)

linters <- all_linters()
``` 

### Failure

``` r
tmp1 <- local_tempfile(lines = "1 + 1", fileext = ".R")
lint(tmp1, linters = linters)
#> /private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpxPFph2/fileca056a1dbb3.R:1:2: style: [implicit_integer_linter] Use 1L or 1.0 to avoid implicit integers.
#> 1 + 1
#> ~^
#> /private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpxPFph2/fileca056a1dbb3.R:1:6: style: [implicit_integer_linter] Use 1L or 1.0 to avoid implicit integers.
#> 1 + 1
#>     ~^
``` 

### Error

```r
tmp2 <- local_tempfile(lines = "1 + ", fileext = ".R")
lint(tmp2, linters = linters)
#> /private/var/folders/xr/v_vddzvs33q5wg7jv9mr__4h0000gn/T/RtmpxPFph2/fileca01d2d9b79.R:1:5: error: [error] unexpected end of input
#> 1 + 
#>     ^
```

### Success

``` r
tmp3 <- local_tempfile(lines = "1L + 1L", fileext = ".R")
lint(tmp3, linters = linters)
```

<sup>Created on 2024-08-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

---

Proposed:

``` r
tmp3 <- local_tempfile(lines = "1L + 1L", fileext = ".R")
lint(tmp3, linters = linters)
#> ℹ No lints found!
```

IndrajeetPatil:
Closes #2640

``` r
library(lintr)
lint(text = "x <- 1L", linters = all_linters())
#> ℹ No lints found.
```

<sup>Created on 2024-08-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -86,7 +86,7 @@
 
 ## Notes
 
-* All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
+* All user-facing messages are now prepared using the `{cli}` package (#2418, #2643, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
 * {lintr} now depends on R version 4.0.0. It already does so implicitly due to recursive upstream dependencies requiring this version; we've simply made that dependency explicit and up-front (#2569, @MichaelChirico).
 
 # lintr 3.1.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,7 @@
 * `make_linter_from_xpath()` errors up front when `lint_message` is missing (instead of delaying this error until the linter is used, #2541, @MichaelChirico).
 * `paste_linter()` is extended to recommend using `paste()` instead of `paste0()` for simply aggregating a character vector with `collapse=`, i.e., when `sep=` is irrelevant (#1108, @MichaelChirico).
 * `expect_no_lint()` was added as new function to cover the typical use case of expecting no lint message, akin to the recent {testthat} functions like `expect_no_warning()` (#2580, @F-Noelle).
+* `lint()` and friends emit a message if no lints are found (#2643, @IndrajeetPatil).
 
 ### New linters
 
@@ -86,7 +87,7 @@
 
 ## Notes
 
-* All user-facing messages are now prepared using the `{cli}` package (#2418, #2643, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
+* All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
 * {lintr} now depends on R version 4.0.0. It already does so implicitly due to recursive upstream dependencies requiring this version; we've simply made that dependency explicit and up-front (#2569, @MichaelChirico).
 
 # lintr 3.1.2
--- a/R/methods.R
+++ b/R/methods.R
@@ -104,7 +104,7 @@ print.lints <- function(x, ...) {
   }
 
   if (length(x) == 0L) {
-    cli_inform(c("i" = "No lints found."))
+    cli_inform(c(i = "No lints found."))
     # Empty lints: clear RStudio source markers
     if (use_rstudio_source_markers) rstudio_source_markers(x)
   }
--- a/R/methods.R
+++ b/R/methods.R
@@ -98,12 +98,15 @@ print.lints <- function(x, ...) {
       lapply(x, print, ...)
     }
 
-    if (isTRUE(settings$error_on_lint)) quit("no", 31L, FALSE) # nocov
-  }
-
-  if (length(x) == 0L) {
-    cli_inform(c("i" = "No lints found."))
-    if (use_rstudio_source_markers) rstudio_source_markers(x) # clear RStudio source markers
+    if (isTRUE(settings$error_on_lint)) {
+      quit("no", 31L, FALSE) # nocov
+    }
+  } else {
+    # Empty lints
+    cli_inform(c(i = "No lints found."))
+    if (use_rstudio_source_markers) {
+      rstudio_source_markers(x) # clear RStudio source markers
+    }
   }
 
   invisible(x)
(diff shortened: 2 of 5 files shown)


R/methods.R
```diff
@@ -101,10 +101,14 @@ print.lints <- function(x, ...) {
     if (isTRUE(settings$error_on_lint)) {
       quit("no", 31L, FALSE) # nocov
     }
-  } else if (use_rstudio_source_markers) {
+  }
+
+  if (length(x) == 0L) {
```
IndrajeetPatil, line 106: Extracting this branch to a separate conditional because the code is already quite nested.

IndrajeetPatil:
Windows failures seem spurious:

```r
> checking top-level files ... NOTE
  Non-standard file/directory found at top level:
    'install.ps1'
```

R/methods.R
AshesITR: This should be an `else` IMO because `!(length(x) > 0)` implies `length(x) == 0`.

IndrajeetPatil: Yes, but I am trying to reduce the nesting here, and it is easy to do so here by separating logic relevant for empty versus non-empty lint cases. 

Otherwise, the code is going to look like this:

```r
  if (length(x) > 0L) {
    inline_data <- x[[1L]][["filename"]] == "<text>"
    if (!inline_data && use_rstudio_source_markers) {
      rstudio_source_markers(x)
    } else if (in_github_actions() && !in_pkgdown()) {
      github_actions_log_lints(x, project_dir = github_annotation_project_dir)
    } else {
      lapply(x, print, ...)
    }

    if (isTRUE(settings$error_on_lint)) {
      quit("no", 31L, FALSE) # nocov
    }
  } else {
    cli_inform(c(i = "No lints found."))
    # Empty lints: clear RStudio source markers
    if (use_rstudio_source_markers) { 
      rstudio_source_markers(x)
    }
  }
```

Instead of 

```r
  if (length(x) > 0L) {
    inline_data <- x[[1L]][["filename"]] == "<text>"
    if (!inline_data && use_rstudio_source_markers) {
      rstudio_source_markers(x)
    } else if (in_github_actions() && !in_pkgdown()) {
      github_actions_log_lints(x, project_dir = github_annotation_project_dir)
    } else {
      lapply(x, print, ...)
    }

    if (isTRUE(settings$error_on_lint)) {
      quit("no", 31L, FALSE) # nocov
    }
  }

  if (length(x) == 0L) {
    cli_inform(c("i" = "No lints found."))
    # Empty lints: clear RStudio source markers
    if (use_rstudio_source_markers) rstudio_source_markers(x)
  }
```

AshesITR: I see no issue with the former. Maybe add a comment to the else, explaining that this handles the empty case?

R/methods.R
```diff
@@ -101,10 +101,14 @@ print.lints <- function(x, ...) {
     if (isTRUE(settings$error_on_lint)) {
       quit("no", 31L, FALSE) # nocov
     }
-  } else if (use_rstudio_source_markers) {
-    # Empty lints: clear RStudio source markers
-    rstudio_source_markers(x)
+  } else {
+    # Empty lints
+    cli_inform(c(i = "No lints found."))
```
MichaelChirico, line 106: this is a classed signal:

```r
class(tryCatch(cli::cli_inform("xxx"), condition=identity))
# [1] "rlang_message" "message"       "condition"
```

Is that appropriate? I might have expected just `cat()` here, I guess {cli} adds nice aesthetics -- are there any analogues for "plain text to terminal"?

(or perhaps a message _is_ appropriate, WDYT?)

NEWS.md
```diff
@@ -86,7 +86,7 @@
 
 ## Notes
 
-* All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
+* All user-facing messages are now prepared using the `{cli}` package (#2418, #2643, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
```
MichaelChirico, line 89: I would give this its own note rather than a simple issue reference -- it's a new feature, independently of {cli} being behind the implementation.

IndrajeetPatil:
> LGTM. The Windows failure looks like it needs reporting as a bug to r-lib/actions.

Done: https://github.com/r-lib/actions/issues/898

R/methods.R
IndrajeetPatil: Yes, we could use `cat()`, but I also think there are benefits to having this as a classed signal over `cat()` (e.g. the messages will show up in logs).

If we don't want a classed signal, either `cli_alert()` or its sibling should work:

``` r
cli::cli_alert_success("No lints found")
#> ✔ No lints found

class(tryCatch(cli::cli_alert_success("No lints found"), condition=identity))
#> [1] "cli_message" "condition"
```

<sup>Created on 2024-08-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

So we can have our pretty aesthetics cake and eat it too.

AshesITR (dismissed):

MichaelChirico:
Still waver a bit on whether signalling a condition is really the right thing to do for code that's performing fine, in fact congratulatory.

But, I don't think it actually matters -- we are a "front end" package & I can't think of any scenario where the difference between `cat()`/`message()` would actually matter for our users.

MichaelChirico (approved):

PR #2646, opened 2024-08-06
Support clickable hyperlinks

### LINKED ISSUE #2645: Support clickable hyperlinks in lintr output if hyperlinks are available (but not markers)
i.e. in Positron, output looks like this
image

It would be great if hyperlinks are available, but not source markers (i.e. `cli::ansi_has_hyperlink_support() && (!requireNamespace("rstudioapi") || !rstudioapi::hasFun("sourceMarkers")` ) then we could just use 

```r
paths <- c("tests/spelling:1:1", "tests/spelling:1:2")
paths2 <- paste0("{.path ", paths, "}")
paths3 <- unlist(lapply(paths2, cli::format_inline))
cat(paths3, sep = "\n")
```

I am however not sure where to put this condition, I tried in 

https://github.com/r-lib/lintr/blob/8d96145aa11d4eabf9b66fa391b57f8cc467a740/R/methods.R#L95-L99

```r

} else if (in_github_actions() && !in_pkgdown()) { 
   github_actions_log_lints(x, project_dir = github_annotation_project_dir) 
 } else if (cli::has_hyperlink_support()) { 
  # Do some modification here?
 } else { 
   lapply(x, print, ...) 
 } 
```

But it didn't seem to work 

This should happen in the print method to make sure no output is problematic.

I think it can happen in `format.lint()`. I may come up with a PR shortly.

olivroy:
fix #2645 

Basically, this checks for clickable hyperlinks. I also saw that some code was still hidden behind `requireNamespace("cli")` but that is no longer required since cli is imported.

To my great surprise, this works out of the box in Positron ! for `lintr::lint_package()`
image
Before, there were no hyperlinks (lintr 3.1.2)
image

Lmk if you'd like me to add NEWS for this!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -88,7 +88,7 @@
 ## Notes
 
 * All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
-* Error messages and linters contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
+* File locations in lints and error messages contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
 * {lintr} now depends on R version 4.0.0. It already does so implicitly due to recursive upstream dependencies requiring this version; we've simply made that dependency explicit and up-front (#2569, @MichaelChirico).
 
 # lintr 3.1.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -88,6 +88,7 @@
 ## Notes
 
 * All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
+* Error messages and linters contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
 * {lintr} now depends on R version 4.0.0. It already does so implicitly due to recursive upstream dependencies requiring this version; we've simply made that dependency explicit and up-front (#2569, @MichaelChirico).
 
 # lintr 3.1.2
--- a/R/methods.R
+++ b/R/methods.R
@@ -36,12 +36,10 @@ build_line_ref <- function(x) {
     as.character(x$column_number)
   )
 
-  create_clickable_link <- cli::ansi_has_hyperlink_support()
-  if (create_clickable_link) {
-    line_ref <- cli::format_inline("{.path {line_ref}}")
+  if (!cli::ansi_has_hyperlink_support()) {
+    return(line_ref)
   }
-
-  line_ref
+  cli::format_inline("{.path {line_ref}}")
 }
 
 #' @export
--- a/R/methods.R
+++ b/R/methods.R
@@ -8,28 +8,9 @@ format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
   )
   emph <- cli::style_bold
 
-  has_hyperlink_support <- cli::ansi_has_hyperlink_support()
-  if (has_hyperlink_support) {
-    fmt_start <- "{.path "
-    fmt_end <- "}"
-    fmt_f <- cli::format_inline
-  } else {
-    fmt_start <- ""
-    fmt_end <- ""
-    fmt_f <- paste0
-  }
-
+  line_ref <- build_line_ref(x)
   annotated_msg <- paste0(
-    emph(
-      fmt_f(
-        fmt_start,
-        x$filename, ":",
-        as.character(x$line_number), ":",
-        as.character(x$column_number),
-        fmt_end
-      ),
-      ": "
-    ),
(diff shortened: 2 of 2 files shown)


R/methods.R
```diff
@@ -1,26 +1,34 @@
...
+    fmt_f <- paste0
   }
 
   annotated_msg <- paste0(
     emph(
-      x$filename, ":",
-      as.character(x$line_number), ":",
-      as.character(x$column_number), ": ",
-      sep = ""
+      fmt_f(
```
MichaelChirico, line 24: I think this puts too much logic inside `format.lint`, could you add a helper for the hyperlink instead? e.g. (pseudo-code):

```r
styled_line_ref <- build_line_ref(x)
styled_type <- color(...)
annotated_msg <- glue("{line_ref}: {styled_type} [{linter}] {emph(message)}")
```

olivroy: Thanks for the suggestion, I also thought something felt off. I think it looks less hackish now?

R/methods.R
```diff
@@ -40,6 +29,22 @@ format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
...
+build_line_ref <- function(x) {
+  line_ref <- paste0(
+    x$filename, ":",
+    as.character(x$line_number), ":",
+    as.character(x$column_number)
+  )
+
+  create_clickable_link <- cli::ansi_has_hyperlink_support()
+  if (create_clickable_link) {
+    line_ref <- paste0("{.path ", line_ref, "}")
```
AshesITR, line 41: This looks brittle to me. How will this handle closing curly braces in filenames?

R/methods.R
```diff
@@ -40,6 +29,22 @@ format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
...
 }
 
+build_line_ref <- function(x) {
+  line_ref <- paste0(
+    x$filename, ":",
+    as.character(x$line_number), ":",
+    as.character(x$column_number)
+  )
+
+  create_clickable_link <- cli::ansi_has_hyperlink_support()
```
olivroy, line 39:
```suggestion
  create_clickable_link <- cli::ansi_has_hyperlink_support() && !grepl("\\{|\\}", x$filename)
```

@AshesITR, you are right, it would error, but I have never seen a file name with `{` in them. Could add this check to prevent this.

R/methods.R
```diff
@@ -40,6 +29,22 @@ format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
...
+  if (create_clickable_link) {
+    line_ref <- paste0("{.path ", line_ref, "}")
+    line_ref <- cli::format_inline(line_ref)
```
olivroy, line 42:
```suggestion
    line_ref <- cli::format_inline("{.path {line_ref}}")
```

Actually, this would be sufficient!

R/methods.R
olivroy: Seems like this works with this approach cc0cb5e

Although I just noticed that this works

```r
x <- "file{e.R"
cli::format_inline("{.path {x}}") |> cat()
```

While this doesn't work 

```r
cli::format_inline("{.path file{e.R}")
```

olivroy:
Just reinstalled this PR and confirm that a file with `{` will not throw an error!

AshesITR:
LGTM now, please add yourself to a NEWS bullet to take credit. Thanks!

NEWS.md
```diff
@@ -88,6 +88,7 @@
 ## Notes
 
 * All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
+* Error messages and linters contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
```
AshesITR, line 91:
```suggestion
* File locations in error messages and lints contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
```

MichaelChirico: Do we expect this to work in any IDE? I guess in terminal it tries to open the file with `EDITOR` (sorry can't test right now)

R/methods.R
```diff
@@ -40,6 +29,21 @@ format.lint <- function(x, ..., width = getOption("lintr.format_width")) {
...
 }
 
+build_line_ref <- function(x) {
+  line_ref <- paste0(
+    x$filename, ":",
+    as.character(x$line_number), ":",
+    as.character(x$column_number)
+  )
+
+  create_clickable_link <- cli::ansi_has_hyperlink_support()
```
MichaelChirico, line 39: Suggestion: this might read nicer as:

```r
if (!cli::ansi_has_hyperlink_support()) {
  return(line_ref)
}
cli::format_inline("{.path {line_ref}}")
```

WDYT?
```

AshesITR (approved):

PR #2659, opened 2024-09-06
Include `{data.table}` example in `redundant_equals_linter()` docs

### LINKED ISSUE #2657: Include `{data.table}` example with and without lint for `redundant_equals_linter()` docs
Since the lint message highlights this case so prominently, there should also be a corresponding example in the docs:

``` r
library(lintr)

lint(text = "dt[x == FALSE, y]", linters = redundant_equals_linter())
#> <text>:1:4: warning: [redundant_equals_linter] Using == on a logical vector is redundant. Well-named logical vectors can be used directly in filtering. For data.table's `i` argument, wrap the column name in (), like `DT[(is_treatment)]`.
#> dt[x == FALSE, y]
#>    ^~~~~~~~~~

lint(text = "dt[!(x), y]", linters = redundant_equals_linter())
#> ℹ No lints found.
```

<sup>Created on 2024-09-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
closes #2657

### DIFF
--- a/R/redundant_equals_linter.R
+++ b/R/redundant_equals_linter.R
@@ -35,6 +35,7 @@
 #'   linters = redundant_equals_linter()
 #' )
 #'
+#' # in data.table semantics, dt[x] is a join, dt[(x)] is a subset
 #' lint(
 #'   text = "dt[!(x), y]",
 #'   linters = redundant_equals_linter()
--- a/R/redundant_equals_linter.R
+++ b/R/redundant_equals_linter.R
@@ -20,7 +20,7 @@
 #' )
 #'
 #' lint(
-#'   text = "dt[x == FALSE, y]",
+#'   text = "dt[is_tall == FALSE, y]",
 #'   linters = redundant_equals_linter()
 #' )
 #'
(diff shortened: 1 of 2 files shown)


R/redundant_equals_linter.R
```diff
@@ -30,6 +35,11 @@
 #'   linters = redundant_equals_linter()
 #' )
 #'
+#' lint(
+#'   text = "dt[!(x), y]",
```
MichaelChirico, line 39: I prefer keeping `!` inside `()` myself:

```suggestion
#'   text = "dt[(!x), y]",
```

MichaelChirico: I also think a comment is warranted here...

MichaelChirico, line 39: I prefer keeping `!` inside `()` myself:

```suggestion
#' # in data.table semantics, dt[x] is a join, dt[(x)] is a subset
#' lint(
#'   text = "dt[!(x), y]",
```

MichaelChirico: I might also prefer something more practical like `dt[is_tall = FALSE, y]` --> `dt[(!is_tall), y]`.

IndrajeetPatil: Thanks, Michael. I think it's a good idea to include a comment and also use a more realistic column name.

MichaelChirico (approved):
Great!

MichaelChirico (approved):

PR #2660, opened 2024-09-06
Test to check that all linters are covered by tags

IndrajeetPatil:


tests/testthat/test-linter_tags.R
```diff
@@ -234,3 +234,11 @@ test_that("available_linters gives precedence to included tags", {
     available_linters(tags = "deprecated", exclude_tags = NULL)
   )
 })
+
+test_that("all non-deprecated linters are covered across tags", {
+  tags <- setdiff(available_tags(), "deprecated")
+  tagged_linters <- lapply(tags, function(tag) names(linters_with_tags(tag))) # nolint: undesirable_function_linter.
```
IndrajeetPatil, line 240: The output length is going to be different across tags, so not sure how to use `vapply()` instead of `lapply()`.

AshesITR:
This test could work by just looking at the `tags` column of `available_linters()`.
INM this is equivalent to `all(lengths(available_linters()$tags) > 0L)`.

IndrajeetPatil:
> all(lengths(available_linters()$tags) > 0L)

That's much cleaner! Thanks.

AshesITR (approved):

PR #2669, opened 2024-10-03
Fix missing character issue in JOSS's PDF output

IndrajeetPatil:
cf. https://github.com/openjournals/joss-reviews/issues/7240#issuecomment-2387723681

### DIFF
--- a/paper/paper.Rmd
+++ b/paper/paper.Rmd
@@ -59,7 +59,8 @@ knitr::opts_chunk$set(
 library(lintr)
 
 withr::local_options(list(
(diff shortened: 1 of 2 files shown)


MichaelChirico (approved):
Approving but not submitting in case another fix (e.g. a more suitable font) is decided

IndrajeetPatil:
I doubt the upstream JOSS issue is going to get any love anytime soon. So I'd prefer to use this workaround for now.

MichaelChirico (approved):

PR #2676, opened 2024-10-26
Draft: New linter for complex conditional expressions

### LINKED ISSUE #1830: New linter to suggest moving complex conditional expressions to boolean function/variable?
## Preamble

Complex conditional expressions can make the code difficult to read because they interrupt the "flow" of the main program.

If one instead extracts them to a boolean function/variable, the new conditional expression much more readable and, as a side effect, the new function/variable introduces a reusable abstraction in the codebase.

## Example

Actual

```r
if (looks_like_a_duck(x) && 
    swims_like_a_duck(x) && 
    quacks_like_a_duck(x)) {
  ...
}
```

Suggested-1: boolean function

```r
is_duck <- function(x) {
  looks_like_a_duck(x) &&
    swims_like_a_duck(x) &&
    quacks_like_a_duck(x)
}

if (is_duck(x)) {
  ...
}
```

Suggested-1: boolean variable

```r
is_duck <- 
  looks_like_a_duck(x) &&
  swims_like_a_duck(x) &&
  quacks_like_a_duck(x)

if (is_duck) {
  ...
}
```

Note that this is the simplest cases, but one can imagine that the expression depend on different objects, and so the boolean function can be multi-parameter (e.g. `is_duck(x, y, z)`).

## Configurable

As for what constitutes a _complex_ conditional expression can be configurable. The default will be 2L. That is, any expression with more than two operands will set off this linter. But users can set this to a higher or a lower value (e.g. if someone prefers to _always_ use a boolean function in a conditional expression to improve readability).

IndrajeetPatil:
closes #1830 

Draft PR for early feedback.

---

### Examples

``` r
library(lintr)

lint(
  text = "if (a && b) NULL",
  linters = complex_conditional_linter()
)
#> ℹ No lints found.

lint(
  text = "if (a && b || c) NULL",
  linters = complex_conditional_linter()
)
#> <text>:1:5: warning: [complex_conditional_linter] Complex conditional with more than 1 logical operator(s). Consider extracting into a boolean function or variable for readability and reusability.
#> if (a && b || c) NULL
#>     ^~~~~~~~~~~

lint(
  text = "if (a && b || c) NULL",
  linters = complex_conditional_linter(2)
)
#> ℹ No lints found.

lint(
  text = "
      if (x > 0 && y < 10 && z == TRUE) {
        while (a && b && c) {
          print('double complex')
        }
      }
  ",
  linters = complex_conditional_linter()
)
#> <text>:2:11: warning: [complex_conditional_linter] Complex conditional with more than 1 logical operator(s). Consider extracting into a boolean function or variable for readability and reusability.
#>       if (x > 0 && y < 10 && z == TRUE) {
#>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
#> <text>:3:16: warning: [complex_conditional_linter] Complex conditional with more than 1 logical operator(s). Consider extracting into a boolean function or variable for readability and reusability.
#>         while (a && b && c) {
#>                ^~~~~~~~~~~
```

<sup>Created on 2024-10-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>

Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.1 (2024-06-14)
#>  os       macOS 15.0.1
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Europe/Berlin
#>  date     2024-10-26
#>  pandoc   3.4 @ /usr/local/bin/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package      * version    date (UTC) lib source
#>  backports      1.5.0      2024-05-23 [1] CRAN (R 4.4.0)
#>  cli            3.6.3      2024-06-21 [1] CRAN (R 4.4.0)
#>  data.table     1.16.0     2024-08-27 [2] CRAN (R 4.4.1)
#>  digest         0.6.37     2024-08-19 [1] CRAN (R 4.4.1)
#>  evaluate       1.0.1      2024-10-10 [1] CRAN (R 4.4.1)
#>  fansi          1.0.6      2023-12-08 [1] CRAN (R 4.4.0)
#>  fastmap        1.2.0      2024-05-15 [1] CRAN (R 4.4.0)
#>  fs             1.6.4      2024-04-25 [1] CRAN (R 4.4.0)
#>  glue           1.8.0      2024-09-30 [1] CRAN (R 4.4.1)
#>  htmltools      0.5.8.1    2024-04-04 [1] CRAN (R 4.4.0)
#>  knitr          1.48       2024-07-07 [1] CRAN (R 4.4.0)
#>  lazyeval       0.2.2      2019-03-15 [1] CRAN (R 4.4.0)
#>  lifecycle      1.0.4      2023-11-07 [1] CRAN (R 4.4.0)
#>  lintr        * 3.1.2.9000 2024-10-26 [1] local
#>  magrittr       2.0.3      2022-03-30 [1] CRAN (R 4.4.0)
#>  pillar         1.9.0      2023-03-22 [1] CRAN (R 4.4.0)
#>  pkgconfig      2.0.3      2019-09-22 [1] CRAN (R 4.4.0)
#>  reprex         2.1.1      2024-07-06 [2] CRAN (R 4.4.1)
#>  rex            1.2.1      2021-11-26 [1] CRAN (R 4.4.0)
#>  rlang          1.1.4      2024-06-04 [1] CRAN (R 4.4.0)
#>  rmarkdown      2.28       2024-08-17 [1] CRAN (R 4.4.0)
#>  rstudioapi     0.16.0     2024-03-24 [2] CRAN (R 4.4.0)
#>  sessioninfo    1.2.2      2021-12-06 [2] RSPM
#>  tibble         3.2.1      2023-03-20 [1] CRAN (R 4.4.0)
#>  utf8           1.2.4      2023-10-22 [1] CRAN (R 4.4.0)
#>  vctrs          0.6.5      2023-12-01 [1] CRAN (R 4.4.0)
#>  withr          3.0.1      2024-07-31 [1] CRAN (R 4.4.1)
#>  xfun           0.48       2024-10-03 [1] CRAN (R 4.4.1)
#>  xml2           1.3.6      2023-12-04 [1] CRAN (R 4.4.0)
#>  xmlparsedata   1.0.5      2021-03-06 [1] CRAN (R 4.4.0)
#>  yaml           2.3.10     2024-07-26 [1] CRAN (R 4.4.1)
#> 
#>  [1] /Users/indrajeetpatil/Library/R/arm64/4.4/library
#>  [2] /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library
#> 
#> ──────────────────────────────────────────────────────────────────────────────
```

</details>


R/complex_conditional_linter.R
```diff
@@ -0,0 +1,105 @@
...
+#'  ...
+#' }
+#' ```
+#'
+#' In addition to improving code readability, extracting complex conditional expressions
+#' has the added benefit of introducing a reusable abstraction.
+#'
+#' @param threshold Integer. The maximum number of logical operators (`&&` or `||`)
+#'   allowed in a conditional expression. The default is `1L`, meaning any conditional expression
+#'   with more than one logical operator will be flagged.
```
IndrajeetPatil, line 49: @MichaelChirico, @AshesITR WDYT? 

I can't make up my mind if the threshold should represent the number of **operators** or **operands**. 
I have chosen the former because I _feel_ that it's an easier and user-friendly way to detect the complexity of the conditional expression, but I am not wedded to the idea.

MichaelChirico:
I don't think `a && b || c` is complex at all TBH, besides the mixing of precedences (but presumably `(a && b) || c` would also throw a lint here?), which I think is a separate issue.

I'm more comfortable calling `nontrivial_expr(a) && nontrivial_expr(b) || nontrivial_expr(c)` complex, but just plain names don't seem to add complexity IMO.

Defining the metric for "complexity" here is going to be quite hard -- are there other languages that have implemented something similar, or is there some CS literature we can turn to for some more generic definitions?

IndrajeetPatil:
> Defining the metric for "complexity" here is going to be quite hard -- are there other languages that have implemented something similar, or is there some CS literature we can turn to for some more generic definitions?

You are correct that this is a difficult question, but I don't think we need to have a consensus definition of what counts as "complex" conditional because that's an inherently subjective notion. The only thing we need to figure out is a good default (the same way we did for cyclocomp linter). And, since this is a configurable linter, even if we adopt a threshold that users feel is too restrictive, they can easily change this in config.

I think the current default of `threshold = 1L` is too aggressive, but we can easily change it to `threshold = 2L` (or higher):

```r
lint(
  text = "if (a && b || c) NULL",
  linters = complex_conditional_linter(2)
)
#> ℹ No lints found.
```

So the question for us to resolve is what should be the default threshold? 
How did we decide on `15L` as the default threshold for McCabe complexity linter?

---

I would vote for `threshold = 2L`; any conditional expression with more than 2 operators (and 3 logical operands) can benefit from simplification, IMO. 

Here is an example from our codebase:

```r
if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line) || e$message == "unexpected end of input")) {
  ...
}
```

which lints with this new linter and can be modified something along these lines:

```r
is_expression_valid <- (is.na(e$line) || !nzchar(e$line) || e$message == "unexpected end of input")
if (inherits(e, "lint") && is_expression_valid) {
  ...
}
```

---

P.S. As for linters in other programming languages, I can't think of any (neither `ruff` nor `eslint` have anything similar; the closest ones are about [complex conditional in type stubs](https://docs.astral.sh/ruff/rules/complex-assignment-in-stub/) or [mixing operators](https://eslint.style/rules/js/no-mixed-operators)).

IndrajeetPatil:
I don't understand why the `repo-meta-tests` workflow is failing. There are already metadata tests for the new linter.

AshesITR:
What happens when you run `.dev/lint_metadata_tests.R` locally?

From my understanding the script replaces all line numbers by 2^31 - 1 during `Lint()` construction and assumes the test file for each linter `the_linter` is named `test-the_linter.R`.

IndrajeetPatil:
@AshesITR Thanks!! That was it. Changing the filename fixed the issue.

IndrajeetPatil:
@MichaelChirico, @AshesITR, @Bisaloo Anyone have time to review this?

It has been sitting for a long time; I have already addressed the initial feedback.

NEWS.md
```diff
@@ -75,6 +75,10 @@
 * The description in `?paste_linter` of `allow_file_path=` has been corrected (#2675, @MichaelChirico). In particular, `allow_file_path="never"` is the most strict form, `allow_file_path="always"` is the most lax form.
 * `comment_token` is removed from settings. This was a vestige of the now-defunct support for posting GitHub comments.
 
+## New linters
```
MichaelChirico, line 82: the NEWS merge appears broken

R/complex_conditional_linter.R
```diff
@@ -0,0 +1,111 @@
...
+#' )
+#'
+#' @evalRd rd_tags("complex_conditional_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+complex_conditional_linter <- function(threshold = 2L) {
+  stopifnot(is.numeric(threshold), length(threshold) == 1L, threshold >= 1L)
+  threshold <- as.integer(threshold)
+
+  xpath <- glue::glue("//expr[
```
MichaelChirico, line 82: Let's avoid `//expr`, it looks like `//IF | //WHILE` will do better

R/complex_conditional_linter.R
```diff
@@ -0,0 +1,111 @@
...
+    parent::expr[IF or WHILE]
+    and
+    preceding-sibling::*[1][self::OP-LEFT-PAREN]
+    and
+    following-sibling::*[1][self::OP-RIGHT-PAREN]
+    and
+    count(descendant-or-self::*[AND2 or OR2]) > {threshold}
```
MichaelChirico, line 89: It's a bit strange to me to consider these of different complexity:

```r
if (any(x | y, na.rm = TRUE)) { ... }
if (any(x, na.rm = na.rm || y)) { ... }
```

One "fix" would be to focus on the operators _strictly related to the condition at hand_, i.e. only those "outmost" enough in the expressions to directly affect the `if (cond)` outcome. It's hard to describe, but some examples:

```r
if (x || y)
#     ^ yes
if (x || (y && z))
#     ^ yes ^ yes
if (foo(x && y))
#         ^ no
if (foo(x && y) && z)
#         ^ no  ^ yes
```

Just generally, I am still hung up on whether we have a sufficiently good+objective way to measure "complexity".

lintr.Rproj
```diff
@@ -1,4 +1,5 @@
 Version: 1.0
+ProjectId: ab4c2695-5b43-4cc6-8ea0-0daf83afd8a3
```
MichaelChirico, line 2: spurious?

PR #2679, opened 2024-10-27
re_matches_logical helper ensures logical output of re_matches_logical

### LINKED ISSUE #2678: todo_comment_linter(except_regex=) fails with capture groups
```r
lint("# abc\n", todo_comment_linter('abc', except_regex = "(abc)"))
# Error in `h()`:
# ! Linter `linter()` failed in /tmp/Rtmp6XgIZm/file2adee8120fb1df:
# invalid argument type
# Run `rlang::last_trace()` to see where the error occurred.
```

The error is really confusing too. It's easy to forget to use `?:` when writing a regex (and it's not clear that's a good requirement). Primarily this is due to `rex::re_matches()` changing output type.

MichaelChirico:
Closes #2678.

This changes what was an error in `expect_lint()` into functioning code. On balance I think consistency is good here; the downside is previously user providing capture group _accidentally_ would be met with an error steering them to improving their inputs.


MichaelChirico:
All of the fixes are for dev-only features, so didn't update any NEWS.

AshesITR:
I thunk we should add a bullet stating that some linters, and `expect_lint()` now support capture groups in their regex.

AshesITR (changes requested):
Let's add to NEWS that support for regex with capture groups was added for some linters.

MichaelChirico:
> Let's add to NEWS that support for regex with capture groups was added for some linters.

OK, PTAL. The thing is, the changes to `return_linter()` and `todo_comment_linter()` are dev-only, so a NEWS item doesn't make sense. The change for `object_usage_linter()` is not user-visible.

NEWS.md
```diff
@@ -94,6 +94,7 @@
 * All user-facing messages are now prepared using the `{cli}` package (#2418, @IndrajeetPatil). All messages have been reviewed and updated to be more informative and consistent.
 * File locations in lints and error messages contain clickable hyperlinks to improve code navigation (#2645, #2588, @olivroy).
 * {lintr} now depends on R version 4.0.0. It already does so implicitly due to recursive upstream dependencies requiring this version; we've simply made that dependency explicit and up-front (#2569, @MichaelChirico).
+* `unreachable_code_linter()` is less strict about whether `allow_comment_regex=` contains capture groups (#2678, @MichaelChirico).
```
MichaelChirico, line 97:
```suggestion
* Some code with parameters accepting regular expressions is less strict about whether there are capture groups  (#2678, @MichaelChirico). In particular, this affects `unreachable_code_linter(allow_comment_regex=)` and `expect_lint(checks=)`.
```

AshesITR (approved):

PR #2681, opened 2024-10-30
Add utils to check for lint and error classes

IndrajeetPatil:
A small step towards lint clean-up that will eventually be needed by https://github.com/r-lib/lintr/pull/2676.

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -86,7 +86,7 @@ get_source_expressions <- function(filename, lines = NULL) {
   source_expression$content <- get_content(source_expression$lines)
   parsed_content <- get_source_expression(source_expression, error = function(e) lint_parse_error(e, source_expression))
 
-  if (inherits(e, "lint") && (is.na(e$line) || !nzchar(e$line) || e$message == "unexpected end of input")) {
+  if (is_lint(e) && (is.na(e$line) || !nzchar(e$line) || e$message == "unexpected end of input")) {
     # Don't create expression list if it's unreliable (invalid encoding or unhandled parse error)
     expressions <- list()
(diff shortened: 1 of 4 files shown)


R/get_source_expressions.R
```diff
@@ -502,7 +502,7 @@ get_source_expression <- function(source_expression, error = identity) {
     error = error
   )
 
-  if (inherits(parsed_content, c("error", "lint"))) {
+  if (is_error(parsed_content) || is_lint(parsed_content)) {
```
AshesITR, line 505: This should be `&&` IINM.

R/get_source_expressions.R
```diff
@@ -513,7 +513,7 @@ get_source_expression <- function(source_expression, error = identity) {
...
```
AshesITR, line 516: ditto

R/get_source_expressions.R
IndrajeetPatil: The original code is conveying that either of those classes is fine, or at least that's what `inherits()` entails:

``` r
x <- tibble::tibble()
y <- data.frame()

class(x)
#> [1] "tbl_df"     "tbl"        "data.frame"
class(y)
#> [1] "data.frame"

inherits(x, c("tbl_df", "data.frame"))
#> [1] TRUE
inherits(y, c("tbl_df", "data.frame"))
#> [1] TRUE
```

<sup>Created on 2024-11-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

AshesITR: Ah, thanks for clarifying.

AshesITR (approved):

PR #2684, opened 2024-11-05
Consolidate suboptimal usages of `which()` under one linter

### LINKED ISSUE #1544: Consolidate suboptimal usages of `which()` under one linter
To lint for instances of `which(x == min(x))` or `which(x == max(x))`, which should be avoided in favour of more efficient `which.min()` or `which.max()`, respectively.

Ref: https://csgillespie.github.io/efficientR/performance.html

Screenshot 2022-09-20 at 15 39 43

IndrajeetPatil:
Closes #1544


AshesITR:
Isn't `which.min(x)` only the first index of a minimum value?

```r
x <- c(1, 1)
which(x == min(x))
which.min(x)
```

IndrajeetPatil:
Oh damn, you are right! 
I didn't consider that using `which(x == min(x))` to find _all_ minimum elements, and not just the (first) _one_, is also a valid and intentional usage. 

I am not sure how to proceed then. Just looking at this code then we can't guess the intent, and it's not possible to suggest an alternative reliable.

AshesITR:
I would forfeit this linter unless we find a more efficient and still equivalent function. It is algorithmically possible with a single pass of `x`, but I'm not aware of any implementations.

PR #2688, opened 2024-11-22
Emphasize S3-S4 difference in recommendation for how to fix class_equals_linter()

MichaelChirico:
Per experience here:

https://github.com/Rdatatable/data.table/pull/6603#discussion_r1854842225

I think it's good to more strongly recommend `is()` only for S4 classes.

### DIFF
--- a/R/class_equals_linter.R
+++ b/R/class_equals_linter.R
@@ -48,9 +48,10 @@ class_equals_linter <- function() {
     bad_expr <- xml_find_all(xml_calls, xpath)
 
     operator <- xml_find_chr(bad_expr, "string(*[2])")
-    lint_message <- sprintf(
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Looks like there's on ongoing outage for `r-lib/actions` on `macos: latest`:

https://github.com/r-lib/actions/issues/948

AshesITR:
How do you feel about a more subtle nudge by mentioning `is(x, "s4-class")`?

MichaelChirico:
It crossed my mind, but OTOH I don't see a big benefit to being overly parsimonious with the lint description, and the comment I received was that they had totally glanced over `inherits()`. `"s4-class"` _could_ be enough, though. If you feel the current description is definitely too wordy I can make the change.

AshesITR (approved):
It's fine both ways for me.

PR #2690, opened 2024-12-03
Register Enhances S3 methods more selectively

MichaelChirico:
Cribbing this from https://github.com/Rdatatable/data.table/pull/6589 -- our current approach always tries `requireNamespace()`. IMO it's better to avoid that for the very-weak `Enhances` dependency if we can.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -341,7 +341,7 @@ enhances_method_registration <- function(generic, s3class, implementation, owner
 
   reset_settings()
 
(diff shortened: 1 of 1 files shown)


AshesITR:
Isn't there `@rawNamespace` for manual entries?

MichaelChirico:
Right, I'm just not sure if it's supported directly or we need to resort to `@rawNamespace`.

PR #2693, opened 2024-12-04
Use exportS3Method for suggests and enhances packages

olivroy:
Fix #2690 

@MichaelChirico 

The advice was updated about S3 methods in R-packages as well  

https://github.com/hadley/r-pkgs/issues/1082


MichaelChirico:
I actually think the current usage of {data.table} is _precisely_ what's intended by the very weak `Enhances` dependency. I wouldn't move it to `Suggests`.

MichaelChirico:
Excellent, thanks!

MichaelChirico (approved):

PR #2698, opened 2024-12-09
Extend assignment_operator to optionally allow '=' assignments

### LINKED ISSUE #2441: Extend assignment_linter() to enforce '=' by default for assignment

MichaelChirico:
Closes #2441. Tests are partially copied from #2521 -- for implementation I'm starting from scratch rather than trying to re-work it based on the discussion there. Thanks @J-Moravec for getting things started & the fruitful discussion!

Pausing here for, well, even more discussion :)

I'm getting hung up on the interaction of the proposed argument with the other arguments, particularly `allow_right_assign` and `allow_pipe_assign`.

If we have `(allow_right_assign=FALSE, top_level_operator="any")`, that does make `x -> 2` a bit ambiguous, right?

I can think of two options:

 1. Subsume the other options into this new one by allowing `top_level_operator = c("<-", "->", "%<>%")` to be equivalent to the current `(allow_right_assign=TRUE, allow_pipe_assign=TRUE)`. That looks pretty transparent to me, but presents some back-compatibility issues.
 2. Ditch `any` in favor of `either`, i.e. _either_ `<-` or `=`.

WDYT @IndrajeetPatil @AshesITR?

### DIFF
--- a/R/assignment_linter.R
+++ b/R/assignment_linter.R
@@ -7,6 +7,9 @@
 #' @param allow_right_assign Logical, default `FALSE`. If `TRUE`, `->` and `->>` are allowed.
 #' @param allow_trailing Logical, default `TRUE`. If `FALSE` then assignments aren't allowed at end of lines.
 #' @param allow_pipe_assign Logical, default `FALSE`. If `TRUE`, magrittr's `%<>%` assignment is allowed.
+#' @param top_level_operator Character, default `"<-"`, matching the style guide recommendation for assignments with
+#'   `<-`. When `"="`, all top-level assignments must be done with `=`; when `"any"`, there is no enforcement of
+#'   top-level assignment operator.
 #'
 #' @examples
 #' # will produce lints
@@ -25,6 +28,11 @@
 #' lint(
 #'   text = "x %<>% as.character()",
 #'   linters = assignment_linter()
+#' )
+#'
+#' lint(
+#'   text = "x <- 1",
+#'   linters = assignment_linter(top_level_operator = "=")
 #' )
 #'
 #' # okay
@@ -64,6 +72,11 @@
 #'   linters = assignment_linter(allow_pipe_assign = TRUE)
 #' )
 #'
+#' lint(
+#'   text = "x = 1",
+#'   linters = assignment_linter(top_level_operator = "=")
+#' )
+#'
 #' @evalRd rd_tags("assignment_linter")
 #' @seealso
 #' - [linters] for a complete list of linters available in lintr.
@@ -73,7 +86,10 @@
 assignment_linter <- function(allow_cascading_assign = TRUE,
                               allow_right_assign = FALSE,
                               allow_trailing = TRUE,
-                              allow_pipe_assign = FALSE) {
+                              allow_pipe_assign = FALSE,
+                              top_level_operator = c("<-", "=", "any")) {
+  top_level_operator <- match.arg(top_level_operator)
+
   trailing_assign_xpath <- paste(
     collapse = " | ",
     c(
--- a/tests/testthat/test-assignment_linter.R
+++ b/tests/testthat/test-assignment_linter.R
@@ -196,14 +196,18 @@ test_that("multiple lints throw correct messages", {
 test_that("assignment operator can be toggled", {
   eq_linter <- assignment_linter(top_level_operator = "=")
   any_linter <- assignment_linter(top_level_operator = "any")
+  lint_message <- rex("Use =, not")
 
   expect_lint("a = 1", NULL, eq_linter())
   expect_lint("a = 1", NULL, any_linter())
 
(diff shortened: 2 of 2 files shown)


AshesITR:
From an interface standpoint I think `top_level_operator` should be called `assignment_operator` and `"any"` shouldn't be an option.

Maybe it's worth exploring what `=` assignment style should look like for inner assignments and reduce flexibility a bit when it comes to them.

IndrajeetPatil:
I agree that we shouldn't support "any" as an option. That is effectively turning off the linter, which they can already do via the config file.

MichaelChirico:
> That is effectively turning off the linter, which they can already do via the config file.

Well, not exactly. It would still retain the remaining [options](https://lintr.r-lib.org/reference/assignment_linter.html) -- even if we subsume all of `allow_cascading_assign`, `allow_right_assign`, `allow_pipe_assign` into one `operator=` argument, there's still the `allow_trailing=` argument that's independent of the assignment operator.

But maybe `operator=NULL` is the better way to say "any operator".

> I think top_level_operator should be called assignment_operator

I didn't like the redundancy of `assignment_linter(assignment_operator=...)`.

> Maybe it's worth exploring what = assignment style should look like for inner assignments and reduce flexibility a bit when it comes to them.

I really wanted to punt here to make progress on this linter in the next release & enable using `=` even at some basic level. The discussion to resolving that in #2521 stalled a few times over 10 months. Not requiring `<-` is relatively highly requested; beyond the existing 👍 there I know we'd use it in {data.table} and I think [{mlr} would use it too](https://github.com/mlr-org/mlr3/blob/b4c4360445faa547a246d9a973a76bb9ca21f783/.lintr#L4).

Using anything but `<-` for implicit assignments is already pretty niche, and implicit assignments are rare to begin with. It probably makes most sense to add a similar `operator=` argument there with similar behavior, but for implicit assignments. Glancing at whether it's prudent to subsume `implicit_assignment_linter()` into `assignment_linter()`, I'd say no -- it's not immediately clear how we'd cleanly include the [three existing arguments to `implicit_assignment_linter()`](https://lintr.r-lib.org/reference/implicit_assignment_linter.html) in the `assignment_linter()` API.

MichaelChirico:
> I like the sound of `operator = c("<-", "=", NA)`.
> 
> NULL can't be included in `c()` IINM.

what would NA mean? I was thinking 'NULL<->any operator' so it wouldn't need to be mixed with other strings & could be handled separately.

AshesITR:
`NA` would mean any (i.e. disable to-level checks) but it allows exposition of the argument, as well as checking via `match.arg()`. I see no benefit for using `NULL` instead - the signature would have to hide this option. `NA` to disable this part of the linter seems somewhat more intuitive than `"any"`.

MichaelChirico:
I see, basically it's to facilitate `match.arg()`, makes sense. OTOH, it's not clear how we'd implement it clearly.

 - `operator = c("<-", "=", ..., NA); match.arg(operator)` --> the current behavior. But doesn't allow `length(operator) > 1L`.
 - `operator = c("<-", "=", ..., NA); match.arg(operator, several.ok=TRUE)` --> allow `length(operator)>1L`, but doesn't match current behavior.
 - `operator = "<-"; match.arg(operator, c("<-", "=", ..., NA), several.ok=TRUE)` --> allow `length(operator)>1L`, match current behavior, _but_ the signature is no longer transparent about what operators are allowed.

I think `NULL` is still an option, then -- AFAICT there's no way to put `NA` (or `NULL`) in the signature for the behavior we want. And `NULL` cleans up the implementation a little (`is.null(operator)` vs. `length(operator) == 1L && is.na(operator)` or `identical(operator, NA)` but then `NA_character_` won't work, something I always found awkward about `testthat::expect_error(x, NA)`).

I'm not sure there's much expositional difference between stating "`NULL` means 'any operator'" vs. "`NA` [logical or any type] means 'any operator'".

MichaelChirico:
In {base}, there's most saliently `match(nomatch=NA)`, `sort(na.last=NA)`, `nchar(keepNA=NA)`, but 28 hits overall:

```r
NULL # ensure .Last.value is NULL
getNamespaceExports('base') |>
  sapply(function(fn) {
    f <- get(fn)
    if (!is.function(f)) return(FALSE)
    f |>
      formals() |>
      sapply(\(arg) is.atomic(arg) && length(arg) == 1L && is.na(arg)) |>
      any()
  }) |>
  which() |>
  length()
```

AshesITR:
Good point re the finicky behavior of `match.arg()` with `NA`. Given that, I think `"any"` is a good name for the permissive option.

MichaelChirico:
Superseded by #2711

PR #2704, opened 2025-01-08
Bump minimum backports version to ensure support for ...names()

karawoo:
lintr uses the `...names()` function via backports on R versions <4.1.0. This function was [added](https://github.com/r-lib/backports/blob/main/NEWS.md#backports-140) to backports in version 1.4.0, so older versions of backports don't work. This PR bumps up the minimum version of backports in the DESCRIPTION.


MichaelChirico:
great pull, thanks!

MichaelChirico (approved):

PR #2711, opened 2025-01-28
assignment_linter() gains operator= argument to allow, e.g. '=' assignments

### LINKED ISSUE #2441: Extend assignment_linter() to enforce '=' by default for assignment

MichaelChirico:
Supersedes #2698 and #2521. h/t again @J-Moravec for getting the ball rolling here.

Closes #2441.

This also un-couples the linted operator from the `allow_trailing` behavior; lints generated by the operator used are now independent of lints generated by line-trailing assignment operators:

```r
## current 'main'
lint("a <-
1", assignment_linter(allow_trailing=FALSE))
# <text>:1:3: style: [assignment_linter] Assignment <- should not be trailing at the end of a line.
# a <-
#   ^~
lint("a =
1", assignment_linter(allow_trailing=FALSE))
# <text>:1:3: style: [assignment_linter] Use <-, not =, for assignment.
# a =
#   ^

## current branch (assignment-operator)
lint("a <-
1", assignment_linter(allow_trailing=FALSE))
# <text>:1:3: style: [assignment_linter] Assignment <- should not be trailing at the end of a line.
# a <-
#   ^~
lint("a =
1", assignment_linter(allow_trailing=FALSE))
# <text>:1:3: style: [assignment_linter] Use one of <-, <<- for assignment, not =.
# a =
#   ^
# <text>:1:3: style: [assignment_linter] Assignment = should not be trailing at the end of a line.
# a =
#   ^
```

Having `<<-` allowed by default is back-compatible, but does strike me as odd -- should we also plan to remove this from the defaults?


IndrajeetPatil (approved):

MichaelChirico:
Thanks @IndrajeetPatil! I'd also like to get feedback from @AshesITR in this case since he commented a lot on the earlier PRs & this will be released soon afterwards.

IndrajeetPatil:
> Having <<- allowed by default is back-compatible, but does strike me as odd -- should we also plan to remove this from the defaults?

Yeah, I think it'll be good to remove it from the defaults.

R/assignment_linter.R
```diff
@@ -64,63 +72,113 @@
...
+#' )
+#'
 #' @evalRd rd_tags("assignment_linter")
 #' @seealso
 #' - [linters] for a complete list of linters available in lintr.
 #' - <https://style.tidyverse.org/syntax.html#assignment-1>
 #' - <https://style.tidyverse.org/pipes.html#assignment-2>
 #' @export
-assignment_linter <- function(allow_cascading_assign = TRUE,
+assignment_linter <- function(operator = c("<-", "<<-"),
```
AshesITR, line 86: Positional changes could affect downstream dependencies if the linter is configured using positional arguments. We should at least check for `is.logical(operator)` to stop with a helpful error message.

R/assignment_linter.R
```diff
@@ -64,63 +72,113 @@
...
-      "Avoid the assignment pipe %s; prefer using <- and %%>%% separately."
+      op_text <- xml_text(op_expr)
+      op_lint_message_fmt <- rep(
+        sprintf(
+          "Use %s for assignment, not %%s.",
+          if (length(operator) > 1L) paste("one of", toString(operator)) else operator
+        ),
+        length(op_text)
+      )
+      op_lint_message_fmt[op_text %in% c("<<-", "->>")] <-
```
AshesITR, line 161: This path will produce a confusing message for `operator = c("<-", "<<-")` when encountering `"->>"`.

R/assignment_linter.R
```diff
@@ -64,63 +72,113 @@
...
+        "Replace %s by assigning to a specific environment (with assign() or <-) to avoid hard-to-predict behavior."
+      op_lint_message_fmt[op_text == "%<>%"] <-
+        "Avoid the assignment pipe %s; prefer using <- and %%>%% separately."
```
AshesITR, line 164: This will be inconsistent advice if `operator = "="`.

tests/testthat/test-assignment_linter.R
AshesITR, line 1: For the test structure, maybe check for deprecation warnings once (maybe with smell tests that the automatic inference of `operator` works correctly) and proceed testing only the new spec with all (old) tests?

That should make full removal easier down the line.

AshesITR (commented):
Good stuff, just a few kinks that could be ironed out.

tests/testthat/test-assignment_linter.R
MichaelChirico: Thanks. Took that approach mainly out of laziness :)

Now, I duplicate the tests of now-deprecated behavior: (1) replace the old test with one using the `operator=` approach (2) copy all the tests that now throw warnings to their own `test_that()`

R/assignment_linter.R
MichaelChirico: Thanks for flagging. I don't see anyone using positional calls:

https://github.com/search?q=lang%3AR+%2F%5Cbassignment_linter%5B%28%5D%5B%5Ea%29%5D%2F+-path%3A.Rd&type=code

And I think 1st is the correct position for the new argument. Still, it can't hurt to add the safety check you mention.

R/assignment_linter.R
MichaelChirico: Good point... I think the sane approach here is to only add this clarification if neither `<<-` nor `->>` is in `operator`, PTAL.

R/assignment_linter.R
```diff
@@ -64,63 +72,116 @@
...
 #' - <https://style.tidyverse.org/pipes.html#assignment-2>
 #' @export
-assignment_linter <- function(allow_cascading_assign = TRUE,
+assignment_linter <- function(operator = c("<-", "<<-"),
+                              allow_cascading_assign = TRUE,
                               allow_right_assign = FALSE,
                               allow_trailing = TRUE,
                               allow_pipe_assign = FALSE) {
-  trailing_assign_xpath <- paste(
+  if (!missing(allow_cascading_assign) || is.logical(operator)) {
```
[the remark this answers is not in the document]
MichaelChirico: True. I think it's a pretty adversarial case though. Doubt anyone will be affected :)

tests/testthat/test-assignment_linter.R
```diff
@@ -186,9 +245,190 @@ test_that("multiple lints throw correct messages", {
...
+      list(message = "Use <- for assignment, not ->", line_number = 4L),
+      list(message = "Avoid the assignment pipe %<>%", line_number = 5L)
+    ),
+    assignment_linter(operator = "<-")
+  )
+})
+
+test_that("assignment operator can be toggled", {
+  eq_linter <- assignment_linter(operator = "=")
+  any_linter <- assignment_linter(operator = "any")
```
AshesITR, line 257: Maybe add a test that lints with this config, too?

AshesITR (commented):
Basically GTG. Nice work, thanks!

AshesITR:
> > Having <<- allowed by default is back-compatible, but does strike me as odd -- should we also plan to remove this from the defaults?
> 
> 
> 
> Yeah, I think it'll be good to remove it from the defaults.

Is there advice in the style guide that motivates this change? If so, I'd be fine with shipping another breaking change related to this linter, given that we do impact analysis first.
Otherwise I see no clear cause to change our default behavior.

MichaelChirico:
> > > Having <<- allowed by default is back-compatible, but does strike me as odd -- should we also plan to remove this from the defaults?
> > 
> > 
> > Yeah, I think it'll be good to remove it from the defaults.
> 
> Is there advice in the style guide that motivates this change? If so, I'd be fine with shipping another breaking change related to this linter, given that we do impact analysis first. Otherwise I see no clear cause to change our default behavior.

Ah, I always assumed there was such advice, but there's not. I think there should be:

https://github.com/tidyverse/style/issues/239

But you're right we shouldn't introduce a breaking change at this late hour if it doesn't contradict the style guide.

On the other hand, {lintr}'s own defaults already do discourage `<<-`: `<<-` is a default undesirable operator.

So at a minimum, we have some internal contradiction in our defaults.

I filed #2717 as a follow-up -- we should at least resolve the conflict in the next release regardless of if there's movement in the upstream style guide report.

tests/testthat/test-assignment_linter.R
MichaelChirico: Well, this specific config won't throw any lints as of now.

I added a test of `operator="any", allow_trailing=FALSE`, though.

AshesITR (approved):

PR #2715, opened 2025-02-06
Make $xml_find_function_calls() land on parent::expr instead

### LINKED ISSUE #2431: Should `xml_find_function_calls()` land on the parent expr instead?
> WDYT about `xml_find_function_calls()` landing on `//SYMBOL_FUNCTION_CALL/parent::expr`? It seems like most XPaths start there.

_Originally posted by @MichaelChirico in https://github.com/r-lib/lintr/pull/2357#discussion_r1424794241_
            
---

Alternatively, it could be an option that is `TRUE` by default.  Or with levels, `land_on = c("parent", "self", "parent::parent")`. I think that would make the issue of documentation easier.

MichaelChirico:
Closes #2431.

### DIFF
--- a/R/boolean_arithmetic_linter.R
+++ b/R/boolean_arithmetic_linter.R
@@ -36,20 +36,19 @@ boolean_arithmetic_linter <- function() {
   one_expr <- "(LT or GE) and expr[NUM_CONST[text() = '1' or text() = '1L']]"
   length_xpath <- glue("
   parent::expr
-    /parent::expr
     /parent::expr[
       expr[SYMBOL_FUNCTION_CALL[text() = 'length']]
       and parent::expr[ ({zero_expr}) or ({one_expr})]
     ]
   ")
   sum_xpath <- glue("
-  parent::expr
-    /parent::expr[
-      expr[
-        expr[SYMBOL_FUNCTION_CALL[text() = 'grepl']]
-        or (EQ or NE or GT or LT or GE or LE)
-      ] and parent::expr[ ({zero_expr}) or ({one_expr})]
+  parent::expr[
+    expr[
+      expr[SYMBOL_FUNCTION_CALL[text() = 'grepl']]
+      or (EQ or NE or GT or LT or GE or LE)
     ]
+    and parent::expr[ ({zero_expr}) or ({one_expr})]
+  ]
   ")
 
   Linter(linter_level = "expression", function(source_expression) {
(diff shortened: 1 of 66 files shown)


R/consecutive_assertion_linter.R
```diff
@@ -35,24 +35,21 @@ consecutive_assertion_linter <- function() {
...
+    not(SYMBOL_SUB[text() = 'msg'])
+    and not(following-sibling::expr[1]/SYMBOL_SUB[text() = 'msg'])
+    and expr[1]/SYMBOL_FUNCTION_CALL = {next_expr}/expr[1]/SYMBOL_FUNCTION_CALL
+  ]")
 
   Linter(linter_level = "file", function(source_expression) {
     # need the full file to also catch usages at the top level
     stopifnot_calls <- source_expression$xml_find_function_calls("stopifnot")
     assert_that_calls <- source_expression$xml_find_function_calls("assert_that")
+    # browser()
```
MichaelChirico, line 52:
```suggestion
```

R/boolean_arithmetic_linter.R
```diff
@@ -36,20 +36,19 @@ boolean_arithmetic_linter <- function() {
...
-        or (EQ or NE or GT or LT or GE or LE)
-      ] and parent::expr[ ({zero_expr}) or ({one_expr})]
+  parent::expr[
+    expr[
+      expr[SYMBOL_FUNCTION_CALL[text() = 'grepl']]
+      or (EQ or NE or GT or LT or GE or LE)
     ]
+    and parent::expr[ ({zero_expr}) or ({one_expr})]
+  ]
   ")
```
MichaelChirico, line 52:
```suggestion
  ]")
```

MichaelChirico:
@AshesITR / @IndrajeetPatil PTAL at this today if you can. I need to submit an update to CRAN by tomorrow.

If you prefer, I can just put a small patch on CRAN to meet the deadline & allow a little longer for review here.

AshesITR:
The change is okay for me on the grounds that we could remove many more `parent::expr` than we had to add `SYMBOL_FUNCTION_CALL`. 

Have you checked which of `self::expr` and `self::*` is faster? I assume it's `self::*` which would justify its usage despite being slightly less clear IMO.

AshesITR (approved):

MichaelChirico:
I'm 90% sure we've benchmarked `*` to be faster, but it's worth double-checking. Not worth pausing the PR over, though: #2718

PR #2716, opened 2025-02-06
New branch_exit_calls= argument for unnecessary_nesting_linter

### LINKED ISSUE #2325: Add exception in unnecessary_nesting_linter() for paired stop()/warning() branches
Hopefully the to-be-squashed commit 35ab376d3ec1204c02dd6da220178bb02b3b336c survives as a reference for the implementation.

The description of the issue is as follows:

```r
if (A) {
  stop("A is bad!")
} else {
  warning("!A is worrying, but not catastrophic!")
}
```

Is a common pattern that _could_ in principle be unnested:

```r
if (A) {
  stop("A is bad!")
}
warning("!A is worrying, but not catastrophic!")
```

But this format is a bit more awkward / doesn't leverage the clear `if`/`else` structure to make it clear what's happening.

We should do the following:

 1. Decide when exactly this reasoning applies. The example above is very clear, but what about if there's intervening code before/after `warning()` in the latter branch?
 2. Decide how to expose this to users. In principle there could be two parameters to this function [1] "exit" calls like `stop()`/`q()` that are _guaranteed_ to terminate a branch [2] "signal" calls that can be "companion" calls to an "exit call", e.g. `warning()`

MichaelChirico:
Closes #2325

### DIFF
--- a/R/unnecessary_nesting_linter.R
+++ b/R/unnecessary_nesting_linter.R
@@ -14,8 +14,12 @@
 #'   testthat requires a braced expression in its `code` argument. The other defaults
 #'   similarly compute on expressions in a way which is worth highlighting by
 #'   em-bracing them, even if there is only one expression, while [switch()] is allowed
-#'   for its use as a control flow analogous to `if`/`else`.
-#'
+#'   for its use as a control flow analogous to `if`/`else`.]
+#' @param branch_exit_calls Character vector of functions which are considered
+#'   as "exiting" a branch for the purpose of recommending removing nesting in a branch
+#'   _lacking_ an exit call when the other branch terminates with one. Calls which
+#'   always interrupt or quit the current call or R session,
+#'   e.g. [stop()] and [q()], are always included.
 #' @examples
 #' # will produce lints
 #' code <- "if (A) {\n  stop('A is bad!')\n} else {\n  do_good()\n}"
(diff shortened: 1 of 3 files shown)


tests/testthat/test-unnecessary_nesting_linter.R
```diff
@@ -139,9 +140,21 @@ test_that("unnecessary_nesting_linter blocks if/else with one exit branch", {
...
+    if (A) {
+      stop('An error')
+    } else {
+      warning('A warning')
+    }
+  ")
+  expect_lint(stop_warning_lines, lint_msg, linter)
+
+  # Optionally consider 'warning' as an exit call --> no lint
+  expect_lint(stop_warning_lines, NULL, unnecessary_nesting_linter(branch_exit_calls = 'warning'))
```
MichaelChirico, line 157:
```suggestion
  expect_lint(stop_warning_lines, NULL, unnecessary_nesting_linter(branch_exit_calls = "warning"))
```

R/unnecessary_nesting_linter.R
```diff
@@ -14,8 +14,12 @@
...
 #'   similarly compute on expressions in a way which is worth highlighting by
 #'   em-bracing them, even if there is only one expression, while [switch()] is allowed
-#'   for its use as a control flow analogous to `if`/`else`.
-#'
+#'   for its use as a control flow analogous to `if`/`else`.]
+#' @param branch_exit_calls Character vector of functions which are considered
+#'   as "exiting" a branch for the purpose of recommending removing nesting in a branch
+#'   _lacking_ an exit call when the other branch terminates with one. Calls which
+#'   always interrupt or quit the current call or R session,
+#'   e.g. [stop()] and [q()], are always included.
```
IndrajeetPatil, line 22: It might be a good idea to include all elements of the relevant vector (`default_branch_exit_calls`) in the docs. 

Not everyone is going to be able to recall all the functions that interrupt or quit the current call or R session.

IndrajeetPatil (approved):

R/unnecessary_nesting_linter.R
MichaelChirico: I am a bit wary of this. It's a bit fragile / easy to get out of sync with the true set of calls, and extremely easy to figure out what the defaults are by glancing at the code.

I will treat this as an implementation detail & wait for a user request here before proceeding.

PR #2719, opened 2025-02-11
Create a second r-devel devcontainer, set it as default

MichaelChirico:
Following the example of {bit64}. This will help create CRAN releases from devcontainer according to CRAN policy on testing with latest r-devel.


MichaelChirico:
Superseded by #2726

PR #2726, opened 2025-02-12
3.2.0 as submitted to CRAN

MichaelChirico:
Here are the last-minute changes I made to tidy up the submission.

Merging without review to reflect what was submitted with fidelity. That also means it subsumes #2719 so I'll close that too.


.Rbuildignore
```diff
@@ -27,7 +27,10 @@ testthat-problems[.]rds$
 ^_pkgdown\.yaml$
 ^docs$
 ^pkgdown$
-^vignettes/[^-]+.gif$
+# large files not needed for CRAN
+^man/figures/logo[.]png$
+# keep '-still.gif' compact versions
+^vignettes/.*([^l]|[^l]l|[^i]ll|[^t]ill|[^s]till)[.]gif$
```
AshesITR, line 33: Is negative lookbehind supported?
`^vignettes/.*(?<!still)[.]gif$`

MichaelChirico: It should be as the match is with `perl=TRUE`:

https://github.com/r-devel/r-svn/blob/8b6c4d6d4be6b238c1b569506672270af13b783e/src/library/tools/R/build.R#L86-L87

PR #2729, opened 2025-02-12
Remove error about incorrect ways to pass linters

MichaelChirico:
This is progressing another deprecation as stated.


tests/testthat/test-lint.R
```diff
@@ -156,14 +156,15 @@ test_that("exclusions work with custom linter names", {
 })
 
 test_that("old compatibility usage errors", {
+  err <- rex::rex("Expected `", anything, "` to be a function of class <linter>")
```
IndrajeetPatil, line 159: We typically use `error_msg` for this. Better to stick to that pattern :)

IndrajeetPatil (approved):

PR #2755, opened 2025-02-14
Lint() always has message class character, not an object

### LINKED ISSUE #2740: pipe_consistency_linter() returns lint object where message has class 'glue'
This causes some inconsistency, e.g. when `data.table::rbindlist()` attempts to combine the `as.data.frame()` results of different files, it can error because of mismatched attributes:

https://github.com/r-lib/lintr/blob/580156dd5a441f2c279953f4be62941acf33505d/R/pipe_consistency_linter.R#L55

MichaelChirico:
Closes #2740

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -21,7 +21,7 @@ linters: all_linters(
       c(
         message = "use cli::cli_inform()",
         warning = "use cli::cli_warn()",
-        stop = "use cli::cli_abort()",
+        stop = "use cli::cli_abort()"
       ),
       symbol_is_undesirable = FALSE
     ),
--- a/.lintr
+++ b/.lintr
@@ -37,6 +37,8 @@ linters: all_linters(
     # We now require R>=4.0.0
     strings_as_factors_linter = NULL,
     todo_comment_linter = NULL,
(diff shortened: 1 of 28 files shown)


.lintr
```diff
@@ -11,15 +11,20 @@ linters: all_linters(
...
-      message = "use cli::cli_inform()",
-      warning = "use cli::cli_warn()",
-      stop = "use cli::cli_abort()",
       normalizePath = "use normalize_path()"
     )),
+    undesirable_function_call_linter = undesirable_function_linter(
+      c(
+        message = "use cli::cli_inform()",
+        warning = "use cli::cli_warn()",
+        stop = "use cli::cli_abort()",
```
MichaelChirico, line 24:
```suggestion
        stop = "use cli::cli_abort()"
```

AshesITR:
Why not gracefully `unclass()` length-1 strings?

MichaelChirico:
> Why not gracefully `unclass()` length-1 strings?

my thinking is it's good discipline to avoid using glue inside the lint() loop

AshesITR (approved):

Issue #2737, opened 2025-02-12
Clean-up: use expect_no_lint() where appropriate


MichaelChirico:
`expect_no_lint()` is a nice alternative to `expect_lint(., NULL, .)`. We should use it where possible.

This will be a relatively large lift: there are about 1,500 call sites to `expect_lint()` with `NULL`:

```r
linter = make_linter_from_xpath("//SYMBOL_FUNCTION_CALL[text() = 'expect_lint']/parent::expr/following-sibling::expr/NULL_CONST", "xxx")
lints = lint_package(linters = linter())
length(lints)
# [1] 1536
```

We could also introduce a linter for this, but I reckon it's too localized to warrant a new export. We can add it as a custom local linter to our suite I reckon.

AshesITR:
Maybe we can do a staged effort for this again by applying on the lint-changed-files workflow only until it's resolved?

PR #2761, opened 2025-02-18
Lint calls to undesirable operators in prefix form

### LINKED ISSUE #1910: undesirable_operator_linter could optionally lint in functional form
Consider:

```r
lintr::lint("`%%`(x, y)\n", lintr::undesirable_operator_linter(c("%%" = NA)))
```

This doesn't throw a lint because `%%` is a call, not an operator. But it seems reasonable to expect this code to lint (perhaps optionally), e.g. if I'm a package developer trying to deprecate my old `%old-grapes%` in favor of `%new-grapes%`. Then any usage of the old function should be pointed out if I'm offering a tool for my downstreams to self-diagnose their exposure.

Of course, we could use `undesirable_function_linter(c("%%" = NA))` (well, actually it seems there's a bug there), but users needn't think of that.

MichaelChirico:
Closes #1910. Progress on #2737.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,7 @@
 
 * `brace_linter()`' has a new argument `function_bodies` (default `"multi_line"`) which controls when to require function bodies to be wrapped in curly braces, with the options `"always"`, `"multi_line"` (only require curly braces when a function body spans multiple lines), `"not_inline"` (only require curly braces when a function body starts on a new line) and `"never"` (#1807, #2240, @salim-b).
 * `seq_linter()` recommends using `seq_along(x)` instead of `seq_len(length(x))` (#2577, @MichaelChirico).
-* `undesirable_operator_linter()` lints operators in functional form, e.g. `` `%%`(x, 2)`` (#1910, @MichaelChirico). Disable this by setting `call_is_undesirable=FALSE`.
+* `undesirable_operator_linter()` lints operators in prefix form, e.g. `` `%%`(x, 2)`` (#1910, @MichaelChirico). Disable this by setting `call_is_undesirable=FALSE`.
 
 ### Lint accuracy fixes: removing false positives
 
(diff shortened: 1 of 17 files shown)


R/undesirable_operator_linter.R
```diff
@@ -8,6 +8,8 @@
 #'   If `NA`, no additional information is given in the lint message. Defaults to
 #'   [default_undesirable_operators]. To make small customizations to this list,
 #'   use [modify_defaults()].
+#' @param call_is_undesirable Logical, default `TRUE`. Should function-style usage of
+#'   the provided operators also generate a lint?
```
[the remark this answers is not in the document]
MichaelChirico: I'd prefer to use a more canonical source for terminology if possible. I didn't see anything definitive in the R manuals, _but_ this does use "prefix notation", and it's indeed a more general programming term, so I think it's OK:

https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Direct-manipulation-of-language-objects
https://en.wikipedia.org/wiki/Polish_notation

(I think "Polish notation" is too niche a description)

NEWS.md
```diff
@@ -19,6 +19,7 @@
 
 * `brace_linter()`' has a new argument `function_bodies` (default `"multi_line"`) which controls when to require function bodies to be wrapped in curly braces, with the options `"always"`, `"multi_line"` (only require curly braces when a function body spans multiple lines), `"not_inline"` (only require curly braces when a function body starts on a new line) and `"never"` (#1807, #2240, @salim-b).
 * `seq_linter()` recommends using `seq_along(x)` instead of `seq_len(length(x))` (#2577, @MichaelChirico).
+* `undesirable_operator_linter()` lints operators in functional form, e.g. `` `%%`(x, 2)`` (#1910, @MichaelChirico). Disable this by setting `call_is_undesirable=FALSE`.
```
IndrajeetPatil, line 22: The changelog entry should also use the same phrasing: "prefix-style usage"

MichaelChirico, line 22:
```suggestion
* `undesirable_operator_linter()` lints operators in prefix form, e.g. `` `%%`(x, 2)`` (#1910, @MichaelChirico). Disable this by setting `call_is_undesirable=FALSE`.
```

IndrajeetPatil (approved):

PR #2762, opened 2025-02-19
Correct documentation of show_progress (no longer utils::txtProgressBar)

MichaelChirico:


AshesITR (approved):

R/lint.R
```diff
@@ -121,9 +121,9 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
 #' @param exclusions exclusions for [exclude()], relative to the package path.
 #' @param pattern pattern for files, by default it will take files with any of the extensions
 #' .R, .Rmd, .qmd, .Rnw, .Rhtml, .Rrst, .Rtex, .Rtxt allowing for lowercase r (.r, ...).
-#' @param show_progress Logical controlling whether to show linting progress with a simple text
-#'   progress bar _via_ [utils::txtProgressBar()]. The default behavior is to show progress in
-#'   [interactive()] sessions not running a testthat suite.
+#' @param show_progress Logical controlling whether to show linting progress with
+#'   [cli::cli_progress_along()]. The default behavior is to show progress in [interactive()] sessions
```
IndrajeetPatil, line 125: My 2 cents: Including such implementation details in user-facing documentation is not a good idea. The users care about showing or displaying the progress bar, not how we decided to implement it.

MichaelChirico: It's a fair point, and this PR proves it's easy to get out of sync with the implementation.

OTOH, it _could_ wind up in the API if we want to expose some flexibility to the display.

I'm also not familiar with {cli}, but there could be some `options()` that affect the display -- knowing which {cli} function is used can tell them which options are relevant.

So I'm a bit indifferent -- if you're motivated to change, I'll approve the PR.

PR #2773, opened 2025-02-21
Reference 'expression', not 'function', in metadata

### LINKED ISSUE #2680: cyclocomp_linter() message should be customized to 'function'/'expression'
`cyclocomp()` can measure complexity of functions _or_ expressions, but the message always says "function":

https://github.com/r-lib/lintr/blob/5bc4dbe4396e558725542cdca7dc9599a2dd1d2b/R/cyclocomp_linter.R#L48

This might be confusing if the lint hits a non-function expression:

```r
lint(linters = cyclocomp_linter(5L), "
for (i in 1:10)
  for (j in 1:10)
    for (k in 1:10)
      i*j*k
")
# <text>:2:1: style: [cyclocomp_linter] Reduce the cyclomatic complexity of this function from 10 to at most 5.
# for (i in 1:10)
# ^
```

MichaelChirico:
Closes #2680. Progress on #2737 

Also move the check outside the actual lint loop; better to fail up-front if the linter is not usable.

### DIFF
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -4,9 +4,10 @@ on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master]
 
-name: test-coverage
+name: test-coverage.yaml
+
+permissions: read-all
 
 jobs:
   test-coverage:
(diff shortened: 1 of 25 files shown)


AshesITR (dismissed):

MichaelChirico:
Argh, the change to put `requireNamespace()` in the factory is much gnarlier than anticipated.

I'll file a follow-up & revert all but the tiny change here.

AshesITR (approved):

PR #2781, opened 2025-02-24
Emphasize that config paths are relative to the config itself

### LINKED ISSUE #2780: Exclusions must be given relative to the config file, not the package -- intended?
```r
options(lintr.linter_file = tmp<-tempfile())
with_tempdir({
  cat(sprintf("Linting from %s\n", getwd()))

  writeLines('a=1', 'a.R')
  writeLines('a=1', 'b.R')

  cat("First pass:\n")
  writeLines('exclusions: list("a.R")', tmp)
  print(lint_dir())

  cat("Second pass:\n")
  A.R <- normalizePath("a.R")
  writeLines(paste0('exclusions: list("', A.R, '")'), tmp)
  print(lint_dir())
})
```

Has output:

```
Linting from /tmp/RtmpRm034g/fileac1f1d3f2a83
First pass:
a.R:1:2: style: [assignment_linter] Use one of <-, <<- for assignment, not =.
a=1
 ^
a.R:1:2: style: [infix_spaces_linter] Put spaces around all infix operators.
a=1
 ^
b.R:1:2: style: [assignment_linter] Use one of <-, <<- for assignment, not =.
a=1
 ^
b.R:1:2: style: [infix_spaces_linter] Put spaces around all infix operators.
a=1
 ^
Second pass:
b.R:1:2: style: [assignment_linter] Use one of <-, <<- for assignment, not =.
a=1
 ^
b.R:1:2: style: [infix_spaces_linter] Put spaces around all infix operators.
a=1
 ^
```

That is, the path relative to the directory where `lint_dir()` is run, 'a.R', is not picked up as being excluded in the first path, hence we see lints on a.R there.

By contrast, when we give the _full_ absolute path to a.R in the second pass, it's correctly excluded.

Is this intended behavior? If so I think it needs to be more clearly documented.

It would certainly be clearer in my use case to keep using paths relative to the directory where `lint()` is to be run:

 - {data.table} keeps .ci/.lintr.R out of the top-level directory to keep that a bit tidier
 - but paths in the config must use `../` to apply correctly: https://github.com/Rdatatable/data.table/blob/f78044ebc63cc70a7ef67607b66fc8158766be86/.ci/.lintr.R#L114

MichaelChirico:
Closes #2780

### DIFF
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -29,7 +29,9 @@
 #'   "   For convenience, a vector can be used in place of a list whenever it would not introduce ambiguity, e.g.",
 #'   "   a character vector of files to exclude or a vector of lines to exclude.",
 #'   "",
-#'   "   Note also that all paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   Note also that:",
+#'   "   1. All paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   2. All paths are taken as relative _to the config file itself_, not the path where `lint()` is run.",
 #'   NULL
 #' )
 #'
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -31,7 +31,8 @@
 #'   "",
 #'   "   Note also that:",
 #'   "   1. All paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
-#'   "   2. All paths are taken as relative _to the config file itself_, not the path where `lint()` is run.",
+#'   "   2. For `exclusions` provided _via_ a config file, all paths are taken as relative _to the config file itself_,",
+#'   "      not the path where `lint()` is run.",
 #'   NULL
 #' )
 #'
--- a/R/exclude.R
+++ b/R/exclude.R
@@ -31,8 +31,8 @@
 #'   "",
 #'   "   Note also that:",
 #'   "   1. All paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
-#'   "   2. For `exclusions` provided _via_ a config file, all paths are taken as relative _to the config file itself_,",
-#'   "      not the path where `lint()` is run.",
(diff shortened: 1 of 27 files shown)


R/exclude.R
```diff
@@ -29,7 +29,9 @@
 #'   "   For convenience, a vector can be used in place of a list whenever it would not introduce ambiguity, e.g.",
 #'   "   a character vector of files to exclude or a vector of lines to exclude.",
 #'   "",
-#'   "   Note also that all paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   Note also that:",
+#'   "   1. All paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   2. All paths are taken as relative _to the config file itself_, not the path where `lint()` is run.",
```
AshesITR, line 34: This doesn't make sense for exclude passed as an argument. Maybe "All paths specified in config files are taken as relative to ..."?

MichaelChirico: Ah, subtle.

MichaelChirico, line 34:
```suggestion
#'   "   2. All paths specified in config files are taken as relative _to the config file itself_, not the path where `lint()` is run.",
```

R/exclude.R
```diff
@@ -29,7 +29,10 @@
 #'   "   For convenience, a vector can be used in place of a list whenever it would not introduce ambiguity, e.g.",
 #'   "   a character vector of files to exclude or a vector of lines to exclude.",
 #'   "",
-#'   "   Note also that all paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   Note also that:",
+#'   "   1. All paths are interpreted as globs ([Sys.glob()]), so that e.g. `*` does pattern expansion.",
+#'   "   2. For `exclusions` provided _via_ config, all paths are taken as relative _to the config file itself_,",
```
AshesITR, line 34:
```suggestion
#'   "   2. For `exclusions` provided _via_ a config file, all paths are taken as relative _to the config file itself_,",
```

AshesITR (approved):

PR #2785, opened 2025-02-24
Move the cyclocomp namespace check into the factory

### LINKED ISSUE #2777: Move requireNamespace() check for cyclocomp_linter() to the factory?
Currently, the validity check for `cyclocomp_linter()` happens inside the lint loop:

https://github.com/r-lib/lintr/blob/e673ee34a4f750130b2c603cff6b81d5aa6bd48f/R/cyclocomp_linter.R#L25-L33

I think it makes more sense to do so outside the loop, but simply moving it breaks many things in our test suite when {cyclocomp} is not available.

Here's one test in particular that gives me pause:

https://github.com/r-lib/lintr/blob/e673ee34a4f750130b2c603cff6b81d5aa6bd48f/tests/testthat/test-linter_tags.R#L55-L74

Are we OK with just doing `skip_if_not_installed('cyclocomp')` here? Should we keep the `requireNamespace()` check where it is?

MichaelChirico:
Closes #2777

Two notable changes accompany this:

 - Rather than _error_ up-front if {cyclocomp} is unavailable, we _warn_ & return a "dumb" linter that just always & immediately finds no lints.
 - `linters_with_tags()` (_via_ `call_linter_factory()`) _warns_ instead of _errors_ when a linter fails to build. Again we return the "dumb" linter in this case. This is not strictly needed thanks to only warning from the factory, but I think it's a better design anyway.

Mostly in tests I've opted anyway for `skip()`s, rather than an awkward construct where we optionally `expect_warning()`, or swap between `expect_warning()`/`expect_no_warning()`, or ...

This exercise also flushed out some more tests that are implicitly reading our main project .lintr file, fixing which increases their hermeticity.

### DIFF
--- a/R/cyclocomp_linter.R
+++ b/R/cyclocomp_linter.R
@@ -28,7 +28,7 @@ cyclocomp_linter <- function(complexity_limit = 15L) {
       "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}.",
       i = "Please install the needed {.pkg cyclocomp} package."
     ))
-    return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of {.pkg cyclocomp} package")))
+    return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of the {.pkg cyclocomp} package")))
   }
   # nocov end
   Linter(linter_level = "expression", function(source_expression) {
--- a/R/cyclocomp_linter.R
+++ b/R/cyclocomp_linter.R
@@ -24,10 +24,11 @@
 cyclocomp_linter <- function(complexity_limit = 15L) {
   # nocov start
   if (!requireNamespace("cyclocomp", quietly = TRUE)) {
-    cli::cli_abort(c(
-      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}.",
+    cli::cli_warn(c(
+      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}. Returning a null linter.",
       i = "Please install the needed {.pkg cyclocomp} package."
     ))
+    return(Linter(function(.) NULL))
   }
   # nocov end
   Linter(linter_level = "expression", function(source_expression) {
--- a/R/cyclocomp_linter.R
+++ b/R/cyclocomp_linter.R
@@ -25,7 +25,7 @@ cyclocomp_linter <- function(complexity_limit = 15L) {
   # nocov start
   if (!requireNamespace("cyclocomp", quietly = TRUE)) {
     cli::cli_warn(c(
-      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}. Returning a null linter.",
+      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}.",
       i = "Please install the needed {.pkg cyclocomp} package."
     ))
     return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of {.pkg cyclocomp} package")))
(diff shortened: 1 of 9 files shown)


AshesITR:
Warnings are easy to overlook - not the biggest fan of this approach.
Maybe warn in the constructor but still error out in the implementation?

R/cyclocomp_linter.R
```diff
@@ -22,16 +22,16 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 cyclocomp_linter <- function(complexity_limit = 15L) {
+  # nocov start
+  if (!requireNamespace("cyclocomp", quietly = TRUE)) {
+    cli::cli_warn(c(
+      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}. Returning a null linter.",
```
AshesITR, line 28:
```suggestion
      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}.",
```

R/cyclocomp_linter.R
```diff
@@ -22,16 +22,16 @@
...
+      i = "Please install the needed {.pkg cyclocomp} package."
+    ))
+    return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of {.pkg cyclocomp} package")))
```
AshesITR, line 31:
```suggestion
```

Removing this would allow for this interaction:

```r
linters <- all_linters()
# warning
install.packages("cyclocomp")
all_lints <- lint_dir(linters = linters)
```

WDYT?

MichaelChirico: Maybe `if (interactive())`? In CI, which I guess is where the plurality of `lint*` execution happens, we'll just get ignored warnings, right? User might not even recognize there's an issue?

AshesITR: Without this early return, CI would error out on the linter running, no?
After all, it would call `cyclocomp::cyclocomp()` without checking.

MichaelChirico: Can't check now, I'm not sure if `try_silently()` will swallow that. Anyway the error from the "broken linter" here seems clearer + friendlier.

AshesITR: That's true. Given `lint()` re-runs the linter creation code by default, maybe my approach is over-engineered.
Let's stick with the hard-coded broken error message and see if someone complains about it.

R/with.R
```diff
@@ -187,10 +187,11 @@ call_linter_factory <- function(linter_factory, linter_name, package) {
   linter <- tryCatch(
     linter_factory(),
     error = function(e) {
-      cli_abort(
```
AshesITR, line 190: Revert this?

MichaelChirico: Maybe we should return a "disabled" linter here too? Probably I'm overthinking it. Reverting for now.

R/with.R
```diff
@@ -187,10 +187,11 @@ call_linter_factory <- function(linter_factory, linter_name, package) {
...
-        "Could not create linter with {.fun {package}::{linter_name}}.",
+      cli_warn(
+        "Could not create linter with {.fun {package}::{linter_name}}. Returning a null linter.",
         parent = e
       )
+      Linter(function(.) list())
```
MichaelChirico, line 194:
```suggestion
      cli_abort(
        "Could not create linter with {.fun {package}::{linter_name}}.",
        parent = e
      )
```

R/cyclocomp_linter.R
```diff
@@ -22,16 +22,16 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 cyclocomp_linter <- function(complexity_limit = 15L) {
+  # nocov start
+  if (!requireNamespace("cyclocomp", quietly = TRUE)) {
+    cli::cli_warn(c(
+      "Cyclocomp complexity is computed using {.fn cyclocomp::cyclocomp}.",
+      i = "Please install the needed {.pkg cyclocomp} package."
+    ))
+    return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of {.pkg cyclocomp} package")))
```
AshesITR, line 31:
```suggestion
    return(Linter(function(.) cli_abort("cyclocomp_linter is disabled due to lack of the {.pkg cyclocomp} package")))
```

AshesITR (approved):

PR #2788, opened 2025-02-24
Teach object_usage_linter() about .env pronoun

### LINKED ISSUE #1472: Dplyr: not detecting variables use with .env in mutate
Hi!, well, this is weird, seems lintr is not detecting when var are used inside mutate with ```.env```, I have others ```.env``` in filter and works fine, the code is from examples of how to use roads networks in R.

```
  source_nodes <- nodes %>%
    dplyr::filter(.data$start_end == "start") %>%
    dplyr::pull(.data$nodeID)

  target_nodes <- nodes %>%
    dplyr::filter(.data$start_end == "end") %>%
    dplyr::pull(.data$nodeID)

  edges <- edges %>%
    dplyr::mutate(from = .env$source_nodes, to = .env$target_nodes)
```

Logs:
```
warning: [object_usage_linter] local variable ‘source_nodes’ assigned but may not be used
  source_nodes <- nodes %>%
  ^~~~~~~~~~~~
warning: [object_usage_linter] local variable ‘target_nodes’ assigned but may not be used
  target_nodes <- nodes %>%
```

Thx!

MichaelChirico:
Closes #1472.

Incomplete for now -- checking feedback about the proposed approach.

The key is we might generate false negatives whenever `.env` is used by non-tidyverse authors, which I'm attempting to mitigate by exposing `"rlang"` to `interpret_extensions` in the same way as we expose `"glue"`.

AFAICT the alternative would be to dynamically check if {rlang} is loaded, which seems messier. My initial thought was to do something like "check we're in a `mutate()` call", which may be the most common situation, but as this is an {rlang} export the number of functions in which this could be referenced is ~unlimited.

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -301,9 +301,13 @@ extract_env_symbols <- function(fun_assignment) {
     "
     .//SYMBOL[text() = '.env']
       /parent::expr
-      /parent::expr[OP-DOLLAR]
-      /SYMBOL
+      /following-sibling::OP-DOLLAR
+      /following-sibling::SYMBOL
+    | .//SYMBOL[text() = '.env']
+      /parent::expr
+      /following-sibling::LBB
+      /following-sibling::expr[STR_CONST]
     "
   )
-  xml_text(env_names)
+  get_r_string(env_names)
 }
--- a/tests/testthat/test-object_usage_linter.R
+++ b/tests/testthat/test-object_usage_linter.R
@@ -874,7 +874,7 @@ test_that("dplyr's .env-specified objects are marked as 'used'", {
         df %>%
           dplyr::mutate(
             from = rlang::.env$source,
-            to = rlang::.env$target
+            to = rlang::.env[['target']]
           )
       }
     "),


AshesITR:
Re configuration interface I'd think about something like `interpret_extensions = c("glue", "rlang")` for future API stability. Think about `cli` (~glue) or `data.table` integrations in the future.

MichaelChirico:
OK SG. I've updated to initiate a deprecation cycle of `interpret_glue` too.

R/object_usage_linter.R
```diff
@@ -29,7 +35,23 @@
...
+  if (!is.null(interpret_glue)) {
+    lintr_deprecated(
+      "interpret_glue",
+      '"glue" in interpret_extensions',
+      version = "3.3.0",
+      type = "Argument",
+      signal = "warning"
+    )
+
+    interpret_extensions <- (if (interpret_glue) union else setdiff)(interpret_extensions, "glue")
```
AshesITR, line 48: That's golfy ^^
Since `TRUE` is a default here, I think we can get away with

```suggestion
    if (!interpret_glue) {
      interpret_extensions <- setdiff(interpret_extensions, "glue")
    }
```

WDYT?

MichaelChirico: That doesn't cover `object_usage_linter(interpret_extensions = "rlang")` right? We'd have to ignore the previous default for `interpret_glue`.

That may be fine, since using `interpret_extensions=` _should_ imply that you've already adapted to the new argument. But the symmetry I opted for is easier to reason about. We can just "uncrack" the conciseness a bit.

AshesITR: Doesn't the current approach also force setting `interpret_glue = FALSE`?

`if (!mssing(interpret_glue) && !interpret_glue) ...`
would be a working condition, right?

MichaelChirico: PTAL. I hear you, but the resulting code is IMO harder to read, I think right now it's very transparent.

NEWS.md
```diff
@@ -10,6 +10,7 @@
    + `linter=` argument of `Lint()`.
    + `with_defaults()`.
    + Linters `closed_curly_linter()`, `open_curly_linter()`, `paren_brace_linter()`, and `semicolon_terminator_linter()`.
+* Argument `interpret_glue` to `object_usage_linter()` is deprecated in favor of the more generalized `interpret_extensions`, in which `"glue"` is present by default (#1472, @MichaelChirico). See the description below.
```
AshesITR, line 13:
```suggestion
* Argument `interpret_glue` to `object_usage_linter()` is deprecated in favor of the more general `interpret_extensions`, in which `"glue"` is present by default (#1472, @MichaelChirico). See the description below.
```

NEWS.md
```diff
@@ -23,6 +24,7 @@
 * `indentation_linter()` handles `for` un-braced for loops correctly (#2564, @MichaelChirico).
 * Setting `exclusions` supports globs like `knitr*` to exclude files/directories with a pattern (#1554, @MichaelChirico).
 * `object_name_linter()` and `object_length_linter()` apply to objects assigned with `assign()` or generics created with `setGeneric()` (#1665, @MichaelChirico).
+* `object_usage_linter()` gains argument `interpret_extensions` to govern which false positive-prone common syntaxes should be checked for used objects (#1472, @MichaelChirico). Currently `"glue"` (renamed from earlier argument `interpret_glue`) and `"rlang"` are supported. The latter newly covers usage of the `.env` pronoun like `.env$key`, where `key` was previously missed as being a used variable.
```
AshesITR, line 27:
```suggestion
* `object_usage_linter()` gains argument `interpret_extensions` to govern which false positive-prone common syntax should be checked for used objects (#1472, @MichaelChirico). Currently `"glue"` (renamed from earlier argument `interpret_glue`) and `"rlang"` are supported. The latter newly covers usage of the `.env` pronoun like `.env$key`, where `key` was previously missed as being a used variable.
```

R/object_usage_linter.R
```diff
@@ -261,3 +287,23 @@ get_imported_symbols <- function(xml) {
...
+  ))
+}
+
+extract_env_symbols <- function(fun_assignment) {
+  env_names <- xml_find_all(
+    fun_assignment,
+    "
+    .//SYMBOL[text() = '.env']
+      /parent::expr
+      /parent::expr[OP-DOLLAR]
```
AshesITR, line 304: Do you feel it's worth supporting LBB as well?

NEWS.md
MichaelChirico: I think singular here would imply only one element is supported, so "syntaxes" is right. Even "syntax(es)" is not ideal IMO since the default already has >1.

R/object_usage_linter.R
MichaelChirico: https://github.com/search?q=lang%3AR+%2F%5B%5Ea-z0-9._%5D%5B.%5Denv%5C%5B%5C%5B%5B%22%27%5D%2F&type=code

Yea, there's enough hits there & I don't think it's much harder. Will do.

AshesITR (approved):

PR #2792, opened 2025-02-25
Capture parser warnings as lints

### LINKED ISSUE #2065: Convert parser warnings to lints?
We already convert parser errors to lints (`type = "error"`), but what about parser warnings?

I came across such a warning when trying to lint `r-devel`, it came from a test file where the lint was intentional:

```r
lintr::lint("tests/simple-true.R", linters = list())
# Warning messages:
# 1: In parse(text = source_expression$content, srcfile = source_expression,  :
#   non-integer value 1e-3L qualified with L; using numeric value
# 2: In parse(text = source_expression$content, srcfile = source_expression,  :
#   integer literal 1.L contains unnecessary decimal point
# 3: In parse(text = source_expression$content, srcfile = source_expression,  :
#   non-integer value 1e-3L qualified with L; using numeric value
# 4: In parse(text = source_expression$content, srcfile = source_expression,  :
#   integer literal 1.L contains unnecessary decimal point
```

Should those be returned as lints (`type = "warning"`) instead?

MichaelChirico:
Closes #2065


R/get_source_expressions.R
```diff
@@ -157,6 +155,56 @@ lint_parse_error <- function(e, source_expression) {
...
+#' @noRd
+lint_parse_warnings <- function(w, parsed_content, source_expression) {
+  if (!length(w)) {
+    return(w)
+  }
+  flatten_lints(lapply(unique(w), lint_parse_warning, parsed_content, source_expression))
+}
+
+#' The set of parser warnings seems pretty stable, but as
+#'   long as they don't generate sourceref hints, we're
```
AshesITR, line 171: Is there an upstream issue to track this?

R/get_source_expressions.R
```diff
@@ -495,13 +543,20 @@ get_single_source_expression <- function(loc,
 get_source_expression <- function(source_expression, error = identity) {
   parse_error <- FALSE
 
-  parsed_content <- tryCatch(
-    parse(
-      text = source_expression$content,
-      srcfile = source_expression,
-      keep.source = TRUE
+  env <- parent.frame() # nolint: object_usage_linter. Used below.
+  parsed_content <- withCallingHandlers(
```
AshesITR, line 547: Any reason to nest the handlers?

MichaelChirico: There's no other way? `withCallingHandlers()` doesn't work with error conditions AFAIK...

https://github.com/search?q=lang%3AR+%2FwithCallingHandlers%5C%28%2F+%2Ferror%5Cs*%3D%2F+-%22tryCatch%22&type=code

(maybe I'm misreading)

AshesITR: Can't `tryCatch()` handle warnings?

R/get_source_expressions.R
AshesITR: Thanks. Should add to the comment for future upgrades.

R/get_source_expressions.R
MichaelChirico: Yes & no, the problem is we don't get the restart mechanism in `tryCatch()`, so parsing would effectively terminate. The current approach still returns intact parse data & can do further linting.

https://adv-r.hadley.nz/conditions.html

Actually that's a good idea for a further test -- "regular" lints are still caught under parser warning(s).

AshesITR: Thanks for elaborating. The rationale (need invokeRestart(), tryCatch doesn't provide it) could be summarized in a comment.

MichaelChirico: I'm a bit torn, obv comments are rarely harmful, but OTOH this is the canonical way to do "catch errors & all warnings" in R, though it may not be common (and it's something that still regularly trips me up).

We already use this twice in the codebase:

https://github.com/r-lib/lintr/blob/e01791f4e02195fc6ab39b9ff785de5a288b74db/R/settings.R#L125-L142
https://github.com/r-lib/lintr/blob/e01791f4e02195fc6ab39b9ff785de5a288b74db/R/settings.R#L154-L166

And here's some other packages' usage:

https://github.com/Rdatatable/data.table/blob/4f84d3c72a08ecdd9965b3cb8e493cabfd80492b/R/test.data.table.R#L431-L437
https://github.com/r-lib/testthat/blob/30f5b119875852005f2a02e6adf96276dbd1ef26/R/test-that.R#L188-L206
https://github.com/r-lib/testthat/blob/30f5b119875852005f2a02e6adf96276dbd1ef26/R/try-again.R#L18-L36

WDYT?

AshesITR: The fact that both of us trip on this sometimes suggests to me that a comment is helpful 😊

MichaelChirico: OK I'm not sure how much detail is appropriate. Just linked the Advanced R chapter.

AshesITR (approved):

PR #2795, opened 2025-02-26
Catch anyDuplicated() equivalents from dplyr, data.table

### LINKED ISSUE #2481: Extend length(unique()) logic in any_duplicated_linter() for dplyr,data.table syntaxes
Examples:

```r
DT[, length(unique(col)) == .N]
DT[, uniqueN(col) == .N
DT %>% filter(length(unique(col)) == n())
```

MichaelChirico:
Closes #2481. Progress on #2737.

There is some further refinement that could be done here, but I think this is progress enough for now.

### DIFF
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -161,7 +161,13 @@ any_duplicated_linter <- function() {
     distinct_lint_message_fmt[!is.na(xml_find_first(distinct_expr, uses_dplyr_xpath))] <-
       "anyDuplicated(x) == 0L is better than %s(x) == n()."
 
-    distinct_call <- xml_find_first(distinct_expr, distinct_call_xpath)
+    distinct_lint_message <- sprintf(distinct_lint_message_fmt, xp_call_name(distinct_expr))
+    distinct_lints <- xml_nodes_to_lints(
+      distinct_expr,
+      source_expression = source_expression,
+      lint_message = distinct_lint_message,
+      type = "warning"
+    )
 
     c(any_duplicated_lints, length_unique_lints, distinct_lints)
   })
--- a/R/any_duplicated_linter.R
+++ b/R/any_duplicated_linter.R
@@ -35,7 +35,7 @@
 #' @export
 any_duplicated_linter <- function() {
   any_duplicated_xpath <- "
-  following-sibling::expr[1][expr[1][SYMBOL_FUNCTION_CALL[text() = 'duplicated']]]
+  following-sibling::expr[1][expr[1]/SYMBOL_FUNCTION_CALL[text() = 'duplicated']]
     /parent::expr[
       count(expr) = 2
       or (count(expr) = 3 and SYMBOL_SUB[text() = 'na.rm'])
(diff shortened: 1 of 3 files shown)


MichaelChirico:
Bumping for review :)

MichaelChirico:
Adding @Bisaloo, how's your bandwidth for doing some reviews? :)

R/any_duplicated_linter.R
```diff
@@ -56,32 +56,70 @@ any_duplicated_linter <- function() {
   //{ c('EQ', 'NE', 'GT', 'LT') }
```
Bisaloo, line 66: Not directly caused by this PR but shouldn't this XPath land on the parent expr?

``` r
lintr::lint(
    text = "x[, length(unique(col)) == .N]",
    linters = lintr::any_duplicated_linter()
)
#> <text>:1:5: warning: [any_duplicated_linter] anyDuplicated(x) == 0L is better than length(unique(x)) == .N
#> x[, length(unique(col)) == .N]
#>     ^~~~~~~~~~~~~~~~~~~
```

<sup>Created on 2025-06-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/any_duplicated_linter.R
```diff
@@ -35,7 +35,7 @@
 #' @export
 any_duplicated_linter <- function() {
```
Bisaloo, line 46: Should the docs and examples be expanded as well?

Bisaloo (commented):
The logic seems sound and the tests cover the potential cases. I couldn't test it in the wild because I didn't find any CRAN packages using these patterns.

R/any_duplicated_linter.R
MichaelChirico: Indeed: #2860

MichaelChirico:
> I didn't find any CRAN packages using these patterns.

Well, I found a similar pattern that's _not_ caught here 😅

https://github.com/cran/dedupewider/blob/dec0899cc557229a0c1c63b7eeb80fb106cce26d/R/na_move.R#L100

R/any_duplicated_linter.R
MichaelChirico: Sure, good idea. I actually am not sure what a "canonical" (non-contrived) example using `n_distinct()` would look like TBH, so omitted for now.

Bisaloo:
To be more explicit, I'd recommend to expand

https://github.com/r-lib/lintr/blob/7e04393d67e19082d8461aef868545a7758581c6/R/any_duplicated_linter.R#L7-L8

as:

```r
 #' Also match usage like `length(unique(x$col)) == nrow(x)`, which can 
 #'   be replaced by `anyDuplicated(x$col) == 0L`, as well as `n_distinct(col) == n()`
 #'   and `uniqueN(col) == .N`, which can be replaced by `anyDuplicated(col) == 0L`.
```

MichaelChirico:
I think "like" is enough here. Actually, I _wouldn't_ recommend replacing `uniqueN(x$col) == .N` with `anyDuplicated(x$col)`, instead I'd recommend `anyDuplicated(x, by="col")`. There might be something similar in tidyverse to replace `n_distinct()`. And there might be other extensions to logic for other common equivalent patterns in the future -- rather than listing them all out (and then needing to maintain that list), I think the documentation can be generic and say patterns "like" `length(unique(x))` and cover all frameworks. Then the lint message can handle the business of being more specific.

PR #2799, opened 2025-02-27
Ignore multi-string contents optionally in line_length_linter

### LINKED ISSUE #856: Turn off line length linter within string literals (optionally?)
We've got a very common scenario where string literals are used to contain big SQL queries.

Crucially, the style guide for SQL & R differ in that the former's max line width is 100, the latter 80.

Moreover, to skip only one such line we've got to somewhat awkwardly inject the nolint tag inside the SQL:

```
query <- "
  SELECT SUM(CASE WHEN some_column > 5 THEN 5 ELSE some_column END) AS output_column -- # nolint
"
```

It seems to me the best solution here is to simply ignore string literal "bodies", perhaps with an option.

Happy to implement this if others agree.

MichaelChirico:
Closes #856. Progress on #2737

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,7 @@
 * `indentation_linter()` handles `for` un-braced for loops correctly (#2564, @MichaelChirico).
 * Setting `exclusions` supports globs like `knitr*` to exclude files/directories with a pattern (#1554, @MichaelChirico).
 * `object_name_linter()` and `object_length_linter()` apply to objects assigned with `assign()` or generics created with `setGeneric()` (#1665, @MichaelChirico).
+* `line_length_linter()` has a new argument `ignore_string_bodies` (defaulting to `FALSE`) which governs whether the contents of multi-line string bodies should be linted (#856, @MichaelChirico). We think the biggest use case for this is writing SQL in R strings, especially in cases where the recommended string with for SQL & R differ.
 
 ### Lint accuracy fixes: removing false positives
 
--- a/R/line_length_linter.R
+++ b/R/line_length_linter.R
@@ -2,7 +2,10 @@
 #'
 #' Check that the line length of both comments and code is less than `length`.
 #'
-#' @param length maximum line length allowed. Default is 80L (Hollerith limit).
+#' @param length Maximum line length allowed. Default is `80L` (Hollerith limit).
+#' @param ignore_string_bodies Logical, default `FALSE`. If `TRUE`, the contents
+#'   of string literals are ignored. The quotes themselves are included, so this
+#'   mainly affects wide multiline strings, e.g. SQL queries.
 #'
 #' @examples
 #' # will produce lints
@@ -22,13 +25,19 @@
 #' - [linters] for a complete list of linters available in lintr.
 #' - <https://style.tidyverse.org/syntax.html#long-lines>
 #' @export
-line_length_linter <- function(length = 80L) {
+line_length_linter <- function(length = 80L, ignore_string_bodies = FALSE) {
   general_msg <- paste("Lines should not be more than", length, "characters.")
 
   Linter(linter_level = "file", function(source_expression) {
     # Only go over complete file
-    line_lengths <- nchar(source_expression$file_lines)
+    line_lengths <- as.integer(nchar(source_expression$file_lines))
     long_lines <- which(line_lengths > length)
+
+    if (ignore_string_bodies) {
+      in_string_body_idx <-
+        is_in_string_body(source_expression$full_parsed_content, length, long_lines)
+      long_lines <- long_lines[!in_string_body_idx]
+    }
 
     Map(
       function(long_line, line_length) {
@@ -46,4 +55,40 @@ line_length_linter <- function(length = 80L) {
       line_lengths[long_lines]
     )
   })
+}
+
+is_in_string_body <- function(parse_data, max_length, long_idx) {
+  str_idx <- parse_data$token == "STR_CONST"
+  if (!any(str_idx)) {
+    return(rep(FALSE, length(long_idx)))
+  }
+  str_data <- parse_data[str_idx, ]
+  if (all(str_data$line1 == str_data$line2)) {
+    return(rep(FALSE, length(long_idx)))
+  }
+  # right delimiter just ends at 'col2', but 'col1' takes some sleuthing
+  str_data$line1_width <- nchar(vapply(
+    strsplit(str_data$text, "\n", fixed = TRUE),
+    function(x) x[1L],
+    FUN.VALUE = character(1L),
+    USE.NAMES = FALSE
+  ))
+  str_data$col1_end <- str_data$col1 + str_data$line1_width
+  vapply(
+    long_idx,
+    function(line) {
+      # strictly inside a multi-line string body
+      if (any(str_data$line1 < line & str_data$line2 > line)) {
+        return(TRUE)
+      }
+      on_line1_idx <- str_data$line1 == line
+      if (any(on_line1_idx)) {
+        return(max(str_data$col1_end[on_line1_idx]) <= max_length)
+      }
+      # use parse data to capture possible trailing expressions on this line
+      on_line2_idx <- parse_data$line2 == line
+      any(on_line2_idx) && max(parse_data$col2[on_line2_idx]) <= max_length
+    },
+    logical(1L)
+  )
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -241,8 +241,12 @@ get_r_string <- function(s, xpath = NULL) {
       s <- xml_find_chr(s, sprintf("string(%s)", xpath))
     }
   }
-  # parse() skips "" elements --> offsets the length of the output,
-  #   but NA in --> NA out
+  r_string_from_parse_text(s)
+}
+
+# parse() skips "" elements --> offsets the length of the output,
+#   but NA in --> NA out
+r_string_from_parse_text <- function(s) {
   is.na(s) <- !nzchar(s)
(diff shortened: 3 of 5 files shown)


R/line_length_linter.R
```diff
@@ -22,14 +25,20 @@
...
 #' - <https://style.tidyverse.org/syntax.html#long-lines>
 #' @export
-line_length_linter <- function(length = 80L) {
+line_length_linter <- function(length = 80L, ignore_string_bodies = FALSE) {
   general_msg <- paste("Lines should not be more than", length, "characters.")
 
   Linter(linter_level = "file", function(source_expression) {
     # Only go over complete file
-    line_lengths <- nchar(source_expression$file_lines)
+    line_lengths <- as.integer(nchar(source_expression$file_lines))
```
AshesITR, line 33: Why is this necessary?

R/line_length_linter.R
```diff
@@ -47,3 +56,39 @@ line_length_linter <- function(length = 80L) {
     )
   })
 }
+
+is_in_string_body <- function(parse_data, max_length, long_idx) {
+  str_idx <- parse_data$token == "STR_CONST"
+  if (!any(str_idx)) {
+    return(rep(FALSE, length(long_idx)))
```
AshesITR, line 63:
```suggestion
    return(rep_len(FALSE, length(long_idx)))
```

R/line_length_linter.R
```diff
@@ -47,3 +56,39 @@ line_length_linter <- function(length = 80L) {
...
+  }
+  str_data <- parse_data[str_idx, ]
+  if (all(str_data$line1 == str_data$line2)) {
+    return(rep(FALSE, length(long_idx)))
```
AshesITR, line 67:
```suggestion
    return(rep_len(FALSE, length(long_idx)))
```

AshesITR (commented):
Looking at the tests I'm wondering whether all you need is supress lints if the lint line satisfies `line1 < line < line2` for some `//STR_CONST`?

R/line_length_linter.R
MichaelChirico: strips names, IINM currently it's doing look-up by name that should be faster using integer index directly instead.

R/line_length_linter.R
MichaelChirico: I don't think there's any benefit, `rep_len()` is a good replacement for `rep(length.out=)`, in this case I think it's identical.

R/line_length_linter.R
AshesITR: So `unname()` would do the trick?
Did you benchmark it? I would expect integer index accessed to always ignore names even if they are present.

MichaelChirico: Hmm I didn't expect that but you're right:

```r
x = setNames(nm=1:100)
cnm = setNames(nm = sample(100, 20))
inm = as.integer(cnm)
microbenchmark(times = 1e4, x[inm], x[cnm])
# Unit: nanoseconds
#    expr min  lq     mean median  uq    max neval cld
#  x[inm] 570 640 792.5278    650 670 131151 10000   a
#  x[cnm] 590 650 795.9336    669 680  48071 10000   a

# disassociate the name from the value
rnm = setNames(sample(100, 20), sample(100, 20))
head(rnm)
# 34  6 68 48 53 36 
# 62 94 37 54 58 70 

x[rnm] # matches integer value
#  62  94  37  54  58  70  22  57 100  95   5  34   4  51  98  83  79   3  87  97 
#  62  94  37  54  58  70  22  57 100  95   5  34   4  51  98  83  79   3  87  97 
x[names(rnm)] # matches names
# 34  6 68 48 53 36 75 76 88 49 67 52 32 21 98 84 70 25 46 24 
# 34  6 68 48 53 36 75 76 88 49 67 52 32 21 98 84 70 25 46 24 
```

Thanks!

MichaelChirico, line 33:
```suggestion
    line_lengths <- nchar(source_expression$file_lines)
```

MichaelChirico:
> Looking at the tests I'm wondering whether all you need is supress lints if the lint line satisfies `line1 < line < line2` for some `//STR_CONST`?

there's also the issue of how to deal with lines like this:

```r
x <- 'long string starting on a short
line which also has long strings in
the other parts of the body'
```

Before this rule, there's 3 lints, but under `ignore_string_bodies=TRUE` here, there are 0

R/line_length_linter.R
MichaelChirico: if anything, `rep_len()` is minimally worse:

```r
n = 100
microbenchmark(times = 1e5, rep(FALSE, n), rep_len(FALSE, n))
# Unit: nanoseconds
#               expr min   lq      mean median   uq      max neval cld
#      rep(FALSE, n) 560  670  949.4929    700  760  1251262 1e+05  a 
#  rep_len(FALSE, n) 910 1020 1922.5406   1051 1131 25629813 1e+05   b
```

MichaelChirico:
Bumping for review :)

NEWS.md
```diff
@@ -48,6 +48,7 @@
 * `object_usage_linter()` lints missing packages that may cause false positives (#2872, @AshesITR)
 * New argument `include_s4_slots` for the `xml_find_function_calls()` entry in the `get_source_expressions()` to govern whether calls of the form `s4Obj@fun()` are included in the result (#2820, @MichaelChirico).
 * `sprintf_linter()` lints `sprintf()` and `gettextf()` calls when a constant string is passed to `fmt` (#2894, @Bisaloo).
+* `line_length_linter()` has a new argument `ignore_string_bodies` (defaulting to `FALSE`) which governs whether the contents of multi-line string bodies should be linted (#856, @MichaelChirico). We think the biggest use case for this is writing SQL in R strings, especially in cases where the recommended string with for SQL & R differ.
```
MichaelChirico:
```suggestion
* `line_length_linter()` has a new argument `ignore_string_bodies` (defaulting to `FALSE`) which governs whether the contents of multi-line string bodies should be linted (#856, @MichaelChirico). We think the biggest use case for this is writing SQL in R strings, especially in cases where the recommended string width for SQL & R differ.
```

R/line_length_linter.R
```diff
@@ -2,7 +2,10 @@
 #'
 #' Check that the line length of both comments and code is less than `length`.
 #'
-#' @param length maximum line length allowed. Default is 80L (Hollerith limit).
+#' @param length Maximum line length allowed. Default is `80L` (Hollerith limit).
+#' @param ignore_string_bodies Logical, default `FALSE`. If `TRUE`, the contents
+#'   of string literals are ignored. The quotes themselves are included, so this
+#'   mainly affects wide multiline strings, e.g. SQL queries.
```
Bisaloo, line 8:
```suggestion
#'   only affects wide multiline strings, e.g. SQL queries.
```

?

Looking at the code, this seems clearer & more precise.

Bisaloo, line 8: Could you add some minimal examples for this new argument please?

R/line_length_linter.R
```diff
@@ -47,3 +56,39 @@ line_length_linter <- function(length = 80L) {
...
+      if (any(on_line1_idx)) {
+        return(max(str_data$col1_end[on_line1_idx]) <= max_length)
+      }
+      # use parse data to capture possible trailing expressions on this line
+      on_line2_idx <- parse_data$line2 == line
+      any(on_line2_idx) && max(parse_data$col2[on_line2_idx]) <= max_length
+    },
+    logical(1L)
+  )
+}
```
Bisaloo, line 142: I **think** this is correct but I'm really struggling to wrap my head around this function, which concerns me a little bit for maintainability.

In particular, all current tests pass with this much simpler version of the vapply() call

```suggestion
is_in_string_body <- function(parse_data, max_length, long_idx) {
  str_idx <- parse_data$token == "STR_CONST"
  if (!any(str_idx)) {
    return(rep(FALSE, length(long_idx)))
  }
  str_data <- parse_data[str_idx, ]
  if (all(str_data$line1 == str_data$line2)) {
    return(rep(FALSE, length(long_idx)))
  }
  vapply(
    long_idx,
    function(line) {
      # strictly inside a multi-line string body
      any(str_data$line1 < line & str_data$line2 > line)
    },
    logical(1L)
  )
}
```

Are we missing tests?

Bisaloo (commented):
I had to come back to this several time because the current version of `is_in_string_body()` is quite complex. I think part of it is the switch between vectorized and vapply logic and the easily missed changes between str_idx/long_idx as well as str_data/parse_data.

R/line_length_linter.R
MichaelChirico: I ran the two versions on 1,700 files with long lines and found a number of differences.

```
queryresults <- queryresults[grepl("-", BoldingPattern, fixed = TRUE)] # Remove baselines
    dplyr::select(all_of(c(paste(col_info, "_holdback", sep = ""), add_col))) %>%
```

It looks like basically the simpler one skips on any line with a string?

MichaelChirico: Added equivalent tests

R/line_length_linter.R
MichaelChirico: Added a new example related to this -- what I'm trying to convey is mainly that the closing quote counts towards the line length where it's found.

```
1234567890
----------
string <- "
 too long!
"
```

The less common case is that the opening `"` here will also trigger a lint (with max width 10)

```r
lint("string <- '\n  too long\n'", line_length_linter(10, ignore_string_bodies=TRUE))
# <text>:1:11: style: [line_length_linter] Lines should not be more than 10 characters. This line is 11 characters.
# string <- '
# ~~~~~~~~~~^
```

PR #2801, opened 2025-02-27
Remove some tests of internal API

MichaelChirico:
Closes #1692.

PR #2802, opened 2025-02-27
Add ignore_order= argument to expect_lint()

MichaelChirico:
For #2801, I'm trying to rewrite this to use `expect_lint()`, it generates 192 lints, quite cumbersome to ensure the lints are all in the expected order.

I am tempted to make `ignore_order=TRUE` the default, actually, I'm not sure how meaningful it is to test the order in which lints are produced. But that would possibly be a breaking change.


R/expect_lint.R
```diff
@@ -59,11 +62,11 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
 
   lints <- lint(file, ...)
   n_lints <- length(lints)
-  lint_str <- if (n_lints) paste(c("", lints), collapse = "\n") else ""
+  lint_fmt <- function() if (n_lints > 0L) paste(c("", lints), collapse = "\n") else ""
```
AshesITR, line 65: Why do we need this?

MichaelChirico: pretty edge-casey, but if there are 100s of lints and this is `expect_no_lint()`, we don't need to generate this big string, we only report "N lints when 0 were expected".

AshesITR: I don't see a code path where it's skipped?
The `is.null(checks)` path includes the list of lints at the end of the message, no?

MichaelChirico: Indeed... 🤦

AshesITR (dismissed):

MichaelChirico:
Merging because it was approved and the review staled by a merge conflict. Repo rules no longer stale reviews in such cases.

PR #2805, opened 2025-03-03
Parse settings prior to reading file to ensure Encoding is set

### LINKED ISSUE #2803: settings$encoding used before set
https://github.com/r-lib/lintr/blob/a1a2cdaa935bd8cadc5ed57b7f91eab56e71bf73/R/lint.R#L48-L62

`get_lines()` is run before `read_settings()` but it relies on `settings$encoding`:

https://github.com/r-lib/lintr/blob/a1a2cdaa935bd8cadc5ed57b7f91eab56e71bf73/R/lint.R#L685-L691
https://github.com/r-lib/lintr/blob/a1a2cdaa935bd8cadc5ed57b7f91eab56e71bf73/R/utils.R#L161

I seem to recall running into issues moving `read_settings()` earlier in `lint()`, but this situation seems incorrect.

This came up in #2801 -- when trying to move away from manually running `lintr:::read_settings()` in the test, we fail because `get_source_expressions()` doesn't know the encoding, and neither does moving to accessing the file by `lint()` instead help:

https://github.com/r-lib/lintr/blob/a1a2cdaa935bd8cadc5ed57b7f91eab56e71bf73/tests/testthat/test-get_source_expressions.R#L106-L110

MichaelChirico:
Closes #2803


AshesITR (dismissed):

IndrajeetPatil (approved):

bastistician:
The NEWS don't reveal it but this change means that `~/.lintr` settings are no longer read when linting `inline_data`, e.g., `lintr::lint(<R-code-as-string>)`, which is what Emacs-ESS uses to lint code in R buffers. So with the development version of **lintr** I'd only see notes from the `default_linters`.

MichaelChirico:
Thanks, would you mind filing this as a new bug?

PR #2808, opened 2025-03-03
brace_linter() covers test_that() without braces

### LINKED ISSUE #2292: Encourage braced expression in `test_that()` calls
More details in https://github.com/r-lib/styler/issues/1154

Not sure if we can accommodate this in any of the existing linters.

MichaelChirico:
Closes #2292


AshesITR (commented):
Interestingly I found two odd usages of test_that on GitHub https://github.com/search?q=%2Ftest_that%5C(%5C%7B%2F&type=code

1. Piping the test name.
2. Supplying the description as a keyword argument.

Both break the positional assumption.

WDYT about supporting one or both?

MichaelChirico:
> Interestingly I found two odd usages of test_that on GitHub https://github.com/search?q=%2Ftest_that%5C(%5C%7B%2F&type=code
> 
> 1. Piping the test name.
> 2. Supplying the description as a keyword argument.
> 
> Both break the positional assumption.
> 
> WDYT about supporting one or both?

I don't think it's worth the investment. If eventually we have a good general framework for handling odd cases like that, we can include it in a sweep of all linters. But I also don't see much demand for that besides from us maintainers recognizing we're being less-than-perfect on this front.

I see 28 hits there vs. >300,000 for just `test_that[(]['"]`, and at least one of them is Lionel presumably himself doing some edge case testing :)

MichaelChirico:
Bumping for review :)

MichaelChirico:
Sorry, I'm not sure what I'm looking at 😅 Your example doesn't parse, and the PR you shared has too much other things going on :)

olivroy:
Sorry, I typed too fast on my phone. See this view instead https://github.com/obrl-soil/h3jsr/pull/11/files?file-filters%5B%5D=.R&show-deleted-files=true&show-viewed-files=true

MichaelChirico:
Can't say I've seen that one before 🫨

PR #2810, opened 2025-03-04
Drop our internal %|||% helper

### LINKED ISSUE #2768: Decide if we actually need our version of %||% or it should just the base version
From #2767, we'll lint for encouraging `%||%`, thus I'm replacing our old version with `%|||%` for simplicity. But ideally we don't need to maintain that extra complexity -- it's just a project to investigate if those other uses actually require the different implementation:

https://github.com/r-lib/lintr/blob/07c55846c1189740b7ac7a83b4b53203e23940ca/R/utils.R#L1-L7

That has been there basically since day one of the package (11 years ago).

MichaelChirico:
Closes #2768

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1,4 +1,5 @@
-# backports approach doesn't work since we need this before .onLoad() in names2()
+# TODO(R>=4.4.0): remove this.
+# NB: {backports} approach doesn't work since we need this before .onLoad() in names2()
 if (!exists("%||%", "package:base")) {
   `%||%` <- function(x, y) if (is.null(x)) y else x # nolint: coalesce_linter.
 }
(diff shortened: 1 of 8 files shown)


MichaelChirico:
I don't know why this isn't working:

https://github.com/r-lib/lintr/blob/700607ef29915690e5fe234539f27433c5e3da4b/R/zzz.R#L312-L313

R/utils.R
```diff
@@ -1,10 +1,6 @@
-# TODO(#2768): possibly just use %||% instead
-`%|||%` <- function(x, y) {
-  if (is.null(x) || length(x) == 0L || (is.atomic(x[[1L]]) && is.na(x[[1L]]))) {
-    y
-  } else {
-    x
-  }
+# backports approach doesn't work since we need this before .onLoad() in names2()
+if (!exists("%||%", "package:base")) {
```
AshesITR, line 2: Mark this with a TODO to delete once support for 4.1 is dropped?

AshesITR (dismissed):

R/utils.R
MichaelChirico: 4.4, but yes, good idea

PR #2190, opened 2023-09-19
Consistently lint OP-LAMBDA where FUNCTION is

MichaelChirico:

PR #2818, opened 2025-03-06
Start a fuzzing suite to test for consistency of lints

MichaelChirico:
Part of #2191. Progress on #2737. Continuation of #2190.

For this initial PR, I've only added a rule about `FUNCTION` (`function(...) { ... }`) and `OP-LAMBDA` (`\(...) { ... }`) equivalency.

Hopefully my approach here is extensible for other rules (I tried for that). My plan is for this to only run on `main`, though it's certainly fast enough to run on every PR.

The upshot is, it smoked out three linters with inconsistencies already!


MichaelChirico:
@DavisVaughan / @lionel-, I'm curious if any of your recent work on {air}/{treesitter.r} & friends could be re-used here.

The idea is we want to make random injections/edits to the R AST (and then do stuff), my approach here felt quite manual/labor-intensive.

DavisVaughan:
The closest thing I can think of is `parser_reparse()` in {treesitter}
https://davisvaughan.github.io/r-tree-sitter/reference/parser-parse.html

It performs an incremental reparse and hands you back a new tree with the edit you made. It's quite verbose in terms of what you have to provide (bytes and points) but this is a tree-sitter limitation (and I'm sure they had some reason for it).

For Air, I don't think Rowan trees are really intended to be edited. At least, there doesn't seem to be any easy to use API for that (and plus Rowan isn't exposed in any R package right now anyways)

MichaelChirico:
Interesting, yea the incremental re-parse is very neat, but more useful in an IDE setting. Here the latency of re-parse is not much concern as the edits are only a tiny fraction of the action run time.

> For Air, I don't think Rowan trees are really intended to be edited. At least, there doesn't seem to be any easy to use API for that (and plus Rowan isn't exposed in any R package right now anyways)

Hmm, I guess in my mind something like {styler} / {air} is well-suited here if it exposes a friendly API for transformers, i.e. instead of the usual helpful rule like 'make this line <80 characters', we have a more adversarial rule like "add comments in random but syntactically valid places". I not be understanding where in the stack behind the format-on-save tool such an analog would sit.

lionel-:
> For Air, I don't think Rowan trees are really intended to be edited. At least, there doesn't seem to be any easy to use API for that (and plus Rowan isn't exposed in any R package right now anyways)

I may be misunderstanding what you mean but AFAIK the red trees in Rowan are modifiable. I believe this is used for refactorings in rust-analyzer.

MichaelChirico:
Bumping for review :)

.github/workflows/ast-fuzz.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: "release"
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+
+      - name: Ensure equivalent code generates equivalent lints
+        run: |
+          callr::rscript(".dev/ast_fuzz_test.R")
```
AshesITR, line 29: Why spawn two Rscript processes?

R/expect_lint.R
```diff
@@ -47,15 +47,8 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en") {
...
-    file <- tempfile()
-    on.exit(unlink(file), add = TRUE)
-    local({
-      con <- base::file(file, encoding = "UTF-8")
-      on.exit(close(con))
-      writeLines(content, con = con, sep = "\n")
-    })
-  }
+  if (is.null(file)) on.exit(unlink(file), add = TRUE)
+  file <- maybe_write_content(file, content)
```
AshesITR: This could use a comment explaining that this is a fuzzer entrypoint.

.github/workflows/ast-fuzz.yaml
MichaelChirico: Good point. I copy-pasted it from here:

https://github.com/r-lib/lintr/blob/ae4046ee92a0cd7b56972f01f30d3dcce9365358/.github/workflows/repo-meta-tests.yaml#L32-L33

R/expect_lint.R
MichaelChirico: Here, or in ast_fuzz_test.R? The latter seems more natural to me. But maybe I'm missing your point.

.github/workflows/ast-fuzz.yaml
MichaelChirico: Let's fix it everywhere at once: #2873

R/expect_lint.R
AshesITR: Here. The control flow does two back-to-back null-checks on `file`, which has no obvious reason without knowing about the fuzzer.

MichaelChirico: Well, the second `is.null()` check is inside `maybe_write_content()`, so I wouldn't look at this code in a vacuum and say "that doesn't look right".

it's two steps because it's not easy to set an `on.exit()` hook in a parent frame (well, we could: https://yihui.org/en/2017/12/on-exit-parent/ but I'm not sure it's worth the complexity)

AshesITR: To me I'd at least wonder why the null check wasn't removed from the call and the call moved into the if branch.

I was thinking something like

`# this line is replaced by the fuzzer in .dev/ast_fuzz_test.R`

MichaelChirico: We might be able to have it both ways if you can suggest an alternative -- IIRC my original thinking was this was an independently OK way to rewrite the code here. For now I have added the comment.

.dev/maybe_fuzz_content.R
```diff
@@ -0,0 +1,62 @@
...
+  pd <- tryCatch(getParseData(suppressWarnings(parse(f, keep.source = TRUE))), error = identity)
+  if (inherits(pd, "error")) {
+    return(invisible())
+  }
+
+  reparse <- FALSE
+  lines <- readLines(f)
+  for (fuzzer in list(function_lambda_fuzzer)) {
+    if (reparse) {
+      pd <- getParseData(parse(f, keep.source = TRUE))
```
AshesITR, line 52: Is there a way to reparse directly from `updated_lines` without a round trip to disk?

.dev/maybe_fuzz_content.R
```diff
@@ -0,0 +1,62 @@
...
+  }
+  lines
+}
+
+# we could also consider just passing any test where no fuzzing takes place,
+#   i.e. letting the other GHA handle whether unfuzzed tests pass as expected.
+apply_fuzzers <- function(f) {
+  # skip errors for e.g. Rmd files, and ignore warnings.
+  #   We could use get_source_expressions(), but with little benefit & much slower.
+  pd <- tryCatch(getParseData(suppressWarnings(parse(f, keep.source = TRUE))), error = identity)
```
AshesITR: What encoding does `parse()` assume if given a string path?

MichaelChirico: Does encoding matter if we're only parsing internal files? For those few tests that touch encoding, I'd expect `#nofuzz` to just not bother with it.

.dev/maybe_fuzz_content.R
MichaelChirico: It _might_ be possible, though it could impose some limits on the type of fuzzes that are allowed. Without looking to deeply, I'd assume we could do so if we're only editing a valid subtree and replacing it with another valid subtree, whereas now we allow generic text replacement.

That said, I didn't see any real performance issue, so I'm not sure there's much to gain from adding this complication. I have in mind things in `get_source_expressions()` that operate from the parse data.frame, which require a lot of code:

https://github.com/r-lib/lintr/blob/afb363c22e95dcbc7daffc93967dc5a6ab3cfba8/R/get_source_expressions.R#L679

.dev/maybe_fuzz_content.R
AshesITR: I wouldn't bother too much either, but hard-code `UTF-8` just to be safe.

MichaelChirico:
@AshesITR anything else left to fix in this branch before proceeding?

MichaelChirico:
NP, the GHA-only parts are not so crucial since they're not user-facing -- only matters insofar as it generates correct fixes in the R code. We can fine-tune the dev-facing stuff at our leisure.

PR #2819, opened 2025-03-07
Add a second fuzzer for pipe equivalency

MichaelChirico:
Part of #2191 -- continuing #2818 with a demonstration of a second fuzzing rule. Also progress on #2737.

One worry is that eventually we'll get so many `# nofuzz` rules as to start creating false negatives, but I think we can worry about that more later -- our suite is quite large.

This fuzzer did not expose any issues with our linters (yay!)

### DIFF
--- a/.dev/maybe_fuzz_content.R
+++ b/.dev/maybe_fuzz_content.R
@@ -67,11 +67,11 @@ apply_fuzzers <- function(f) {
     updated_lines <- fuzzer(pd, lines)
     if (is.null(updated_lines) || identical(unedited, updated_lines)) next # skip some I/O if we can
     writeLines(updated_lines, f)
-    # check if our attempted edit introduced some error
+    # check if our attempted edit introduced some error; skip applying this fuzzer only if so
     pd <- error_or_parse_data(f)
     if (inherits(pd, "error")) {
-      writeLines(unedited, f)
-      return(invisible())
+      writeLines(lines, f)
+      next
     }
     lines <- readLines(f)
   }


tests/testthat/test-pipe_consistency_linter.R
```diff
@@ -1,5 +1,7 @@
+# nofuzz start
 test_that("pipe_consistency skips allowed usage", {
   skip_if_not_r_version("4.1.0")
+
```
MichaelChirico:
```suggestion
```

tests/testthat/test-pipe_consistency_linter.R
```diff
@@ -133,10 +134,7 @@ test_that("pipe_consistency_linter works with %>% argument", {
     linter
   )
 
-  expect_no_lint(
-    "1:3 %>% mean() %>% as.character()",
-    linter
-  )
+  expect_no_lint("1:3 %>% mean() %>% as.character()", linter)
```
MichaelChirico:
```suggestion
  expect_no_lint(
    "1:3 %>% mean() %>% as.character()",
    linter
  )
```

tests/testthat/test-implicit_assignment_linter.R
```diff
@@ -335,15 +315,18 @@ test_that("implicit_assignment_linter blocks disallowed usages in function calls
...
   linter <- implicit_assignment_linter()
 
-  expect_lint("data %>% mutate(a := b)", NULL, linter)
-  expect_lint("dt %>% .[, z := x + y]", NULL, linter)
-  expect_lint("data %<>% mutate(a := b)", NULL, linter)
+  expect_no_lint("data %>% mutate(a := b)", linter)
+  expect_no_lint(
+    "dt %>% .[, z := x + y]",
+    linter
+  )
```
MichaelChirico:
```suggestion
  expect_no_lint("dt %>% .[, z := x + y]", linter)
```

.dev/maybe_fuzz_content.R
```diff
@@ -14,48 +14,66 @@ maybe_fuzz_content <- function(file, lines) {
...
-    if (reparse) {
-      pd <- getParseData(parse(f, keep.source = TRUE))
-      lines <- readLines(f)
-    }
+  unedited <- lines <- readLines(f)
+  for (fuzzer in list(function_lambda_fuzzer, pipe_fuzzer)) {
     updated_lines <- fuzzer(pd, lines)
-    reparse <- !is.null(updated_lines)
-    if (!reparse) next # skip some I/O if we can
+    if (is.null(updated_lines)) next # skip some I/O if we can
```
AshesITR:
```suggestion
    if (is.null(updated_lines) || identical(unedited, updated_lines)) next # skip some I/O if we can
```

.dev/maybe_fuzz_content.R
```diff
@@ -14,48 +14,66 @@ maybe_fuzz_content <- function(file, lines) {
...
     writeLines(updated_lines, f)
+    # check if our attempted edit introduced some error
+    pd <- error_or_parse_data(f)
+    if (inherits(pd, "error")) {
```
AshesITR, line 72: This will abort fuzzing if any fuzzer caused a parse error, no?

tests/testthat/test-commented_code_linter.R
```diff
@@ -88,19 +88,17 @@ test_that("commented_code_linter can detect operators in comments and lint corre
...
-    expect_lint(paste("i", op, "1", collapse = ""), NULL, linter)
-    expect_lint(paste("# something like i", op, "1", collapse = ""), NULL, linter)
+    expect_no_lint(paste("i", op, "1", collapse = ""), linter)
+    expect_no_lint(paste("# something like i", op, "1", collapse = ""), linter)
     expect_lint(paste("# i", op, "1", collapse = ""), lint_msg, linter)
   }
-})
 
-test_that("commented_code_linter can detect operators in comments and lint correctly", {
   skip_if_not_r_version("4.1.0")
```
AshesITR: Why skip in the middle of a test?

tests/testthat/test-pipe_continuation_linter.R
```diff
@@ -161,12 +158,14 @@ local({
         test_data <- diamonds %>% head(10) %>% tail(5)
       })
     "),                      "three inside test_that()",
-    trim_some("
-      {
-        x <- a %>% b %>% c
-        y <- c %>% b %>% a
-      }
-    "),                      "two different single-line pipelines",
+    trim_some(
```
AshesITR: What is this diff for?

.dev/maybe_fuzz_content.R
MichaelChirico: Yes, but that's exactly the point -- it makes writing the fuzzers easier if we don't need to over-complicate the logic. `%>%` ↔️ `|>` is a great example -- they can just be text-swapped most of the time, but for those cases where it's not true, we just give up at this point. Writing the fuzzer itself to be completely fool-proof is way harder.

One thing to address what I believe is your concern would be to add unit tests for the fuzzers, but I don't see a need for that at this point.

tests/testthat/test-pipe_continuation_linter.R
```diff
@@ -161,12 +158,14 @@ local({
...
+      "
+        {
+          x <- a %>% b %>% c
+          y <- c %>% b %>% a
+        }
+      "
+    ),                      "two different single-line pipelines",
```
MichaelChirico:
```suggestion
    trim_some("
      {
        x <- a %>% b %>% c
        y <- c %>% b %>% a
      }
    "),                      "two different single-line pipelines",
```

tests/testthat/test-pipe_continuation_linter.R
```diff
@@ -161,6 +158,7 @@ local({
         test_data <- diamonds %>% head(10) %>% tail(5)
       })
     "),                      "three inside test_that()",
+    trim_some(
```
MichaelChirico:
```suggestion
```

.dev/maybe_fuzz_content.R
AshesITR: Oh, sorry for causing confusion.
My suggestion would be to write `lines` instead of `unedited`, so that earlier fuzzers in the list still get a chance to fuzz the code. That way, only the failing fuzzer will be skipped instead of resetting to the original file.

MichaelChirico: Oh, got you. I _think_ the edit does that -- back up to the current "known-working" version of the file from before the current fuzzer is applied, then step to the `next` fuzzer.

AshesITR (approved):

PR #2832, opened 2025-03-12
Improve the nofuzz system to allow specific exclusions

MichaelChirico:
Continues the chain from #2818 to #2829 (the last of which this is currently based against).

I wasn't satisfied with the approach to just comment out & ignore large swaths of the test suite needed to implement `# nofuzz` through now, especially after the `comment_injection_fuzzer` turned out to require such large `# nofuzz` regions, but I was stuck on how to proceed with specific exclusions.

I think the approach taken here is pretty satisfactory: as a preprocessing step, we convert the `# nofuzz` markup into new code in the suite that signals activated/deactivated fuzzers for each execution of `expect_lint()`.

The approach could be improved -- it can be a bit of a pain to debug, e.g. when `expect_lint(file = .)` was erroring due to encoding issues, the error causes the test to exit and thus fails to run the `activate_fuzzers()` step, leading to cascading confusing errors it's hard to pin down. And there's no simple workaround here, as e.g. using a `withr::defer()` approach to ensure the `activate_fuzzers()` step is run is not straightforward.

But I think there's enough good progress here (and it turned up a few more true positive fixes!) that it is ready for review.

### DIFF
--- a/.dev/ast_fuzz_test.R
+++ b/.dev/ast_fuzz_test.R
@@ -76,30 +76,36 @@ get_str <- \(x) tail(unlist(strsplit(x, ": ", fixed = TRUE)), 1L)
 test_restorations <- list()
 for (test_file in list.files("tests/testthat", pattern = "^test-", full.names = TRUE)) {
   test_lines <- readLines(test_file)
-  one_expr_idx <- grep("# nofuzz:", test_lines)
+  one_expr_idx <- grep("# nofuzz", test_lines)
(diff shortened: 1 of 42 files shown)


MichaelChirico:
There is some useful feedback in there, I will address it soon.

MichaelChirico:
Fixed items 1 & 2.

For 3, I'm not sure a `->` swapper is worth the overhead it will take to put `->` in an always-valid way. Anyway, filed #3005.

For 4, we can revisit this if it becomes an annoyance.

Issue #1974, opened 2023-06-07
Make `backport_linter()` more informative by default


Bisaloo:
The default for the `r_version` argument of `backport_linter()` is `getRversion()`, which, in most case, won't provide any useful insights.

A possibly better default in the case of R package, or any project with a `DESCRIPTION` file, would be to extract the stated minimum R version and use it.

This means that `backport_linter()` would now automatically warn you if you use a function that has been defined in an earlier R version that what you declared in `DESCRIPTION`.

This can be done with a handful of base R function:

```r
depends <- read.dcf("DESCRIPTION", "Depends")

depends_list <- trimws(unlist(strsplit(",")))

depends_r <- grep("R ", depends_list, value = TRUE)
min_r_version <- gsub("^R \\(>=?\\s(.+)\\)", "\\1", depends_r)
```

An alternative that seems less generic but still more informative than the default could be to use the tidyverse policy of supporting `oldrel-4`. This can make sense since the rest of the lintr already defaults to the tidyverse policy.

AshesITR:
The way it's currently set up, `backport_linter()` in combination with an `expect_lint_free()` test will make sure that all versions the package is tested on in CI are fine.

(on a related note: we need to bump the R current version to 4.3; currently that is still considered devel)

Bisaloo:
Fair point, but:

- from what I've seen, it's more common for people to use the `lint.yaml` or `lint-changed-file.yaml` workflows from r-lib/actions than adding `expect_lint_free()` to their tests.
- in terms of efficiency, the setup you're describing uses quite a lot more resources since `lint_package()` will run multiple times in a single CI run. Are there any benefits for this apart from changing the `r_version` value in `backport_linter()`, which could also be addressed by the solution proposed here? I don't expect other linters to report different results across versions but I might be missing something.

AshesITR:
I can definitely see the benefit of what you are proposing. Fetching the info from DESCRIPTION seems like a good idea when available.

Then again, `backport_linter()` is non-default so anyone using it must explicitly add it and thus see the `r_version=` option.

We could add an `"auto"` option to choose an R version for the user.
But how should this behave by default?

Bisaloo:
>  But how should this behave by default?

Do you mean when the `DESCRIPTION` doesn't specify a minimum R version?
At the moment, in my organization, I fall back to `getRversion()` in this situation. It seems like the most conservative (but maybe not the best) approach.

> We could add an "auto" option to choose an R version for the user.

[...]

PR #2833, opened 2025-03-12
Make `backport_linter(r_version = )` default and options more explicit

Bisaloo:
While thinking about #1974, I noticed some features of `backport_linter()` were not explicitly documented. This minor PR aims to help discoverability of this useful feature.

### DIFF
--- a/R/backport_linter.R
+++ b/R/backport_linter.R
@@ -21,6 +21,11 @@
 #'   linters = backport_linter("3.2.0")
 #' )
 #'
+#' lint(
+#'   text = "deparse1(expr)",
+#'   linters = backport_linter("3.6.0")
+#' )
+#'
 #' # okay
 #' lint(
 #'   text = "trimws(x)",
@@ -29,24 +34,12 @@
 #'
 #' lint(
 #'   text = "str2lang(x)",
-#'   linters = backport_linter("4.0.0")
-#' )
-#'
-#' lint(
-#'   text = "str2lang(x)",
 #'   linters = backport_linter("3.2.0", except = "str2lang")
 #' )
 #'
 #' # Version aliases instead of numbers can also be passed to `r_version`
-#' # lints
 #' lint(
-#'   text = "str2lang('x')",
-#'   linters = backport_linter("3.5.0")
-#' )
-#'
-#' # ok
-#' lint(
-#'   text = "str2lang('x')",
+#'   text = "deparse1(expr)",
 #'   linters = backport_linter("release")
 #' )
 #'
--- a/R/backport_linter.R
+++ b/R/backport_linter.R
@@ -3,7 +3,8 @@
 #' Check for usage of unavailable functions. Not reliable for testing r-devel dependencies.
 #'
 #' @param r_version Minimum R version to test for compatibility. Defaults to
-#'  the R version currently in use.
+#'  the R version currently in use. The version can be specified as a version
+#'  number, or as a version alias (such as `"devel"`, `"oldrel"`, `"oldrel-1"`).
 #' @param except Character vector of functions to be excluded from linting.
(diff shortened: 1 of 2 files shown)


R/backport_linter.R
```diff
@@ -35,6 +37,11 @@
 #'   linters = backport_linter("3.2.0", except = "str2lang")
 #' )
 #'
+#' lint(
+#'   text = "mean(x)",
+#'   linters = backport_linter("release")
```
MichaelChirico, line 42: generally we try and structure the examples as lint/ok pairs, could you add one that lints on an old version?

MichaelChirico (commented):
great idea & thanks!

R/backport_linter.R
Bisaloo: The problem here is that examples that currently lint for, e.g., `"oldrel-3"`, will no longer lint in the future, when `"oldrel-3"` is recent enough and includes the new function.

A solution could be to go quite far back to give some years of leeway before the example starts to be inaccurate, but it will still stop linting at some point:

```
#' lint(
#'   text = "array2DF(x)",
#'   linters = backport_linter("oldrel-10")
#' )
```

Do you have a better suggestion?

MichaelChirico: Why not

```r
# lints
lint(
  text = "str2lang('x')",
  linters = backport_linter("3.5.0")
)

# ok
lint(
  text = "str2lang('x')",
  linters = backport_linter("release")
)
```

Bisaloo: Does [`cafe5e8` (#2833)](https://github.com/r-lib/lintr/pull/2833/commits/cafe5e8028b2c53af019b3bf3d24b99925ce09df) correspond to what you have in mind?

MichaelChirico: yes, but (and sorry about the inconvenience) not precisely. Sorry, I should have been clearer.

We format examples as a block of `will produce lints` then a block of corresponding `okay` usages, so you should move the first example up in the `@examples` section.

Bisaloo: I think where I got confused is then: there is already an equivalent example in `will produce lints`. Should I still add the new one? Or just the new one with `"release"` in `okay`?

MichaelChirico:
(quick aside: @Bisaloo, as maintainer, you have permission to create branches directly on this repo -- please do so, it makes review _slightly_ easier in some cases than working on a fork)

R/backport_linter.R
MichaelChirico: I see what you mean. Two things:

 1. I didn't realize there's already a `str2lang()` example, let's use something different for more clarity :face_with_peeking_eye:
 2. The examples are already "unpaired" with 3 "ok" and 2 "lints" examples. @IndrajeetPatil I guess in ea1685b6f4eaeebb2312c324de0ff688719bbed4 we omitted the "lints" correspondence for the new example? Let's just subsume the 2nd & 3rd examples, I think...

MichaelChirico: PTAL at my suggested fix, feel free to merge if that's OK!

MichaelChirico (approved):

R/backport_linter.R
Bisaloo: LGTM. 

I can't merge because 2 required checks didn't run and I don't have permissions to bypass requirements. I think the names of the required checks just need to be updated.

MichaelChirico: thanks for flagging. IDK what's going on there. I tried updating but not sure it had any effect.

MichaelChirico:
@IndrajeetPatil + @AshesITR I just changed the repo settings such that reviews aren't dismissed by "new reviewable commits", I think we've found way too much review toil from simple "merge/rebase" when several PRs are pending.

Of course, use your best judgment on when you think the reviewer should take another look at changes coming after an update.

PR #2834, opened 2025-03-12
Create cbind_dataframe_linter()

### LINKED ISSUE #2596: New linter to recommend `list2DF()` base R function
As an alternative to `do.call(cbind.data.frame, x)`.
``` r
x <- list(
  ind = seq_along(letters),
  let = letters
)

do.call(cbind.data.frame, x)
#>    ind let
#> 1    1   a
#> 2    2   b
#> 3    3   c
#> 4    4   d
#> 5    5   e
#> 6    6   f
#> 7    7   g
#> 8    8   h
#> 9    9   i
#> 10  10   j
#> 11  11   k
#> 12  12   l
#> 13  13   m
#> 14  14   n
#> 15  15   o
#> 16  16   p
#> 17  17   q
#> 18  18   r
#> 19  19   s
#> 20  20   t
#> 21  21   u
#> 22  22   v
#> 23  23   w
#> 24  24   x
#> 25  25   y
#> 26  26   z
```

``` r

list2DF(x)
#>    ind let
#> 1    1   a
#> 2    2   b
#> 3    3   c
#> 4    4   d
#> 5    5   e
#> 6    6   f
#> 7    7   g
#> 8    8   h
#> 9    9   i
#> 10  10   j
#> 11  11   k
#> 12  12   l
#> 13  13   m
#> 14  14   n
#> 15  15   o
#> 16  16   p
#> 17  17   q
#> 18  18   r
#> 19  19   s
#> 20  20   t
#> 21  21   u
#> 22  22   v
#> 23  23   w
#> 24  24   x
#> 25  25   y
#> 26  26   z
```

``` r

y <- list(
  var1 = runif(1e6),
  var2 = rnorm(1e6),
  var3 = rnorm(1e6),
  var4 = rnorm(1e6)
)

  do.call(cbind.data.frame, y),
  list2DF(y),
  check = "identical"
)
#> Unit: microseconds
#>                          expr     min       lq      mean   median       uq
#>  do.call(cbind.data.frame, y) 125.398 131.2090 158.49506 139.6595 166.9540
#>                    list2DF(y)   6.227   7.1795  10.40463   8.3165  10.6615
#>      max neval
#>  344.463   100
#>   67.065   100
```

<sup>Created on 2024-06-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Some hits in the wild: https://github.com/search?q=org%3Acran+%2Fdo%5C.call%5C%28cbind%5C.data%5C.frame%2F&type=code

Are there some edge cases one needs to be aware of? Happy to submit a PR for this.

Bisaloo:
Fix #2596


MichaelChirico:
quick feedback: let's name it `list2df_linter()`. suppose there are other lint rules we come up with in the future that result in the recommendation to use `list2DF()` -- we'd still want them in this same linter.

similarly, we have `nzchar_linter()` and `lengths_linter()` where the linter name is derived from the recommended function.

R/list2df_linter.R
```diff
@@ -0,0 +1,37 @@
+#' Recommend direct usage of `data.frame()` to create a data.frame from a list
+#'
+#' It is possible to create a data.frame from a list of columns with `data.frame()`
+#' or `list2DF()` (since \R 4.0.0 and if recycling is not required), rather than
```
MichaelChirico, line 4: the exposition here is not clear, how about

> `list2DF()` is the preferred way to create a data.frame from a list of equal-length vectors. If you need recycling, use data.frame() instead.

R/list2df_linter.R
```diff
@@ -0,0 +1,37 @@
...
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "list2df(x)",
+#'   linters = list2df_linter()
+#' )
+#'
+#' lint(
+#'   text = "data.frame(list(letters, LETTERS))",
```
MichaelChirico, line 21: I think we try and keep examples paired where possible. so how about

```r
# lint
lint("do.call(cbind.data.frame, list(a = 1, b = 1:10))", list2df_linter())
# ok
lint("data.frame(list(a = 1, b = 1:10))", list2df_linter())
```

R/list2df_linter.R
```diff
@@ -0,0 +1,37 @@
...
+#'   linters = list2df_linter()
+#' )
+#'
+#' @evalRd rd_tags("list2df_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+list2df_linter <- make_linter_from_function_xpath(
+  function_names = "do.call",
+  xpath = "./following-sibling::expr[1]/SYMBOL[text() = 'cbind.data.frame']",
```
MichaelChirico, line 30: it should be easy to catch `do.call("cbind.data.frame"` too:

https://github.com/search?q=lang%3AR%20%2Fdo%5B.%5Dcall%2F%20%2F%22cbind%5B.%5Ddata%5B.%5Dframe%22%2F&type=code

Though I think it will require a "more traditional" linter structure than `make_linter_from_function_xpath()` since we'll need `get_r_string()` to reliable compute the `STR_CONST` value.

R/list2df_linter.R
```diff
@@ -0,0 +1,37 @@
...
+  lint_message = paste0(
+    "Instead of `do.call(cbind.data.frame, lst)`, use `data.frame(lst)`, ",
+    "or `list2DF(lst)` if available and recyclying is not required"
```
MichaelChirico, line 33: "if available" I don't think we need to qualify that really, {lintr} already depends on R>=4.0.0 so `list2DF()` is always available to anyone running {lintr}:

https://github.com/r-lib/lintr/blob/a2b8d6fec967e037e736ba81eb7fca83333438c7/DESCRIPTION#L26-L27

MichaelChirico:
Had a go at refining the wording here too, PTAL

MichaelChirico:
Strange, I get this error in r-devel

```
lint('utils/R/str.R', list2df_linter())
# Error in `lint()`:
# ! Linter `linter()` failed in /home/michael/svn/R-devel/src/library/utils/R/str.R:
# Caused by error in `xml_nodeset()`:
# ! Expecting an external pointer: [type=NULL]
# Run `rlang::last_trace()` to see where the error occurred.
```

MichaelChirico:
Looks like it's due to a lambda in `do.call()`:

```r
lint("do.call(function(x) x, l)\n", list2df_linter())
# Error in `lint()`:
# ! Linter `linter()` failed in /tmp/RtmpTKzzIM/file1393690d1c43:
# Caused by error in `xml_nodeset()`:
# ! Expecting an external pointer: [type=NULL]
# Run `rlang::last_trace()` to see where the error occurred.
```

Bisaloo:
The edits look good to me. Thanks!

I can indeed still reproduce the `do.call()` issue with the latest R-devel but it seems unrelated to this PR. What would you recommend here?

MichaelChirico:
> it seems unrelated to this PR

Hmm, how do you figure? it's being generated by {lintr} code and only from this new linter.

Bisaloo:
> Hmm, how do you figure? it's being generated by {lintr} code and only from this new linter.

No, you're right. For some reason, I was under the impression this happened at the parsing stage, before reaching lintr, but it was incorrect.

I have added a test for this, fixed the immediate issue, and added an extra `!is.na()` before the equality testing to avoid similar issues with more complex expressions I didn't imagine.

tests/testthat/test-list2df_linter.R
```diff
@@ -0,0 +1,38 @@
...
+  expect_lint("do.call('cbind.data.frame', x)", lint_message, linter)
+})
+
+test_that("lints vectorize", {
+  lint_message <- rex::rex("use `data.frame(lst)`")
+
+  expect_lint(
+    trim_some("{
+      cbind(a, b)
+      do.call(cbind.data.frame, x)
```
MichaelChirico, line 29: maybe add an anonymous function version here too to ensure the logic handling that vectorizes?

R/list2df_linter.R
```diff
@@ -0,0 +1,64 @@
...
+list2df_linter <- function() {
+
+  docall_nolambda_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'do.call']
+    /parent::expr
+    /following-sibling::expr[1][SYMBOL or STR_CONST]
+  "
+
+  Linter(linter_level = "expression", function(source_expression) {
+    xml <- source_expression$xml_parsed_content
```
MichaelChirico, line 46: we should still be able to use `$xml_find_function_calls("do.call")`, then add a follow-up `xml_find_all()` with an XPath _relative to_ the `SYMBOL_FUNCTION_CALL/parent::expr`, i.e. in your current commit, that would just be `following-sibling::expr[1][SYMBOL or STR_CONST]`.

R/list2df_linter.R
```diff
@@ -0,0 +1,64 @@
...
+    xml_calls <- xml_find_all(xml, docall_nolambda_xpath)
+    docall_fun <- get_r_string(xml_calls)
+
+    bad_expr <- xml_calls[
+      !is.na(docall_fun) & docall_fun == "cbind.data.frame"
```
MichaelChirico, line 51: I think it's good to be safe, but now that you've landed on `SYMBOL or STR_CONST`, it should definitely not be `NA`, so it might be better to be consistent with other linters & omit the `is.na()` check.

The problem before was you did not include any condition on `SYMBOL or STR_CONST` in the initial XPath search, so `get_r_string()` returned NA for those nodes that didn't match. By including that condition up-front, we guarantee `get_r_string()` lands as intended.

R/list2df_linter.R
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' @evalRd rd_tags("list2df_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+list2df_linter <- function() {
+
+  docall_nolambda_xpath <- "
+  //SYMBOL_FUNCTION_CALL[text() = 'do.call']
+    /parent::expr
+    /following-sibling::expr[1][SYMBOL or STR_CONST]
```
MichaelChirico, line 42: right now, the linter will land on the first argument, is that the best place? Or should we include the whole `do.call` expression? (not immediately clear to me, though I lean towards the latter)

```
do.call(cbind.data.frame, ...)
        ^ ~ current    |
^         alternative        |
```

Bisaloo: I agree. I have done this by adding an `xml_find_all(, "./parent")` before returning. Please let me know if this is not the preferred option.

MichaelChirico (approved):

Bisaloo:
Hi @MichaelChirico, since I have made a couple of (minor) changes since you approved, I am not completely sure if I should go ahead and merge this.

PR #2856, opened 2025-05-07
Add grepv() awareness to the package

### LINKED ISSUE #2648: Reference new grepv() in regex_subset_linter() once it makes an R release
https://github.com/r-lib/lintr/blob/5990da6a347de8b87566b6d7516e307497ce0680/R/regex_subset_linter.R#L4

Just noticed R will start shipping `grepv()` as a shorthand for `grep(ptn, x, value=TRUE)`:

https://github.com/r-devel/r-svn/blob/874fa1008e6cb090c7c898c422d703ca0026d600/doc/NEWS.Rd#L39-L40

We can reference that once it makes an R release.

### LINKED ISSUE #2855: lintr is not grepv-aware
R 4.5.0 ships `grepv()`, a nice wrapper of `grep(value=TRUE)`.

 - `fixed_regex_linter()` needs to recognize fixed regexes passed to `grepv()`
 - `regex_subset_linter()` should recommend it over `grep(value=TRUE)`

MichaelChirico:
Closes #2855. Closes #2648. Progress on #2737


R/regex_subset_linter.R
```diff
@@ -72,7 +72,7 @@ regex_subset_linter <- function() {
       grep_expr,
       source_expression = source_expression,
       lint_message =
-        "Prefer grep(pattern, x, ..., value = TRUE) over x[grep(pattern, x, ...)] and x[grepl(pattern, x, ...)].",
+        "Prefer grepv(pattern, x, ...) over x[grep(pattern, x, ...)] and x[grepl(pattern, x, ...)].",
```
Bisaloo, line 75: Can we maybe customize this error message based on the R version since lintr is still compatible with R 4.0 to 4.4, which don't know about `grepv()`?

Bisaloo (commented):
Potentially in a follow up PR, but if we start recommending `grepv()` (which I agree with), it would be good to also lint `grep(, value = TRUE)`.

R/regex_subset_linter.R
MichaelChirico: I wouldn't go that way, since the process running {lintr} can be different from the DESCRIPTION dependency, or any other number of incompatibilities between the report and the actual intention.

WDYT about just adding

> Use {backports} if your code might also run on R versions before 4.5.0.

Bisaloo: Sounds good in theory but we would have to wait until https://github.com/r-lib/backports/pull/88 is merged and a new version of backports released.

Maybe the safest option is to be more verbose:

> Prefer grepv(pattern, x, ...), or grep(pattern, x, value = TRUE) if your code might also run on R versions before 4.5.0, over x[grep(pattern, x, ...)] and x[grepl(pattern, x, ...)].

MichaelChirico: Adjusted, PTAL.

Bisaloo (approved):

Bisaloo:
Also fixes https://github.com/r-lib/lintr/issues/2648 I think.

PR #2859, opened 2025-05-30
Add new R 4.5.0 functions to backport_linter & update relative version names

Bisaloo:
I haven't added an item to `NEWS.md` because I'm not completely sure if it should be there, since it's part of the regular housekeeping process.


MichaelChirico:
> I haven't added an item to `NEWS.md` because I'm not completely sure if it should be there, since it's part of the regular housekeeping process.

Borderline case, could go either way. I checked the history and NEWS was omitted in 9dd4f115de67aae4e1806c4f169176c306579c90, so I'm fine without it.

PR #2861, opened 2025-06-04
Clarify error message for backport_linter()

Bisaloo:
Maybe that's just me but I was slightly confused by the error message used in `backport_linter()` since functions released in, e.g. R 3.5.0 are available in (at least some version of) R >= 3.0.0.

I am proposing a slight rephrase which is in my opinion much more explicit.


R/backport_linter.R
```diff
@@ -74,7 +74,7 @@ backport_linter <- function(r_version = getRversion(), except = character()) {
     bad_versions <- unname(backport_index[all_names])
 
     lint_message <- sprintf(
-      "%s (R %s) is not available for dependency R >= %s.",
+      "%s (R %s) is not available for all versions of R >= %s.",
```
IndrajeetPatil: How about keeping it succinct?

```suggestion
      "%s (R %s) is not available for R versions >= %s.",
```

Bisaloo: I believe "all" is key. Its absence is what confused me. Because I was thinking 3.2.0 >= 3.0.0 so it seemed odd to read

> trimws (R 3.2.0) is not available for dependency R >= 3.0.0.

MichaelChirico: What about

```suggestion
      "%s (R %s) is not available for requested dependency R %s.",
```

> trimws (R 3.2.0) is not available for requested dependency R 3.0.0.

Or perhaps, putting it in parentheses

```suggestion
      "%s (R %s) is not always available for requested dependency (R >= %s).",
```

Bisaloo: What about 

> "%s (R %s) is not always available for requested dependency (R >= %s)."

then?

But maybe I'm getting stuck on a "me" thing. If you both believe the current phrasing is good enough, I'm happy with keeping it and closing this PR.

MichaelChirico: no I think there's definitely room for improvement here. I only see "always" in your version as different from my second suggestion, how should I interpret that?

Bisaloo: > I only see "always" in your version as different from my second suggestion, how should I interpret that?

I would put always in both suggestions. I might prefer slightly the second form because it immediately creates a connection with the format used in `DESCRIPTION` but I don't feel very strongly at all.

Issue #2858, opened 2025-05-30
Gitlab / CodeClimate report


lschneiderbauer:
Gitlab supports the consumption of a so called "codequality report" artifact, so that it can display static code checker results in a structured way, e.g. in merge requests, see https://docs.gitlab.com/ci/testing/code_quality/#import-code-quality-results-from-a-cicd-job.

This reports needs to be in a [specific JSON format](https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format) which is a subset of the [CodeClimate format](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types).

Linter like

- [ESLint](https://docs.gitlab.com/ci/testing/code_quality/#eslint)
- [Stylelint](https://docs.gitlab.com/ci/testing/code_quality/#stylelint)
- [MyPy](https://docs.gitlab.com/ci/testing/code_quality/#mypy)
- [Flake8](https://docs.gitlab.com/ci/testing/code_quality/#flake8)
- [Pylint](https://docs.gitlab.com/ci/testing/code_quality/#pylint)
- [Ruff](https://docs.gitlab.com/ci/testing/code_quality/#ruff)
- [golangci-lint](https://docs.gitlab.com/ci/testing/code_quality/#golangci-lint)
- [PMD Copy/Paste Detector](https://docs.gitlab.com/ci/testing/code_quality/#pmd-copypaste-detector)

already support that format.

Would you be interested in a contribution that adds support in the form of a function that consumes an "lints" object and returns a JSON string in the mentioned format?

MichaelChirico:
makes sense. we already have something similar:

https://lintr.r-lib.org/reference/sarif_output.html

MichaelChirico:
I think it depends on what would work best for the tooling, so I leave that up to you.

MichaelChirico:
Closed by #2862

PR #2862, opened 2025-06-07
add `gitlab_output()`: a report format digestible by gitlab

lschneiderbauer:
Addresses #2858.

I tested on gitlab whether gitlab compains about the format by pushing a locally generated report: https://gitlab.com/lschneiderbauer-group/lintr-test/-/merge_requests/1
It seems to work fine.


lschneiderbauer:
@MichaelChirico I would appreciate your feedback.

R/lint.R
```diff
@@ -649,6 +649,72 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
...
+  # location.positions.begin.line
+  #       
+  # severity             String   The severity of the violation, can be one of info, minor, major, critical, or blocker.
+
+  # Gitlab format as R data structure
+  res <-
+    lapply(
+      lints,
+      function(lint) {
+        list(
```
MichaelChirico: Just a suggestion: it might be slightly nicer to use `with(lint, list(description = message, check_name = linter, ...))` to save a few characters.

R/lint.R
```diff
@@ -649,6 +649,72 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
...
+          description = lint$message,
+          check_name = lint$linter,
+          fingerprint = digest::sha1(lint$line),
```
MichaelChirico: is there any difference between `digest::sha1` and `digest::digest(algo="sha1")`? The latter is used elsewhere:

https://github.com/r-lib/lintr/blob/d5905943b8103b4b60e694ae9cbc8421b13a9b28/R/cache.R#L39

R/lint.R
```diff
@@ -649,6 +649,72 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
   write(jsonlite::toJSON(sarif, pretty = TRUE, auto_unbox = TRUE), filename)
 }
 
+#' Gitlab Report for lint results
```
MichaelChirico: nit: the official name is `GitLab`, not `Gitlab`:

https://en.wikipedia.org/wiki/GitLab

R/lint.R
lschneiderbauer: According to the docs:
> The main difference with digest(x, algo = "sha1") is that sha1() will give the same hash on 32-bit and 64-bit systems.

I think that is irrelevant in this case. I changed it to `digest::digest()` to not deviate from the usage already in place.

R/lint.R
lschneiderbauer: I adjusted that everywhere.

R/lint.R
```diff
@@ -649,6 +649,75 @@ sarif_output <- function(lints, filename = "lintr_results.sarif") {
...
+                  begin = line_number
+                )
+              ),
+            severity =
+              switch(type,
+                style = "info",
+                error = "blocker",
+                warning = "major",
+                "info"
+              )
```
MichaelChirico:
```suggestion
            location = list(
              path = filename,
              lines = list(
                begin = line_number
              )
            ),
            severity = switch(type,
              style = "info",
              error = "blocker",
              warning = "major",
              "info"
            )
```

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,30 @@ test_that("Linter() input is validated", {
...
+        description = "Put spaces around all infix operators.",
+        check_name = "infix_spaces_linter",
+        fingerprint = "eb7cc117e8616bd8170fe6aa29e8b0ae849ac6c7",
+        location =
+          list(
+            path = "<text>",
+            lines = list(
+              begin = 1L
+            )
+          ),
```
MichaelChirico:
```suggestion
        location = list(path = "<text>", lines = list(begin = 1L)),
```

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,24 @@ test_that("Linter() input is validated", {
...
+    jsonlite::read_json(tmpfile),
+    list(
+      list(
+        description = "Put spaces around all infix operators.",
+        check_name = "infix_spaces_linter",
+        fingerprint = "eb7cc117e8616bd8170fe6aa29e8b0ae849ac6c7",
+        location = list(path = "<text>", lines = list(begin = 1L)),
+        severity = "info"
+      )
+    )
```
MichaelChirico:
```suggestion
    list(list(
      description = "Put spaces around all infix operators.",
      check_name = "infix_spaces_linter",
      fingerprint = "eb7cc117e8616bd8170fe6aa29e8b0ae849ac6c7",
      location = list(path = "<text>", lines = list(begin = 1L)),
      severity = "info"
    ))
```

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,22 @@ test_that("Linter() input is validated", {
 test_that("typo in argument name gives helpful error", {
   expect_error(lint("xxx", litners = identity), "Found unknown arguments in `...`: `litners`")
 })
+
+
+test_that("gitlab_output() writes expected report", {
+  skip_if_not_installed("jsonlite")
+
+  tmpfile <- withr::local_tempfile()
+  gitlab_output(lint(text = "x<-1"), filename = tmpfile)
```
MichaelChirico: maybe better to specify the linter explicitly, in tests we often find the settings wind up out of whack

```suggestion
  gitlab_output(lint(text = "x<-1", linters = infix_spaces_linter()), filename = tmpfile)
```

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,22 @@ test_that("Linter() input is validated", {
...
+
+test_that("gitlab_output() writes expected report", {
+  skip_if_not_installed("jsonlite")
+
+  tmpfile <- withr::local_tempfile()
+  gitlab_output(lint(text = "x<-1", linters = infix_spaces_linter()), filename = tmpfile)
+
+  expect_identical(
+    jsonlite::read_json(tmpfile),
+    list(list(
```
MichaelChirico, line 267: This `list(list())` is slightly 🤔 to me, I guess it's WAI because the return value should be a `list`, each element of which is a `list` with lint metadata.

Do you think a second test with 2 lints would make that more obvious (or just expand this test to generate 2 lints)?

lschneiderbauer: Yes, the expected JSON format is an array of objects, each object corresponding to a lint. `jsonlite::toJSON()` happens to convert a list of named lists to an array of objects, and so an array with a single object corresponds to a `list(list())` in R.

The test is a particular case with only a single lint. I will add another test with two lints as well as a test with zero lints.

lschneiderbauer: > The test is a particular case with only a single lint. I will add another test with two lints as well as a test with zero lints.

I added those in https://github.com/r-lib/lintr/pull/2862/commits/559800dd045d948bbc7d6afc3b10dbf2a8112e1f.
Anything else?

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,49 @@ test_that("Linter() input is validated", {
...
+test_that("gitlab_output() writes expected report", {
+  skip_if_not_installed("jsonlite")
+
+  tmpfile <- withr::local_tempfile()
+
+  # zero lints: we expect an empty json array
+  gitlab_output(lint(text = "", linters = infix_spaces_linter()), filename = tmpfile)
+  expect_match(
+    readLines(tmpfile),
+    "\\s*\\[\\s*\\]\\s*"
```
MichaelChirico:
```suggestion
    R"(\s*\[\s*\]\s*)"
```

tests/testthat/test-lint.R
```diff
@@ -233,3 +233,49 @@ test_that("Linter() input is validated", {
...
+      description = "Put spaces around all infix operators.",
+      check_name = "infix_spaces_linter",
+      fingerprint = "eb7cc117e8616bd8170fe6aa29e8b0ae849ac6c7",
+      location = list(path = "<text>", lines = list(begin = 1L)),
+      severity = "info"
+    ))
+  )
+
+  # two lints
+  gitlab_output(lint(text = c("x<-1", "x<-1"), linters = infix_spaces_linter()), filename = tmpfile)
```
MichaelChirico: sorry, one last request

```suggestion
  gitlab_output(lint(text = c("x<-1", "y<-2"), linters = infix_spaces_linter()), filename = tmpfile)
```

That way we'll get more than one unique value of `fingerprint` across the different tests.

PR #2863, opened 2025-06-21
lint(text=) finds local settings again

### LINKED ISSUE #2847: `lintr::lint(text=)` no longer reads personal settings
(This doesn't affect a release yet, but is an issue of the recently merged #2805, raised here per request of @MichaelChirico.) 

The recent change from #2805 means that `~/.lintr` settings are no longer read when linting text input (`inline_data`), e.g., `lintr::lint(<R-code-as-string>)`, which is what Emacs-ESS uses to lint code in R buffers. So with the current development version of **lintr** I'd only see notes from the `default_linters`.

Here is an example where the non-default `duplicate_argument_linter()` is added in the settings:

```R
library("lintr")
cat("linters: linters_with_defaults(duplicate_argument_linter())\n", file = tf <- tempfile())
options("lintr.linter_file" = tf)
lint(text = "list(x = 1, x = 2)")  # parse_settings is TRUE by default
## i No lints found.
```

In **lintr** 3.2.0, this did produce a "lint".

MichaelChirico:
Closes #2847, cc @bastistician

PS, I originally pushed this directly to `main` by mistake as b548c789f23efaf6404a50c57bf29a2c29b3a5cd; reverted as dc7a612bb44da150e0e1ef4f6b9a48b23a61210b to put this through normal review

### DIFF
--- a/R/lint.R
+++ b/R/lint.R
@@ -45,9 +45,8 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings = 
 
   needs_tempfile <- missing(filename) || re_matches(filename, rex(newline))
   inline_data <- !is.null(text) || needs_tempfile
-  parse_settings <- !inline_data && isTRUE(parse_settings)
 
-  if (parse_settings) {
+  if (isTRUE(parse_settings)) {
     read_settings(filename)
     on.exit(reset_settings(), add = TRUE)
   }
--- a/R/settings.R
+++ b/R/settings.R
@@ -65,6 +65,7 @@
 read_settings <- function(filename, call = parent.frame()) {
   reset_settings()
 
+  # doing lint(text=) should read settings from the current directory, required e.g. for Emacs
   if (missing(filename)) filename <- "./any_local_file"
   config_file <- find_config(filename)
   default_encoding <- find_default_encoding(filename)
--- a/R/settings.R
+++ b/R/settings.R
@@ -65,6 +65,7 @@
 read_settings <- function(filename, call = parent.frame()) {
   reset_settings()
 
+  if (missing(filename)) filename <- "./any_local_file"
   config_file <- find_config(filename)
   default_encoding <- find_default_encoding(filename)
   if (!is.null(default_encoding)) {
--- a/tests/testthat/test-lint.R
+++ b/tests/testthat/test-lint.R
@@ -233,3 +233,11 @@ test_that("Linter() input is validated", {
 test_that("typo in argument name gives helpful error", {
   expect_error(lint("xxx", litners = identity), "Found unknown arguments in `...`: `litners`")
 })
+
+test_that("settings are picked up under lint(text=)", {
+  .lintr <- withr::local_tempfile(lines = "linters: list(assignment_linter())")
+  withr::local_options(lintr.linter_file = .lintr)
+
+  # lint '=', but not the operator spacing
+  expect_length(lint(text = "a=1"), 1L)
+})
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,7 @@
 ## Bug fixes
 
 * `Lint()`, and thus all linters, ensures that the returned object's `message` attribute is consistently a simple character string (and not, for example, an object of class `"glue"`; #2740, @MichaelChirico).
-* Files with encoding inferred from settings read more robustly under `lint(parse_settings = TRUE)` (#2803, @MichaelChirico).
+* Files with encoding inferred from settings read more robustly under `lint(parse_settings = TRUE)` (#2803, @MichaelChirico). Thanks also to @bastistician for detecting a regression caused by the initial change for users of Emacs (#2847).
 
 ## Changes to default linters
 


MichaelChirico:
Hm, I _think_ this is having an interaction with the lack of {cyclocomp}. IINM package examples of `lint(text=.)` are trying to use our `./.lintr` config, which requires having {cyclocomp}. Testing if 159234d31d17d658935cbb6cccc4a7e0967ffac4 works.

.lintr
```diff
@@ -38,6 +38,7 @@ linters: all_linters(
       `<<-` = NULL
     )),
     unnecessary_concatenation_linter(allow_single_expression = FALSE),
+    cyclocomp_linter = if (requireNamespace("cyclocomp", quietly = TRUE)) cyclocomp_linter(),
```
AshesITR: This we should avoid I think. It could inadvertently disable the linter for its core purpose (linting lintr) if the dev dependencies aren't correct elsewhere.

MichaelChirico: True... any ideas? The problem is that both `R-CMD-check-hard` and `R-CMD-check` GHA use this same config, but the former can't work without `cyclocomp_linter()`.

AshesITR: I think it was a good idea to not parse settings by default for `lint(text=)`.
If we want to go back on that, maybe add a step in R-CMD-check-hard that removes our `.lintr` file entirely before running?

MichaelChirico: I thought so too, however, I don't know how else to restore the Emacs behavior described in the issue.

AshesITR: If I understand the (7 year old) ess-lint code correctly, it gives all arguments to lint() by position.
We could soft-deprecate parsing settings if the code is pushed in via `file=` instead of `text=`.
WDYT?

MichaelChirico: What's the deprecation path, exactly? Would it suffice for tools like `ess-lint` to explicitly supply `parse_settings=TRUE`? Or such tools are required to do full-file parsing instead?

AshesITR: I think explicitly requesting `parse_settings = TRUE` is a good way forward.

bastistician:
Two comments from the outside:

1. The use of `parse_settings=FALSE` in tests doesn't seem to be a bad idea, citing from `vignette("lintr")` (emphasis mine):

   > If only the specified settings should be changed, and the remaining settings should be taken directly from the defaults, the argument `parse_settings = FALSE` can be added to the function calls. This will suppress reading of the .lintr configuration. **This is particularly useful for tests** which should not exclude example files containing lints while the package-level `.lintr` excludes those files because the lints are intentional.

2. For the purpose of `ess-r-flymake`, it seems to be sufficient to keep respecting `options("lintr.linter_file")` when that gives an absolute path, even when linting "inline" code. Maybe this can simplify things and avoid the proposed `read_settings()` kludge `filename <- "./any_local_file"`?

.lintr
MichaelChirico:
```suggestion
```

R/settings.R
```diff
@@ -65,6 +65,8 @@
 read_settings <- function(filename, call = parent.frame()) {
   reset_settings()
 
+  # doing lint(text=) should read settings from the current directory, required e.g. for Emacs, #2847
+  if (missing(filename)) filename <- "./any_local_file"
```
MichaelChirico:
```suggestion
```

tests/testthat/test-lint.R
```diff
@@ -279,3 +278,11 @@ test_that("gitlab_output() writes expected report", {
...
   )
 })
+
+test_that("settings are picked up under lint(text=)", {
+  .lintr <- withr::local_tempfile(lines = "linters: list(assignment_linter())")
+  withr::local_options(lintr.linter_file = .lintr)
+
+  # lint '=', but not the operator spacing
+  expect_length(lint(text = "a=1"), 1L)
+})
```
MichaelChirico, line 297:
```suggestion
```

.lintr
MichaelChirico: ~With the PR to fix ESS, I have this PR structured now as simply a NEWS entry mentioning a known breaking change, PTAL.~

bastistician:
As it stands this PR is merely a NEWS change, so AFAICS will *not* fix #2847. Argument `parse_settings` is already true by default and setting it to `TRUE` explicitly (as recommended in the edited NEWS) makes no difference for 
```r
parse_settings <- !inline_data && isTRUE(parse_settings)
```
inside `lint()`, which evaluates to `FALSE` for `text`-based linting. Am I overlooking something?

MichaelChirico:
You're right:

```r
setwd(tempdir())

options(lintr.linter_file = tmp.lintr<-tempfile())
write.dcf(list(linters = 'list(assignment_linter("="))'), tmp.lintr)

writeLines("a <- 1", tmpr <- tempfile())
lint(tmpr, parse_settings=TRUE)
# /tmp/RtmpkTPVgV/file27119c6c48e30e:1:3: style: [assignment_linter] Use = for assignment, not <-.
# a <- 1
#   ^~
lint(text = "a <- 1", parse_settings=TRUE)
# ℹ No lints found.
lint("a <- 1\n", parse_settings=TRUE)
# ℹ No lints found.
```

R/settings.R
```diff
@@ -65,6 +65,8 @@
...
   config_file <- find_config(filename)
   default_encoding <- find_default_encoding(filename)
```
Bisaloo: I don't love this. Could we maybe do something like:

```suggestion
  if (missing(filename)) {
    location <- "."
  } else {
    location <- filename
  }
  config_file <- find_config(location)
  default_encoding <- find_default_encoding(location)
```

Almost the same but it avoids the confusing use of a non-existing file.

R/lint.R
```diff
@@ -45,7 +45,7 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
 
   needs_tempfile <- missing(filename) || re_matches(filename, rex(newline))
   inline_data <- !is.null(text) || needs_tempfile
-  parse_settings <- !inline_data && isTRUE(parse_settings)
+  parse_settings <- (!inline_data || !missing(parse_settings)) && isTRUE(parse_settings)
```
Bisaloo: Slightly easier to parse but functionally equivalent if I'm not mistaken:

```suggestion
  if (missing(parse_settings)) {
    parse_settings <- !inline_data
  }
```

If not missing, we trust what the user explicitly passed. If missing, we know that the `parse_settings` argument is `TRUE` (default) so only the `inline_data` value matters.

I'm still concerned this is an unintuitive situation. We end up with a different behaviour when relying on the implicit `parse_settings = TRUE` set as the default, and writing explicitly `parse_settings = TRUE`.

bastistician: I agree. It would be more transparent and self-documenting to use `parse_settings = !inline_data` as the formal function argument, taking advantage of lazy evaluation. Referencing other arguments only, as in `parse_settings = is.null(text)` would be better still, but that misses the optional text-mode interpretation of the first argument.

R/settings.R
bastistician: Nice. Regarding the "required e.g. for Emacs" comment: Strictly speaking, a local `filename` is not currently required for ESS because the `ess-r-setup-flymake` hook that runs when entering `ess-r-mode` automatically looks for a `.lintr` config file by itself and sets the absolute path as `options("lintr.linter_file")` before calling `lint(text=, parse_settings=TRUE)` in a sub-process. That option [takes precedence](https://lintr.r-lib.org/articles/lintr.html#other-configuration-options) over the `filename`-based search in `lintr:::find_config`.

R/lint.R
MichaelChirico: Great suggestion, thanks!

MichaelChirico:
Thanks both. I think the PR is in a much better state now.

R/lint.R
```diff
@@ -111,6 +91,33 @@ lint <- function(filename, linters = NULL, ..., cache = FALSE, parse_settings =
   zap_temp_filename(res, needs_tempfile)
 }
 
+lint_impl_ <- function(linters, lint_cache, filename, source_expressions) {
```
MichaelChirico, line 94: we could revert this -- it's only done to keep the cyclomatic complexity down, which had ballooned a bit at the earlier changes, now simplified again.

otoh, it does seem like a natural place for a helper; leaving it in for now.

MichaelChirico:
@AshesITR WDYT about the suggestion to require UTF-8? My sense is you are the developer with the most non-UTF-8 experience.

AshesITR:
In some environments (looking at you, Windows), enforcing UTF-8 requires admin privileges especially when using terminals. I think it's worth keeping an escape hatch so that users running in such an environment aren't left out of their luck. As to default settings I would encourage assuming UTF-8.

PR #2882, opened 2025-07-16
Create download_file_linter()

### LINKED ISSUE #2866: New linter suggestion: `download.url(mode = "w")`
> `download.url(mode = "w")`. On a windows OS, when `mode = "w"`, this can create a corrupt file.

Spin-off from https://github.com/ropensci-review-tools/goodpractice/issues/63 @njtierney

Bisaloo:
First attempt to fix #2866 (cc @maelle).

A potential improvement would be to not lint when `method = "wget"` or `method = "curl"`, as `mode` is ignored in this situation according to `?download.file`.


R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
+#' Recommend usage of a portable `mode` value in [download.file()]
+#'
+#' `mode = "w"` or `mode = "a"` in `download.file()` can generate broken files
+#' on Windows. Instead, `?download.file()` recommends the usage of `mode = "wb"`
```
MichaelChirico:
```suggestion
#' on Windows. Instead, `[utils::download.file()]` recommends the usage of `mode = "wb"`
```

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
+#' Recommend usage of a portable `mode` value in [download.file()]
+#'
+#' `mode = "w"` or `mode = "a"` in `download.file()` can generate broken files
```
MichaelChirico:
```suggestion
#' `mode = "w"` (the default) or `mode = "a"` in `download.file()` can generate broken files
```

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
...
+#'   linters = download_file_linter()
+#' )
+#'
+#' lint(
+#'   text = "download.file(x = my_url, mode = 'w')",
+#'   linters = download_file_linter()
+#' )
+#'
+#' lint(
+#'   text = "download.file(x = my_url, , , , 'w')",
```
MichaelChirico: seems strange someone would use `x=` but then supply several empty arguments to get `'w'` positionally, rather than just supply everything positionally

```suggestion
#'   text = "download.file(my_url, , , , 'w')",
```

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
...
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = "download.file(x = my_url, mode = 'wb')",
+#'   linters = download_file_linter()
+#' )
+#'
+#' lint(
+#'   text = "download.file(x = my_url, , , , 'wb')",
```
MichaelChirico:
```suggestion
#'   text = "download.file(my_url, , , , 'wb')",
```

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
...
+      xml_calls,
+      "parent::expr[1][
+        not(SYMBOL_SUB[text() = 'mode'])
+        and count(OP-COMMA) < 4
+      ]"
+    )
+
+    xml_calls_bad_mode_by_name <- xml_find_all(
+      xml_calls,
+      glue(
```
MichaelChirico: Avoid `glue()` if it's not actually needed as there's some overhead, unless this is done in the factory generator (i.e., the body of `download_file_linter` as opposed to here in the lambda inside `Linter()`).

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
...
+  Linter(linter_level = "expression", function(source_expression) {
+    xml_calls <- source_expression$xml_find_function_calls("download.file")
+
+    # 3 options:
+    # - no explicit value for mode
+    # - mode = "a" or mode = "w"
+    # - similar as previous but specified by position
+    xml_calls_no_mode <- xml_find_all(
+      xml_calls,
+      "parent::expr[1][
```
MichaelChirico: I reckon it will still be preferable to make this all into a single XPath rather than use three `xml_find_all()` invocations, possibly also giving a custom message for the `missing(mode)` case since it may not be clear to readers that they're relying on a bad default.

R/download_file_linter.R
```diff
@@ -0,0 +1,83 @@
...
+      xml_calls,
+      glue(
+        "parent::expr[1][
+          SYMBOL_SUB[text() = 'mode'
+          and following-sibling::expr[1]/STR_CONST[not(contains(text(), 'b'))]]
+        ]"
+      )
+    )
+
+    xml_calls_bad_mode_by_position <- xml_find_all(
```
MichaelChirico: Unfortunately positional matching is not so easy, c.f.

```r
download.file(cacheOK=TRUE, destfile, method, quiet, x='http://...')
```

Will throw a lint in the current logic based on whether the URL has a 'b' in it.

Given how messy positional matching is, unles we have good evidence of people actually using the argument positionally, I would leave this for future work for now.

inst/lintr/linters.csv
```diff
@@ -18,6 +18,7 @@ consecutive_assertion_linter,style readability consistency
 consecutive_mutate_linter,consistency readability configurable efficiency
 consecutive_stopifnot_linter,defunct
 cyclocomp_linter,style readability best_practices configurable
+download_file_linter,common_mistakes best_practices
```
MichaelChirico: `robustness` too right?

MichaelChirico:
> A potential improvement would be to not lint when `method = "wget"` or `method = "curl"`, as `mode` is ignored in this situation according to `?download.file`.

There are a few linters that do similar things you could consult, I'm not sure which is the simplest (`inner_combine_linter()` and `paste_linter()` came to mind but they're both pretty complicated in different ways):

https://github.com/search?q=repo%3Ar-lib%2Flintr%20%2FSYMBOL_SUB%2F&type=code

man/download_file_linter.Rd
```diff
@@ -0,0 +1,38 @@
+% Generated by roxygen2: do not edit by hand
+% Please edit documentation in R/download_file_linter.R
+\name{download_file_linter}
+\alias{download_file_linter}
+\title{Recommend usage of a portable \code{mode} value in \code{\link[=download.file]{download.file()}}}
```
MichaelChirico: Somewhat unusual to see a link in a title, can't say I've come across that often...

In [~508K `\title{}` fields on CRAN](https://github.com/search?q=org%3Acran+path%3A%2F%5C.Rd%24%2F+%2F%5Cn%5C%5Ctitle%5C%7B%2F&type=code), I see

 - [~19K `\code{...}` usages](https://github.com/search?q=org%3Acran%20path%3A%2F%5C.Rd%24%2F%20%2F%5C%5Ctitle%5C%7B.*%5C%5Ccode%2F&type=code)
 - [~3K `\link{...}` usages](https://github.com/search?q=org%3Acran+path%3A%2F%5C.Rd%24%2F+%2F%5C%5Ctitle%5C%7B.*%5C%5Clink%2F&type=code)

Given that we also link it in `\description{}` below, and the `[utils:download.file]` qualification was apparently lost in the `\title{}`. let's drop that link in the `\title{}`.

R/download_file_linter.R
```diff
@@ -0,0 +1,65 @@
+#' Recommend usage of a portable `mode` value in [download.file()]
```
MichaelChirico:
```suggestion
#' Recommend usage of a portable `mode` value for downloading files
```

tests/testthat/test-download_file_linter.R
```diff
@@ -0,0 +1,50 @@
...
+  expect_lint("download.file(x, mode = 'a')", lint_message, linter)
+
+  # 'wb' passed to different argument
+  expect_lint("download.file(x, mode = 'w', method = 'internal', quiet = TRUE, 'wb')", lint_message, linter)
+  expect_lint("download.file(cacheOK = TRUE, destfile, method, quiet, x = 'wb')", lint_message, linter)
+})
+
+test_that("lints vectorize", {
+  lint_message <- rex::rex("download.file() should use mode = 'wb' (or 'ab')")
+
```
MichaelChirico: not used

```suggestion
```

tests/testthat/test-download_file_linter.R
```diff
@@ -0,0 +1,50 @@
...
+  expect_lint(
+    trim_some(
+      "{
+      download.file(x, mode = 'w')
+      download.file(y, mode = 'a')
+    }"
+    ),
```
MichaelChirico:
```suggestion
    trim_some("{
      download.file(x, mode = 'w')
      download.file(y, mode = 'a')
    }"),
```

MichaelChirico (approved):
Looking great! One minor revision needed.

Feel free to merge and handle the `method=` nuances as a follow-up.

PR #2885, opened 2025-07-22
Create all_equal_linter()

### LINKED ISSUE #2610: New linter for if(all.equal(...))
`all.equal()` has a strange return pattern -- it's `TRUE` on "success", but typically character (at least in all the base methods) on "failure".

Therefore it's inappropriate to test `if(all.equal(...))` -- instead we should use `if(isTRUE(all.equal(...)))`.

Here's a couple hundred hits in CRAN packages using `if(all.equal(...))`:

https://github.com/search?q=org%3Acran%20lang%3AR%20%2Fif%5Cs*%5C(%5Cs*all%5B.%5Dequal%5C(%2F&type=code

Note that this advice is very explicit in `?all.equal`:

> Do not use `all.equal` directly in `if` expressions—either use `isTRUE(all.equal(....))` or `identical` if appropriate.

Filed as a new linter since no existing linter comes to mind that would accommodate this. Closest is maybe `vector_logic_linter()` but it's too far a stretch IMO.

Bisaloo:
Fix #2610


R/all_equal_linter.R
MichaelChirico, line 1: Shall we cover `attr.all.equal` too?

> Either TRUE (NULL for attr.all.equal) or a vector of [mode](https://stat.ethz.ch/R-manual/R-devel/library/base/help/mode.html) "character" describing the differences between target and current.

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
+#' Warn about invalid usage of `all.equal()`
+#'
+#' [all.equal()] returns `TRUE` in the absence of differences but return a
+#' character string (not `FALSE`) in the presence of differences.
+#' Usage of `all.equal()` without wrapping it is `isTRUE()` in `if` clauses, or
```
MichaelChirico:
```suggestion
#' Usage of `all.equal()` without wrapping it in `isTRUE()` in `if` clauses, or
```

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
...
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+all_equal_linter <- function() {
+
+  Linter(linter_level = "expression", function(source_expression) {
+    all_equal_calls <- source_expression$xml_find_function_calls("all.equal")
+
+    in_if <- xml_find_all(
+      all_equal_calls,
+      "parent::expr[preceding-sibling::IF]"
```
MichaelChirico: `WHILE` is also affected right?

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
+#' Warn about invalid usage of `all.equal()`
+#'
+#' [all.equal()] returns `TRUE` in the absence of differences but return a
+#' character string (not `FALSE`) in the presence of differences.
+#' Usage of `all.equal()` without wrapping it is `isTRUE()` in `if` clauses, or
+#' preceded by the negation operator `!`, are thus likely to generate unexpected
+#' errors if the compared objects have differences.
+#' An alternative is to use `identical()` to compare vector of strings or when
```
MichaelChirico, line 8: I wouldn't advertise this:

 - `isTRUE()` is specifically designed to overcome shortcomings of `identical(x, TRUE)`
 - I guess it's rare to do something like `identical(all.equal(x, y), "Mean relative difference ...")`

There may be some advantages to `identical()` but I don't think we need to cite them here, too detailed

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
...
+#' )
+#'
+#' # okay
+#' lint(
+#'   text = 'if (isTRUE(all.equal(a, b))) message("equal")',
+#'   linters = all_equal_linter()
+#' )
+#'
+#' lint(
+#'   text = '!identical(a, b)',
```
MichaelChirico, line 35: I would just use `isTRUE()`, I think that's the safest drop-in alternative

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
...
+  Linter(linter_level = "expression", function(source_expression) {
+    all_equal_calls <- source_expression$xml_find_function_calls("all.equal")
+
+    in_if <- xml_find_all(
+      all_equal_calls,
+      "parent::expr[preceding-sibling::IF]"
+    )
+    negated <- xml_find_all(
+      all_equal_calls,
+      "parent::expr[preceding-sibling::OP-EXCLAMATION]"
```
MichaelChirico: If it's not going to get a different `lint_message` (which I guess is not needed in this case), let's just combine XPaths:

```xpath
parent::expr[preceding-sibling::IF or preceding-sibling::OP-EXCLAMATION]
```

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
+#' Warn about invalid usage of `all.equal()`
+#'
+#' [all.equal()] returns `TRUE` in the absence of differences but return a
+#' character string (not `FALSE`) in the presence of differences.
+#' Usage of `all.equal()` without wrapping it is `isTRUE()` in `if` clauses, or
+#' preceded by the negation operator `!`, are thus likely to generate unexpected
+#' errors if the compared objects have differences.
+#' An alternative is to use `identical()` to compare vector of strings or when
+#' exact equality is expected.
```
MichaelChirico, line 9: BTW `stopifnot()` has some built-in logic for handling `all.equal()` checks, c.f.

```
stopifnot(all.equal(1, 1.001))
# Error: 1 and 1.001 are not equal:
#   Mean relative difference: 0.001
```

I might actually lint this code while we're at it:

```r
stopifnot(isTRUE(all.equal(x, y)))
```

Since it gives a worse error:

```r
stopifnot(isTRUE(all.equal(1, 1.001)))
# Error: isTRUE(all.equal(1, 1.001)) is not TRUE
```

Some hits:

https://github.com/search?q=lang%3AR%20%2Fstopifnot%5C(%2F%20%2FisTRUE%5C(all%5C.equal%5C(%2F&type=code

We can do that as a follow-up if you'd like; in this PR at least I'd want to have some mention of `stopifnot()`.

R/all_equal_linter.R
MichaelChirico: I see, this corresponds to the clarified note in the `lint_message`... maybe just add a comment above this example with the same clarification about not needing tolerance?

tests/testthat/test-all_equal_linter.R
```diff
@@ -0,0 +1,34 @@
+test_that("all_equal_linter() skips allowed usages", {
+  linter <- all_equal_linter()
+
+  # Only when used in if
+  expect_no_lint("all.equal(a, b)", linter)
```
MichaelChirico, line 5: Add another test with `tolerance=` proving extra arguments don't matter?

tests/testthat/test-all_equal_linter.R
```diff
@@ -0,0 +1,34 @@
...
+
+  expect_no_lint("if (isTRUE(all.equal(a, b))) message('equal')", linter)
```
MichaelChirico, line 8: may as well test the `!isTRUE(all.equal(...))` case as well?

R/all_equal_linter.R
MichaelChirico: We might as well test `isFALSE(all.equal())` here too, it gets some usage:

https://github.com/search?q=lang%3AR+%2FisFALSE%5C%28all%5C.equal%5C%28%2F&type=code

tests/testthat/test-all_equal_linter.R
```diff
@@ -0,0 +1,34 @@
...
+})
+
+test_that("all_equal_linter() blocks simple disallowed usages", {
+  linter <- all_equal_linter()
+  lint_message <- rex::rex("Wrap all.equal() in isTRUE()")
+
+  expect_lint("if (all.equal(a, b)) message('equal')", lint_message, linter)
```
MichaelChirico, line 20: ditto for an example with `tolerance=`

R/all_equal_linter.R
Bisaloo: I would say probably not. Since it returns `NULL` in the case of no differences, I would expect that it's already wrapped in `is.null()` and thus safe.

R/all_equal_linter.R
Bisaloo: I think it became clearer later in the review but the idea was not to recommend wrapping `all.equal()` in `identical()`. I agree `isTRUE()` is a better solution here.

But I have seen the following thought process happen:

``` r
x <- letters

if (x == letters) {
  message("success")
}
#> Error in if (x == letters) {: the condition has length > 1

# Oh no, it doesn't work. How can I compare that all elements of x are equal to
# elements of letters

if (all.equal(x, letters)) {
  message("success")
}
#> success
```

<sup>Created on 2025-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In reality, in these cases, the author was really looking for `identical()`, not `all.equal()`.

I have the feeling this is what happened in most cases where `all.equal()` compares character vectors (see https://github.com/search?q=org%3Acran%20%2Fall%5C.equal%5C(names%5C(%2F&type=code). This might be a good new case for this linter in a follow up PR.

Bisaloo: Would you have a suggestion to phrase this idea more clearly?

R/all_equal_linter.R
Bisaloo: Yes, `isFALSE(all.equal())` is definitely wrong since it'll always return `FALSE`. I wonder if that'd fit better under a `istrue_linter()` that check cases where authors wanted `!isTRUE()` rather than `isFALSE()`.

Bisaloo: Similarly for `stopifnot()`, great catch but I wonder if it should live here or in `stopifnot_all_linter()` :thinking:

R/all_equal_linter.R
```diff
@@ -0,0 +1,58 @@
...
+    dangerous_unwrapped_all_equal <- xml_find_all(
+      all_equal_calls,
+      "parent::expr[
+        preceding-sibling::OP-EXCLAMATION
+        or preceding-sibling::IF
+        or preceding-sibling::WHILE
+      ]"
+    )
+
+    xml_nodes_to_lints(
```
Bisaloo: If we don't end up covering more cases in this linter, that'd be a good candidate for `make_lint_from_function_xpath()`.

R/all_equal_linter.R
MichaelChirico: Yea, I see what you mean. WDYT about trying to customize the message -- exclude mention of `identical` if we can detect it's definitely not appropriate?

```r
setdiff(names(formals(all.equal.numeric)), names(formals(all.equal.character)))
# [1] "tolerance"   "scale"       "countEQ"     "formatFUN"   "check.class" "giveErr" 
```

R/all_equal_linter.R
MichaelChirico: `isFALSE(all.equal(...))` is pretty rare, I guess it doesn't warrant its own linter. Let's put it here.

I also think the `stopifnot()` bit belongs here, since it's about incorrect usage of `all.equal()`, whereas `stopifnot_all_linter()` is about incorrect usage of `stopifnot()`. WDYT @AshesITR / @IndrajeetPatil ?

R/all_equal_linter.R
MichaelChirico: True, good point, I was underthinking it. I only see one case of `<foo>(attr.all.equal(...))` that's not `is.null`, and it's `c()` which could also be valid.

https://github.com/search?q=lang%3AR+%2F%28%5B%5El%5D%7C%5B%5El%5Dl%7C%5B%5Eu%5Dll%7C%5B%5En%5Dull%7C%5B%5E.%5Dnull%7C%5B%5Es%5D%5B.%5Dnull%7C%5B%5Ei%5Ds%5B.%5Dnull%29%5C%28attr%5C.all%5C.equal%5C%28%2F&type=code

R/all_equal_linter.R
Bisaloo: I view `stopifnot_all_linter()` as "`stopifnot(check)` has actually more bells and whistles than `if (check) stop()`". You can pass it `all.equal()` without `isTRUE()` the same way that you can pass it a logical vector without `all()`,

So it seems like a good fit there.

R/all_equal_linter.R
Bisaloo: I think saying that `identical()` should never be used on numeric vectors still goes a bit too far. In some cases, it might make sense to still use it on numeric vectors. Only the context & the author's intention can determine this.

OTOH, if `tolerance=` is specified, then yes, it's likely `all.equal()` is what the author wants.

NEWS.md
```diff
@@ -51,6 +51,7 @@
 
 ### New linters
 
+* `all_equal_linter()` warns about incorrect use of `all.equal()` in `if` clauses or preceded by `!` (#2885, @Bisaloo).
```
MichaelChirico, line 54: revisit the wording here

R/all_equal_linter.R
MichaelChirico: "never" is too far, but those cases are very rare (basically, low-level libraries only); those authors should know they can `#nolint`, which gives an extra signal to the reader that yes, we really do mean bitwise equality for this double.

(anyway, the wording here is good now)

R/all_equal_linter.R
MichaelChirico: One other thing to consider is that `conjunct_test_linter()` already has some logic related to `stopifnot()`:

https://github.com/r-lib/lintr/blob/dc8f14996fc303d40f120eea6d56fb54187dbc9c/R/conjunct_test_linter.R#L110

If we put the new `all.equal()` lint inside `stopifnot_all_linter()`, I'd want to rename it to just `stopifnot_linter()`, which would also create a tension with `conjunct_test_linter()` about where the check on `stopifnot(A && B)` should live. Fun :)

Not something that needs to be addressed in this PR: #2905

R/all_equal_linter.R
```diff
@@ -0,0 +1,93 @@
...
+#'
+#' @evalRd rd_tags("all_equal_linter")
+#' @seealso [linters] for a complete list of linters available in lintr.
+#' @export
+all_equal_linter <- function() {
+
+  Linter(linter_level = "expression", function(source_expression) {
+    all_equal_calls <- source_expression$xml_find_function_calls("all.equal")
+
+    dangerous_unwrapped_all_equal <- xml_find_all(
```
MichaelChirico, line 52: the logic is slightly off -- we get lints here in r-devel:

https://github.com/r-devel/r-svn/blob/6d2efbc65c5ae1c15ce4f1e25e2fcf5ddc3752d2/src/library/base/R/all.equal.R#L225-L226
https://github.com/r-devel/r-svn/blob/6d2efbc65c5ae1c15ce4f1e25e2fcf5ddc3752d2/src/library/base/R/eigen.R#L38-L41

I guess we're not being careful about which `expr` under `IF` is being linted:

```r
if (A) all.equal(x, y)
```

MichaelChirico:
Found some true positives :)

https://github.com/ycroissant/plm/pull/71

PR #2888, opened 2025-07-23
Apply condition_call_linter() only on string literals

Bisaloo:
Fix #2647

PR #2890, opened 2025-07-24
Support cases of stop(sprintf( in condition_call_linter

Bisaloo:
Follow up from #2888

### DIFF
--- a/R/condition_call_linter.R
+++ b/R/condition_call_linter.R
@@ -57,13 +57,19 @@
 #' @export
 condition_call_linter <- function(display_call = FALSE) {
   call_xpath <- glue::glue("
-    following-sibling::expr/STR_CONST
+    following-sibling::expr[
+      STR_CONST
+      or expr/SYMBOL_FUNCTION_CALL[text() = 'sprintf' or text() = 'gettextf']
+    ]
     and following-sibling::SYMBOL_SUB[text() = 'call.']
       /following-sibling::expr[1]
       /NUM_CONST[text() = '{!display_call}']
   ")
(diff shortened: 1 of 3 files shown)


R/condition_call_linter.R
```diff
@@ -57,13 +57,19 @@
 #' @export
 condition_call_linter <- function(display_call = FALSE) {
   call_xpath <- glue::glue("
-    following-sibling::expr/STR_CONST
+    following-sibling::expr[
+      STR_CONST
+      or expr/SYMBOL_FUNCTION_CALL[text() = 'sprintf' or text() = 'gettextf']
```
MichaelChirico: may as well put `paste()` and `paste0()` here? Maybe even `format()`?

R/condition_call_linter.R
```diff
@@ -57,13 +57,19 @@
...
+    ]
     and following-sibling::SYMBOL_SUB[text() = 'call.']
       /following-sibling::expr[1]
       /NUM_CONST[text() = '{!display_call}']
   ")
   no_call_xpath <- "
-    following-sibling::expr/STR_CONST
+    following-sibling::expr[
+      STR_CONST
+      or expr/SYMBOL_FUNCTION_CALL[text() = 'sprintf' or text() = 'gettextf']
```
MichaelChirico: since it's getting more complicated with the above suggestion, let's factor this out into something we can `glue()` in & share the logic

MichaelChirico (approved):
LGTM, feel free to merge after addressing my small feedback. Thanks!

Issue #2183, opened 2023-09-18
Constant sprintf() calls


Bisaloo:
I see the following unnecessary `sprintf()` across the codebase I maintain:

```r
sprintf("hello")

sprintf("%s", "hello")
```

In other words:

- `sprintf()` does not have any arguments passed to `...` 
- `sprintf()` `fmt` is `"%s"` **and** the next argument in a `STR_CONST`. 

These two examples could be replaced by a more readable and more efficient `"hello"`.

I believe the existing `sprintf_linter()` infrastructure and in particular the logic around piped calls could be re-used here. If you believe this would be a good improvement, I can give it a try.

AshesITR:
I think `sprintf_linter()` is a good place (and `paste_linter()` for the other case. There is no coercion going on in `sprintf("%s", "a")`.

MichaelChirico:
Large number of hits on GitHub:

https://github.com/search?q=%2F%5B%5Ea-z0-9._%5Dsprintf%5B%28%5D%5B%22%27%5D%5B%5E%27%22%5D*%5B%27%22%5D%5B%29%5D%2F+lang%3AR&type=code

MichaelChirico:
There's some subtlety here, it's not quite trivial to implement:

```r
sprintf("%s")         # already lints b/c '%s' is not matched
sprintf("abc")        # needs to lint because there's no template at all
sprintf("def", "ghi") # already lints b/c 'ghi' doesn't correspond to anything
```

i.e. it's not as simple as "find `sprintf()` calls with just one argument", it's got to first subset among those calls that don't already generate a lint.

Not particularly difficult, but requires a bit more care.

AshesITR:
Another edge case to be aware of is `sprintf("%%")` and `sprintf("%%", "")`. The former could lint (replace by `"%"`). The latter already lints.

`contains(text(), '%')` is too simple.

PR #2894, opened 2025-07-26
Update sprintf_linter() to lint constant strings

Bisaloo:
Covers some cases from #2183.

As opposed to what seemed implied in some of the comments, I went with treating the constant strings case before checking if the call is invalid to make the maximum possible as static analysis before evaluating calls.


R/sprintf_linter.R
```diff
@@ -104,15 +101,39 @@ sprintf_linter <- function() {
...
+      get_r_string(sprintf_calls, "OP-LEFT-PAREN/following-sibling::expr[1]/STR_CONST")
+    )
+
+    fmt <- ifelse(!is.na(fmt_by_name), fmt_by_name, fmt_by_pos)
+    constant_fmt <- !is.na(fmt) & !grepl("%[^%]", fmt)
+
+    constant_fmt_lint <- xml_nodes_to_lints(
+      sprintf_calls[constant_fmt],
+      source_expression = source_expression,
+      lint_message = "A constant string is used and the sprintf call can be removed",
```
Bisaloo: This message should be improved to use the actual function, which can be gettextf, rather than sprintf.

tests/testthat/test-sprintf_linter.R
```diff
@@ -23,7 +22,10 @@ patrick::with_parameters_test_that(
     linter <- sprintf_linter()
     unused_arg_msg <- if (getRversion() >= "4.1.0") "one argument not used by format" else NULL
 
-    expect_lint(paste0(call_name, "('hello', 1)"), unused_arg_msg, linter)
+    expect_lint(paste0(call_name, "('hello', 1)"), "constant", linter)
+
+    expect_lint(paste0(call_name, "('hello')"), "constant", linter)
+    expect_lint(paste0(call_name, "('%%')"), "constant", linter)
```
MichaelChirico: let's add more examples with more complicated escapes like

```
%%s
%%%
%%%s
%%%%
%%%%s
```

What should be the behavior for each of these?

R/sprintf_linter.R
```diff
@@ -104,15 +101,44 @@ sprintf_linter <- function() {
...
+
+    fct_name <- xml_text(xml_find_all(
+      sprintf_calls,
+      "expr/SYMBOL_FUNCTION_CALL[1]/text()"
+    ))
+
+    constant_fmt_lint <- xml_nodes_to_lints(
+      sprintf_calls[constant_fmt],
+      source_expression = source_expression,
+      lint_message = sprintf("A constant string is used and the %s call can be removed", fct_name[constant_fmt]),
```
MichaelChirico: nit: the preferred lint message format is "action, reason", here is "reason, action"

R/sprintf_linter.R
```diff
@@ -104,15 +101,44 @@ sprintf_linter <- function() {
...
     sprintf_calls <- xml_find_all(xml_calls, call_xpath)
+    in_pipeline <- !is.na(xml_find_first(sprintf_calls, in_pipe_xpath))
+
+    fmt_by_name <- get_r_string(
+      sprintf_calls,
+      "SYMBOL_SUB[text() = 'fmt']/following-sibling::expr[1]/STR_CONST"
+    )
+    fmt_by_pos <- ifelse(
+      in_pipeline,
+      get_r_string(sprintf_calls, "preceding-sibling::*[2]/STR_CONST"),
```
MichaelChirico: I reckon this will conflict with #2902

MichaelChirico (commented):
I think it's ready to go, just another round of small tweaks requested. Thanks!

R/sprintf_linter.R
```diff
@@ -104,15 +103,44 @@ sprintf_linter <- function() {
...
+      in_pipeline,
+      get_r_string(sprintf_calls, "preceding-sibling::*[not(self::COMMENT)][2]/STR_CONST"),
+      get_r_string(sprintf_calls, "OP-LEFT-PAREN/following-sibling::expr[1]/STR_CONST")
+    )
 
-    sprintf_warning <- vapply(sprintf_calls, capture_sprintf_warning, character(1L))
+    fmt <- ifelse(!is.na(fmt_by_name), fmt_by_name, fmt_by_pos)
+    constant_fmt <- !is.na(fmt) & !grepl("%", gsub("%%", "", fmt, fixed = TRUE), fixed = TRUE)
+
+    fct_name <- xml_text(xml_find_all(
```
MichaelChirico: I think this needs to be `xml_find_first()`, which is guaranteed to return a same-size vector. I _think_ `xml_find_all()` happens to work fine in this case, but `xml_find_first()` should be faster and is anyway safer.

Even better is probably to use `xp_call_name()`

https://github.com/r-lib/lintr/blob/38db1acd5460379661dca96947eeb68b0119489e/R/xp_utils.R#L79

MichaelChirico (approved):
Great, thanks!

PR #2899, opened 2025-07-26
Fuzz comments

MichaelChirico:
Part of #2191 -- continuing #2821 (which is included here & against which this is based, for now). Also progress on https://github.com/r-lib/lintr/issues/2737.

 Closes #2824.  Closes #2825.  Closes #2826. Closes #2827.

Wow, this one really spiralled out of control & became quite a quixotic endeavor in the end.

A lot of it is really edge-casey, but better to have the fix than not. A surprising number of legit issues / bugs (i.e. not false positive/negative, but {lintr} errors based on faulty assumptions).

Probably, some pieces of this belong in dedicated PRs separate from this one.

Leaving this open for now, apparently it makes things a bit easier.

PR #2900, opened 2025-07-26
Handle comments in unreachable_code_linter

MichaelChirico:
Split off from #2899, this one is large enough to get its own PR. Also progress on #2737.

### DIFF
--- a/R/unreachable_code_linter.R
+++ b/R/unreachable_code_linter.R
@@ -76,33 +76,55 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 unreachable_code_linter <- function(allow_comment_regex = getOption("covr.exclude_end", "# nocov end")) {
+  # nolint next: object_usage_linter. Used in glue() in statically-difficult fashion to detect.
   expr_after_control <- "
     (//REPEAT | //ELSE | //FOR)/following-sibling::expr[1]
     | (//IF | //WHILE)/following-sibling::expr[2]
   "
+
+  unreachable_expr_cond_ws <- "
+  following-sibling::*[
+    not(self::OP-RIGHT-BRACE or self::OP-SEMICOLON or self::ELSE or preceding-sibling::ELSE)
+    and (not(self::COMMENT) or @line2 > preceding-sibling::*[not(self::COMMENT)][1]/@line2)
+  ][1]"
+  # when a semicolon is present, the condition is a bit different due to <exprlist> nodes
+  unreachable_expr_cond_sc <- "
+  parent::exprlist[OP-SEMICOLON]
+    /following-sibling::*[
+      not(self::OP-RIGHT-BRACE)
+      and (not(self::COMMENT) or @line1 > preceding-sibling::exprlist/expr/@line2)
+    ][1]
+  "
+
   # NB: use not(OP-DOLLAR) to prevent matching process$stop(), #1051
-  xpath_return_stop <- glue("
+  xpath_return_stop_fmt <- "
   (
     {expr_after_control}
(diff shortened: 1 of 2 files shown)


tests/testthat/test-unreachable_code_linter.R
```diff
@@ -55,44 +55,43 @@ test_that("unreachable_code_linter works in sub expressions", {
...
-      }
-    }
-  ")
+    "),
+    linter
+  )
 
-  expect_lint(
-    lines,
+   expect_lint(
```
MichaelChirico:
```suggestion
  expect_lint(
```

R/unreachable_code_linter.R
```diff
@@ -76,33 +76,55 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 unreachable_code_linter <- function(allow_comment_regex = getOption("covr.exclude_end", "# nocov end")) {
+  # nolint next: object_usage_linter. Used in glue() in statically-difficult fashion to detect.
```
AshesITR: Maybe we can both remove the lint and break up the code into more digestible parts by only glueing once like

```r
glue::glue("
  ({expr_after_control} | {last_function_call})/{unreachable_expr_ws} |
  ({expr_after_control} | {last_function_call})/{unreachable_expr_sc} |
  ({expr_after_control})/{next_or_break}/{unreachable_expr_ws} |
  ({expr_after_control})/{next_or_break}/{unreachable_expr_sc}
")
```

WDYT?

MichaelChirico: I think I did what you had in mind -- merge the logic for the [return/stop] and [next/break] cases, then use lint customization to give the right message in every case. That includes improving the customization beyond what existed already to mention `return()` / `stop()` specifically.

MichaelChirico: Checking back :)

R/unreachable_code_linter.R
```diff
@@ -80,32 +80,48 @@ unreachable_code_linter <- function(allow_comment_regex = getOption("covr.exclud
...
-    /following-sibling::*[
-      not(self::OP-RIGHT-BRACE or self::OP-SEMICOLON)
-      and (not(self::COMMENT) or @line2 > preceding-sibling::*[1]/@line2)
-    ][1]
-  ")
-  xpath_next_break <- glue("
-  ({expr_after_control})
-    /expr[NEXT or BREAK]
+
+  unreachable_expr_cond_ws <- "
```
MichaelChirico, line 85:
```suggestion
  # normal case: expression after terminal call is on the next line
  unreachable_expr_cond_ws <- "
```

R/unreachable_code_linter.R
```diff
@@ -80,32 +80,48 @@ unreachable_code_linter <- function(allow_comment_regex = getOption("covr.exclud
...
+  following-sibling::*[
+    not(self::OP-RIGHT-BRACE or self::OP-SEMICOLON or self::ELSE or preceding-sibling::ELSE)
+    and (not(self::COMMENT) or @line2 > preceding-sibling::*[not(self::COMMENT)][1]/@line2)
+  ][1]"
+  # when a semicolon is present, the condition is a bit different due to <exprlist> nodes
```
MichaelChirico:
```suggestion
  # robustness case: expression after terminal call is on the same line (after a semicolon),
  #   wherefore the condition is a bit different due to <exprlist> nodes
```

MichaelChirico:
So as not to get stuck indefinitely due to low reviewer bandwidth, I'm trying out LLM-assisted reviews.

I'm not sure the right contours for this yet, and want to keep iterating on the prompts to do so well. Some principles I have in mind:

 - Only resort to this on PRs stuck in queue for a while (O(weeks))
 - Keep some transparency about the model+prompt used
 - Allow some time (O(days)) after the LLM review for human(s) to chime in before accepting the change

PR #2901, opened 2025-07-26
Logic to make linters robust to adversarial comments

MichaelChirico:
Split of from #2899. Progress on #2737.

I tried to include here those linters that only fail under "adversarial" commenting, which is a bit subjective, the idea is to basically only consider cases of comments that are not particularly common in practice.

### DIFF
--- a/R/object_usage_linter.R
+++ b/R/object_usage_linter.R
@@ -61,13 +61,21 @@ object_usage_linter <- function(interpret_glue = NULL, interpret_extensions = c(
   # NB: the repeated expr[2][FUNCTION] XPath has no performance impact, so the different direct assignment XPaths are
   #   split for better readability, see PR#1197
   # TODO(#1106): use //[...] to capture assignments in more scopes
-  xpath_function_assignment <- "
-    expr[LEFT_ASSIGN or EQ_ASSIGN]/expr[2][FUNCTION or OP-LAMBDA]
-    | expr_or_assign_or_help[EQ_ASSIGN]/expr[2][FUNCTION or OP-LAMBDA]
-    | equal_assign[EQ_ASSIGN]/expr[2][FUNCTION or OP-LAMBDA]
-    | //SYMBOL_FUNCTION_CALL[text() = 'assign']/parent::expr/following-sibling::expr[2][FUNCTION or OP-LAMBDA]
-    | //SYMBOL_FUNCTION_CALL[text() = 'setMethod']/parent::expr/following-sibling::expr[3][FUNCTION or OP-LAMBDA]
-  "
+  fun_node <- "FUNCTION or OP-LAMBDA"
+  xpath_function_assignment <- glue("
+    expr[LEFT_ASSIGN or EQ_ASSIGN]/expr[2][{fun_node}]
+    | expr_or_assign_or_help[EQ_ASSIGN]/expr[2][{fun_node}]
+    | equal_assign[EQ_ASSIGN]/expr[2][{fun_node}]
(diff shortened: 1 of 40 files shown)


R/object_usage_linter.R
```diff
@@ -61,13 +61,22 @@ object_usage_linter <- function(interpret_glue = NULL, interpret_extensions = c(
...
+  fun_node <- "FUNCTION or OP-LAMBDA"
+  xpath_function_assignment <- glue("
+    expr[LEFT_ASSIGN or EQ_ASSIGN]/expr[2][{fun_node}]
+    | expr_or_assign_or_help[EQ_ASSIGN]/expr[2][{fun_node}]
+    | equal_assign[EQ_ASSIGN]/expr[2][{fun_node}]
+    | //SYMBOL_FUNCTION_CALL[text() = 'assign']/parent::expr/following-sibling::expr[2][{fun_node}]
+    | //SYMBOL_FUNCTION_CALL[text() = 'setMethod']/parent::expr/following-sibling::expr[3][{fun_node}]
+  ")
+
+  # code like:content
```
AshesITR:
```suggestion
  # code like:
```

R/utils.R
```diff
@@ -85,20 +85,20 @@ names2 <- function(x) {
   names(x) %||% rep("", length(x))
 }
 
-get_content <- function(lines, info) {
+get_content <- function(lines, info, known_safe = TRUE) {
```
AshesITR: The argument name is not really clear.
Maybe invert and name it `needs_braces`?

MichaelChirico:
Bumping :)

tests/testthat/test-if_switch_linter.R
```diff
@@ -207,17 +215,17 @@ test_that("max_branch_lines= and max_branch_expressions= arguments work", {
...
     max_lines4_linter
   )
-  expect_lint(five_expr_three_lines_lines, NULL, max_expr2_linter)
-  expect_lint(five_expr_three_lines_lines, NULL, max_expr4_linter)
+  expect_no_lint(five_expr_three_lines_lines, max_expr2_linter)
+  expect_no_lint(five_expr_three_lines_lines, max_expr4_linter)
 })
 
-test_that("max_branch_lines= and max_branch_expressions= block over-complex switch() too", {
+test_that("max_branch_lines= and max_branch_expressions= block over-complex switch() too", { # nofuzz
```
Bisaloo, line 228: It is possible to add a note next to # nofuzz comments explaining why they are required?

R/utils.R
```diff
@@ -85,20 +85,20 @@ names2 <- function(x) {
...
-        as.integer(xml_attr(info, attr))
-      })
+      info <- lapply(xml2::xml_attrs(info), as.integer)
     }
 
     lines <- lines[seq(info$line1, info$line2)]
     lines[length(lines)] <- substr(lines[length(lines)], 1L, info$col2)
     lines[1L] <- substr(lines[1L], info$col1, nchar(lines[1L]))
   }
+  if (needs_braces) lines <- c("{", lines, "}")
```
Bisaloo, line 108: Can we also add a comment here indicating when this argument can be helpful?

tests/testthat/test-if_switch_linter.R
MichaelChirico: Yes but let's try to do that after #2832 where we at least get a more specific reason for each `#nofuzz`. Often it makes the reasoning more obvious when the fuzzer being avoided is named.

R/utils.R
MichaelChirico: Added a comment, there was already some explanation in `object_usage_linter()`, I'm not sure where it's best to write out everything, but some duplication here is helpful. I also updated the comment given the upstream bug was addressed (as `WONTFIX`).

PR #2907, opened 2025-07-30
Implement "auto" r_version value in backport_linter()

### LINKED ISSUE #1974: Make `backport_linter()` more informative by default

Bisaloo:
Opening as a draft for a first general impression.

If this seems good, I still need to:

- [ ] Add a `NEWS.md` item
- [ ] Add tests

Should this be the new default value for `backport_linter(r_version=)`?

Fix #1974.


R/settings_utils.R
```diff
@@ -112,3 +112,20 @@ pkg_name <- function(path = find_package()) {
   nm <- read.dcf(file.path(path, "DESCRIPTION"), fields = "Package")[1L]
   if (!is.na(nm)) nm
 }
+
+min_r_version <- function(path = find_package(".")) {
+  if (is.null(path)) {
+    return(getRversion())
```
AshesITR: This is basically disabling the linter if no DESCRIPTION is found. I would expect a warning message detailing how to properly configure backports_linter.

Bisaloo: `getRversion()` is the currently default so I don't really think of it as disabling the linter, but more as reverting to the old default behaviour.

Does [`b68d97e` (#2907)](https://github.com/r-lib/lintr/pull/2907/commits/b68d97e78ba1ad07db967325ae2b9e9a06916c3a) help?

MichaelChirico:
LMK if you'd like this to be included in the pending release (I'll probably submit this weekend)

PR #2911, opened 2025-07-30
Don't warn if cyclocomp is missing but it's explicitly excluded

### LINKED ISSUE #2909: cyclocomp warning when not using cyclocomp_linter
https://github.com/r-lib/lintr/commit/fdb001c4a5e63dced8d123c1d6f8527e41fdec21 moved the cyclocomp namespace check which now means having a .lintr.R file like the following
```R
linters <- all_linters(
    indentation_linter(indent = 4L),
    cyclocomp_linter = NULL
)
```
causes a package check to become somewhat noisy and confusing, e.g.

```R
lintr::lint_package()
#> Warning message:
#> Warning encountered while loading config:
#> Caused by warning:
#> ! Cyclocomp complexity is computed using `cyclocomp::cyclocomp()`.
#> ℹ Please install the needed cyclocomp package. 
```
Only noticed this change today whilst playing with the dev version of lintr.  Any chance this could be improved?

MichaelChirico:
Closes #2909. A bit ugly, not super satisfied with the fix.

It may be possible to defer the `call_linter_factory` step until `modify_defaults()`, but that would change the behavior of both `linters_with_tags()` and `modify_defaults()`.

### DIFF
--- a/R/with.R
+++ b/R/with.R
@@ -97,10 +97,16 @@ linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "dep
   }
   tagged_linters <- list()
 
+  dots <- list(...)
+  exclude_idx <- vapply(dots, is.null, logical(1L))
+  excluded_linters <- names(dots)[exclude_idx]
+  dots <- dots[!exclude_idx]
+
   for (package in packages) {
     pkg_ns <- loadNamespace(package)
     ns_exports <- getNamespaceExports(pkg_ns)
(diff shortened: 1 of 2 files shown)


MichaelChirico:
This totally breaks how `guess_names()` works & I don't see an easy fix that way.

MichaelChirico:
Closing for now, I also don't see an easy way to do this by modifying `modify_defaults()` because that expects to receive evaluated factories, allowing evaluation there contradicts our earlier deprecation of that functionality. There might be a way to get there with `delayedAssign()`...

PR #2912, opened 2025-07-30
Delay calling linter factories until modify_defaults

### LINKED ISSUE #2909: cyclocomp warning when not using cyclocomp_linter

MichaelChirico:
Closes #2909 

This still has some kinks to iron out, but feels more promising than the earlier approach.

### DIFF
--- a/R/with.R
+++ b/R/with.R
@@ -34,8 +34,8 @@ modify_defaults <- function(defaults, ...) {
   if (missing(defaults)) {
     cli_abort("{.arg defaults} is a required argument, but is missing.")
   }
-  if (!is.list(defaults) || !all(nzchar(names2(defaults)))) {
-    cli_abort("{.arg defaults} must be a named list, not {.obj_type_friendly {defaults}}.")
+  if (!(is.list(defaults) || is.environment(defaults)) || !all(nzchar(names2(defaults)))) {
+    cli_abort("{.arg defaults} must be a named list or environment, not {.obj_type_friendly {defaults}}.")
   }
   vals <- list(...)
   nms <- names2(vals)
@@ -53,7 +53,8 @@ modify_defaults <- function(defaults, ...) {
   }
 
   is.na(vals) <- nms == vals
-  defaults[nms] <- vals
+  for (ii in seq_along(nms)) defaults[[nms[ii]]] <- vals[[ii]]
+  if (is.environment(defaults)) defaults <- as.list(defaults)
 
   res <- defaults[!vapply(defaults, is.null, logical(1L))]
   res <- res[platform_independent_order(names(res))]
@@ -95,7 +96,7 @@ linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "dep
   if (!is.character(tags) && !is.null(tags)) {
     cli_abort("{.arg tags} must be a character vector, or {.code NULL}, not {.obj_type_friendly {tags}}.")
   }
-  tagged_linters <- list()
+  tagged_linter_env <- new.env()
 
   for (package in packages) {
     pkg_ns <- loadNamespace(package)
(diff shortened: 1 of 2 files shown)


tests/testthat/test-with.R
```diff
@@ -118,3 +118,12 @@ test_that("all_linters respects ellipsis argument", {
...
 })
+
+test_that("Excluding cyclocomp linter avoids a warning", {
+  local_mocked_bindings(
+    requireNamespace = function(pkg, ...) pkg != "cyclocomp" || base::requireNamespace(pkg, ...)
+  )
+
+  expect_silent(all_linters(cyclocomp_linter = NULL))
+  expect_silent(linters_with_tags("configurable", cyclocomp_linter = NULL))
+})
```
AshesITR, line 136: Add an expect_warning case too?

R/with.R
```diff
@@ -112,18 +113,14 @@ linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "dep
...
       }
-      linter_factories <- mget(available$linter, envir = pkg_ns)
-      linters <- Map(
-        call_linter_factory,
-        linter_factory = linter_factories,
-        linter_name = names(linter_factories),
-        MoreArgs = list(package = package)
-      )
-      tagged_linters <- c(tagged_linters, linters)
+      for (linter in available$linter) {
```
MichaelChirico: Note to self -- it looks like `delayedAssign()` doesn't snapshot the value of `linter` that exists at this line, instead, when it's actually evaluated, it winds up with the version of `linter` that exists in this environment later, namely, that's `tail(available$linter, 1)`.

So to proceed with this we'd need to do something like `local()` to force the symbol to evaluate as expected later.

MichaelChirico:
It looks like Gemini's suggested alternative approach is in fact what I'd originally tried :)

So I had a follow-up:

----

Prompt:

My memory is something to that effect was tried earlier in this branch. Please read through the successive commit diffs to get further context on what else was tried.
  (Note that you'll be missing the context on the associated GitHub Actions, so you can't tell if a given commit actually passed tests)

----

Response:

You are correct. Having reviewed the commit history, it's clear that the solution evolved. The local() approach was an intermediate step to fix a subtle bug in
  delayedAssign's variable capture within the loop, and the final implementation using a helper function is a further refinement for clarity.

  Here is my updated review:

  ---

  Thank you for the excellent work and for the thoughtful evolution of this solution. The use of `delayedAssign` is a very clever approach to deferring the package
  checks.

  My initial thought was to use a `local()` block to handle the variable scoping issue with `delayedAssign` inside the loop. However, after reviewing the commit history, I
  see that you had already implemented that exact solution in commit b90b3c5f and then refined it further.

  The final implementation in 1051124e which abstracts this logic into the `lazily_assign_linter_` helper is a definite improvement. It elegantly sidesteps the scoping
  problem by leveraging function argument evaluation, which is more idiomatic and self-documenting than the `linter <- linter` trick required inside the `local()` block.
  The detailed comment explaining the rationale is also very helpful for future maintainers.

  The tests are thorough and correctly verify both the intended fix and the preservation of the warning in normal cases.

  This is a well-crafted solution to a tricky problem. I have no further suggestions. Great job.

----

PR #2919, opened 2025-08-04
feat: argument to disable linting of `(` for auto-printing

### LINKED ISSUE #2916: FR: add argument to disable linting of `(` for auto-printing  in implicit_assignment_linter
Would it be possible to gain a parameter in the `implicit_assignment_linter` to allow for implicit printing? At present, I see no way to disable this, e.g

``` r
lintr::lint(
    text = "(a <- 1)", 
    linters = lintr::implicit_assignment_linter()
)
#> <text>:1:2: warning: [implicit_assignment_linter] Call print() explicitly instead of relying on implicit printing behavior via '('.
#> (a <- 1)
#>  ^~~~~~

lintr::lint(
    text = "(a <- 1)", 
    linters = lintr::implicit_assignment_linter(except = "`(`")
)
#> <text>:1:2: warning: [implicit_assignment_linter] Call print() explicitly instead of relying on implicit printing behavior via '('.
#> (a <- 1)
#>  ^~~~~~
```

TimTaylor:
Adds an argument to `implicit_assignment_linter` to disable linting of `(` for auto-printing.
closes #2916

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,7 +51,7 @@
 * `use_lintr()` adds the created `.lintr` file to the `.Rbuildignore` if run in a package (#2926, initial work by @MEO265, finalized by @Bisaloo).
 * `implicit_assignment_linter()` gains argument `allow_print` to disable lints for the use of `(` for auto-printing (#2919, @TimTaylor).
 * `length_test_linter()` is extended to check incorrect usage of `nrow()`, `ncol()`, `NROW()`, `NCOL()` (#2933, @mcol).
-* `implicit_assignment_linter()` gains argument `allow_print` to disable lints for the use of `(` for auto-printing (#2919, @TimTaylor).
+* `implicit_assignment_linter()` gains argument `allow_paren_print` to disable lints for the use of `(` for auto-printing (#2919, @TimTaylor).
 
 ### New linters
 
--- a/R/implicit_assignment_linter.R
+++ b/R/implicit_assignment_linter.R
@@ -9,7 +9,7 @@
 #' @param allow_scoped Logical, default `FALSE`. If `TRUE`, "scoped assignments",
 #'   where the object is assigned in the statement beginning a branch and used only
 #'   within that branch, are skipped.
-#' @param allow_print Logical, default `FALSE`. If `TRUE`, using `(` for auto-printing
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -48,6 +48,7 @@
 * `object_usage_linter()` lints missing packages that may cause false positives (#2872, @AshesITR)
 * New argument `include_s4_slots` for the `xml_find_function_calls()` entry in the `get_source_expressions()` to govern whether calls of the form `s4Obj@fun()` are included in the result (#2820, @MichaelChirico).
 * `sprintf_linter()` lints `sprintf()` and `gettextf()` calls when a constant string is passed to `fmt` (#2894, @Bisaloo).
+* `implicit_assignment_linter()` gains argument `allow_print` to disable lints for the use of `(` for auto-printing (#2919, @TimTaylor).
```
MichaelChirico, line 52: I think `allow_implicit_print` may be better, WDYT? My only hesitation is doubling "implicit" in the linter & argument names...

R/implicit_assignment_linter.R
```diff
@@ -116,6 +130,9 @@ implicit_assignment_linter <- function(except = c("bquote", "expression", "expr"
     bad_expr <- xml_find_all(xml, xpath)
 
     print_only <- !is.na(xml_find_first(bad_expr, "parent::expr[parent::exprlist and *[1][self::OP-LEFT-PAREN]]"))
+    if (allow_print) {
+      bad_expr <- bad_expr[!print_only]
```
MichaelChirico, line 134: This works, I think it will be ideal to work this in to the up-front XPath logic, rather than including those hits in the search only to drop them.

Have a try and if you get stuck we can mark it as a follow-up issue.

TimTaylor: I thought that initially but then I realised you need to do the `print_only` for the error message anyway so this seemed to be the most KISS approach. WDY?

NEWS.md
TimTaylor: `allow_bracket_print`?

I'm already wary of `implicit` as a description as it already doesn't reflect how I think of my usage of this lint ... basically "stop people accidentally using `<-` in function calls".  I'd prefer not to use the term more if possible.

MichaelChirico: Let's call it `allow_paren_print`

PR #2921, opened 2025-08-05
Lint `identical(x, sort(x))` in `sort_linter()`

### LINKED ISSUE #2908: Recommend `!is.unsorted()` instead of `identical(x, sort(x))` in `sort_linter()`
Not a super common pattern:

https://github.com/search?q=language%3AR+%2Fidentical%5C%28%5Cw%2B%2C+sort%5C%28%5Cw%2B%5C%29%5C%29%2F&type=code

But there is a huge opportunity for optimization here:

``` r
x <- runif(1e5)

  identical(x, sort(x)),
  !is.unsorted(x),
  check = "identical"
)
#> Unit: microseconds
#>                   expr      min       lq       mean   median       uq      max
#>  identical(x, sort(x)) 4010.383 4313.883 4636.98955 4496.855 4761.682 7040.232
#>        !is.unsorted(x)   40.104   41.166   46.29157   42.666   46.807   86.509
#>  neval
#>    100
#>    100

x <- sort(x)

  identical(x, sort(x)),
  !is.unsorted(x),
  check = "identical"
)
#> Unit: nanoseconds
#>                   expr  min     lq    mean median     uq   max neval
#>  identical(x, sort(x)) 3612 3732.0 4094.85 3872.5 4033.0 22994   100
#>        !is.unsorted(x)  662  722.5  803.60  773.0  800.5  3131   100
```

<sup>Created on 2025-07-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Should probably not lint if `na.last = TRUE` or `FALSE` or `decreasing = TRUE` in `sort()`.

Bisaloo:
I initially tried to re-use the existing `sorted_xpath` but I think customizing the error message was then too messy and unclear.

Fix #2908

### DIFF
--- a/R/sort_linter.R
+++ b/R/sort_linter.R
@@ -85,11 +85,10 @@ sort_linter <- function() {
   sorted_xpath <- "
   parent::expr[not(SYMBOL_SUB)]
     /parent::expr[
-      (EQ or NE)
+      (EQ or NE or expr/SYMBOL_FUNCTION_CALL/text() = 'identical')
       and expr/expr = expr
     ]
   "
-
 
   arguments_xpath <-
     ".//SYMBOL_SUB[text() = 'method' or text() = 'decreasing' or text() = 'na.last']"
--- a/R/sort_linter.R
+++ b/R/sort_linter.R
@@ -85,7 +85,14 @@ sort_linter <- function() {
   sorted_xpath <- "
   parent::expr[not(SYMBOL_SUB)]
     /parent::expr[
-      (EQ or NE or expr/SYMBOL_FUNCTION_CALL/text() = 'identical')
+      (EQ or NE)
+      and expr/expr = expr
+    ]
+  "
+  sorted_identical_xpath <- "
+  parent::expr[not(SYMBOL_SUB)]
+    /parent::expr[
+      expr/SYMBOL_FUNCTION_CALL[text() = 'identical']
       and expr/expr = expr
     ]
   "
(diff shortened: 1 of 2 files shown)


tests/testthat/test-sort_linter.R
```diff
@@ -108,15 +108,19 @@ test_that("sort_linter blocks simple disallowed usages", {
...
   expect_lint("x == sort(x)", sorted_msg, linter)
+  expect_lint("identical(sort(x), x)", sorted_msg, linter)
 
   # inverted version
   expect_lint("sort(x) != x", unsorted_msg, linter)
+  expect_lint("!identical(x, sort(x))", unsorted_msg, linter)
 
   # expression matching
   expect_lint("sort(foo(x)) == foo(x)", sorted_msg, linter)
+  expect_lint("identical(foo(x), sort(foo(x)))", sorted_msg, linter)
```
MichaelChirico, line 153: Let's add some more tests against false positives, like `identical(sort(x), y)`, `identical(sort(foo(x)), x)`, and a test that this catches expressions: `identical(sort(foo(x)), foo(x))`

R/sort_linter.R
```diff
@@ -89,7 +89,13 @@ sort_linter <- function() {
       and expr/expr = expr
     ]
   "
-
+  sorted_identical_xpath <- "
+  parent::expr[not(SYMBOL_SUB)]
+    /parent::expr[
+      expr/SYMBOL_FUNCTION_CALL[text() = 'identical']
+      and expr/expr = expr
```
MichaelChirico: We may need more baroque handling here, _a la_ #2901 and `strip_comments_from_subtree()`. Try test cases like:

```r
identical(x, sort( # comment
  x
))
identical(foo(x), sort(foo( # comment
  x
)))
```

Bisaloo: Do I need to add these tests explicitly or is this now handled by the fuzzer?

MichaelChirico: right, the fuzzer should pick up if there's an issue (now that it's merged)

PR #2926, opened 2025-08-18
Add .lintr to .Rbuildignore in use_lintr()

### LINKED ISSUE #1805: `use_lintr()` should add `.lintr` to `.Rbuildignore`

Bisaloo:
Inspired from #2396 but smaller in scope and simpler. In particular:

- we don't try to hide the warning about the missing newline. If users have a misformatted .Rbuildignore, that's outside the scope of this package
- we don't try to see if the created config file is already Rbuildignored via another regex (e.g., `.*` in `.Rbuildignore`). `use_lintr()` should feel like a usethis function IMO and usethis is not doing this. We simply avoid replicating an existing line so we check if the regex to add is already present

Other difference:

- this PR delegates the task of finding the package root to the existing `find_package()` helper 

Would fix #1805.

Does this seem sensible?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -48,6 +48,7 @@
 * `object_usage_linter()` lints missing packages that may cause false positives (#2872, @AshesITR)
 * New argument `include_s4_slots` for the `xml_find_function_calls()` entry in the `get_source_expressions()` to govern whether calls of the form `s4Obj@fun()` are included in the result (#2820, @MichaelChirico).
 * `sprintf_linter()` lints `sprintf()` and `gettextf()` calls when a constant string is passed to `fmt` (#2894, @Bisaloo).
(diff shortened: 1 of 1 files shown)


R/use_lintr.R
```diff
@@ -42,5 +42,32 @@ use_lintr <- function(path = ".", type = c("tidyverse", "full")) {
...
   write.dcf(the_config, config_file, width = Inf)
+
+  pkg_path <- find_package(path)
+
+  if (is.null(pkg_path)) {
+    return(invisible(config_file))
+  }
+
+  rbuildignore_path <- file.path(pkg_path, ".Rbuildignore")
+  rel_path <- xfun::relative_path(config_file, pkg_path)
```
Bisaloo, line 53: I tried to do this with different approaches:

- with `substr()` as in the initial PR
- with `grep()`

But tests on CI showed that this approach is too simple in many cases to be able to deal with symlinks or complex filesystems.

As mentioned in the commit message, xfun is already an indirect dependency anyways.

Issue #2929, opened 2025-09-15
roxygen2 change causing update in manual link resolution


MichaelChirico:
https://github.com/r-lib/lintr/blob/33efecf7f4aee58fbc9d4f2e1f3fda5ae908ed42/R/any_is_na_linter.R#L3

See https://github.com/r-lib/roxygen2/issues/1725

How do we want to handle this? For now, it's causing the `repo-meta-tests` check to fail.

 - We could disable the check that roxygen2 comments & man/ contents are in sync
 - We could manually specify `base::anyNA()` in the manual
 - We could manually write out the `\code{\link[=anyNA]{anyNA}()}` Rd link
 - We could un-link `anyNA()` and just leave it like `` `anyNA()` `` (Rd: `\code{anyNA()}`)

Since there are a number of options I'm first filing this as an issue. At a glance using `base::anyNA()` will be the best, possibly with a TODO to revert it in the future.

PR #2930, opened 2025-09-15
Update roxygen2 to 7.3.3

MichaelChirico:
For #2929 (not closing for now while we wait for confirmation upstream that this is WAI)


MichaelChirico:
I don't have the latest roxygen2 at the moment so fixing the `@examplesIf` mismatch will have to wait for now.

MichaelChirico:
Merging to un-break CI.

PR #2937, opened 2025-09-29
Partial fix for upcoming testthat release

hadley:
`expect_success()` and `expect_failure()` now test that you have exactly one success/failure and zero failures/successes. I can't quite figure out why the tests are still failing here; maybe it's something to do with recycling? I'm happy to help but I unfortunately I don't know enough about the lintr internals to figure out what's going wrong here.

We're planning to submit testthat to CRAN on Nov 10.

### DIFF
--- a/R/expect_lint.R
+++ b/R/expect_lint.R
@@ -70,7 +70,7 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en", igno
 
   if (n_lints != length(checks)) {
     msg <- sprintf(wrong_number_fmt, n_lints, length(checks), lint_str)
-    return(testthat::expect(FALSE, msg))
+    return(testthat::fail(msg))
   }
 
   if (ignore_order) {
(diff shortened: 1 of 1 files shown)


R/expect_lint.R
```diff
@@ -69,8 +71,7 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en", igno
   checks[] <- lapply(checks, fix_names, "message")
 
   if (n_lints != length(checks)) {
-    msg <- sprintf(wrong_number_fmt, n_lints, length(checks), lint_str)
-    return(testthat::expect(FALSE, msg))
+    return(testthat::fail(sprintf(wrong_number_fmt, n_lints, length(checks), lint_str)))
```
MichaelChirico, line 74: @hadley do we need `return()` around `fail()`? It always interrupts right?

MichaelChirico: Hm, I guess not from glancing at the source

MichaelChirico:
Thanks @hadley! Can I check if this is back-compatible? We have `testthat (>= 3.2.1),`, I just confirmed it works on 3.2.3 at least

MichaelChirico:
> I can't quite figure out why the tests are still failing here

AFAICT you were only failing the {lintr} check, the main thing was by switching from `Map()` to a nested `for` loop it increased the cyclomatic complexity of the implementation enough to trigger a lint. I refactored the nested loops into a helper instead. Bonus -- we could drop the `itr_env` (replacing older `<<-` usage) in the new approach which I also find cleaner.

Thanks for the proactive fix! Your release time is perfectly fine. We are overdue for a release anyway.

MichaelChirico (approved):

R/expect_lint.R
```diff
@@ -85,42 +86,47 @@ expect_lint <- function(content, checks, ..., file = NULL, language = "en", igno
...
+      }
+      check_field <- check[[field]]
+      value <- lint[[field]]
+      ok <- if (field == "message") {
+        re_matches_logical(value, check_field)
+      } else {
+        isTRUE(all.equal(value, check_field))
+      }
+      if (!ok) {
+        return(testthat::fail(sprintf(
```
MichaelChirico, line 122: For posterity:

The issue here is that `fail()` doesn't interrupt execution, and essentially we have:

```r
myExpectation <- function(x) {
  maybeFail()
  succeed()
}

maybeFail <- function(x) {
  if (x < 0) fail('negative')
}

expect_failure(myExpectation(-1))
```

So as written (and my own guess) we might think "once `fail()` is reached, `succeed()` can't be" but that's not true.

hadley: Do you think that this could be clearer in https://testthat.r-lib.org/articles/custom-expectation.html ? Maybe I need to say something like this ? 

> It's easy to think that `fail()` is similar to `stop()` or `abort()`, because that's how it behaves when you run it interactively in the console. But it can't work this way in tests because tests need to keep running even after failure so that they can report all the failures at once. So in a real testing setup, `fail()` and `pass()` work very similarly: they both signal a condition that testhat records and then reports to the user.

MichaelChirico: Yes that's quite helpful! The examples in the vignette hint at this by always having `if`/`else` structure to `pass`/`fail` . But as you mentioned, it's deceptively similar to `stop`/`abort`, so useful to call out the difference.

PR #2947, opened 2025-10-16
Don't lint T and F if followed by `[`

### LINKED ISSUE #2944: T_and_F_symbol_linter seems too eager
The linter flags cases in which a variable is called `T`:
```
style: [T_and_F_symbol_linter] Use TRUE instead of the symbol T.
    main = expression(paste0(T[n], ",", T[x], " curves")),
```

In this specific case, perhaps the linter should notice the presence of `[`, which makes it unlikely that `T` referred to `TRUE`.

mcol:
Fixes #2944.

### DIFF
--- a/R/with.R
+++ b/R/with.R
@@ -92,6 +92,9 @@ modify_defaults <- function(defaults, ...) {
 #' }
 #' @export
 linters_with_tags <- function(tags, ..., packages = "lintr", exclude_tags = "deprecated") {
+  if (missing(tags)) {
+    cli_abort("{.arg tags} was not specified. Available tags: {available_tags()}")
+  }
   if (!is.character(tags) && !is.null(tags)) {
     cli_abort("{.arg tags} must be a character vector, or {.code NULL}, not {.obj_type_friendly {tags}}.")
   }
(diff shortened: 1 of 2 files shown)


R/T_and_F_symbol_linter.R
```diff
@@ -36,7 +36,7 @@ T_and_F_symbol_linter <- function() { # nolint: object_name.
     (text() = 'T' or text() = 'F')
     and not(parent::expr[OP-DOLLAR or OP-AT])
     and (
-      not(ancestor::expr[OP-TILDE])
+      not(ancestor::expr[OP-TILDE or OP-LEFT-BRACKET])
```
MichaelChirico: let's catch `[[` too

MichaelChirico: Also, I don't think this should be on the `anchestor::` axis -- please add a test like

```
x[, cols, drop = F]
```

mcol: I've addressed your comments to the best of my XPath knowledge. Feel free to take over if the query can be better formulated.  :)

tests/testthat/test-T_and_F_symbol_linter.R
```diff
@@ -32,6 +34,8 @@ test_that("T_and_F_symbol_linter blocks disallowed usages", {
   expect_lint("DF$bool <- T", msg_true, linter)
   expect_lint("S4@bool <- T", msg_true, linter)
   expect_lint("sum(x, na.rm = T)", msg_true, linter)
+  expect_lint("x[T]", msg_true, linter)
```
MichaelChirico, line 37: Great test. Indeed using `T` (as `TRUE`) to index a vector is valid, and we'd be better off linting it as here.

PR #2950, opened 2025-10-17
Use expect_no_lint() instead of expect_lint(., NULL)

mcol:
This is a small part of #2737 (5 files), which should be manageable to review. If it's not desired, just close it.


mcol:
Thanks Michael. I've removed changes to files that were touched in any form by PRs opened since January, and gave a cursory glance to older ones (without seeing other conflicts). This is much smaller but better than nothing! :)

PR #2952, opened 2025-10-17
Lint nrow, ncol, NROW and NCOL with logical expressions

### LINKED ISSUE #2933: Include `nrow()` and `ncol()` in `length_test_linter()`
Would it make sense to also cover `nrow(x == )` and `ncol(x == )` in length_test_linter()`?

I was recently bitten by this and I don't think there is any more reasons this should be valid than for `length(x == )`.

https://github.com/search?q=%2F\bn(row|col)\([^\(\)\[\]]%2B%3D%3D\s*\d%2B\)%2F+language%3AR&type=code

mcol:
I've taken a stab to fix #2933. I'm still unsure about the wording in the help page, at the moment I've only applied minimal changes. Also in terms of tests, I've used `patrick` in one case, but manually added some other no_lint tests, as I was not sure that it was worth the effort. I'm open to suggestions!

### DIFF
--- a/tests/testthat/test-length_test_linter.R
+++ b/tests/testthat/test-length_test_linter.R
@@ -13,49 +13,42 @@ test_that("skips allowed usages", {
   expect_no_lint("NCOL(DF[key == val, cols])", linter)
 })
 
-local({
-  linter <- length_test_linter()
+patrick::with_parameters_test_that(
+  "blocks simple disallowed usages",
+  {
+    linter <- length_test_linter()
+    lint_msg_stub <- sprintf("Checking the %s of a logical vector is likely a mistake. Did you mean ", fun)
+    expect_lint(
+      paste0(fun, "(x == 0)"),
+      rex::rex(lint_msg_stub, "`", fun, "(x) == 0`?"),
+      linter
+    )
+    expect_lint(
+      paste0(fun, "(x == y)"),
+      rex::rex(lint_msg_stub, "`", fun, "(x) == y`?"),
+      linter
+    )
+    expect_lint(
+      paste0(fun, "(x + y == 2)"),
+      rex::rex(lint_msg_stub, "`", fun, "(x+y) == 2`?"),
+      linter
+    )
+  },
+  fun = c("length", "nrow", "ncol", "NROW", "NCOL")
+)
 
-  patrick::with_parameters_test_that(
-    "blocks simple disallowed usages",
-    {
-      lint_msg_stub <- sprintf("Checking the %s of a logical vector is likely a mistake. Did you mean ", fun)
-      expect_lint(
-        paste0(fun, "(x == 0)"),
-        rex::rex(lint_msg_stub, "`", fun, "(x) == 0`?"),
-        linter
-      )
-      expect_lint(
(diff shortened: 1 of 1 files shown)


R/length_test_linter.R
```diff
@@ -1,4 +1,4 @@
-#' Check for a common mistake where length is applied in the wrong place
+#' Check for a common mistake where length/nrow/ncol/NROW/NCOL is applied in the wrong place
```
MichaelChirico:
```suggestion
#' Check for a common mistake where a size check like 'length' is applied in the wrong place
```

MichaelChirico: And mention nrow/ncol/NROW/NCOL in the description below, as well as adding one new example for any of those.

R/length_test_linter.R
```diff
@@ -26,13 +26,14 @@ length_test_linter <- function() {
...
-    xml_calls <- source_expression$xml_find_function_calls("length")
+    xml_calls <- source_expression$xml_find_function_calls(c("length", "nrow", "ncol", "NROW", "NCOL"))
     bad_expr <- xml_find_all(xml_calls, xpath)
 
+    matched_function <- xp_call_name(bad_expr)
     expr_parts <- vapply(lapply(bad_expr, xml_find_all, "expr[2]/*"), xml_text, character(3L))
     lint_message <- sprintf(
-      "Checking the length of a logical vector is likely a mistake. Did you mean `length(%s) %s %s`?",
-      expr_parts[1L, ], expr_parts[2L, ], expr_parts[3L, ]
+      "Checking the length of a logical vector is likely a mistake. Did you mean `%s(%s) %s %s`?",
```
MichaelChirico: Maybe it should be "Checking the %s of a logical vector" too?

tests/testthat/test-length_test_linter.R
```diff
@@ -1,33 +1,64 @@
...
+          linter
+      )
+      expect_lint(
+          paste0(fun, "(x + y == 2)"),
+          rex::rex(lint_msg_stub, "`", fun, "(x+y) == 2`?"),
+          linter
+      )
+    },
+    fun = funs,
+    .test_name = names(fun)
```
MichaelChirico: I _think_ we can drop `.test_name` here and just use `fun = c("length", ...)`

tests/testthat/test-length_test_linter.R
```diff
@@ -1,33 +1,64 @@
...
+  )
 })
 
 local({
   ops <- c(lt = "<", lte = "<=", gt = ">", gte = ">=", eq = "==", neq = "!=")
+  funs <- rep(c(length = "length", nrow = "nrow", ncol = "ncol", NROW = "NROW", NCOL = "NCOL"),
+              each = length(ops))
+  ops <- rep(ops, length(unique(funs)))
```
MichaelChirico: this looks like `expand.grid()`, right? You can use `.cases=` in `with_parameters_test_that()` here, c.f.

https://github.com/r-lib/lintr/blob/aff43f12390b9a093652585ba9d79e695e55bec8/tests/testthat/test-list_comparison_linter.R#L9-L18

Issue #2943, opened 2025-10-16
cyclocomp linter is not actionable


mcol:
I received this message but I don't have a clue on what I should be doing:
```
style: [cyclocomp_linter] Reduce the cyclomatic complexity of this function from 32 to at most 15.
```
Perhaps just some generic advice should be reported alongside that message. Even better would be to lint the places that are considered highly complex and thus would help reduce the score.

MichaelChirico:
Yea, I agree it's not the most actionable. I wonder how helpful it would be to do a leave-one-out computation like:

`for (j in seq_along(body(FUN))) { cylcocomp(exclude_expr(FUN, jj)) }` to at least point to the top-level expression that adds the most to the complexity.

One issue is that `cyclocomp_linter()` is already one of the slowest in the suite, this would substantially increase the cost.

The most generic advice would be something like:

> Consider replacing high-complexity sections like loops and branches with helper functions.

MichaelChirico:
I think probably the best solution is to suggest the user run a {cyclocomp} diagnostic themselves to understand how to proceed... but there is no such helper today!

In fact, it turns out I already made this FR 2 years ago :) https://github.com/gaborcsardi/cyclocomp/issues/24

PR #2953, opened 2025-10-19
Make the message from cyclocomp_linter() more actionable

mcol:
Addresses the easy part of #2943.

### DIFF
--- a/R/paren_body_linter.R
+++ b/R/paren_body_linter.R
@@ -36,7 +36,6 @@ paren_body_linter <- make_linter_from_xpath(
       or preceding-sibling::OP-LAMBDA
       or preceding-sibling::IF
       or preceding-sibling::WHILE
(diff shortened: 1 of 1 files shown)


R/cyclocomp_linter.R
```diff
@@ -45,8 +45,9 @@ cyclocomp_linter <- function(complexity_limit = 15L) {
       column_number = source_expression[["column"]][1L],
       type = "style",
       message = sprintf(
-        "Reduce the cyclomatic complexity of this expression from %d to at most %d.",
-        complexity, complexity_limit
+        "The cyclomatic complexity of this expression is %d, which exceeds the limit of %d. %s",
```
MichaelChirico: I prefer the original wording for the first part -- we try and structure the lint message like "Action, reason", which I think is better accomplished by the old wording.

MichaelChirico (approved):

PR #2955, opened 2025-10-21
Don't lint expression(paste(., sep = "")) in paste_linter

### LINKED ISSUE #2945: paste_linter should not warn inside expression() in a plot
This suggestion is actually misleading, as only `paste()` and not `paste0()` is allowed when `expression()` is used within a plot, such as to generate title or axis labels:
```R
warning: [paste_linter] paste0(...) is better than paste(..., sep = "").
      main = expression(paste("Plateau test ", L[n], ",", L[x], " curves", sep =
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The rationale is that only a certain number of functions and operators are valid inside `expression()`:
https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/plotmath.html

mcol:
Fixes #2945.


R/paste_linter.R
```diff
@@ -75,6 +77,11 @@
 #' )
 #'
 #' lint(
+#'   text = 'expression(paste("a", "b", sep = ""))',
```
MichaelChirico: this should be in the "will produce lints" section; our approach generally is to "pair" linting expressions with their "clean" equivalents across the "top" and bottom (under "okay") parts of the documentation.

(it's a bit messy for this file in particular because `paste_linter()` does a lot)

R/paste_linter.R
```diff
@@ -121,7 +128,11 @@ paste_linter <- function(allow_empty_sep = FALSE,
   following-sibling::SYMBOL_SUB[text() = 'sep' and following-sibling::expr[1][STR_CONST]]
     /parent::expr
   "
-
+  expression_paste_sep_xpath <- "
+  following-sibling::SYMBOL_SUB[text() = 'sep' and following-sibling::expr[1][STR_CONST]
+    and parent::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'expression']]
```
MichaelChirico:
```suggestion
  following-sibling::SYMBOL_SUB[
    text() = 'sep'
    and following-sibling::expr[1][STR_CONST]
    and parent::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'expression']
  ]
```

R/paste_linter.R
```diff
@@ -184,7 +198,16 @@ paste_linter <- function(allow_empty_sep = FALSE,
       paste_sep_value <- get_r_string(paste_sep_expr, xpath = "./SYMBOL_SUB[text() = 'sep']/following-sibling::expr[1]")
     }
 
-    if (!allow_empty_sep) {
+    ## check if we are inside an expression()
+    expression_paste_sep_expr <- xml_find_all(paste_calls, expression_paste_sep_xpath)
```
MichaelChirico, line 202: this looks off -- what happens for something like this?

```r
c(
  expression(paste(x, y, sep="")),
  paste(x, y, sep=""),
  expression(paste(a, b, sep="")),
  paste(a, b, sep=""),
  expression(c(paste(d, e, sep=""), paste(f, g), paste(h, i, sep="")))
)
```

(generally `xml_find_all()` is 🚩 to me since it is not 1:1 -- for `y = xml_find_all(x, ...)`, `y[i]` cannot predictably be associated with `x[i]` -- usually within linters we want to use `xml_find_first()`)

MichaelChirico:
I asked gemini for a review, which looks basically correct.

---

## Correctness & Robustness Issues

### 1. Nested Expressions (Missed Edge Case / False Positive)
The PR implementation only detects `paste()` calls that are *direct* arguments of `expression()`. If the `paste()` call is nested inside another plotmath function within `expression()`, it is not detected and the linter will incorrectly suggest using `paste0()`.

**Example of missed edge case:**
```R
expression(italic(paste("a", sep = "")))
```
In this case, `paste` is a child of `italic`, which is a child of `expression`.
* **PR Behavior**: Lints this as `paste0(...) is better than paste(..., sep = "").` (False Positive, since `paste0` will fail in plotmath).
* **Reason**: The PR uses the following XPath condition to check for `expression()`:
  `parent::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'expression']`
  This only checks the immediate parent call.

### 2. Multiple Lints in Same Expression (False Negatives)
The logic in the linter body uses an `if-else` structure that can cause valid lints to be skipped if they co-occur with an `expression(paste(...))` call in the same parsed expression (e.g., inside a block `{ ... }`).

**PR Code:**
```R
    if (length(expression_paste_sep_expr) > 0L) {
      optional_lints <- c(optional_lints, xml_nodes_to_lints(
        expression_paste_sep_expr[1L], # Bug: Only lints the first one
        ...
      ))
    } else if (!allow_empty_sep) {
      # Bug: Skipped if length(expression_paste_sep_expr) > 0
      optional_lints <- c(optional_lints, xml_nodes_to_lints(
        paste_sep_expr[!nzchar(paste_sep_value)],
        ...
      ))
    }
```

**Example of skipped lint:**
```R
{
  paste("a", sep = "")             # Should lint to paste0
  expression(paste("b", sep = "")) # Should lint to "inside expression..."
}
```
* **PR Behavior**: ONLY lints the `expression(...)` call. The first `paste("a", sep = "")` is completely ignored.
* **Reason**: Because `expression_paste_sep_expr` is not empty, the code enters the `if` block and skips the `else if` block entirely. Additionally, only the first node (`[1L]`) of `expression_paste_sep_expr` is linted.

---

## Efficiency

The PR uses `xml_find_all()` instead of the internal helper `xml_find_all_()`. The helper should be preferred as it uses a pre-defined empty namespace to optimize lookup performance.

---

## Proposed Fix

To resolve these issues, we need to:
1. Use `ancestor-or-self::` in the XPath to handle nesting.
2. Make the XPaths for "normal paste" and "expression paste" mutually exclusive so they can be processed independently without double-linting.
3. Remove the `else if` logic and process all matched nodes.

### Diff of the Fix

The following changes have been tested and verified to fix both bugs:

```diff
diff --git a/R/paste_linter.R b/R/paste_linter.R
index 89ecc8bf..e200ef86 100644
--- a/R/paste_linter.R
+++ b/R/paste_linter.R
@@ -120,8 +120,21 @@ paste_linter <- function(allow_empty_sep = FALSE,
   check_file_paths <- allow_file_path %in% c("double_slash", "never")
 
   paste_sep_xpath <- "
-  following-sibling::SYMBOL_SUB[text() = 'sep' and following-sibling::expr[1][STR_CONST]]
+  following-sibling::SYMBOL_SUB[
+    text() = 'sep'
+    and following-sibling::expr[1][STR_CONST]
+    and not(parent::expr/ancestor-or-self::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'expression'])
+  ]
+    /parent::expr
+  "
+
+  expression_paste_sep_xpath <- "
+  following-sibling::SYMBOL_SUB[
+    text() = 'sep'
+    and following-sibling::expr[1][STR_CONST]
+    and parent::expr/ancestor-or-self::expr/preceding-sibling::expr/SYMBOL_FUNCTION_CALL[text() = 'expression']
+  ]
     /parent::expr
   "
 
@@ -184,7 +197,16 @@ paste_linter <- function(allow_empty_sep = FALSE,
       paste_sep_value <- get_r_string(paste_sep_expr, xpath = "./SYMBOL_SUB[text() = 'sep']/following-sibling::expr[1]")
     }
 
-    if (!allow_empty_sep) {
+    ## check if we are inside an expression()
+    expression_paste_sep_expr <- xml_find_all_(paste_calls, expression_paste_sep_xpath)
+    optional_lints <- c(optional_lints, xml_nodes_to_lints(
+      expression_paste_sep_expr,
+      source_expression = source_expression,
+      lint_message = "inside expression(...), paste does not accept a 'sep' argument.",
+      type = "warning"
+    ))
+
+    if (!allow_empty_sep) {
       optional_lints <- c(optional_lints, xml_nodes_to_lints(
         paste_sep_expr[!nzchar(paste_sep_value)],
         source_expression = source_expression,
```

---

## Test Sufficiency

The tests added in the PR were insufficient as they did not cover nested expressions or multiple lints.

The following test cases should be added to `tests/testthat/test-paste_linter.R` to ensure correctness (these are included in the diff above):

```R
  # Test nested expression (Bug 2)
  expect_lint(
    "expression(italic(paste('a', sep = '')))",
    rex::rex("inside expression(...), paste does not accept a 'sep' argument."),
    paste_linter()
  )

  # Test multiple lints in same expression (Bug 1)
  expect_lint(
    "{\n  paste('a', sep = '')\n  expression(paste('b', sep = ''))\n}",
    list(
      rex::rex('paste0(...) is better than paste(..., sep = "").'),
      rex::rex("inside expression(...), paste does not accept a 'sep' argument.")
    ),
    paste_linter()
  )
```

MichaelChirico:
Superseded by #3054, sorry for the delay and thanks for the fix!

PR #2958, opened 2025-10-22
Note potential danger if argument can be NA in scalar_in_linter()

### LINKED ISSUE #2956: scalar_in_linter provides incorrect hint if argument may be NA
Consider this example: 

```R
lint(text = 'if (list(...)$may_not_exist %in% "arg") {}', linter = scalar_in_linter())

# ::warning file=<text>,line=1,col=5::file=<text>,line=1,col=5,[scalar_in_linter] Use comparison operators
# (e.g. ==, !=, etc.) to match length-1 scalars instead of %in%. Note that comparison operators preserve
# NA where %in% does not.
```

Here, replacing `%in%` with `==` would make the code crash if the list element doesn't exist, as we would get a `NULL == "arg"`, which fails with "argument is of length zero".

I don't think that we can safely guess when an argument may or may never be `NULL`. Probably the only action is to mention this possibility in the message, something like:

> Use comparison operators (e.g. ==, !=, etc.) to match length-1 scalars instead of %in% if you are sure that the term to the left can never be NULL. Note that comparison operators preserve NA where %in% does not.

mcol:
This implements @MichaelChirico's phrasing suggestions to fix #2956.

### DIFF
--- a/R/scalar_in_linter.R
+++ b/R/scalar_in_linter.R
@@ -1,7 +1,10 @@
 #' Block usage like x %in% "a"
 #'
 #' `vector %in% set` is appropriate for matching a vector to a set, but if
-#'   that set has size 1, `==` is more appropriate.
+#'   that set has size 1, `==` is more appropriate. However, if `vector` has
+#'   also size 1 and can be `NA`, the use of `==` should be accompanied by extra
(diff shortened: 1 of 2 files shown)


MichaelChirico (approved):
Thanks!

PR #2961, opened 2025-10-28
Fix duplicate entry for undesirable operators

clerousset:
Removed duplicate mention of the '<-' operator and its implications.


Bisaloo:
You need to re-run `devtools::document()`, which is why the `repo-meta-tests` check is failing. Could you do this please? This is then ready to be merged.

Bisaloo (approved):

PR #2962, opened 2025-10-29
feat: argument to disable linting of `(` for auto-printing

TimTaylor:
closes r-lib#2916

supersedes r-lib#2919. Main difference from before I borked that PR is the renaming of the argument from `allow_print` to `allow_paren_print` as suggested in https://github.com/r-lib/lintr/pull/2919/files#r2350158023.

### DIFF
--- a/R/implicit_assignment_linter.R
+++ b/R/implicit_assignment_linter.R
@@ -9,8 +9,8 @@
 #' @param allow_scoped Logical, default `FALSE`. If `TRUE`, "scoped assignments",
 #'   where the object is assigned in the statement beginning a branch and used only
 #'   within that branch, are skipped.
-#' @param allow_paren_print Logical, default `FALSE`. If `TRUE`, using `(` for auto-printing
-#'   at the top-level is not linted.
+#' @param allow_paren_print Logical, default `FALSE`. If `TRUE`, assignments using
+#'   `(` for auto-printing at the top-level are not linted.
 #'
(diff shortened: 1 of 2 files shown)


R/implicit_assignment_linter.R
```diff
@@ -9,6 +9,8 @@
 #' @param allow_scoped Logical, default `FALSE`. If `TRUE`, "scoped assignments",
 #'   where the object is assigned in the statement beginning a branch and used only
 #'   within that branch, are skipped.
+#' @param allow_paren_print Logical, default `FALSE`. If `TRUE`, using `(` for auto-printing
+#'   at the top-level is not linted.
```
MichaelChirico: Let's tighten this further, since this was never linted here:

```r
foo = \(x) invisible(x)
foo(1)
(foo(1))
# [1] 1
```

Only `(` for printing of assignments is included under `!allow_paren_print`

TimTaylor: How about
```r
#' @param allow_paren_print Logical, default `FALSE`. If `TRUE`, assignments using `(` for auto-printing 
#'   at the top-level are not linted.
```

MichaelChirico: great!

TimTaylor: Cool. pushed the tweak.

MichaelChirico (approved):

PR #2969, opened 2025-11-20
Revert "brace_linter() covers test_that() without braces (#2808)"

MichaelChirico:
This reverts commit b9c4205cfb178af8cd8aa5a432cc1696262c2905.

As noted, {testthat} itself no longer requires braces: https://github.com/r-lib/lintr/issues/2292


Bisaloo (approved):
Should the NEWS item be removed as well?

PR #2979, opened 2025-11-27
Progress deprecations of function arguments

MichaelChirico:


NEWS.md
```diff
@@ -1,6 +1,9 @@
 # lintr (development version)
 
-Nothing yet :)
+## Deprecations & breaking changes
+
+* Arguments `allow_cascading_assign=`, `allow_right_assign=`, and `allow_pipe_assign=` to `assignment_linter()` are now removed.
+* Argument `interpret_glue` to `object_usage_linter()`, marked deprecated in the previous release, is now defunct.
```
Bisaloo: Is it worth repeating here what the recommended way to replace these arguments is? Since the docs (for `object_usage_linter()`) and the error message (for `assignment_linter()`) have been removed.

MichaelChirico: It doesn't hurt to add the extra sentence; I feel more compelled to do so for the `interpret_glue` case, which is freshly deprecated. For `assignment_linter()`, it's been an error to try those arguments since the 3.2.0 release.

PR #2988, opened 2025-12-06
Drop skip_if_not_r_version for skip_unless_r

MichaelChirico:
This is from the very recent (November '25) {testthat} release.

We might just leave this PR pending for a while to let that version bake, OTOH, {lintr} was also just released with no immediate plans for its own CRAN push.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,7 +45,7 @@ Suggests:
     rlang,
     rmarkdown,
     rstudioapi (>= 0.2),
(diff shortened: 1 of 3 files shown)


Bisaloo (approved):
I don't have big concerns about requiring more recent versions of packages as long as they don't bump the recursive minimum R version dependency. So all good here.

On top of this, testthat is just in `Suggests` so the risk is minimal.

PR #3004, opened 2026-01-14
Allow scalar logic in anonymous functions within subsetting calls

### LINKED ISSUE #2935: `vector_logic_linter()` false positive with `&&` in mapping/apply
Example below uses `&&` within a call to `vapply()` which prevents warnings/errors

``` r
packageVersion("lintr")
#> [1] '3.2.0.9000'
library(lintr)

example <- expression(
  tibble::tibble(
    id = 1:4,
    x = list(a = 1:2, b = 3:4, c = NULL, d = "a")
  ) |>
    dplyr::filter(
      vapply(x, function(i) is.numeric(i) && mean(i) > 1, NA)
    )
)

lint(
  text = as.character(example),
  linters = list(vector_logic_linter())
)
#> <text>:1:124: warning: [vector_logic_linter] Use `&` in subsetting expressions.
#> dplyr::filter(tibble::tibble(id = 1:4, x = list(a = 1:2, b = 3:4, c = NULL, d = "a")), vapply(x, function(i) is.numeric(i) && mean(i) > 1, NA))
#>                                                                                                                            ^~
```

<sup>Created on 2025-09-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

emmanuel-ferdman:
## PR Summary
Using `&&` or `||` inside an anonymous function passed to `vapply()`, `sapply()`, or similar within `filter()`/`subset()` is correct because each iteration processes a single element. The linter was incorrectly flagging these as needing `&`/`|`. This adds an XPath exclusion for `FUNCTION` and `OP-LAMBDA` ancestors, matching the pattern already used in `library_call_linter`.

Closes #2935

### DIFF
--- a/R/vector_logic_linter.R
+++ b/R/vector_logic_linter.R
@@ -81,10 +81,14 @@ vector_logic_linter <- function() {
         or self::expr/SYMBOL_FUNCTION_CALL[text() = 'expect_true' or text() = 'expect_false']
       ]
     ]
-    and not(ancestor::expr[
-      preceding-sibling::expr[last()][SYMBOL_FUNCTION_CALL[not(text() = 'expect_true' or text() = 'expect_false')]]
-      or preceding-sibling::OP-LEFT-BRACKET
-    ][not(descendant-or-self::expr[FUNCTION or OP-LAMBDA])])
+    and not(
+      ancestor::expr[
+        preceding-sibling::expr[last()][SYMBOL_FUNCTION_CALL[not(text() = 'expect_true' or text() = 'expect_false')]]
+        or preceding-sibling::OP-LEFT-BRACKET
+      ][
+        not(descendant-or-self::expr[FUNCTION or OP-LAMBDA])
+      ]
+    )
     and not(parent::expr/expr[
       STR_CONST
       or expr/SYMBOL_FUNCTION_CALL[text() = 'as.raw' or text() = 'as.octmode' or text() = 'as.hexmode']


R/vector_logic_linter.R
```diff
@@ -99,6 +99,7 @@ vector_logic_linter <- function() {
       and not(preceding-sibling::expr[last()]/SYMBOL_FUNCTION_CALL[not(text() = 'subset' or text() = 'filter')])
       and not(preceding-sibling::OP-LEFT-BRACKET)
       and not(preceding-sibling::*[not(self::COMMENT)][2][self::SYMBOL_SUB and text() = 'circular'])
+      and not(ancestor::expr[FUNCTION or OP-LAMBDA])
```
MichaelChirico, line 107: this may not be complicated enough:

```r
subset(x,
  sapply(col, function(x) {
  if (A | B) {
    do_ab(x)
  } else {
    do_other(x)
  })
)
```

emmanuel-ferdman: Good catch. I added `[not(descendant-or-self::expr[FUNCTION or OP-LAMBDA])]` to the exclusion so that `if (A | B)` inside a function definition still gets linted, while things like `if (any(x | y))` stay excluded. Added tests for your exact example and a few variations (lambda syntax, while loops, nested aggregators). Let me know if you spot anything else.

tests/testthat/test-vector_logic_linter.R
```diff
@@ -151,6 +151,44 @@ test_that("subsetting logic handles nesting", {
   expect_no_lint("subset(x, y < x[y > 0, drop = AA && BB, y])", linter)
 })
 
+test_that("scalar logic in anonymous functions within filter/subset is allowed", {
+  linter <- vector_logic_linter()
+
+  expect_no_lint("filter(x, vapply(y, function(i) is.numeric(i) && mean(i) > 1, NA))", linter)
```
MichaelChirico, line 157: I notice all of the `expect_no_lint()` cases drop `{` in the anonymous function, while all of the `expect_lint()` cases have `{` in the bodies, is that intentional? If not, I would make sure to use at least one of each (with/without `{`) in both cases.

R/vector_logic_linter.R
```diff
@@ -83,7 +83,7 @@ vector_logic_linter <- function() {
     and not(ancestor::expr[
       preceding-sibling::expr[last()][SYMBOL_FUNCTION_CALL[not(text() = 'expect_true' or text() = 'expect_false')]]
       or preceding-sibling::OP-LEFT-BRACKET
-    ])
+    ][not(descendant-or-self::expr[FUNCTION or OP-LAMBDA])])
```
MichaelChirico: I'm finding it pretty hard to reason about what this condition means, exactly -- it's like `not(ancestor::expr[<cond>][not(...)])` -- the double negation isn't easy to parse.

Can that condition be absorbed into `<cond>` instead? like

```xpath
not(ancestor::expr[
  preceding-sibling::expr[last()][SYMBOL_FUNCTION_CALL[not(text() = 'expect_true' or text() = 'expect_false')]]
  or preceding-sibling::OP-LEFT-BRACKET
  or descendant-or-self::expr[FUNCTION or OP-LAMBDA]
])
```

If that doesn't work, a comment above the XPath explaining the complicated condition is warranted.

tests/testthat/test-vector_logic_linter.R
emmanuel-ferdman: Sure, added a test case for each.

R/vector_logic_linter.R
emmanuel-ferdman: Tried your suggestion but it fails - moving the lambda check into `or`, changes the meaning from "every call ancestor has a lambda" to "skip if there's a call OR a lambda", which wrongly stops linting `if (A & B)` inside lambdas (e.g. `filter(df, sapply(col, function(x) { if (A & B) x }))`). Also tried other XPath alternatives (flipping conditions, using `count()`, ancestor checks) but all failed - Looks like `not(A[not(B)])` is the standard idiom for "every A has B". Added a comment explaining this.

PR #3012, opened 2026-01-26
Gather all failures in roxygen_test.R

### LINKED ISSUE #2931: roxygen_test should gather all failures and report them as a block
In #2930, we're observing that the CI is not failing in the friendlist way -- because the GHA errors on the first man-roxygen2 inconsistency it finds, we have to find such issues iteratively instead of fixing them in a batch.

https://github.com/r-lib/lintr/blob/33efecf7f4aee58fbc9d4f2e1f3fda5ae908ed42/.dev/roxygen_test.R#L46

Ideally we can refactor this GHA so that `check_roxygenize_idempotent` returns a data structure instead of just erroring, then at the end of the script report all the issues at once.

MichaelChirico:
Closes #2931. Co-written with Gemini.

### DIFF
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Upload test results
         if: failure()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: coverage-test-failures
           path: ${{ runner.temp }}/package
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -14,7 +14,8 @@ Authors@R: c(
     person("Indrajeet", "Patil", email = "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531")),
     person("Hugo", "Gruson", role = "aut",
-           comment = c(ORCID = "0000-0002-4094-1476"))
+           comment = c(ORCID = "0000-0002-4094-1476")),
+    person("Emmanuel", "Ferdman", role = "aut")
   )
 Description: Checks adherence to a given style, syntax errors and possible
     semantic issues.  Supports on the fly checking of R code edited with
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,8 @@
 
 ### Lint accuracy fixes: removing false positives
 
+* `if_switch_linter()` no longer produces a false positive when comparing to empty strings (`""`, `''`, or raw strings like `R"()"`), which cannot be used as `switch()` case names (#2835, @emmanuel-ferdman).
+* `undesirable_operator_linter(call_is_undesirable = FALSE)` now correctly skips prefix notation like `` `:::`(pkg, fun) `` (#2999, @emmanuel-ferdman).
 * `unnecessary_nesting_linter()` treats `=` assignment the same as `<-` for several pieces of logic (#2245 and #2829, @MichaelChirico).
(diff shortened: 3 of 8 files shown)


.dev/roxygen_test.R
```diff
@@ -8,7 +8,14 @@ stopifnot(file.copy("man", tempdir(), recursive = TRUE))
...
 
-.Last <- function() unlink(old_dir, recursive = TRUE)
+.Last <- function() {
+  if (!dir.exists(old_dir)) {
+    return(invisible())
+  }
+  unlink(new_dir, recursive = TRUE)
+  file.copy(old_dir, ".", recursive = TRUE)
+  unlink(old_dir, recursive = TRUE)
+}
```
Bisaloo, line 18: I don't understand how this `.Last()` is used since I cannot see it called anywhere.

.dev/roxygen_test.R
```diff
@@ -38,16 +67,24 @@ check_roxygenize_idempotent <- function(LOCALE) {
...
+
+  !any_failed
 }
 
 # Run the check in a few locales to ensure there's no idempotency issues w.r.t. sorting, too
+all_locales_passed <- TRUE
 for (LOCALE in c("C", "en_US.utf8", "hu_HU.utf8", "ja_JP.utf8")) {
-  check_roxygenize_idempotent(LOCALE)
+  all_locales_passed <- check_roxygenize_idempotent(LOCALE) && all_locales_passed
+}
```
Bisaloo: The switch between looking at "what failed" in `check_roxygenize_idempotent()` vs "what passed" here is not the most straightforward to resonate about IMO.

.dev/roxygen_test.R
MichaelChirico: `.Last` is like `on.exit()` for R sessions:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/quit.html

.dev/roxygen_test.R
MichaelChirico: Yea I think that's me being too cute to save 2 lines.

Bisaloo: No, sorry, I wasn't clear. I think what you changed was perfectly reasonable.

I highlighted the fact we use `any_failed` inside `check_roxygenize_idempotent()` to `all_locales_passed` outside. It requires some mental gymnastics when thinking about the code, to make the boolean negation.

I would use `all_passed` (rather than `any_failed`) inside `check_roxygenize_idempotent()` or `any_locale_failed` in the `for` loop here.

Does this make sense?

(More generally, and slightly off-topic, I would likely generally consider that using `!` in an `return()` is likely a code smell for readability. Maybe a linter could make sense.)

MichaelChirico: > Does this make sense?

Got it, yea makes sense.

> I would likely generally consider that using ! in an return() is likely a code smell for readability. Maybe a linter could make sense.

My first sense is "code smell, but not necessarily lintable". E.g. I had your comment in mind when writing something like this this week:

```r
CheckTableExists <- function(...) {
  ...
  result <- tryCatch(CheckTableCall(), error=identity)
  return(!inherits(result, "error"))
}
```

 - `CheckTableExists()` feels much more natural than something like `CheckTableAbsent()`
 - I find this more readable than something like `result <- tryCatch({ CheckTableCall(); TRUE }, error=\(.) FALSE)`
 - I don't see much benefit in "cheating the linter" to move `!` elsewhere

So I'd want to see a more thorough check of some existing codebases/scripts with proposed improvements before moving to an implementation here.

Issue #2999, opened 2025-12-29
undesirable_operator_linter(call_is_undesirable=FALSE) is not actually implemented


MichaelChirico:
#2761 added the parameter but it doesn't do anything.

PR #3018, opened 2026-02-27
Fix false positive in `if_switch_linter()` for empty strings

### LINKED ISSUE #2835: if_switch_linter false positive when testing == ""
As seen in {DBItest}:

https://github.com/r-dbi/DBItest/blob/00060442200eb230350a308c37e30f797a371f95/R/spec-meta-bind-.R#L136-L153

```r
if (kind == "") {
  ...
} else if (kind == "1") {
  ...
} else if (kind == "name") {
  ...
} else {
  ...
}
```

This has no `switch()` equivalent because `""` would be a 0-length variable name.

emmanuel-ferdman:
### PR Summay
`if_switch_linter()` was flagging if-else chains that compare against empty strings (`""` or `''`), but we can't actually use empty strings as `switch()` case names as R throws an error. This PR skips those cases so the linter only suggests `switch()` when it's actually valid.

Fixes #2835

### DIFF
--- a/tests/testthat/test-if_switch_linter.R
+++ b/tests/testthat/test-if_switch_linter.R
@@ -98,6 +98,20 @@ test_that("multiple lints have right metadata", {
       } else if (y == 'C') {
         do_C()
       }
+      if (z1 == 'a') {
+        do_a()
+      } else if (z2 == 'b') {
+        do_b()
+      } else if (z3 == 'c') {
+        do_c()
+      }
+      if (a == '1') {
+        do_1()
+      } else if (a == '2') {
+        do_2()
+      } else if (a == '') {
+        do_blank()
+      }
     }"),
     list(
       list(lint_msg, line_number = 2L),


R/if_switch_linter.R
```diff
@@ -159,7 +159,7 @@
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
 if_switch_linter <- function(max_branch_lines = 0L, max_branch_expressions = 0L) {
-  equal_str_cond <- "expr[1][EQ and expr/STR_CONST]"
+  equal_str_cond <- "expr[1][EQ and expr/STR_CONST[string-length(text()) > 2]]"
```
MichaelChirico, line 162: unfortunately this isn't a robust solution -- consider

```
x == R"()"
```

very unusual, yes, but we prefer to just get it right up-front. the right fix should use get_r_string().

tests/testthat/test-if_switch_linter.R
```diff
@@ -19,6 +19,26 @@ test_that("if_switch_linter skips allowed usages", {
...
+  # empty string comparisons can't use switch()
+  expect_no_lint("if (x == '') 1 else if (x == 'a') 2 else if (x == 'b') 3", linter)
+  expect_no_lint('if (x == "") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
+  expect_no_lint("if (x == 'a') 1 else if (x == '') 2 else if (x == 'b') 3", linter)
+  expect_no_lint("if (x == 'a') 1 else if (x == 'b') 2 else if (x == '') 3", linter)
+  expect_no_lint("if (x == 'a') 1 else if (x == 'b') 2 else if (x == 'c') 3 else if (x == '') 4", linter)
+})
+
+test_that("if_switch_linter handles raw empty strings", {
+  skip_unless_r(">= 4.0.0")
```
MichaelChirico: we require 4.1.0 so no need for this

https://github.com/r-lib/lintr/blob/6afb72de7d73a805cfc7f02ba24091f178ccd983/DESCRIPTION#L28

R/if_switch_linter.R
```diff
@@ -246,3 +244,33 @@ if_switch_linter <- function(max_branch_lines = 0L, max_branch_expressions = 0L)
...
+    if (is.na(else_if)) break
+    current <- else_if
+    str_node <- xml_find_first(current, "IF/following-sibling::expr[1][EQ]/expr/STR_CONST")
+    if (!is.na(str_node)) str_nodes <- c(str_nodes, list(str_node))
+  }
+  str_nodes
+}
+
+chain_vars_equal <- function(expr) {
+  var_nodes <- list()
```
MichaelChirico: why not

```suggestion
  var_nodes <- character()
```

R/if_switch_linter.R
```diff
@@ -246,3 +244,33 @@ if_switch_linter <- function(max_branch_lines = 0L, max_branch_expressions = 0L)
...
+    else_if <- xml_find_first(current, "ELSE/following-sibling::expr[IF]")
+    if (is.na(else_if)) break
+    current <- else_if
+    str_node <- xml_find_first(current, "IF/following-sibling::expr[1][EQ]/expr/STR_CONST")
+    if (!is.na(str_node)) str_nodes <- c(str_nodes, list(str_node))
+  }
+  str_nodes
+}
+
+chain_vars_equal <- function(expr) {
```
MichaelChirico: add a quick comment for these helpers explaining their purpose, the names are not immediately evocative enough 👍

MichaelChirico: glancing, this is very close to `get_chain_strings`, is this equivalent to something like `length(unique(xml_text(get_chain_strings(expr)))) == 1`?

tests/testthat/test-if_switch_linter.R
```diff
@@ -19,6 +19,24 @@ test_that("if_switch_linter skips allowed usages", {
...
+
+test_that("if_switch_linter handles raw empty strings", {
+  linter <- if_switch_linter()
+
+  # raw empty strings can't use switch() either
+  expect_no_lint('if (x == R"()") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
+  expect_no_lint('if (x == r"[]") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
+  expect_no_lint('if (x == R"{}") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
+  expect_no_lint('if (x == R"--()--") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
+  expect_no_lint('if (x == R"()") 1 else if (x == R"{}") 2 else if (x == R"[]") 3', linter)
```
MichaelChirico: I think we can reduce the # of test cases here

```suggestion
  expect_no_lint('if (x == R"--()--") 1 else if (x == "a") 2 else if (x == "b") 3', linter)
  expect_no_lint('if (x == R"()") 1 else if (x == R"{}") 2 else if (x == R"[]") 3', linter)
```

R/if_switch_linter.R
```diff
@@ -205,23 +202,24 @@ if_switch_linter <- function(max_branch_lines = 0L, max_branch_expressions = 0L)
...
+    bad_expr_clean <- strip_comments_from_subtree(bad_expr)
+    expr_all_equal <- vapply(bad_expr_clean, chain_vars_equal, logical(1L))
+    bad_expr <- bad_expr[expr_all_equal]
+
+    # Exclude empty strings, which can't be used as switch() case names
+    has_empty <- vapply(bad_expr, function(expr) {
+      str_nodes <- get_chain_strings(expr)
+      any(vapply(str_nodes, function(n) !nzchar(get_r_string(n)), logical(1L)))
+    }, logical(1L))
+    bad_expr <- bad_expr[!has_empty]
```
MichaelChirico: I might negate the naming here, now we have inner `!` and outer `!` later, I think:

```suggestion
    nonempty <- vapply(bad_expr, function(expr) {
      str_nodes <- get_chain_strings(expr)
      all(vapply(str_nodes, function(n) nzchar(get_r_string(n)), logical(1L)))
    }, logical(1L))
    bad_expr <- bad_expr[nonempty]
```

NEWS.md
```diff
@@ -11,6 +11,7 @@
 * Excluding `cyclocomp_linter()` in `available_linters()` or `linters_with_tags()`, which requires the weak dependency {cyclocomp}, no longer emits a warning (#2909, @MichaelChirico).
 * `repeat_linter()` no longer errors when `while` is in a column to the right of `}` (#2828, @MichaelChirico).
 * `undesirable_operator_linter(call_is_undesirable = FALSE)` now correctly skips prefix notation like `` `:::`(pkg, fun) `` (#2999, @emmanuel-ferdman).
+* `if_switch_linter()` no longer produces a false positive when comparing to empty strings (`""`, `''`, or raw strings like `R"()"`), which cannot be used as `switch()` case names (#2835, @emmanuel-ferdman).
```
MichaelChirico: Our NEWS is a bit unique in structure. `Bug fixes` really only apply to things where e.g. `lint()` throws an error or a linter does so. False positive/negative fixes get their own subheadings below:

`### Lint accuracy fixes: removing false positives`

(I see I missed this issue for your earlier PR for #2999 too)

emmanuel-ferdman: Moved both :) Noted for future

R/if_switch_linter.R
emmanuel-ferdman: Looks like the structure is close, but the XPaths target opposite sides of the comparison. I think a shared `get_chain_conditions` walker is possible not sure if its cleaner. Let me know if you want me to test it or something else.

MichaelChirico: I think the function comments will be enough then

emmanuel-ferdman:
@MichaelChirico thanks on all the the guidance on this!

PR #3021, opened 2026-03-08
Replace .gif in README with equivalent code

### LINKED ISSUE #2725: Ship a static .gif file with the CRAN README
Much like we have `*-still.gif` for the vignettes on CRAN, we should also shrink the {lintr} package size by shipping a static version of this image:

https://github.com/r-lib/lintr/blob/main/man/figures/demo.gif

It currently occupies 16% of the package size on its own. I'd rather just point users to the GitHub page/website for an animated version.

That said, it requires more than a few seconds to make the change, since we currently use a fixed .md file; it will probably require switching to a README.Rmd.

https://github.com/r-lib/lintr/blob/39f7a595a339f0677ae135f81d7a7a7b5d086ff4/README.md?plain=1#L10

MichaelChirico:
Closes #2725. Because CRAN ships with one static .md file (that is not re-generated), the ~best solution I found would have involved a new step in the release process for generating the static file to ship.

Not worth it IMO when the proposal here works just as well. This also has the advantage of keeping the output correct to the day it's rendered.


Bisaloo (approved):
FWIW, we could probably also go this way:

- save the content to lint to a tempfile
- include the content to lint in the README with the embed knitr engine:
  ````
  ```{embed, file=pathtofile}
  ```
  ````
- lint the file

This would ensure the displayed code and the linted code never go out of sync.

But it might be overengineering a little bit.

MichaelChirico:
We _do_ have https://github.com/r-lib/lintr/blob/bd016a02869dcec525d516bd63654f5b14cd4b01/inst/example/bad.R... my immediate worry is having the `embed` path work on both the website and the installed package (since `inst/` will get stripped?)

But I think we can use `knit_code$get()` to help here?

Bisaloo:
It's usually fair to assume the package has been installed beforehand when you render the README. `devtools::render_readme()` for example will install the package dev version first and then render the README. So you could use `system.file(package = "lintr")`.

But I also think we can always use the `inst/` path, since CRAN, etc. only use the rendered version README.md. The README.Rmd version is only used by us. [pkgdown docs](https://pkgdown.r-lib.org/reference/build_home.html?q=README#home-page) also state (I think it used to be different):

> If you use index.Rmd or README.Rmd it's your responsibility to knit the document to create the corresponding .md. pkgdown does not do this for you because it only touches files in the doc/ directory.

PR #3025, opened 2026-03-11
Don't trip over capture groups in lint(exclude=) regex

### LINKED ISSUE #2831: Exclude should ignore groups
This works

```r
lintr::lint(
  text = "x = 1",
  linters = lintr::assignment_linter(),
  exclude = "(?:a)|(?:b)"
)
```

But this

```r
lintr::lint(
  text = "x = 1",
  linters = lintr::assignment_linter(),
  exclude = "(a)|(b)"
)
```

raises an error that I think most users will not understand:

```
Error in `[.data.frame`(re_matches(lines, exclude, locations = TRUE),  : 
  undefined columns selected
```

MichaelChirico:
Closes #2831.

This was selected from the milestone and fixed by Gemini (840649677f2266dc46adb0d1c477c00333c838f1 and 23a15bcf8555bace465d212b108bca55cd14eff2).


MichaelChirico:
Review from Gemini, which identified a gap in testing now addressed.

---

## Correctness & Suitability
The fix is **correct** and **suitable**.
- It introduces a helper function `re_matches_locations` (in `R/utils.R`) that uses base R `regexpr` to reliably extract the overall match's start and end locations, regardless of capture groups.
- It replaces calls to `re_matches(..., locations = TRUE)` with `re_matches_locations(...)` in `R/exclude.R` for:
    - `exclude_start`
    - `exclude_end`
    - `exclude_next`
    - `exclude`
- This ensures that user-customized exclusion patterns won't cause crashes if they contain capture groups.

## Efficiency & Robustness
- **Efficiency**: The helper `re_matches_locations` is efficient. It uses base `regexpr` and performs simple vector operations. It avoids the overhead of `rex::re_matches` which does extra work to extract capture group contents which are not needed here.
- **Robustness**: The helper correctly handles non-matches by returning `NA` for both `start` and `end`, maintaining compatibility with how the rest of the exclusion logic handles non-matches.

## Test Evaluation
The new tests in `tests/testthat/test-exclusions.R` are **appropriate** and cover various capture group scenarios.

### Missed Edge Cases (Addressed)
While the fix correctly updates `exclude_start`, `exclude_end`, and `exclude_next`, the original PR tests only covered the `exclude` parameter.
It is possible (though less common) that users might customize `exclude_start`/`exclude_end`/`exclude_next` with capture groups.

I have added tests for these cases in `tests/testthat/test-exclusions.R` to ensure full coverage:
```r
  # exclude_start / exclude_end with capture groups
  expect_lint(
    "x = 1 # nolint start\ny = 2\nz = 3 # nolint end",
    NULL,
    linters = assignment_linter(),
    exclude_start = "(# nolint start)",
    exclude_end = "(# nolint end)"
  )

  # exclude_next with capture groups
  expect_lint(
    "x = 1 # nolint next\ny = 2",
    list(message = "Use <- for assignment, not =.", line_number = 1L),
    linters = assignment_linter(),
    exclude_next = "(# nolint next)"
  )
```
These tests are now part of the test suite and pass successfully.

## Simpler Fix / Alternatives
1. **Fixing `rex` package**: The root cause is in `rex::re_matches`. A cleaner long-term solution might be to update `rex` to always include overall match locations even when capture groups are present (perhaps in `match.start`/`match.end` to avoid collision with group names). However, since `rex` is an external dependency, fixing it in `lintr` via a helper is the most practical and immediate solution.
2. **Forcing non-capturing groups**: Trying to programmatically convert capturing groups to non-capturing groups in user input would be complex and error-prone. The chosen solution of just ignoring them and getting overall match locations is much simpler and robust.

## Conclusion
The PR is high quality and ready to be merged. The test coverage has been expanded to include other exclusion parameters (`exclude_start`, `exclude_end`, `exclude_next`) that were modified.

PR #3026, opened 2026-03-12
Ignore outer unary operators in unnecessary_lambda_linter

### LINKED ISSUE #2742: unnecessary_lambda_linter() misses '!' when suggesting removing a lambda
```r
lint(text = "sapply(x, function(xi) !all.equal(xi, y))", linters = unnecessary_lambda_linter())
# <text>:1:11: warning: [unnecessary_lambda_linter] Pass all.equal directly as a symbol to sapply() instead of wrapping it in an unnecessary anonymous function. For example, prefer lapply(DF, sum) to lapply(DF, function(x) sum(x)).
# sapply(x, function(xi) !all.equal(xi, y))
#           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

We could do `sapply(x, Negate(all.equal), y)`, I think, but that's hardly a strict improvement.

MichaelChirico:
Closes #2742 

There was some discussion about maybe turning this on optionally since _often_ it's the case that such cases would be better of switching the operator "outside the loop".

Ultimately, I'm a bit unconvinced:

 - doesn't apply to `~` for sure
 - doesn't apply to _all_ the "mappers" we cover, e.g. `rapply()`
 - there is a readability argument to preferring the operator close to its operand
 - weird cases like `!` doing coercion that won't work once `sapply()` does its thing and assembles the output
 - also cases where dispatch differs on the "inner" vs "outer" expression

So on balance, not linting is best IMO.

Gemini chose this issue from the milestone and wrote d8cdd8ed304373620c8b0dd6889ed27b56340b61 and then 3ffd8352987a3f698418d0b977e856c1abdcd05e after my prodding.


MichaelChirico:
Review by Gemini uncovered some missed edge cases. I'll think about how to address them.

----

## Detailed Analysis

### Correctness & Robustness

The PR implements the fix by adding the following check to `default_fun_xpath` in [unnecessary_lambda_linter.R](file:///usr/local/google/home/chiricom/git/lintr/R/unnecessary_lambda_linter.R):

```r
and not(OP-EXCLAMATION or OP-PLUS or OP-MINUS or OP-TILDE)
```

This checks if the *direct children* of the lambda body `expr` contain these unary operators.

#### Missed Edge Cases

This direct child check is insufficient if the unary operator is wrapped:

1.  **Parenthesized Unary Operators:**
    ```r
    sapply(x, function(xi) (!foo(xi)))
    sapply(x, function(xi) (-foo(xi)))
    ```
    In these cases, the top-level body `expr` has `OP-LEFT-PAREN` as a child, not the unary operator. The unary operator is a child of the nested `expr`. The PR will **still lint** these cases (false positives), suggesting `Pass foo directly...` which leads to invalid code.

2.  **Braced Unary Operators:**
    ```r
    sapply(x, function(xi) { -foo(xi) })
    ```
    Similar to parentheses, the top-level body has `OP-LEFT-BRACE`. The unary operator is nested. The PR will **still lint** this case.

### Proposed Improvement

To make the linter robust against parentheses and braces, we need to check if there is *any* unary operator wrapping the target function call, regardless of depth of nesting via `()` or `{}`.

We can achieve this by checking if there is any descendant `expr` of the body that:
1.  Contains the unary operator.
2.  Contains the formal argument (indicating it wraps the call we want to simplify).

The following XPath replacement for line 124 in [unnecessary_lambda_linter.R](file:///usr/local/google/home/chiricom/git/lintr/R/unnecessary_lambda_linter.R) fixes all these cases:

```xpath
and not(descendant-or-self::expr[OP-EXCLAMATION or OP-PLUS or OP-MINUS or OP-TILDE][.//SYMBOL = ancestor-or-self::expr[parent::expr[FUNCTION or OP-LAMBDA]][1]/preceding-sibling::SYMBOL_FORMALS])
```

#### Explanation of the improved XPath:
*   `descendant-or-self::expr`: Looks at the body itself and any nested expressions.
*   `[OP-EXCLAMATION or ...]` Filters for those expressions that are unary operations.
*   `[.//SYMBOL = ...]`: Filters for those unary operations that contain the formal argument of the lambda.
    *   `ancestor-or-self::expr[parent::expr[FUNCTION or OP-LAMBDA]][1]`: Finds the body node (the `expr` whose parent is the lambda function node).
    *   `/preceding-sibling::SYMBOL_FORMALS`: Gets the formal argument name (e.g., `xi`).

This ensures that if there is *any* unary operator wrapping the call to the formal argument, we skip linting.

### Test Sufficiency

The PR adds tests for:
*   `sapply(x, function(xi) !all.equal(xi, y))`
*   `sapply(x, \(xi) !all.equal(xi, y))`
*   `purrr::map(x, ~!foo(.x))` (Note: `purrr` mappers are already safe because `purrr_fun_xpath` is more restrictive and doesn't match nested formulas, but the test is good for regression).
*   `sapply(x, function(xi) -foo(xi))`
*   `sapply(x, function(xi) +foo(xi))`
*   `sapply(x, function(xi) ~foo(xi))`

**Recommendation:** Add the following test cases to [test-unnecessary_lambda_linter.R](file:///usr/local/google/home/chiricom/git/lintr/tests/testthat/test-unnecessary_lambda_linter.R) to cover the missed edge cases:

```r
  # parenthesized/braced unary operators
  expect_no_lint("sapply(x, function(xi) (!foo(xi)))", linter)
  expect_no_lint("sapply(x, function(xi) (-foo(xi)))", linter)
  expect_no_lint("sapply(x, function(xi) { -foo(xi) })", linter)
```

And ensure that parenthesized *valid* simplifications still lint:
```r
  expect_lint("sapply(x, function(xi) (foo(xi)))", "Pass foo directly", linter)
  expect_lint("sapply(x, function(xi) { foo(xi) })", "Pass foo directly", linter)
```

## Conclusion

The PR is a good step but incomplete. Implementing the proposed XPath improvement and adding the recommended test cases will make the fix robust and prevent false positives on parenthesized or braced negated lambdas.

MichaelChirico:
Opted for a more direct fix checking `{` or `(` with unary operator, where Gemini wanted the more complicated-but-less-repetitive `descendant`/`ancestor` axis checks.

PR #3047, opened 2026-05-19
Handle unencoded raw text correctly

### LINKED ISSUE #3046: absolute_path_linter() fails for unmarked UTF-8
Reproducible example, assuming a UTF-8 session:

``` r
library("lintr")

tf <- tempfile()
writeLines('x <- c("\u00e4")', tf, useBytes = TRUE)

lint(tf, linters = absolute_path_linter())                   # OK
#> ℹ No lints found.

text_UTF8 <- readLines(tf, encoding = "UTF-8")
lint(text = text_UTF8, linters = absolute_path_linter())     # OK
#> ℹ No lints found.

text_native <- readLines(tf)  # should be fine in a UTF-8 session
lint(text = text_native, linters = absolute_path_linter())
#> Error in `lint_impl_()`:
#> ! Linter `linter()` failed in '/tmp/RtmpTmjsLz/file185f638e6615c':
#> Caused by error in `parse()`:
#> ! <text>:1:5: Unerwartete(s) ')'
#> 1: "ä")
#>         ^
```

Possibly related to #2836.

MichaelChirico:
Closes #3046. Written by Gemini. Would be great to have it tested by those more familiar with wonky encodings (cc @bastistician), but it's basically copying over the logic from `read_lines()`


tests/testthat/test-get_source_expressions.R
```diff
@@ -562,3 +562,20 @@ test_that("parser warnings generate lints", {
     linters = list()
   )
 })
+
+test_that("get_source_expressions() handles unmarked UTF-8 lines correctly", {
+  tf <- withr::local_tempfile()
+  writeLines('x <- "\u00e4"', tf, useBytes = TRUE)
+  text_native <- readLines(tf)
```
bastistician, line 574: This only makes sense in a UTF-8 locale, matching the file encoding.
Otherwise the input encoding would have to be specified, marking non-ASCII strings as UTF-8, which defeats the test.

MichaelChirico: Ah, thanks for confirming my latest commit :)

R/lint.R
```diff
@@ -771,13 +771,17 @@ maybe_append_condition_lints <- function(lints, source_expression, lint_cache, f
...
   } else if (re_matches(filename, rex(newline))) {
-    strsplit(gsub("\n$", "", filename), "\n", fixed = TRUE)[[1L]]
+    lines <- strsplit(gsub("\n$", "", filename), "\n", fixed = TRUE)[[1L]]
   } else {
-    read_lines(filename)
+    lines <- read_lines(filename)
+    encoding <- settings$encoding
   }
+
+  ensure_utf8(lines, encoding)
```
bastistician, line 784: A simpler change would be to assume as usual that any non-ASCII `text` input is either marked or in the native encoding. So you could simply use `enc2utf8()` in the text branches and leave the settings-driven `read_lines` alone.

R/get_source_expressions.R
```diff
@@ -68,7 +68,8 @@ get_source_expressions <- function(filename, lines = NULL) {
   old_lang <- set_lang("en")
   on.exit(reset_lang(old_lang))
 
-  source_expression$lines <- lines %||% read_lines(filename)
```
bastistician, line 71: With the smaller change, this would assign
`if (is.null(lines)) read_lines(filename) else enc2utf8(lines)`

MichaelChirico: an earlier commit was pretty similar to that: https://github.com/r-lib/lintr/pull/3047/changes/ceedb3eb2f7f983319599576dafc34e756537d55

at a glance it looks like only the latter branch is handling encoding, whereas it _is_ handling encoding, just "under the hood". Hence a desire to make encoding handling more transparent by moving it outside of `read_lines()`.

so while I think your suggestion makes a smaller-diff change, I'm still leaning towards the current approach being preferable.

MichaelChirico:
Sharing a Gemini-authored code review.  I don't think it's worth complicating the implemenation of `ensure_utf8()` in the suggested manner unless we actually find this impacts real users.

---

## 1. Summary of Changes

The PR addresses an issue where `lint()` fails with a parse error when given unmarked UTF-8 text input in a UTF-8 locale.

It introduces a helper function `ensure_utf8(lines, encoding)` in `R/utils.R` and applies it at key entry points where source lines are read or processed:
*   `R/exclude.R` (in `parse_exclusions`)
*   `R/get_source_expressions.R` (in `get_source_expressions`)
*   `R/lint.R` (in `get_lines`)

It also refactors `read_lines` in `R/utils.R` to remove the encoding conversion logic, delegating it to `ensure_utf8` called by its users.

---

## 2. Correctness & Suitability

The fix is **correct** and addresses the root cause of the bug, which is a subtle behavior of R's parser:
*   If the input text to `parse()` is **marked UTF-8**, the parser returns **character-based column numbers** in `getParseData()`.
*   If the input text is **unmarked** (encoding is `"unknown"`), even if it contains valid UTF-8 bytes and we are in a UTF-8 locale, `parse()` returns **byte-based column numbers**.

`lintr`'s internal helper `fix_octal_escapes` (in `R/get_source_expressions.R`) uses `substr()` on the source lines. `substr()` in R is character-based (in UTF-8 locales). If `fix_octal_escapes` is triggered and uses byte columns (returned for unmarked input) as character indices for `substr()`, it extracts incorrect text, leading to corruption and subsequent parse errors in linters (like `absolute_path_linter` via `get_r_string`).

By ensuring all input lines are marked UTF-8 *before* parsing, the PR guarantees that `parse()` returns character columns, aligning them with `substr()` and preventing the corruption.

The fix is **suitable** as `lintr` should handle UTF-8 source code consistently.

---

## 3. Robustness & Edge Cases (Critique)

While the fix works for the reported case, the implementation of `ensure_utf8` has robustness issues and missed edge cases:

### Issue A: Mixed Encodings Bug (`any(is_utf8)`)

In `ensure_utf8`:
```r
  is_utf8 <- Encoding(lines) == "UTF-8"
  if (any(is_utf8)) {
    return(lines)
  }
```
*   **Problem**: If the input vector `lines` has mixed encodings (e.g., some elements are marked `"UTF-8"` and others are `"latin1"` or `"unknown"`), `any(is_utf8)` will be `TRUE` if at least one element is UTF-8. The function will then return the vector as-is, **skipping conversion for the non-UTF-8 elements**.
*   **Impact**: If mixed encoding text is passed to `lint(text = ...)`, the non-UTF-8 parts will not be converted, potentially leading to the same parse errors or encoding issues downstream.
*   **Why it was done**: Likely to avoid double-encoding UTF-8 elements if we ran `iconv` on the whole vector with a different source `encoding` (e.g., `iconv(utf8_str, from="latin1", to="UTF-8")` double-encodes).
*   **Better Solution**: Only convert the elements that are NOT already marked UTF-8.

### Issue B: Falsely Marking Failed Conversions as UTF-8

In `ensure_utf8` (non-UTF-8 path):
```r
  lines_conv <- iconv(lines, from = encoding, to = "UTF-8")
  lines[!is.na(lines_conv)] <- lines_conv[!is.na(lines_conv)]
  Encoding(lines) <- "UTF-8"
```
*   **Problem**: If `iconv` fails to convert some lines (resulting in `NA` in `lines_conv`), those lines are NOT replaced in `lines` (they keep their original bytes). However, `Encoding(lines) <- "UTF-8"` is applied to the **entire** vector.
*   **Impact**: Elements that failed conversion (and thus are NOT valid UTF-8) are now explicitly marked as `"UTF-8"`. This is corruption and will cause functions like `nchar()` or `parse()` to fail with "invalid multibyte string" errors later.
*   **Context**: This behavior was inherited from the old `read_lines` implementation, but it should be fixed for robustness.
*   **Better Solution**: Only mark the successfully converted elements as UTF-8, leaving failed ones as they were (or handling them otherwise).

---

## 4. Proposed Improved Implementation

Here is an improved version of `ensure_utf8` that addresses both issues:

```r
ensure_utf8 <- function(lines, encoding = NULL) {
  if (is.null(encoding) || is.na(encoding)) {
    encoding <- ""
  }

  if (encoding == "" && l10n_info()[["UTF-8"]]) {
    encoding <- "UTF-8"
  }

  if (encoding == "UTF-8") {
    Encoding(lines) <- "UTF-8"
    return(lines)
  }

  is_utf8 <- Encoding(lines) == "UTF-8"
  if (all(is_utf8)) {
    return(lines)
  }

  to_conv <- !is_utf8
  if (any(to_conv)) {
    lines_to_conv <- lines[to_conv]
    lines_conv <- iconv(lines_to_conv, from = encoding, to = "UTF-8")
    
    keep <- !is.na(lines_conv)
    Encoding(lines_conv) <- "UTF-8" # Mark successfully converted ones
    
    indices <- which(to_conv)[keep]
    lines[indices] <- lines_conv[keep]
  }
  lines
}
```

*   Uses `all(is_utf8)` instead of `any(is_utf8)` for early return.
*   Subsets `lines` to only convert and update the non-UTF-8 elements (`to_conv`).
*   Only marks the successfully converted elements (`keep`) as UTF-8.

---

## 5. Review of Tests

The new tests are **sufficient and appropriate**:
*   `test-absolute_path_linter.R`: Verifies `absolute_path_linter()` handles unmarked UTF-8 text input without error.
*   `test-get_source_expressions.R`: Verifies `get_source_expressions()` parses unmarked UTF-8 lines correctly and returns correct character-based columns (verified by checking `col1` and `col2` of `STR_CONST`).
*   `test-lint.R`:
    *   Verifies `lint(text=)` handles unmarked UTF-8 text correctly.
    *   Verifies `lint(text=)` handles UTF-8 marked text in a non-UTF-8 locale (simulated by setting `LC_CTYPE = "C"`).

These tests cover the regression and the locale-dependent behavior.

---

## 6. Conclusion

The PR is a good step and fixes the bug, but the implementation of `ensure_utf8` is fragile. It should be refactored to handle mixed encodings and conversion failures more robustly as suggested in Section 4.

PR #3050, opened 2026-05-26
Supply 'ns=empty_ns' to avoid expensive xml_ns() lookup on most xml2 calls

MichaelChirico:
Found by some profiling done by Gemini -- all `xml_*` functions use a default of `ns = xml_ns(x)` which incurs some overhead for us that can be avoided since we don't really need any namepacing for the AST. This is https://github.com/r-lib/xml2/issues/327.

Performance improvement can be pretty dramatic, e.g. we found 5% on {bit64} but 30% on {utils}:

```r
system.time(for (ii in 1:10) lint_package("~/svn/r-devel/src/library/utils"))
# main
#    user  system elapsed 
# 653.397   5.802 662.961 
# PR
#    user  system elapsed 
# 452.236   6.510 463.223
```

It's a pretty large diff to switch to the `_` suffixed version but I think it beats alternatives until we can get a fix in {xml2} itself.


Bisaloo (commented):
Could you also add a new setting to our linter (under `undesirable_function_linter()`) to remind us to use the wrapper functions instead of the original xml2 functions please?

MichaelChirico:
> Could you also add a new setting to our linter (under `undesirable_function_linter()`) to remind us to use the wrapper functions instead of the original xml2 functions please?

I was thinking to do so as a follow-up but you're right it makes more sense to do here.

NEWS.md
```diff
@@ -59,6 +59,7 @@
 ## Notes
 
 * {lintr} now requires R 4.1.0
+* Internal code is refactored to avoid `xml2::xml_ns()` for e.g. 5-30% end-to-end improvement in `lint_package()`
```
Bisaloo:
```suggestion
* Internal code is refactored to avoid `xml2::xml_ns()` for e.g. 5-30% end-to-end speed improvement in `lint_package()`
```

Bisaloo (approved):

MichaelChirico:
I found a few more implicit `ns=xml_ns(x)` usages, especially in `xml_attr()`, so I re-ran a benchmark -- overall we get about 25% improvement over 5 codebases

```
system.time(for (ii in 1:10) { lint_package("bit64"); lint_package("lintr"); lint_package("ggplot2"); lint_package("data.table"); lint_package("~/svn/r-devel/src/library/utils") })
# PR
#     user   system  elapsed 
# 3688.159   55.745 3769.065 
# main
#     user   system  elapsed 
# 4844.868   54.501 4925.063 
```

PR #3051, opened 2026-05-26
Optimize flatten_list for common cases

MichaelChirico:
Found by Gemini during profiling.

Optimize the core `flatten_list` utility to bypass expensive recursive environment creation for the most common inputs:
- Empty lists
- Single objects
- Flat lists of the target class

This gave me a 7% end-to-end improvement:

```
system.time(for (ii in 1:2) { lint_package("bit64"); lint_package("lintr"); lint_package("ggplot2"); lint_package("data.table"); lint_package("~/svn/r-devel/src/library/utils") })

# PR
#    user  system elapsed 
# 705.916  11.071 721.554 
# main
#    user  system elapsed 
# 753.357  11.111 769.909 
```


MichaelChirico:
Gemini review found some shortcomings of this current state. See the edit history here -- I've pushed back on (1) needing new tests here and (2) a recommendation to check `inherits()` before `length()`.

---

## 1. Architectural & Domain Suitability

### Optimal Check Ordering
Placing `if (length(x) == 0L)` before `if (inherits(x, class))` is precisely the correct trade-off for `lintr`. When linters analyze clean code, they generate thousands of empty lists (`list()`). Early-exiting via a primitive C-level length check intercepts the dominant stream of inputs before triggering attribute lookups in `inherits()`.

### Avoiding Test Ossification
Because `flatten_list()` is a private implementation detail executed tens of thousands of times across the existing public test suite (yielding 100% line coverage on the diff), omitting direct unit tests is a sound design choice. Testing the public package contracts rather than private helpers preserves freedom to refactor internal data structures later without breaking rigid unit tests.

---

## 2. Efficiency & Simplicity (The Simpler Fix)

### Eliminating Helper Overhead
The PR currently introduces a package-scoped helper:
```r
maybe_unname <- function(x) if (is.null(names(x))) x else unname(x)
```
This helper is unnecessary. Base R's `unname()` already performs `if (!is.null(names(obj))) names(obj) <- NULL` internally. Furthermore, adding a helper function introduces extra call frame overhead on a heavily traveled path.

### Inline Optimization
Instead of defining `maybe_unname()`, name stripping should be inlined directly inside the successful `is_flat` condition:

```r
if (is_flat) {
  if (!is.null(names(x))) {
    names(x) <- NULL
  }
  return(x)
}
```
This keeps `flatten_list()` completely self-contained, removes namespace pollution, and guarantees zero unnecessary function call overhead on the hot path.

---

## Proposed Streamlined Implementation

Replace the diff in `R/utils.R` with this ultra-clean, self-contained version:

```r
flatten_list <- function(x, class) {
  if (length(x) == 0L) {
    return(list())
  }
  if (inherits(x, class)) {
    return(list(x))
  }
  if (is.list(x)) {
    is_flat <- TRUE
    for (i in seq_along(x)) {
      if (!inherits(x[[i]], class)) {
        is_flat <- FALSE
        break
      }
    }
    if (is_flat) {
      if (!is.null(names(x))) {
        names(x) <- NULL
      }
      return(x)
    }
  }

  outer_env <- new.env(parent = emptyenv())
  outer_env$res <- list()
  outer_env$itr <- 1L
  assign_item <- function(x) {
    if (inherits(x, class)) {
      outer_env$res[[outer_env$itr]] <- x
      outer_env$itr <- outer_env$itr + 1L
    } else if (is.list(x)) {
      lapply(x, assign_item)
    }
  }
  assign_item(x)
  outer_env$res
}
```

PR #3056, opened 2026-06-09
Create portability tag

### LINKED ISSUE #2867: Add portability linter tag?
It would contain

- https://lintr.r-lib.org/reference/nonportable_path_linter.html
- #2865 
- #2866 
- probably others?

Spin-off from https://github.com/ropensci-review-tools/goodpractice/issues/63 @njtierney

Bisaloo:
Fix #2867

cc @maelle

There is some overlap with the robustness tag, which made me realize that the robustness tag scope is not very clear (to me at least). Should I open a follow up issue for this?


MichaelChirico:
some more to consider: backport_linter(), paste_linter() (which has a rule around file.path()), system_file_linter(), pipe_consistency_linter() (which can be used to block native pipe) 

I agree robustness is not perfectly defined, but seq_linter() is a good example that's appropriate for robustness but not portability. maybe we should just remove 'robustness' when what's meant is actually 'portability'?

Bisaloo:
Can you imagine any case of a linter that would have both `robustness` and `portability`?

MichaelChirico:
> Can you imagine any case of a linter that would have both `robustness` and `portability`?

Yes, especially something like `paste_linter()` that enforces a number of related things, or something principally broad like `undesirable_function_linter()`.

MichaelChirico (approved):
Feel free to merge when ready!

MichaelChirico:
LMK if you'd like this to be included in the pending release (I'll probably submit this weekend)

PR #3060, opened 2026-06-16
Handle un-braced ('{') expressions in if/switch correctly in if_switch_linter

### LINKED ISSUE #3042: if_switch_linter counts expressions incorrectly if the argument doesn't use '{'
```r
l = if_switch_linter(max_branch_expressions = 2L)
lint('
  switch(x,
    a = foo(c, d, e),
    b = bar(f, g, h)
  )
', l)
# <text>:2:3: warning: [if_switch_linter] Prefer repeated if/else statements over overly-complicated switch() statements. Disable this with a comment: `# nolint` (one line) or `# lintr: disable` (multiple lines). See go/tidy-style#linting for more.
#   switch(x,
#   ^~~~~~~~~
```

The lint logic is written with `switch(a = { ... }, b = { ... })` in mind, whereas in this case the number of expressions is just 1.

MichaelChirico:
Closes #3042


MichaelChirico:
Gemini review:

---

### Summary of the Issue and Fix
The `if_switch_linter` has parameters `max_branch_lines` and `max_branch_expressions` to control when a branch is considered "too complex" for a `switch` statement (or too complex to be kept in a `switch` statement).

Previously, the linter assumed all branches were braced (`{ ... }`).
- For `max_branch_expressions`, it counted child `expr` nodes. In a braced branch, these are the expressions inside the braces. In a non-braced branch (e.g., `foo(a, b)`), these are the sub-expressions (the function name and its arguments), leading to over-counting.
- For `max_branch_lines`, it calculated the difference between `OP-RIGHT-BRACE` and `OP-LEFT-BRACE` lines. If braces were missing, this check failed or behaved unexpectedly.

The PR fixes this by distinguishing between braced and non-braced branches in the XPath complexity condition.

### Detailed Review

#### 1. Correctness & Suitability
The fix is **correct** and **suitable**. It directly addresses the root cause: the structural difference in the XML AST between braced and non-braced expressions in R.

*   **Expression Counting:**
    ```R
    if (max_branch_expressions > 0L) paste("OP-LEFT-BRACE and count(expr) >", max_branch_expressions)
    ```
    By adding `OP-LEFT-BRACE and`, the linter now only applies the `count(expr)` check to braced expressions.
    *   **Rationale:** A non-braced branch is grammatically limited to a single expression in R. Therefore, it can never exceed `max_branch_expressions` (which must be $\ge 1$ to be active). Thus, skipping the check for non-braced branches is correct.

*   **Line Counting:**
    ```R
    if (max_branch_lines > 0L) {
      c(
        paste("OP-LEFT-BRACE and (OP-RIGHT-BRACE/@line2 - OP-LEFT-BRACE/@line1 > 1 +", max_branch_lines, ")"),
        paste("not(OP-LEFT-BRACE) and (@line2 - @line1 >=", max_branch_lines, ")")
      )
    }
    ```
    This splits the check:
    *   **Braced:** Uses the previous logic (distance between braces).
    *   **Non-braced:** Uses the line span of the branch expression itself (`@line2 - @line1`).
    *   **Formula Verification:**
        *   If `max_branch_lines = 2`.
        *   A 3-line non-braced branch (e.g., `foo(\n1,\n2\n)`) has `@line2 - @line1 = 2`.
        *   `2 >= 2` is `TRUE`, so it is correctly flagged as complex.
        *   A 2-line non-braced branch has `@line2 - @line1 = 1`.
        *   `1 >= 2` is `FALSE`, so it is not flagged.
        This is correct.

#### 2. Test Sufficiency
The new tests in [test-if_switch_linter.R](file:///usr/local/google/home/chiricom/git/lintr/tests/testthat/test-if_switch_linter.R) are **sufficient and appropriate**. They cover:
*   Multi-line non-braced branches in `if` (should not suggest `switch`).
*   Multi-line non-braced branches in `switch` (should suggest `if`).
*   Single-expression non-braced branches in `if` (should suggest `switch`).
*   Single-expression non-braced branches in `switch` (should not suggest `if`).

These tests verify both the false positive and false negative aspects of the bug.

#### 3. Edge Cases
*   **Parenthesized Expressions:** Expressions wrapped in parentheses (e.g., `(x + 1)`) do not have `OP-LEFT-BRACE` and are correctly treated as single expressions.
*   **Empty Braces:** `{}` has `OP-LEFT-BRACE` but `count(expr)` is `0`, so it correctly avoids complexity triggers.
*   **Multi-line Strings:** A single multi-line string constant without braces (e.g., `if (cond) "multi\nline"`) will be correctly evaluated by the `@line2 - @line1` rule.

#### 4. Maintenance Burden & Efficiency
*   **Maintenance:** The change increases the complexity of the XPath generation, but it is well-commented (implicit in the logical structure) and self-contained. The risk of regression is low.
*   **Efficiency:** The XPath engine (`libxml2`) handles these checks efficiently. The performance impact is negligible.

### Conclusion
The PR is a clean, correct, and minimal fix for a legitimate bug that caused false positives. It should be merged.

PR #3062, opened 2026-06-18
Finish switch to R style config

MichaelChirico:
Follow-up to #3061 to actually use R syntax.


MichaelChirico:
Easier to view suppressing whitespace: https://github.com/r-lib/lintr/pull/3062/changes?w=1

PR #3063, opened 2026-06-19
Custom linter for expect_no_lint

MichaelChirico:
Putting it directly in .lintr.R for now.  My sense is this is not something we want to ship+export either.

In {data.table}, we've taken a slightly different approach of putting them in a folder and sourcing them:

https://github.com/Rdatatable/data.table/blob/0b57d38982fb30f70125ebd5eb2be1cc06d988d9/.ci/.lintr.R#L1-L6

Open to starting that way here too.

### DIFF
--- a/.lintr.R
+++ b/.lintr.R
@@ -1,3 +1,9 @@
+expect_no_lint_linter <- make_linter_from_xpath(
+  "//SYMBOL_FUNCTION_CALL[text() = 'expect_lint']
+    /parent::expr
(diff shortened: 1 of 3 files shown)


Bisaloo (approved):
I think it is very useful in its current form and should be merged.

It would miss `"x" |> expect_lint(NULL, linters = linter)` but it doesn't look like we're using this pattern anywhere.

MichaelChirico:
Ideally code reviewers can catch cases like that 😄

Issue #160, opened 2016-07-09
lint_package fails if .lintr does not end with new line


dirkschumacher:
Just observed the following error. If it is a feature and not a bug, then it might make sense to improve the error message. Tested on a Mac.

``` R
cat("linters: with_defaults(closed_curly_linter = NULL)", file = ".lintr")
lintr::lint_package()
```

=> `Error in mapply(function(from, to) paste(lines[from:to], collapse = "\n")`

``` R
cat("linters: with_defaults(closed_curly_linter = NULL)\n", file = ".lintr")
lintr::lint_package()
```

This works as intented.

PR #3070, opened 2026-07-07
Amend test to work before+after r-devel change about read.dcf warning aroud trailing newlines

MichaelChirico:
This became a warning in #2254 and was previously (b4484df854b2e93e66ffc05024f285828c760593) an error.

In turn this came from R itself using `readLines(warn=TRUE)` in `read.dcf()`, which it no longer does ([r90200](https://github.com/r-devel/r-svn/commit/af4426e4c351b935ecce1bdc6597a03a236db89e)).

Leave the test intact as it ensures a newline-less DCF-style config doesn't cause an error, at least (#160).

I don't think a NEWS entry is warranted.

### DIFF
--- a/tests/testthat/test-settings.R
+++ b/tests/testthat/test-settings.R
@@ -1,16 +1,19 @@
 test_that("read_config_file() warns if the config file does not end in a newline", {
   .lintr <- withr::local_tempfile()
-  # temporarily disable warning->error conversion, see below
-  withr::local_options(lintr.linter_file = .lintr, warn = 0L)
+  withr::local_options(lintr.linter_file = .lintr)
   withr::local_dir(withr::local_tempdir())
 
   # cat() not writeLines() to ensure no trailing \n
   cat("linters: linters_with_defaults(brace_linter = NULL)", file = .lintr)
   writeLines("a <- 1", "aaa.R")
+
   # on R 4.7.0+, read.dcf() uses warn=FALSE in internal readLines(), so we
   #   no longer catch a warning here as part of the fix for old #160.
-  #   expect_no_error() ensures this test passes all supported versions.
-  expect_no_error(lint_dir())
+  if (inherits(tryCatch(read.dcf(.lintr), condition = identity), "warning")) {
+    expect_warning(lint_dir(), "Warning encountered while loading config", fixed = TRUE)
+  } else {
+    expect_silent(lint_dir())
+  }
 })
 
 test_that("it gives informative errors if the config file contains errors", {
--- a/tests/testthat/test-settings.R
+++ b/tests/testthat/test-settings.R
@@ -1,3 +1,14 @@
+test_that("read_config_file() warns if the config file does not end in a newline", {
+  .lintr <- withr::local_tempfile()
+  withr::local_options(lintr.linter_file = .lintr)
+  withr::local_dir(withr::local_tempdir())
+
+  # cat() not writeLines() to ensure no trailing \n
+  cat("linters: linters_with_defaults(brace_linter = NULL)", file = .lintr)
+  writeLines("a <- 1", "aaa.R")
+  expect_warning(lint_dir(), "Warning encountered while loading config", fixed = TRUE)
+})
+
 test_that("it gives informative errors if the config file contains errors", {
   .lintr <- withr::local_tempfile(lines = c(
     "linters: linters_with_defaults(",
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Review from Gemini:

---

Here is a thorough review of [PR #3070](https://github.com/r-lib/lintr/pull/3070).

### Summary of the Change
The PR addresses a test failure in tests/testthat/test-settings.R that occurs when running on R-devel (which will become R 4.7.0+). 

The test `read_config_file() warns if the config file does not end in a newline` historically relied on `read.dcf()` (used internally by `lintr` to read config files) warning when a file lacks a trailing newline. In R 4.7.0+, `read.dcf()` has been updated to use `readLines(warn = FALSE)` internally, meaning it no longer emits this warning.

To make the test pass on both older R versions (which warn) and R 4.7.0+ (which do not), the PR:
1.  Changes the expectation from `expect_warning(...)` to `expect_no_error(lint_dir())`.
2.  Sets `warn = 0L` in `withr::local_options()` to prevent the warning on older R versions from being promoted to an error if `warn = 2` (warnings as errors) is set in the environment.

---

### Review Criteria

#### 1. Is it correct?
**Yes, but it can be improved.** 
The proposed change succeeds in making the test pass across all R versions. However, it has a side effect on R < 4.7.0: since the warning is still emitted by `read.dcf` but no longer caught by `expect_warning()`, it bubbles up to the test runner and shows up as a `WARN` in the test results. 

Additionally, by using `expect_no_error()`, we lose the assertion that `lintr` actually correctly handles and bubbles up the warning on R versions where it *is* emitted.

#### 2. Is it well-designed?
**It is a workaround rather than a clean design.**
Using `warn = 0L` to suppress warning-to-error conversion and then using `expect_no_error()` to ignore the resulting warning is a bit hacky. 

A cleaner design would be to use a conditional expectation based on the R version. This keeps the test runs clean (no stray warnings) and maintains the strength of the assertion on older R versions.

**Recommended Alternative:**
```R
  if (getRversion() < "4.7.0") {
    expect_warning(lint_dir(), "Warning encountered while loading config", fixed = TRUE)
  } else {
    expect_no_warning(lint_dir())
  }
```
I have verified locally (on R 4.6.0) that:
*   This conditional approach passes.
*   It does not produce stray warnings in the test run.
*   It works correctly even if the environment has `options(warn = 2)` set, because `expect_warning` handles this internally.

#### 3. Is it an appropriate addition to the codebase in terms of maintenance burden?
**Yes.** The change is minimal and confined to a single test. The maintenance burden is very low. However, the recommended conditional approach is more self-documenting for future maintainers.

#### 4. Is the PR minimal?
**Yes.** It only touches the single affected test.

#### 5. Is a NEWS entry appropriate?
**No.** This is an internal test correction to adapt to upstream R changes. It does not change any user-facing behavior of `lintr`. Therefore, omitting a NEWS entry is correct.

#### 6. Is the implementation clean and efficient?
The PR implementation is simple but, as noted, leaves the test suite with a warning on R < 4.7.0. The conditional approach suggested above is cleaner.

#### 7. Are new tests appropriate, and do they cover reasonable edge cases?
No new tests are needed, as this is fixing an existing test. The modified test still covers the edge case of loading a config file without a trailing newline, ensuring it doesn't crash on any supported R version.

---

### Conclusion & Recommendation
The PR is correct in its goal but the implementation can be improved. 

**Recommendation:** Accept the PR subject to changing the test to use the conditional version comparison:

```diff
 test_that("read_config_file() warns if the config file does not end in a newline", {
   .lintr <- withr::local_tempfile()
-  withr::local_options(lintr.linter_file = .lintr)
+  withr::local_options(lintr.linter_file = .lintr)
   withr::local_dir(withr::local_tempdir())
 
   # cat() not writeLines() to ensure no trailing \n
   cat("linters: linters_with_defaults(brace_linter = NULL)", file = .lintr)
   writeLines("a <- 1", "aaa.R")
-  expect_warning(lint_dir(), "Warning encountered while loading config", fixed = TRUE)
+  # on R 4.7.0+, read.dcf() uses warn=FALSE in internal readLines(), so we
+  #   no longer catch a warning here as part of the fix for old #160.
+  if (getRversion() < "4.7.0") {
+    expect_warning(lint_dir(), "Warning encountered while loading config", fixed = TRUE)
+  } else {
+    expect_no_warning(lint_dir())
+  }
 })
```
This avoids needing `warn = 0L` and keeps the test suite output clean.

PR #3072, opened 2026-07-08
Skip linting eval=FALSE chunks

### LINKED ISSUE #1964: Skip eval=FALSE chunks
Resurrection of #581.

[As noted there](https://github.com/r-lib/lintr/issues/581#issuecomment-732421520), we can't get a fully general solution, as run-time usage of `opts_chunk$set(eval = FALSE)` .

Still, I think it's feasible for us to look at chunk-level options as an improvement over status quo. The key will be using `knitr:::parse_params()`:

https://github.com/yihui/knitr/blob/e8aae19fcc7443b8a212c8dc65ba2b83894f1fe0/R/parser.R#L174

which takes as input a string like `"x, eval = stop('haha')"` and outputs a key-value list like `list(label = "x", eval = quote(stop('haha')))`.

Parsing chunk params is also something we want for #773.

At a glance, using `eval = FALSE` is a _very_ common way to tell `knitr` to skip R chunks:

https://github.com/search?q=filename%3ARmd+%22eval+%3D+FALSE%22

It looks like GH search can't help us find "```r" usage; I also don't see any usage of `opts_chunk$set(eval = FALSE)`:

https://github.com/search?q=filename%3ARmd+%22opts_chunk%24set%22%2C+%22eval+%3D+FALSE%22

So I think this issue is pretty high-impact & value, even if imperfect.

MichaelChirico:
Closes #1964. Initial pass authored by Gemini (I'll try and keep LLM edits pristine and marked as such in the commit log in the PR before ultimately squashing), though you can see it was a bit off on the code quality mark.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,8 @@
 
 ## New and improved features
 
+* Rmd files with explicit `eval=FALSE` chunks are skipped (#1964, @MichaelChirico). Complex cases like `eval=obj` where `obj=FALSE` and `opts_chunk$set(eval = FALSE)` are not yet supported. This obviates the previous workaround requiring such chunks to use plain markdown highlighting gates like ```` ```r ```` (i.e., without {knitr} mark-up).
+
 ### Linter improvements
 
 * General handling of logic around where comments can appear in code has been improved (#2822, @MichaelChirico). In many cases, this is a tiny robustness fix for weird edge cases unlikely to be found in practice, but in others, this improves practical linter precision (reduced false positives and/or false negatives). The affected linters (with annotations for changes noteworthy enough to have gotten a dedicated bug) are:
--- a/R/extract.R
+++ b/R/extract.R
@@ -159,10 +159,6 @@ non_eval_chunk <- function(start, end, lines, pattern) {
   header <- lines[start]
   # essentially knitr:::extract_params_src
   params_src <- trimws(gsub(pattern$chunk.begin, "\\1", header))
-  if (identical(pattern, knitr::all_patterns$md)) {
-    # essentially knitr:::get_chunk_params
-    params_src <- sub("^[a-zA-Z0-9_]+", "", params_src)
-  }
   header_params <- suppressMessages(suppressWarnings(tryCatch(
     xfun::csv_options(params_src),
     error = \(e) list()
--- a/R/extract.R
+++ b/R/extract.R
@@ -172,19 +172,11 @@ chunk_evals_to_false <- function(start, end, lines, pattern) {
   )))
 
   code <- lines[(start + 1L):(end - 1L)]
-  has_partition <- exists(
-    "partition_chunk",
-    envir = asNamespace("knitr"),
-    inherits = FALSE
-  )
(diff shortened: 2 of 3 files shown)


MichaelChirico:
Gemini review:

---

## Summary of Changes
The PR aims to skip linting of R Markdown/Quarto/Sweave chunks that are marked as `eval=FALSE` (or `eval=F`). It updates `R/extract.R` to parse chunk options and filter out these chunks, and adds corresponding tests.

## Review Findings

### 1. Robustness: Parsing Transient Code (Critical)
During background linting in IDEs, files are often in a transient, syntactically malformed state. 
If `xfun::csv_options()` (for header options) or `xfun::divide_chunk()` (for body options) encounter invalid syntax, they will throw errors and halt execution. This can crash the background linter process.

**Recommendation:** Wrap both option parsing calls in `tryCatch` to handle errors gracefully, defaulting to treating the chunk as evaluated (`eval=TRUE`) if parsing fails.

```diff
is_eval_chunk <- function(start, end, lines, pattern) {
  header <- lines[start]
  params_src <- trimws(gsub(pattern$chunk.begin, "\\1", header))
  header_params <- tryCatch(
    suppressWarnings(xfun::csv_options(params_src)),
    error = function(e) NULL
  )

  code <- lines[(start + 1L):(end - 1L)]
  body_params <- tryCatch(
    suppressWarnings(xfun::divide_chunk("r", code)$options),
    error = function(e) NULL
  )

  eval_value <- body_params$eval %||% header_params$eval
  if (identical(eval_value, quote(F))) {
    return(FALSE)
  }
  !isFALSE(eval_value)
}
```

### 2. Design: Avoid Double Negation
The current implementation uses `non_eval_chunk` which returns `TRUE` if the chunk is *not* evaluated. This leads to double negation in the caller:
`eval_keep <- !as.logical(Map(...))`

**Recommendation:** 
1. Rename `non_eval_chunk` to `is_eval_chunk` (returning `TRUE` if it *should* be evaluated).
2. Use `vapply` instead of `Map` for type safety and performance.

Suggested refactoring for `R/extract.R`:

```diff
@@ -73,8 +73,12 @@
   starts <- starts[nonempty_keep]
   ends <- ends[nonempty_keep]
 
-  eval_keep <- !as.logical(Map(\(start, end) non_eval_chunk(start, end, lines, pattern), starts, ends))
-  starts <- starts[eval_keep]
-  ends <- ends[eval_keep]
+  eval_keep <- vapply(
+    seq_along(starts),
+    function(i) is_eval_chunk(starts[i], ends[i], lines, pattern),
+    FUN.VALUE = logical(1)
+  )
+  starts <- starts[eval_keep]
+  ends <- ends[eval_keep]
```

### 3. Tests
The added tests in `tests/testthat/test-knitr_formats.R` are excellent. They cover:
* `.Rmd`, `.qmd`, and `.Rnw` formats.
* Header options (`eval=FALSE`, `eval=F`) and body options (`#| eval: false`).
* Verifying that `eval=TRUE` chunks are still linted.
* Zero-chunk files.

Issue #2012, opened 2023-07-21
sub_linter for using sub() over gsub()


MichaelChirico:
Some regexes can only have one match, e.g. `^str_` doesn't make sense for `gsub()`.

We could lint a few simple such cases and redirect to `sub()`.

AshesITR:
`^a|b$` comes to mind as a special case.

MichaelChirico:
indeed... is it enough to skip anything with `|`?

AshesITR:
How are multiline flags passed to gsub? Is `(?m)` as a prefix the only way?

PR #3073, opened 2026-07-09
namespace_linter() extended for finding redundant '::'/':::' usage for already-imported symbols

### LINKED ISSUE #2081: redundant_import_linter() for using '::' for functions already imported in the NAMESPACE
```r
#' @importFrom pkg foo
pkg::foo()
```

The latter usage of `foo()` could be linted as `pkg::` is not needed.

Such a linter could help backsliding after clean-ups like #2031

MichaelChirico:
Closes #2081. It already found some hits on {lintr} itself!

### DIFF
--- a/R/namespace_linter.R
+++ b/R/namespace_linter.R
@@ -6,6 +6,8 @@
 #'
 #' @param check_exports Check if `symbol` is exported from `namespace` in `namespace::symbol` calls.
 #' @param check_nonexports Check if `symbol` exists in `namespace` in `namespace:::symbol` calls.
+#' @param check_imports Check if `symbol` is already imported from `namespace` in `namespace::symbol` or
+#'   `namespace:::symbol` calls.
 #'
 #' @examples
 #' # will produce lints
@@ -35,10 +37,15 @@
 #'   linters = namespace_linter(check_nonexports = FALSE)
 #' )
 #'
+#' lint(
+#'   text = "stats::sd(c(1, 2, 3))",
+#'   linters = namespace_linter(check_imports = FALSE)
+#' )
+#'
 #' @evalRd rd_tags("namespace_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
+namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE, check_imports = TRUE) {
   Linter(linter_level = "file", function(source_expression) {
     xml <- source_expression$full_xml_parsed_content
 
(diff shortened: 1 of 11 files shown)


R/namespace_linter.R
```diff
@@ -91,6 +91,7 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
     ns_get <- xml_text(ns_nodes) == "::"
     symbol_nodes <- xml_find_all_(ns_nodes, "following-sibling::*[1]")
     symbols <- get_r_string(symbol_nodes)
+    symbols <- gsub("^`(.*)`$", "\\1", symbols)
```
Bisaloo:
```suggestion
    symbols <- sub("^`(.*)`$", "\\1", symbols)
```
?
It would be good to lint these. I often find myself using `gsub()` where I could use `sub()`.

R/namespace_linter.R
```diff
@@ -112,6 +113,29 @@ namespace_linter <- function(check_exports = TRUE, check_nonexports = TRUE) {
...
+      lints <- c(lints, xml_nodes_to_lints(
+        symbol_nodes[is_imported],
+        source_expression = source_expression,
+        lint_message = sprintf(
+          "Don't use `%s` to access %s, which is already imported from %s.",
+          ifelse(ns_get[is_imported], "::", ":::"), symbols[is_imported], packages[is_imported]
+        ),
+        type = "warning"
+      ))
+    }
```
Bisaloo, line 137: Small nit, if I prefix a function with `:::`, I get two lints, which might be slightly confusing:

```
R/store_utils.R:21:23: warning: [namespace_linter] Don't use `:::` to access getAbsolutePath, which is exported from R.utils.
    path <- R.utils:::getAbsolutePath(path, expandTilde = TRUE)
                      ^~~~~~~~~~~~~~~
R/store_utils.R:21:23: warning: [namespace_linter] Don't use `:::` to access getAbsolutePath, which is already imported from R.utils.
    path <- R.utils:::getAbsolutePath(path, expandTilde = TRUE)
                      ^~~~~~~~~~~~~~~
```

Bisaloo (commented):
I tested it on an external package and it's working fine there as well.

One common issue it doesn't lint is when the functions from the current package are prefixed. E.g., `lintr::fct()` calls in lintr codebase. This could be a follow up PR

R/namespace_linter.R
MichaelChirico: Yea... note that you only get one with `check_nonexports = FALSE` (I question whether such an option is necessary, but again an issue for another day).

Is it worth adding a branch to drop those lints if found in this one? There _are_ two distinct issues, but simply removing `R.utils:::` will fix both. OTOH, it's far from the only place where the same change fixes >1 lint. I'm a bit torn.

R/namespace_linter.R
MichaelChirico: True, https://github.com/r-lib/lintr/issues/2012

MichaelChirico: In this case ``gsub("^`|`$", "", x)`` also works, I wonder how they compare...

MichaelChirico: Makes sense -- if there's no `` ` ``, `sub()` is faster since it can terminate immediately, while `gsub()` has to check the end of the string. If there is `` ` ``, `gsub()` is faster since it only has to do two character checks.

`sub()` will be better for the by-far more common case.

R/namespace_linter.R
Bisaloo: Yes, let's go with the current version then. It's probably not worth complexifying the code for this.

Bisaloo (approved):

PR #3074, opened 2026-07-09
Add some SKILLs and agents

MichaelChirico:
These skills were self-generated by examining the progression from first pass in #3072 and #3073 to their current state.

Having skills defined in the repo will help tighten the loop in the future -- 10+ manual commits is a bit much.

Best practices for style in these is still pretty unclear. Welcome to the wild west :)


MichaelChirico:
This is useful enough to merge IMO; more refinements to follow.

PR #3088, opened 2026-07-19
check_file_listing= argument for fixed_regex_linter() to check list.dir(pattern=) [and dir()]

### LINKED ISSUE #3003: Support `list.files()`/`dir()` in `fixed_regex_linter()`
Currently in R-devel (R 4.6.0):

> list.files() now has optional argument fixed to allow literal pattern matching. Suggested and implemented by Duncan Murdoch in [PR#18925](https://bugs.r-project.org/show_bug.cgi?id=18925).

MichaelChirico:
Closes #3003.

 - Not super married to the argument name `check_file_listing`, any other suggestions?
 - I _think_ `TRUE` is a good default, but maybe we should only do that once X% of CRAN packages are requiring 4.6.0?

### DIFF
--- a/R/fixed_regex_linter.R
+++ b/R/fixed_regex_linter.R
@@ -132,15 +132,13 @@ fixed_regex_linter <- function(allow_unescaped = FALSE) {
     ])
   ]
     /following-sibling::expr[
-      (
-        position() = 2 - count({ in_pipe_cond })
-        and STR_CONST
-        and not(EQ_SUB)
-      ) or (
-        STR_CONST
-        and preceding-sibling::*[not(self::COMMENT)][2][
+      STR_CONST and (
+        preceding-sibling::*[not(self::COMMENT)][2][
           self::SYMBOL_SUB[text() = 'pattern' or text() = 'split']
-        ]
+        ] or (
+          position() = 2 - count({ in_pipe_cond })
+          and not(EQ_SUB)
+        )
       )
     ]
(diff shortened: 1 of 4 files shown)


R/fixed_regex_linter.R
```diff
@@ -71,10 +79,16 @@
...
+#'   text = 'list.files(pattern = "RDS", fixed = TRUE)',
+#'   linters = fixed_regex_linter(check_file_listing = TRUE)
+#' )
+#'
 #' @evalRd rd_tags("fixed_regex_linter")
 #' @seealso [linters] for a complete list of linters available in lintr.
 #' @export
-fixed_regex_linter <- function(allow_unescaped = FALSE) {
+# TODO(R>=4.6.0): Deprecate check_file_listing once R 4.6.0 is the minimum supported version.
+fixed_regex_linter <- function(allow_unescaped = FALSE, check_file_listing = TRUE) {
```
Bisaloo, line 91: Is the function of this new `check_file_listing` argument just to allow compatibility with older versions of R?

MichaelChirico: yes, so that you can get the rest of fixed_regex_linter without having to either depend on very recent R or nolint when using list.files

MichaelChirico: the plan is to remove it in ~5 years when the tidyverse R dependency is 4.6.0.

PR #3095, opened 2026-07-23
Catch is.element(class(x), cls) akin to class(x) %in% cls in class_equals_linter

### LINKED ISSUE #2849: possible extension for class_equals_linter
I came across the following pattern for checking class membership:
`is.element("msts", class(object$x))`, which is equivalent to `"msts" %in% class(object$x)`, but not yet checked by the class_equals_linter. Perhaps this could be an additonal check.

MichaelChirico:
Closes #2849


MichaelChirico:
Merging with Gemini review that led to some refinements (see commit history).

PR #3109, opened 2026-08-13
Handle #line directives in get_source_expressions() (#3106)

### LINKED ISSUE #3106: get_source_expressions() doesn't handle #line directives correctly
Really, this is downstream of an issue with `getParseData()`:

https://bugs.r-project.org/show_bug.cgi?id=19137

But we could add it to the set of things that we "fix up" coming from the R parser:

```r
writeLines(con = tmp<-tempfile(), c("{", '#line 1 "foo.R"', "1+1", "}"))
get_source_expressions(tmp)
# Error in expression_xmls[[i]] : subscript out of bounds
```

youdie006:
Fixes #3106.

### Root cause

A `#line` directive resets the R parser's line numbering. Because of that, `getParseData()` can report the `LINE_DIRECTIVE` row with `parent == 0`, so `top_level_expressions()` (which is `which(pc$parent <= 0L)`) counts it as a top-level expression. But `xmlparsedata` does not always surface the directive as a top-level `/exprlist/*` node. The two lists then have different lengths, and `maybe_append_expression_xml()` zips them positionally:

```r
for (i in seq_along(expressions)) {
  expressions[[i]]$xml_parsed_content <- expression_xmls[[i]]  # subscript out of bounds
```

Repro from the issue:

```r
writeLines(con = tmp <- tempfile(), c("{", '#line 1 "foo.R"', "1+1", "}"))
lintr::get_source_expressions(tmp)
#> Error in expression_xmls[[i]] : subscript out of bounds
```

I confirmed the desync directly: for that input `top_level_expressions()` returns 2 rows (`expr` + `LINE_DIRECTIVE`) while `/exprlist/*` returns 1 (`expr`).

### Fix

`LINE_DIRECTIVE` is a parser directive, not an expression, so it should never be paired with an XML expression node. I exclude it on **both** sides so the two lists stay aligned in every configuration:

- `top_level_expressions()`: `which(pc$parent <= 0L & pc$token != "LINE_DIRECTIVE")`
- the XML selection: `/exprlist/*[not(self::LINE_DIRECTIVE)]`

Excluding it on only one side would re-introduce a desync in the plain top-level `#line` case (where `xmlparsedata` *does* list the directive), so both are needed. I verified alignment (and that linting still works and reports physical line numbers) across: a `#line` inside a block, a `#line` at top level, multiple `#line` directives, plain comments, and no directive.

This follows your suggestion in the issue to "add it to the set of things that we fix up coming from the R parser." Happy to adjust if you'd prefer neutralizing the directive earlier in the pipeline instead.

### Test

Added a regression test in `test-get_source_expressions.R`. Red-green verified with `devtools::test(filter = "get_source_expressions")`: before the change the new test errors with `subscript out of bounds`; after, the whole file passes. `lintr::lint()` on the changed files is clean; `NEWS.md` updated.

---
Disclosure: I used AI assistance (Claude) while preparing this change. I root-caused the desync, ran the tests (red-green), and take responsibility for the contribution.

### DIFF
--- a/R/get_source_expressions.R
+++ b/R/get_source_expressions.R
@@ -577,7 +577,7 @@ get_source_expression <- function(source_expression, error = identity) {
   }
 
   source_expression$parsed_content <- parsed_content
-  fix_octal_escapes(fix_tab_indentations(source_expression), source_expression$lines)
+  fix_line_directives(fix_octal_escapes(fix_tab_indentations(source_expression), source_expression$lines))
 }
 
 maybe_append_expression_xml <- function(expressions, xml_parsed_content) {
@@ -682,6 +682,47 @@ top_level_expressions <- function(pc) {
     return(integer(0L))
   }
   which(pc$parent <= 0L)
+}
+
+# `#line` directives are parser directives, not expressions; {lintr} has no reason to treat one as
+#   anything other than a comment, and leaving it labelled LINE_DIRECTIVE means the rare file that
+#   really uses the construct can produce subtly wrong lints.
+# Relabelling alone is not enough. R also declines to treat LINE_DIRECTIVE as comment-like when it
+#   assigns parents, with two consequences: the directive itself is left an orphan (parent 0), and a
+#   neighbouring COMMENT can be attached to the directive instead of skipping over it (parent
+#   -<directive id>). Either way a positionally nested node is reported with parent <= 0, so
+#   top_level_expressions() counts it while xmlparsedata::xml_parse_data() -- which nests purely by
+#   source position -- does not emit a matching /exprlist/* node, and zipping the two errors (#3106).
+#   Fixed in r-devel (PR#19137, https://bugs.r-project.org/show_bug.cgi?id=19137) by making
+#   finalizeData() treat LINE_DIRECTIVE as comment-like:
+#   https://github.com/r-devel/r-svn/commit/5a482292547659cf25163d96ff1ef15b22dc82e8
+# TODO(R>=4.7.0): remove the re-parenting below; the COMMENT relabelling is ours to keep.
+fix_line_directives <- function(pc) {
+  is_directive <- pc$token == "LINE_DIRECTIVE"
+  if (!any(is_directive)) {
+    return(pc)
+  }
+  pc$token[is_directive] <- "COMMENT"
+
+  # Only files using the directive reach here, so plain comments elsewhere keep R's parents.
+  orphans <- which(pc$token == "COMMENT" & pc$parent <= 0L)
+  if (length(orphans) == 0L) {
+    return(pc)
+  }
+  # Encode positions exactly as xml_parse_data() does, so the parent assigned here always agrees
+  #   with where the node actually lands in the XML.
+  max_col <- max(pc$col1, pc$col2) + 1L
+  node_start <- pc$line1 * max_col + pc$col1
+  node_end <- pc$line2 * max_col + pc$col2
+  for (ii in orphans) {
+    encloses <- !pc$terminal & node_start <= node_start[ii] & node_end >= node_end[ii]
+    if (!any(encloses)) {
+      next
+    }
+    # innermost enclosing node, i.e. the one spanning the fewest characters
+    pc$parent[ii] <- pc$id[which(encloses)[which.min((node_end - node_start)[encloses])]]
+  }
+  pc
 }
 
 # workaround for bad parse data bug for octal escapes
--- a/tests/testthat/test-get_source_expressions.R
+++ b/tests/testthat/test-get_source_expressions.R
@@ -584,3 +584,116 @@ test_that("get_source_expressions() handles unmarked UTF-8 lines correctly", {
   expect_identical(str_const$col1, 6L)
   expect_identical(str_const$col2, 8L)
 })
+
+# `#line` directives, #3106. NB R < 4.4.0 discards the parse data for any file containing one --
+# getParseData() returns NULL there, so lintr only ever sees the whole-file expression and the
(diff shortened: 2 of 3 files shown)


MichaelChirico (changes requested):
thanks! I don't think this is the right direction to go. I think we should basically try and overcome the reported R bug and treat LINE_DIRECTIVE like a COMMENT whenever we see it. That's ~basically what the parser itself wants to do -- note that getParseData doesn't attempt to interpret the line/file info at all, for example.

One option is to try and get equivalent behavior to my proposed R patch ex-post in get_source_expressions by fixing up parent as-if it were a COMMENT.

But I think probably the best approach for now is wait-and-see -- I'd like to have R core weigh in on the bug first before fixing a path forward for lintr since this case is a little bit ambiguous.

R/get_source_expressions.R
```diff
@@ -684,6 +684,47 @@ top_level_expressions <- function(pc) {
...
+#   Fixed in r-devel (PR#19137, https://bugs.r-project.org/show_bug.cgi?id=19137) by making
+#   finalizeData() treat LINE_DIRECTIVE as comment-like:
+#   https://github.com/r-devel/r-svn/commit/5a482292547659cf25163d96ff1ef15b22dc82e8
+# TODO(R>=4.7.0): remove the re-parenting below; the COMMENT relabelling is ours to keep.
+fix_line_directives <- function(pc) {
+  is_directive <- pc$token == "LINE_DIRECTIVE"
+  if (!any(is_directive)) {
+    return(pc)
+  }
+  pc$token[is_directive] <- "COMMENT"
```
MichaelChirico, line 705: the code structure is surprising. It seems to me it would be easier to defer this step until the end.

R/get_source_expressions.R
```diff
@@ -684,6 +684,47 @@ top_level_expressions <- function(pc) {
...
+#   really uses the construct can produce subtly wrong lints.
+# Relabelling alone is not enough. R also declines to treat LINE_DIRECTIVE as comment-like when it
+#   assigns parents, with two consequences: the directive itself is left an orphan (parent 0), and a
+#   neighbouring COMMENT can be attached to the directive instead of skipping over it (parent
+#   -<directive id>). Either way a positionally nested node is reported with parent <= 0, so
+#   top_level_expressions() counts it while xmlparsedata::xml_parse_data() -- which nests purely by
+#   source position -- does not emit a matching /exprlist/* node, and zipping the two errors (#3106).
+#   Fixed in r-devel (PR#19137, https://bugs.r-project.org/show_bug.cgi?id=19137) by making
+#   finalizeData() treat LINE_DIRECTIVE as comment-like:
+#   https://github.com/r-devel/r-svn/commit/5a482292547659cf25163d96ff1ef15b22dc82e8
```
MichaelChirico, line 698: It's better practice IMO to refer to the SVN revision: r90434

R/get_source_expressions.R
```diff
@@ -684,6 +684,47 @@ top_level_expressions <- function(pc) {
...
+# TODO(R>=4.7.0): remove the re-parenting below; the COMMENT relabelling is ours to keep.
+fix_line_directives <- function(pc) {
+  is_directive <- pc$token == "LINE_DIRECTIVE"
+  if (!any(is_directive)) {
```
MichaelChirico, line 702: We should also return early if we observe r>= r90434 behavior, i.e., `all(pc$parent[is_directive] > 0L)`, right?

tests/testthat/test-get_source_expressions.R
```diff
@@ -584,3 +584,116 @@ test_that("get_source_expressions() handles unmarked UTF-8 lines correctly", {
...
+    "  identity({",
+    '#line 3 "c.R"',
+    "    1",
+    "  })",
+    "}"
+  ),
+  "next to comments" = c('#line 1 "a.R"', "# hi", "1", "{", "# there", '#line 2 "b.R"', "2", "}")
+)
+
+patrick::with_parameters_test_that(
```
MichaelChirico, line 607: IMO the parameterized test in this case obscures more than it elucidates

tests/testthat/test-get_source_expressions.R
```diff
@@ -584,3 +584,116 @@ test_that("get_source_expressions() handles unmarked UTF-8 lines correctly", {
...
+  .test_name = names(line_directive_cases),
+  lines = unname(line_directive_cases)
+)
+
+test_that("a #line directive on its own attracts no lints (#3106)", {
+  expect_no_lint(paste(line_directive_cases[["at top level"]], collapse = "\n"), linters = default_linters)
+})
+
+test_that("get_source_expressions() treats #line directives as comments (#3106)", {
+  skip_unless_r(">= 4.4.0")
```
MichaelChirico, line 623: why this skip?

tests/testthat/test-get_source_expressions.R
```diff
@@ -584,3 +584,116 @@ test_that("get_source_expressions() handles unmarked UTF-8 lines correctly", {
...
+  # expression the way it attaches a comment, so the expression list used to be one entry longer
+  # than the list of /exprlist/* nodes, and zipping the two errored.
+  tmp <- withr::local_tempfile(lines = line_directive_cases[["inside braces"]])
+
+  src_exprs <- get_source_expressions(tmp)
+
+  # the braced expression, plus the appended full-file expression
+  expect_length(src_exprs$expressions, 2L)
+
+  expr <- src_exprs$expressions[[1L]]
```
MichaelChirico, line 635: better IMO to use `src_exprs$expressions[[length(src_exprs$expressions)]]` which will strengthen the following tests to be global (full file, full XML tree)

PR #3124, opened 2026-08-24
Improve messaging/classification of hanging indents

### LINKED ISSUE #2144: Indentation bug?
I think the `indentation_linter()` has a bug. The following code gives me the error message below on line 6. 
```r
test_that("unreachable_code_linter works in sub expressions", {
  linter <- unreachable_code_linter()
  msg <- rex::rex("Code and comments coming after a return() or stop()")

  expect_lint(
    lines,
    list(
      list(line_number = 4L, message = msg),
      list(line_number = 7L, message = msg),
      list(line_number = 10L, message = msg),
      list(line_number = 15L, message = msg)
    ),
    linter)
})
```
`Hanging indent should be 14 spaces but is 4 spaces.`

See also PR #2141

MichaelChirico:
Closes #2144. WIP, currently unreviewed.

### DIFF
--- a/tests/testthat/test-indentation_linter.R
+++ b/tests/testthat/test-indentation_linter.R
@@ -66,12 +66,24 @@ test_that("indentation linter flags unindented expressions", {
     indentation_linter(indent = 4L)
   )
 
-  # ugly code, but still correctly indented
-  expect_no_lint(
+  expect_lint(
     trim_some("
       list(
            1,
            2)
+    "),
+    list(
+      rex::rex("Indentation should be 2 spaces but is 5 spaces (or start argument on previous line)."),
+      line_number = 2L
+    ),
+    linter
+  )
+
+  expect_no_lint(
+    trim_some("
+      list(
+        1,
+        2)
     "),
     linter
   )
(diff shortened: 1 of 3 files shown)


tests/testthat/test-indentation_linter.R
```diff
@@ -1103,4 +1139,86 @@ test_that("for loop gets correct linting", {
     linter
   )
 })
+
+test_that("closing parentheses on multi-line calls without first-line args work (#2144)", {
```
MichaelChirico, line 1143: @AshesITR / @Bisaloo / @IndrajeetPatil some extra eyes would be nice here -- WDYT?

Basically, we have these edge cases where the code could be _either_ block indented _or_ hanging indented. The current choice is "don't lint" _for `indentation_linter()`_ since what looks off to us here is not a problem with _indentation_ but with _delimiters_. So this code _could_ lint from other rules (it doesn't currently), but `indentation_linter()` leaves it alone.

The alternative would be for a lengthy lint message in all cases like this along the lines of

> Resolve this lint by either moving the first argument after the opening delimiter or by moving the closing delimiter onto the next line

For now, something like that _is_ done for faulty function declaration like:

```r
foo <- function(
  a = 1,
  b = 2) {
  <body>
}
```

Because I think it's a common mistake for the new-ish single-indented function formals rule. But maybe that should also not lint?

WDYT?

Issue #1979, opened 2023-06-25
spaces_inside_linter() doesn't allow spacing for aligned indentation


MichaelChirico:
```r
foo(
  c( 45, 90, 135),
  c(135, 90,  45)
)
````

Consider `c( 45,`, if we try to induce the alignment in this way, it throws `spaces_inside_linter()`, but if we try to induce it as `c(45 ,`, it throws `commas_linter()` instead.

Given that `infix_spaces_linter()` has `allow_multiple_spaces=TRUE` by default, this creates some inconsistency where most of the code allows spacing-for-alignment, except for this one initial entry.

**Comment by AshesITR** [comment]
IINM left-alignment is possible like this:

```r
foo(
  c(45,  90, 135),
  c(135, 90, 45 )
)
```

MichaelChirico:
that's right, but it looks odd to have that one comma out of place when the rest of the lines are aligned by commas. it's more obvious with more entries:

```r
foo(
  c(45,  90, 135,  45, 90, 135),
  c(135, 90,  45, 135, 90,  45)
)
```

what I've actually done for now is 0-pad that entry:

```r
foo(
  c(045, 90, 135,  45, 90, 135),
  c(135, 90,  45, 135, 90,  45)
)
```

AshesITR:
I agree that right-aligned code in your example would be nicer to read.
I'm wondering how we could reliably detect such a use of space for alignment.
Some examples to look at:

```r
f( 42 )
f(  5 )

f(42  )
f( 0.5)

f( 3, 42)
f( 5, 67)
f(89,  1)
```

MichaelChirico:
The logic will definitely be wayyyy more complicated than currently. {styler} has an associated vignette:

https://cran.r-project.org/web/packages/styler/vignettes/detect-alignment.html

This is definitely a nice-to-have, so I changed labels to 'feature' instead of 'false-positive'.

PR #3130, opened 2026-09-23
allow_alignment_calls for commas_linter to match air formatting

### LINKED ISSUE #3053: commas_linter() could allow spaces before commas to match air
Possibly related/subsumes: #2442 

`air` prefers spaces _before_ a comma when doing vertical alignment, where {styler} prefers spaces _after_ a comma, e.g.

```
# Formatted by air
tibble::tribble(
  ~x       , ~y ,
  "another", 1:3 ,
  "b"      , 1211234
)

# Formatted by styler
tibble::tribble(
  ~x,        ~y,
  "another", 1:3,
  "b",       1211234
)
```

I suspect it's possible to just support this by default without an option by checking that, for a given possible lint on 'spaces before', all lines in that call with at least the same line length _also_ have a comma in the same position. It can let through a few false negatives but I think it will be rare.

Alternatively, we could just make this an option like `allow_before` that totally skips checking spacing before commas. This approach will let in a lot more "false negatives" in the sense that a formatter will make a change that the linter did not find.

MichaelChirico:
Closes #3053. Related to #1979.

`air` defaults to allowing `tribble()` and `fcase()` (and soon, `rowwiseDT()`) to use indentation for alignment + readability:

https://github.com/posit-dev/air/pull/525

This PR aims to bring {lintr} in sync with `air` formatting so that no `commas_linter()` hits are found on `air`-compliant 'tables'.

`air` also supports doing so via `# fmt: table`, which is not supported in this PR.

Original author @randy3k using Gemini.


MichaelChirico:
The main question here is API -- is this the right design? For now, it's quite tailored to how `air` does things -- a narrow allowlist of functions supporting this type of formatting. Moreover, our naming style (`allow_alignment_calls`) does not exactly mirror that of `air` (`# fmt: table` and option `default-table`).

So, I'd like to let this simmer and gather feedback from other maintainers before submitting. cc @AshesITR @IndrajeetPatil

